PR #3, opened 2017-05-28
Start on AVX

jackmott:
I got a start on the avx functions so if I can talk another person to come help they will have an easy template to get started.


BurntSushi:
Fantastic, thank you!

PR #5, opened 2017-05-30
avx2 lhf complete

jackmott:
This represents all of the 'low hanging fruit' in avx2, the remaining functions are all tricky, so my commits will slow down, and/or I may need to bug you a bit with questions on the rest.

Alternatively I could move on to the low hanging fruit in avx.rs,  let me know what you would prefer!

### DIFF
--- a/src/x86/avx2.rs
+++ b/src/x86/avx2.rs
@@ -1,4 +1,5 @@
 use v256::*;
+use v128::*;
 use x86::__m256i;
 
 /// Computes the absolute values of packed 32-bit integers in `a`.
@@ -109,9 +110,6 @@ pub fn _mm256_avg_epu16 (a: u16x16, b: u16x16) -> u16x16 {
 pub fn _mm256_avg_epu8 (a: u8x32, b: u8x32) -> u8x32 {
     unsafe { pavgb(a, b) }
 }
-
-
-
 
 // TODO _mm256_blend_epi16
 // TODO _mm_blend_epi32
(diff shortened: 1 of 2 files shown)


src/x86/avx2.rs
```diff
@@ -568,24 +566,326 @@ pub fn _mm256_sad_epu8 (a: u8x32, b: u8x32) -> u64x4 {
...
+pub fn _mm256_sra_epi32(a: i32x8, count: i32x4) -> i32x8 {
+    unsafe { psrad(a, count) }
+}
+
+/// Shift packed 16-bit integers in `a` right by `imm8` while 
+/// shifting in sign bits.
+#[inline(always)]
+#[target_feature = "+avx2"]
+pub fn _mm256_srai_epi16(a: i16x16, imm8: i32) -> i16x16 {
+    unsafe { psraiw(a, imm8) }
```
BurntSushi: Does this (and other intrinsics accepting a `imm8`) work in debug mode?

BurntSushi:
This is awesome! Thanks so much! I kind of wonder whether some of the intrinsics with an `imm8` parameter work in debug mode? Does Clang do anything special with those?

src/x86/avx2.rs
jackmott: seem to yes, they pass the test anyway.  What have you seen in the past with those?

BurntSushi: Nothing specific to this particular intrinsic. The `imm8` parameter is *usually* just an indicator that it needs to be a literal constant. This is where I usually check with Clang's headers to see if they're doing anything tricky. For example, Clang will define an intrinsic with a `#define` (IIRC) if it needs to rely on a parameter being a constant.

jackmott: I see, and when it depends on it being a constant, is when the macro madness has to happen?

BurntSushi: Right. :-) It requires investigative work unfortunately. The fact that it works in debug mode now though is promising... but still worth double checking. I'll do it myself eventually, but not while I'm at work.

jackmott: ok yeah scanning through the clang headers, the ones I've implemented are not #defined.   I'm remembering now that the .NET team are having trouble doing these ones that require constants, because C# has no way to indicate a function parameter is a constant, and no macros to fake it.

BurntSushi: @jackmott Yeah macros + the llvm optimizer saves our butts here. My vague understanding is that folks want Rust to have const parameters some day.

jackmott:
Anything else you need from me before a merge? Just need to find time to review?

BurntSushi:
@jackmott I still think some of these intrinsics need a closer look, but I'm find taking these for now. Thanks for the ping!

PR #16, opened 2017-09-19
Added _mm_unpackhi_ps function

mohanrajendran:
I have added the `_mm_unpackhi_ps` function to `sse.rs`. This is my first pull request. Appreciate your feedback.

### DIFF
--- a/src/x86/sse.rs
+++ b/src/x86/sse.rs
@@ -1,5 +1,7 @@
 use v128::*;
 
+use simd_llvm::simd_shuffle4;
+
 /// Return the square root of packed single-precision (32-bit) floating-point
 /// elements in `a`.
 #[inline(always)]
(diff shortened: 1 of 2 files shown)


src/x86/sse.rs
```diff
@@ -51,7 +61,7 @@ pub fn _mm_movemask_ps(a: f32x4) -> i32 {
 }
 
 #[allow(improper_ctypes)]
-extern {
+extern "C" {
```
BurntSushi: What motivated this change?

src/x86/sse.rs
```diff
@@ -1,5 +1,7 @@
 use v128::*;
 
+use simd_llvm::simd_shuffle4;
```
BurntSushi: Please group this with the other crate imports in alphabetical order. :-) So it should be:

```rust
use simd_llvm::simd_shuffle4;
use v128::*;
```

Thanks!

mohanrajendran: Will do!

src/x86/sse.rs
mohanrajendran: Looks like my rustfmt pass made this change. There is an option to turn this off. I'll do that.

BurntSushi:
Fantastic, thank you!

PR #23, opened 2017-09-19
[tests] run tests only if target supports them

gnzlbg:
- Compile the x86 tests for x86 or x86_64 targets only
- Compile the tests only if the features required to run them are enabled at compile-time (this means one will need to use `RUSTFLAGS="-C target-cpu=native" cargo test` to enable all tests for a particular target architecture.


gnzlbg:
superseeded by #24 .

Issue #12, opened 2017-09-17
add license


BurntSushi:
This repo doesn't currently have a license, but it should be the same dual MIT/AL license as Rust proper.

PR #24, opened 2017-09-19
Implement abm, bmi, bmi2, and tbm intrinsics

gnzlbg:
Implements all ABM, BMI, BMI2, and TBM intrinsics: 

- abm: POPCNT, LZCNT
- bmi: ANDN, BEXTR, BLSI ,BLSMSK, BLSR, TZCNT
- bmi2: BZHI , MULX, PDEP, PEXT
- tbm: BLCFILL, BLCI, BLCIC, BLCMSK, BLCS, BLSFILL, BLSIC, T1MSKC, TZMKS

This PR compiles the tests only for those architectures that support them. That is, to compile and run all the tests that a CPU can target one should use `RUSTFLAGS="-C target-cpu=native" cargo test`. 

Known issues:
- Add TBM's BEXTRI (the only instruction missing) which triggers an LLVM code-gen error #12 .

### DIFF
--- a/src/x86/abm.rs
+++ b/src/x86/abm.rs
@@ -1,0 +1,71 @@
+//! Advanced Bit Manipulation (ABM) instructions
+//! 
+//! That is, POPCNT and LZCNT. These instructions have their own CPUID bits to
+//! indicate support.
+//!
+//! TODO: it is unclear which target feature to use here. SSE4.2 should be good
+//! enough but we might need to use BMI for LZCNT if there are any problems.
+//!
+//! The references are:
+//!
+//! - [Intel 64 and IA-32 Architectures Software Developer's Manual Volume 2: Instruction Set Reference, A-Z](http://www.intel.de/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-instruction-set-reference-manual-325383.pdf).
+//! - [AMD64 Architecture Programmer's Manual, Volume 3: General-Purpose and System Instructions](http://support.amd.com/TechDocs/24594.pdf).
+//!
+//! [Wikipedia](https://en.wikipedia.org/wiki/Bit_Manipulation_Instruction_Sets#ABM_.28Advanced_Bit_Manipulation.29)
+//! provides a quick overview of the instructions available.
+
+
+/// Counts the leading most significant zero bits.
+///
+/// When the operand is zero, it returns its size in bits.
+#[inline(always)]
+#[target_feature = "+sse4.2"]
+pub fn _lzcnt_u32(x: u32) -> u32 { x.leading_zeros() }
+
+/// Counts the leading most significant zero bits.
+///
+/// When the operand is zero, it returns its size in bits.
+#[inline(always)]
+#[target_feature = "+sse4.2"]
+pub fn _lzcnt_u64(x: u64) -> u64 { x.leading_zeros() as u64 }
+
+/// Counts the bits that are set.
+#[inline(always)]
+#[target_feature = "+sse4.2"]
+pub fn _popcnt32(x: u32) -> u32 { x.count_ones() }
+
+/// Counts the bits that are set.
+#[inline(always)]
+#[target_feature = "+sse4.2"]
+pub fn _popcnt64(x: u64) -> u64 { x.count_ones() as u64 }
+
+#[cfg(all(test, target_feature = "sse4.2", any(target_arch = "x86", target_arch = "x86_64")))]
+mod tests {
+    use x86::abm;
+
+    #[test]
+    #[target_feature = "+sse4.2"]
+    fn _lzcnt_u32() {
+        assert_eq!(abm::_lzcnt_u32(0b0101_1010u32), 25u32);
+    }
+
+    #[test]
+    #[target_feature = "+sse4.2"]
+    fn _lzcnt_u64() {
+        assert_eq!(abm::_lzcnt_u64(0b0101_1010u64), 57u64);
+    }
+
+    #[test]
+    #[target_feature = "+sse4.2"]
+    fn _popcnt32() {
+        assert_eq!(abm::_popcnt32(0b0101_1010u32), 4);
+    }
+
+    #[test]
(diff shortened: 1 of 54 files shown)


src/x86/abm.rs
```diff
@@ -0,0 +1,71 @@
...
+//!
+//! [Wikipedia](https://en.wikipedia.org/wiki/Bit_Manipulation_Instruction_Sets#ABM_.28Advanced_Bit_Manipulation.29)
+//! provides a quick overview of the instructions available.
+
+
+/// Counts the leading most significant zero bits.
+///
+/// When the operand is zero, it returns its size in bits.
+#[inline(always)]
+#[target_feature = "+sse4.2"]
```
BurntSushi: Clang seems to use the `lzcnt` target feature: https://github.com/llvm-mirror/clang/blob/a38ba9770a70056f9fdd6c71f819e5db45a105e4/lib/Headers/lzcntintrin.h

Should we be doing the same?

src/x86/abm.rs
```diff
@@ -0,0 +1,71 @@
...
+pub fn _lzcnt_u64(x: u64) -> u64 { x.leading_zeros() as u64 }
+
+/// Counts the bits that are set.
+#[inline(always)]
+#[target_feature = "+sse4.2"]
```
BurntSushi: Clang uses the `popcnt` target feature: https://github.com/llvm-mirror/clang/blob/a38ba9770a70056f9fdd6c71f819e5db45a105e4/lib/Headers/popcntintrin.h

src/x86/bmi.rs
```diff
@@ -0,0 +1,266 @@
...
+//!
+//! The reference is [Intel 64 and IA-32 Architectures Software Developer's
+//! Manual Volume 2: Instruction Set Reference,
+//! A-Z](http://www.intel.de/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-instruction-set-reference-manual-325383.pdf).
+//!
+//! [Wikipedia](https://en.wikipedia.org/wiki/Bit_Manipulation_Instruction_Sets#BMI1_.28Bit_Manipulation_Instruction_Set_1.29)
+//! provides a quick overview of the available instructions.
+
+#[allow(dead_code)]
+extern "platform-intrinsic" {
```
BurntSushi: Yeah so I'm explicitly trying not to use `platform-intrinsic` since I'd like to do away with that machinery. If you take a look at the other files, we link with the llvm intrinsics directly. My guess (based on [llvm intrinsic defs](https://gist.githubusercontent.com/anonymous/a25d3e3b4c14ee68d63bd1dcb0e1223c/raw/fb798be53a7b29f9b28a0c7a9e2b9a16ae915f71/allintrinsics) is that you probably want

```rust
extern {
    #[link_name = "llvm.x86.bmi.bextr.32"]
    fn bextr_u32(x: u32, y: u32) -> u32;
    #[link_name = "llvm.x86.bmi.bextr.64"]
    fn bextr_u32(x: u64, y: u64) -> u64;
}
```

src/x86/bmi2.rs
```diff
@@ -0,0 +1,202 @@
...
+
+#[allow(dead_code)]
+extern "platform-intrinsic" {
+    fn x86_bmi2_bzhi_32(x: u32, y: u32) -> u32;
+    fn x86_bmi2_bzhi_64(x: u64, y: u64) -> u64;
+    fn x86_bmi2_pdep_32(x: u32, y: u32) -> u32;
+    fn x86_bmi2_pdep_64(x: u64, y: u64) -> u64;
+    fn x86_bmi2_pext_32(x: u32, y: u32) -> u32;
+    fn x86_bmi2_pext_64(x: u64, y: u64) -> u64;
+}
```
BurntSushi: Same as for bmi, let's be consistent and link to llvm intrinsics directly.

src/x86/abm.rs
gnzlbg: @BurntSushi I have a comment on the top of the module with TODO mentioning this. IIRC  `lzcnt` is not a feature that `cfg(target_feature)` supports.  

That is, we would need to add support for `lzcnt` there, and in the future run-time detection support. The implementation cost of doing this is small, but all future code that wants to be portable would need to distinguish lzcnt from popcnt and SSE4.2. Whether this is worth it, I don't know, but I'd like to fill an issue in this repo to discuss it.

src/x86/bmi.rs
gnzlbg: @BurntSushi I did not knew this, should be fixed now.

gnzlbg:
@BurntSushi I've moved the `platform-intrinsic`s to use `link_llvm_intrinsics`, and also added all of the TBM instruction set, except for `bextri`, which triggers a codegen error.

src/x86/abm.rs
gnzlbg: Same as for lzcnt.

gnzlbg:
So I've checked the assembly of all generated intrinsics. It would be nice to set up travis-ci to do this automatically on each PR.

src/x86/bmi.rs
BurntSushi: Yeah technically we're still using `platform-intrinsic` for the cross platform simd llvm intrinsics (see `src/simd_llvm.rs`), but I'm trying to avoid them other than that. In particular, the platform intrinsics, IIRC, can't actually represent all of the vendor intrinsics. I think it's simpler to just write them out explicitly in Rust code instead of describing them in JSON.

src/x86/abm.rs
gnzlbg: @BurntSushi I tried to find an alternative, but I couldn't, so I've fixed this.

BurntSushi:
@gnzlbg All right this basically looks good to me! Is this ready to bring in? I will work on setting up the CI and polishing the tests.

gnzlbg:
@BurntSushi yes this is ready. I will implement the ARM intrinsics in a different branch.

BurntSushi:
Awesome thanks!

PR #25, opened 2017-09-19
SSE4.1 dot product instructions

kocsis1david:
This is my first pull request on GitHub.

I implemented the dpps and dppd instructions. They have a imm8 operand so I had to use the `constify_imm8!` macro, but it was in the `sse42.rs` file. I moved this macro to the `x86/macros.rs` file to be able to use it in both `sse42.rs` and `sse41.rs`.


src/x86/sse41.rs
```diff
@@ -1,4 +1,5 @@
 use x86::__m128i;
+use v128::*;
```
BurntSushi: Please sort imports alphabetically. :-)

src/x86/macros.rs
```diff
@@ -0,0 +1,264 @@
+
```
BurntSushi: Please remove this empty line.

src/x86/sse41.rs
```diff
@@ -10,10 +11,42 @@ pub fn _mm_blendv_epi8(
     unsafe { pblendvb(a, b, mask) }
 }
 
+/// Returns the dot product of two f64x2 vectors
+/// Imm8[1:0] is the broadcast mask, and the imm8[5:4] is the condition mask
+/// If a condition mask bit is zero, the corresponding multiplication is replaced by a value of 0.0
+/// If a broadcast mask bit is one, the result of the dot product will be stored in the return value component,
+/// if it is zero then the return component will be zero.
```
BurntSushi: The content here looks good, but could you please reformat the docs so that there is proper punctuation and so that they are wrapped to 79 columns inclusive? Here's how I'd format it with some wording tweaks:

```rust
/// Returns the dot product of two f64x2 vectors.
///
/// `imm8[1:0]` is the broadcast mask, and `imm8[5:4]` is the condition mask.
/// If a condition mask bit is zero, the corresponding multiplication is
/// replaced by a value of `0.0`. If a broadcast mask bit is one, the result of
/// the dot product will be stored in the return value component. Otherwise if
/// the broadcast mask bit is zero then the return component will be zero.
```

src/x86/sse41.rs
```diff
@@ -10,10 +11,42 @@ pub fn _mm_blendv_epi8(
...
+        }
+    }
+    constify_imm8!(imm8, call)
+}
+
+/// Returns the dot product of two f32x4 vectors
+/// Imm8[3:0] is the broadcast mask, and the imm8[7:4] is the condition mask
+/// If a condition mask bit is zero, the corresponding multiplication is replaced by a value of 0.0
+/// If a broadcast mask bit is one, the result of the dot product will be stored in the return value component,
+/// if it is zero then the return component will be zero.
```
BurntSushi: Please format these docs as well.

src/x86/sse41.rs
kocsis1david: It looks better this way, if you don't mind, I will just copy and paste it.

kocsis1david:
I also fixed #[target_feature = "+sse4.2"] in my commit and _mm_blendv_epi8.

PR #28, opened 2017-09-19
Add assertions correct instructions are generated

alexcrichton:
This commit adds a procedural macro which can be used to test instruction
generation in a lightweight way. The intention is that all functions are
annotated with:

    #[cfg_attr(test, assert_instr(maxps))]
    fn foo(...) {
        // ...
    }

and then during `cargo test --release` it'll assert the function `foo` does
indeed generate the instruction `maxps`. This only activates tests in optimized
mode to avoid debug mode inefficiencies, and it uses a literal invocation of
`objdump` and some parsing to figure out what instructions are inside each
function. Finally it also uses the `backtrace` crate to figure out the symbol
name of the relevant function and hook that up to the output of `objdump`.

I added a few assertions in the `sse` module to get some feedback, but curious
what y'all think of this!

### DIFF
--- a/assert-instr/src/lib.rs
+++ b/assert-instr/src/lib.rs
@@ -1,0 +1,247 @@
+#![feature(proc_macro)]
+
+extern crate assert_instr_macro;
+extern crate backtrace;
+extern crate cc;
+extern crate rustc_demangle;
+#[macro_use]
+extern crate lazy_static;
+
+use std::collections::HashMap;
+use std::env;
+use std::process::Command;
+use std::str;
+
+pub use assert_instr_macro::*;
+
+lazy_static! {
+    static ref DISASSEMBLY: HashMap<String, Vec<Function>> = disassemble_myself();
+}
+
+struct Function {
+    instrs: Vec<Instruction>,
+}
+
+struct Instruction {
+    parts: Vec<String>,
+}
+
+fn disassemble_myself() -> HashMap<String, Vec<Function>> {
+    let me = env::current_exe().expect("failed to get current exe");
+
+    if cfg!(target_arch = "x86_64") &&
+        cfg!(target_os = "windows") &&
+        cfg!(target_env = "msvc") {
+        let mut cmd = cc::windows_registry::find("x86_64-pc-windows-msvc", "dumpbin.exe")
+            .expect("failed to find `dumpbin` tool");
+        let output = cmd.arg("/DISASM").arg(&me).output()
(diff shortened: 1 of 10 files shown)


assert-instr/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+    instrs: Vec<Instruction>,
+}
+
+struct Instruction {
+    parts: Vec<String>,
+}
+
+fn disassemble_myself() -> HashMap<String, Vec<Function>> {
+    let me = env::current_exe().expect("failed to get current exe");
+    let output = Command::new("objdump")
```
gnzlbg: How portable is this? Does this work on Windows? Or only Linux/MacOS?

alexcrichton: An excellent question! AFAIK there's *some* method of doing this on all platforms. This'll require some degree of platform-specific code, but it's something I can take care of.

alexcrichton:
@BurntSushi if you're ok with this I can look to setting up CI on Travis/AppVeyor and enabling this there before merging.

alexcrichton:
Ok this now has Travis/AppVeyor configuration and should work on all three platforms. @BurntSushi want to enable Travis/AppVeyor for the repo to ensure this is green before merging?

gnzlbg:
Would it be possible to add rustfmt support as well? I am really not used to care about formatting, but I don't want to rustfmt on the whole crate and possibly screw up everything.

BurntSushi:
@gnzlbg The last time I tried rustfmt, it formatted the code in a way that I really wasn't happy with, particularly in deeply nested cases. But I'm OK with doing it here since most of the code isn't really that nested, and it would be good to enforce uniformity amidst a variety of anticipated contributors. I would like the line length to be set to 79 though. :-)

alexcrichton:
Ok [Linux](https://travis-ci.org/rust-lang-nursery/stdsimd/jobs/277857591) and [Windows](https://ci.appveyor.com/project/rust-lang-libs/stdsimd/build/1.0.9) are green and Travis has a huge backlog of OSX right now, but I've tested locally and it's green. Gonna go ahead and merge so we can get some Windows/Linux CI at least.

gnzlbg:
@BurntSushi I don't like rustfmt much either but I like having to create a different editor config for each project or manually indenting things even less. I think a line length of 79 is fine, now that @alexcrichton has setup CI, it would be ideal if travis would test the format as well, so that PRs get a reminder to run rustfmt.

PR #31, opened 2017-09-20
[runtime] initial run-time feature detection support

gnzlbg:
This is completely untested, but an initial step towards run-time feature support. This is based on @gz and @shepmaster libraries. Should I document this directly in the file, or somewhere else?


src/x86/runtime.rs
```diff
@@ -0,0 +1,120 @@
...
+        if inv_test_bit(ecx, 28) { value = set_bit(value, __Feature::avx as u32); }
+
+        if inv_test_bit(edx, 25) { value = set_bit(value, __Feature::sse as u32); }
+        if inv_test_bit(edx, 26) { value = set_bit(value, __Feature::sse2 as u32); }
+
+        FeatureStorage(value)
+    }
+}
+
+lazy_static! {
```
BurntSushi: Do we need `lazy_static!` here? Can we use [`std::sync::Once`](https://doc.rust-lang.org/std/sync/struct.Once.html) from the standard library so that we don't bring in any external deps?

gnzlbg: I'll look into it, never used `std::sync::Once` before. I just saw that lazy static solved the problem, is part of the nursery, and `no_std`.

alexcrichton: I think we'll actually want to eschew *everything* here and just use raw `AtomicUsize`, as that's the only thing that's actually no_std compatible. Hopefully shouldn't be too hard though! (just storing some bitflags globally)

gnzlbg: `AtomicUsize` sounds better. I've tried to search for the performance of an atomic load but haven't found much. Would it be worth it to try to make the load non-atomic?

shepmaster: > just storing some bitflags globally

How do you distinguish "hasn't been filled yet" from "doesn't support it" (both of which seem like a 0)? Is the plan to have two atomics? Use a bit to denote "has been set"?

alexcrichton: > Would it be worth it to try to make the load non-atomic?

Nah, I wouldn't bother (plus I think this'd be unsound).

> Use a bit to denote "has been queried"?

Indeed!

gnzlbg: > Use a bit to denote "has been queried"?

I'll use `std::usize::MAX` as sentinel value.

gnzlbg: > Nah, I wouldn't bother

Ok, I was thinking about initializing this before `main`.

src/macros.rs
```diff
@@ -267,3 +267,33 @@ macro_rules! define_casts {
         )+
     }
 }
+
+#[macro_export]
+macro_rules! cfg_feature_enabled {
+    ($name:tt) => (
+        {
+            #[cfg(target_feature = $name)] {
```
alexcrichton: Could this use `cfg!` to be a bit more compact?

src/x86/runtime.rs
```diff
@@ -0,0 +1,120 @@
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+#[macro_export]
+macro_rules! __unstable_detect_feature_x86 {
+    ("sse") => { ::stdsimd::vendor::__unstable_detect_feature(::stdsimd::vector::__Feature::sse{}) };
```
alexcrichton: I think the `::stdsimd` here may be better done as `$crate` (weird macro thing...)

src/x86/runtime.rs
```diff
@@ -0,0 +1,120 @@
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+#[macro_export]
+macro_rules! __unstable_detect_feature_x86 {
```
alexcrichton: It may actually be simpler if this is defined on all platforms and just returns `false` for all features?

src/x86/runtime.rs
```diff
@@ -0,0 +1,120 @@
...
+    ("fma") => { ::stdsimd::vendor::__unstable_detect_feature(::stdsimd::vector::__Feature::fma{}) };
+    ("bmi") => { ::stdsimd::vendor::__unstable_detect_feature(::stdsimd::vector::__Feature::bmi{}) };
+    ("bmi2") => { ::stdsimd::vendor::__unstable_detect_feature(::stdsimd::vector::__Feature::bmi2{}) };
+    ("abm") => { ::stdsimd::vendor::__unstable_detect_feature(::stdsimd::vector::__Feature::abm{}) };
+    ("lzcnt") => { ::stdsimd::vendor::__unstable_detect_feature(::stdsimd::vector::__Feature::abm{}) };
+    ("tbm") => { ::stdsimd::vendor::__unstable_detect_feature(::stdsimd::vector::__Feature::tbm{}) };
+    ("popcnt") => { ::stdsimd::vendor::__unstable_detect_feature(::stdsimd::vector::__Feature::popcnt{}) };
+}
+
+/// X86 CPU Feature
```
alexcrichton: Could this be `#[doc(hidden)]` with explicit documentation "please do not use this, it will break"?

src/macros.rs
gnzlbg: You mean `if cfg!(...) { ... }` ?

alexcrichton: Ah I was thinking like:

```rust
cfg!(target_feature = $name) || __unstable_detect_feature!($name)
```

and then the `__unstable_detect_feature` would do all sorts of macro craziness

src/x86/runtime.rs
gnzlbg: I think we can just make it a plain `static mut FEATURES: usize` global. The worst thing that can happen is that some threads write to the variable without synchronization, but they will be writting the exact same value.

alexcrichton: Er no, that's undefined behavior. Let's stick to atomics, and only look to alternate strategies if that proves too slow.

src/macros.rs
gnzlbg: I could do it like that, but then the code is generated for both cases, both branches must type check, etc.

alexcrichton: That's true... 

I guess I'd just favor library readability over microseconds of compile time downstream

src/x86/runtime.rs
gnzlbg: I've added it here, and also below to the other pub `__unstable_...` fn.

src/x86/runtime.rs
gnzlbg: What do you mean? On ARM we won't probably need such a macro because all the feature detection will happen on the parent macro (ARM features can only be portably detected at compile-time).

src/macros.rs
gnzlbg: That is probably a bad habit of mine. I use `#[cfg(...)]` when I want conditional compilation because that's what it does. 

I use `cfg!` to do unconditional compilation of all branches because that's what it does but... that's something I basically almost never want. If you give me a nice if-like feature for conditional compilation I'll use it every time (didn't you write a crate for that?), but until then, they mean different things.

gnzlbg: I also think that we should get rid of the `cfg!` habit when working with target features. By using `cfg!` one is basically inserting unwanted intrinsic calls in dead branches of the code. Sure LLVM should remove those, but call me paranoid if I don't want to trust what happens if LLVM actually sees one of those and then concludes that your whole function must be AVX2 enabled or similar.

src/x86/runtime.rs
gnzlbg: I can refactor this somewhere else, but I think it would be better to make compilation fail on ARM if one tries to do run-time feature detection for, e.g., SSE4.

alexcrichton: I'd just imagine something like:

```rust

#[cfg(x86)]
macro_rules! __unstable_detect_feature { ... }
#[cfg(arm)]
macro_rules! __unstable_detect_feature { ... }
#[cfg(everything_else)]
macro_rules! __unstable_detect_feature { ... }
```

gnzlbg: Ah! I see, I can do that!

gnzlbg: This will look like this though: 

```rust
#[cfg(all(target_arch = "x86", target_arch = "x86_64"))]
macro_rules! __unstable_detect_feature { ... }
#[cfg(all(target_arch = "arm", target_arch = "aarch64"))]
macro_rules! __unstable_detect_feature { ... }
// ...
#[cfg(not(any(target_arch = "x86", target_arch = "x86_64", target_arch = "arm", target_arch = "aarch64", ...)))]
macro_rules! __unstable_detect_feature { ... }
```

gnzlbg: Fixed.

src/macros.rs
```diff
@@ -267,3 +267,35 @@ macro_rules! define_casts {
...
+            }
+        }
+    )
+}
+
+#[cfg(not(any(target_arch = "x86",
+              target_arch = "x86_64")))]
+#[macro_export]
+macro_rules! __unstable_detect_feature {
+    ($t:tt) => { false }
```
alexcrichton: Is this the interface we want to end up having? If you're on x86 I think it's a static error to test for a feature that doesn't exist, so maybe it should be a static error to call this macro on other platforms with unknown features?

src/x86/runtime.rs
```diff
@@ -0,0 +1,121 @@
...
+    ("popcnt") => { $crate::vendor::__unstable_detect_feature($crate::vendor::__Feature::popcnt{}) };
+}
+
+/// X86 CPU Feature
+#[doc(hidden)]  // Please do not use this, it is an implementation detail subject to change.
+#[allow(non_camel_case_types)]
+#[repr(u8)]
+pub enum __Feature {
+    /// SSE (Streaming SIMD Extensions)
+    sse = 0,
```
alexcrichton: Are these discriminants necessary?

src/x86/runtime.rs
```diff
@@ -0,0 +1,121 @@
...
+    if inv_test_bit(ecx, 23) { value = set_bit(value, __Feature::popcnt as u32); }
+    if inv_test_bit(ecx, 28) { value = set_bit(value, __Feature::avx as u32); }
+
+    if inv_test_bit(edx, 25) { value = set_bit(value, __Feature::sse as u32); }
+    if inv_test_bit(edx, 26) { value = set_bit(value, __Feature::sse2 as u32); }
+
+    value
+}
+
+static mut FEATURES: AtomicUsize = AtomicUsize::new(::std::usize::MAX);
```
alexcrichton: This can just be a plain `static` instead of `static mut`, removing the need for `unsafe` below

src/x86/runtime.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    unsafe {
+        asm!("cpuid" : "={ecx}"(ecx), "={edx}"(edx)
+             : "{eax}"(0x00000001u32), "{ecx}"(0 as u32) : :);
+        asm!("cpuid" : "={ebx}"(ebx) : "{eax}"(0x00000007u32), "{ecx}"(0 as u32) : :);
+    }
+
+    let mut value: usize = 0;
+
+    // CPUID call with EAX=7, ECX=0 => Extended Features in EBX and ECX (unneeded):
```
alexcrichton: Can this link to a document online explaining what the bits are?

src/x86/runtime.rs
gnzlbg: Which ones do you mean? Only the SSE or all of them ?

src/macros.rs
gnzlbg: I'd prefer that. I don't know the best way to do this though. I madae this the case on the x86 macro by only including matches for existing features. Should I just leave the macro empty?

src/x86/runtime.rs
alexcrichton: All of them

src/x86/runtime.rs
gnzlbg: Done.

src/x86/runtime.rs
gnzlbg: Done.

src/x86/runtime.rs
gnzlbg: I store whether the feature is detected in an `AtomicUsize`, where each bit represents a feature. This `enum` is used to map feature "names" to bit positions. (I should document this). 

So... I don't understand what you mean. I need to map bit positions to names somewhere. Are you suggesting that I remove the `enum` and use raw numbers for this? Or are you suggesting that I remove the concrete values, that is, `= 3` and use the ones assigned automatically?

gnzlbg: I've documented the whole module as extensively as I could.

alexcrichton: Sorry what I meant was that the explicit `= n` for the discriminant isn't necessary here, and I believe it can be removed (the names can and should stay)

gnzlbg: @alexcrichton If I do that, am I guranteed that the values rustc will assign these discriminants will be in the range [0, 32] ? (I am using `enum as usize` to index the bitset).

src/macros.rs
```diff
@@ -267,3 +267,46 @@ macro_rules! define_casts {
...
+}
+
+/// On ARM features are only detected at compile-time using
+/// cfg(target_feature), so if this macro is executed the
+/// feature is not supported.
+#[cfg(any(target_arch = "arm",
+          target_arch = "aarch64"))]
+#[macro_export]
+macro_rules! __unstable_detect_feature {
+    ($t:tt) => { false }
```
alexcrichton: Could this and the fallback implementation (and x86) do something like:

```rust
macro_rules! __unstable_detect_feature {
    // ...
    ($t:tt) => (compile_error!(concat!("unknown target feature: ", $t)));
}
```

src/x86/runtime.rs
alexcrichton: Indeed you are!

src/macros.rs
gnzlbg: `compile_error!` !!! I did not knew this existed! Sure!

src/x86/runtime.rs
gnzlbg: Ah I did not know this. I thought that if I did not assign the values then rustc was able to use whatever ordering it wants. Is the sentinel value also defined ? I will just remove the `=`s then.

gnzlbg:
So I've tested this on Broadwell and Ivy Bridge and it works. If you want to check it with your CPU just `cargo test -- --nocapture` and one of the tests will output the values of all feature flags.

gnzlbg:
Resolved the conflicts, travis-ci should be green.

alexcrichton:
Ok I'm gonna go ahead and merge this and we can continue to iterate in-tree

PR #36, opened 2017-09-21
Add CI for more platforms

### LINKED ISSUE #13: testing/development strategy for support beyond Intel
Other than the vendor independent interface, this repo only contains vendor intrinsics for Intel. We will of course want to add intrinsics from other vendors on different platforms (like ARM), but I'm not quite sure how to go about it. For the most part, I'd expect it to be a relatively similar process as the Intel intrinsics, but I've held off from doing it because I don't have any systems in place for testing or development of intrinsics on other platforms.

For testing, I assume this means getting CI setup so that it runs on the various platforms that we're targeting.

For development, I'm less sure. I'd like to be able to test, say, ARM intrinsics from the comfort of my local machine. Does this mean running a VM? Or is there something better?

### LINKED ISSUE #33: code-gen tests should be compiled with the appropriate feature flags and platform
We need to extend the new code-gen test machinery to:

- use the features specified in the `#[target_feature = "+...,+..."]` attribute when compiling the test (or otherwise the proper assembly won't be generated and the test will fail, --release is not enough, the python script inferred this from the test name)
- specify a target platform for the tests to be compiled (e.g. the arm tests do not run on my machine anymore, the python script inferred this from the test name)

ping: @alexcrichton

alexcrichton:
This commit adds CI for a few more targets:

* i686-unknown-linux-gnu
* arm-unknown-linux-gnueabihf
* armv7-unknown-linux-gnueabihf
* aarch64-unknown-linux-gnu

The CI here is structured around using a Docker container to set up a test
environment and then QEMU is used to actually execute code from these platforms.
QEMU's emulation actually makes it so we can continue to just use `cargo test`,
as processes can be spawned from QEMU like `objdump` and files can be read (for
libbacktrace). Ends up being a relatively seamless experience!

Note that a number of intrinsics were disabled on i686 because they were failing
tests, and otherwise a few ARM touch-ups were made to get tests passing.

Closes #13
Closes #33

### DIFF
--- a/src/arm/v6.rs
+++ b/src/arm/v6.rs
@@ -3,12 +3,8 @@
 //! The reference is [ARMv6-M Architecture Reference
 //! Manual](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0419c/index.html).
 
-/// Reverse the order of the bytes.
-#[inline(always)]
-#[cfg_attr(test, assert_instr(rev))]
-pub fn _rev_u8(x: u8) -> u8 {
-    x.swap_bytes() as u8
-}
+#[cfg(test)]
+use assert_instr::assert_instr;
 
 /// Reverse the order of the bytes.
 #[inline(always)]
--- a/src/x86/bmi.rs
+++ b/src/x86/bmi.rs
@@ -32,6 +32,7 @@ pub fn _bextr_u32(a: u32, start: u32, len: u32) -> u32 {
 #[inline(always)]
 #[target_feature = "+bmi"]
 #[cfg_attr(test, assert_instr(bextr))]
+#[cfg(not(target_arch = "x86"))]
 pub fn _bextr_u64(a: u64, start: u64, len: u64) -> u64 {
     _bextr2_u64(a, (start & 0xffu64) | ((len & 0xffu64) << 8u64))
 }
@@ -56,6 +57,7 @@ pub fn _bextr2_u32(a: u32, control: u32) -> u32 {
 #[inline(always)]
 #[target_feature = "+bmi"]
 #[cfg_attr(test, assert_instr(bextr))]
+#[cfg(not(target_arch = "x86"))]
 pub fn _bextr2_u64(a: u64, control: u64) -> u64 {
     unsafe { x86_bmi_bextr_64(a, control) }
 }
@@ -88,6 +90,7 @@ pub fn _blsi_u32(x: u32) -> u32 {
 #[inline(always)]
 #[target_feature = "+bmi"]
 #[cfg_attr(test, assert_instr(blsi))]
+#[cfg(not(target_arch = "x86"))] // generates lots of instructions
 pub fn _blsi_u64(x: u64) -> u64 {
     x & x.wrapping_neg()
 }
@@ -104,6 +107,7 @@ pub fn _blsmsk_u32(x: u32) -> u32 {
 #[inline(always)]
 #[target_feature = "+bmi"]
 #[cfg_attr(test, assert_instr(blsmsk))]
+#[cfg(not(target_arch = "x86"))] // generates lots of instructions
 pub fn _blsmsk_u64(x: u64) -> u64 {
     x ^ (x.wrapping_sub(1u64))
 }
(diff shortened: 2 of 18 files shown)


alexcrichton:
cc @gnzlbg you may be quite interested in this!

src/arm/v6.rs
```diff
@@ -3,9 +3,11 @@
 //! The reference is [ARMv6-M Architecture Reference
 //! Manual](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0419c/index.html).
 
+#[cfg(test)]
+use assert_instr::assert_instr;
+
 /// Reverse the order of the bytes.
 #[inline(always)]
-#[cfg_attr(test, assert_instr(rev))]
 pub fn _rev_u8(x: u8) -> u8 {
```
gnzlbg: what was the problem here ?

src/x86/bmi.rs
```diff
@@ -32,6 +32,7 @@ pub fn _bextr_u32(a: u32, start: u32, len: u32) -> u32 {
 #[inline(always)]
 #[target_feature = "+bmi"]
 #[cfg_attr(test, assert_instr(bextr))]
+#[cfg(not(target_arch = "x86"))]
```
gnzlbg: should we maybe do this at the module level (in x86/mod.rs ?)

src/x86/bmi2.rs
```diff
@@ -16,7 +16,8 @@ use assert_instr::assert_instr;
 /// the low half and the high half of the result.
 #[inline(always)]
 // LLVM BUG (should be mulxl): https://bugs.llvm.org/show_bug.cgi?id=34232
-#[cfg_attr(test, assert_instr(imul))]
+#[cfg_attr(all(test, target_arch = "x86_64"), assert_instr(imul))]
+#[cfg_attr(all(test, target_arch = "x86"), assert_instr(mulx))]
```
gnzlbg: what? does this generate the right instruction  (`mulx`) on x87 but not x86_64 ? If so the LLVM bug is incorrect, because there the state of the art is that LLVM never uses `mulx` on 32bit integers. It is not profitable.

src/x86/bmi2.rs
```diff
@@ -31,6 +32,7 @@ pub fn _mulx_u32(a: u32, b: u32) -> (u32, u32) {
 #[inline(always)]
 #[cfg_attr(test, assert_instr(mulx))]
 #[target_feature = "+bmi2"]
+#[cfg(not(target_arch = "x86"))] // calls an intrinsic
```
gnzlbg: we might want to disable bmi2 for x86 at the module level as well

src/x86/tbm.rs
```diff
@@ -75,6 +75,7 @@ pub fn _blcfill_u32(x: u32) -> u32 {
 #[inline(always)]
 #[target_feature = "+tbm"]
 #[cfg_attr(test, assert_instr(blcfill))]
+#[cfg(not(target_arch = "x86"))] // generates lots of instructions
```
gnzlbg: same for tbm

gnzlbg:
I am a bit confused, I thought x86 was 32bit x86, and x86_64 64bit x86. So I would understand removing the 64bit intrinsics on 32bit x86, but it seems that some of the 32bit intrinsics are also affected. What am I missing?

src/arm/v6.rs
alexcrichton: Ah this function is a noop, so all it had was `blx`, a return instruction!

src/x86/bmi2.rs
alexcrichton: Yeah I dunno :(

alexcrichton:
Ah yeah sorry, I should have commented those more effectively, but some of the 64-bit intrinsics generated LLVM errors when emitting on i686/x86

src/arm/v6.rs
gnzlbg: Indeed, can you delete it? It was a mistake. rev is only supported for 16, 32 and 64bit, for 8bits it does not make sense :D CI saved the day!

src/x86/bmi2.rs
gnzlbg: I can't reproduce this with clang `-m32`, weird :/
https://godbolt.org/g/EVPwHw

alexcrichton: Maybe differing versions of LLVM? Or maybe we're doing the wrong target features? You may want to try checking this branch out and fiddling with the attribute as that'll show the disassembly in the test failure

src/arm/v6.rs
alexcrichton: Sure!

alexcrichton:
@BurntSushi I'm gonna go ahead and merge, but if you've got some comments on this, just lemme know :)

PR #38, opened 2017-09-22
Some SSE instructions

kocsis1david:
I've implemented some SSE instructions. The `llvm.x86.sse.add.ps`, `llvm.x86.sse.sub.ps`, etc symbols were unresolved external symbols, so I had to use the `+ - * /` operators for them.


alexcrichton:
Thanks so much @kocsis1david!

PR #39, opened 2017-09-25
Added _mm_unpacklo_ps, _mm_movehl_ps, _mm_movelh_ps and _mm_shuffle_ps

mohanrajendran:


alexcrichton:
Thanks for the PR @mohanrajendran! We recently merged the ability to have attributes like `assert_instr(foo)` to ensure that these intrinsics, generate the right instructions, mind adding that to the intrinsics added here as well?

mohanrajendran:
I am exclusively using `simd_shuffle4` for the new functions that I have added. Which `assert_instr` should I be using? Also is there any resource on what these `"platform-instrinsic"` functions do under the hood?

BurntSushi:
@mohanrajendran Thanks for doing this! Would you might squashing this PR down to a single commit?

> I am exclusively using simd_shuffle4 for the new functions that I have added. Which assert_instr should I be using?

You need to figure out which CPU instruction should be emitted for the vendor intrinsics you're implementing and then write the assert on that instruction. There should be other examples in the repo to take a look at!

> Also is there any resource on how these "platform-instrinsic" functions work under the hood?

Not really. A `platform-intrinsic` is one particular way that `rustc` implements compiler intrinsics. The simd platform intrinsics, in particular, get codegen'd explicitly via LLVM: https://github.com/rust-lang/rust/blob/3df1f7b82d9bede5122ee745cdd4e731abbce892/src/librustc_trans/intrinsic.rs#L967-L1210

alexcrichton:
Oh so typically what's happening here is that these intrinsics correspond to a particular x86 instruction, so we just want to assert that particular instruction is generate. The instruction varies per intrinsic, but for example I tried out the `_mm_shuffle_ps` function here. I'm not personally familiar for which instruction is with which function so I start out by adding:

```
#[cfg_attr(test, assert_instr(random_instr))]
```

and then the error message will spit out the expected disassembly. In this case though it looks like it's printing out a lot of instructions! I wonder if that intrinsic may not be getting lowered correctly?

I tried out `_mm_unpacklo_ps` next and got:

```
disassembly for stdsimd::x86::sse::_mm_unpacklo_ps:
         0: push %rbp
         1: mov %rsp,%rbp
         2: unpcklps %xmm1,%xmm0
         3: pop %rbp
         4: retq
         5: nopl 0x0(%rax)
thread 'x86::sse::assert_instr__mm_unpacklo_ps' panicked at 'failed to find instruction `wut` in the disassembly', assert-instr/src/lib.rs:280:4
```

so for this function I think you'll want to add an `assert_instr(uncklps)` attribute.

As for the `platform-intrinsic` business, that's currently all recognized upstream in rustc itself. All intrinsics are [defined in this library](https://github.com/rust-lang/rust/blob/master/src/librustc_platform_intrinsics/lib.rs) and the `"platform-intrinsic"` ABI basically tries to look up what the corresdponding entry is for that library.

BurntSushi:
For `_mm_shuffle_ps`, [according to Intel](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_shuffle_ps&expand=4791), it should be emitting a `shufps` instruction.

alexcrichton:
@mohanrajendran did you intend to close?

mohanrajendran:
Moved to #43 after squashing commits.

BurntSushi:
@mohanrajendran FYI it's cool to force push your PR branch after squashing.

PR #42, opened 2017-09-26
[DO NOT MERGE] Add subtraction, rounding avx instrs

AdamNiederer:
Hi all,
I've implemented subpd256, subps256, roundpd256, ceilpd256, floorpd256, and their associated tests. It's still a bit of a work in progress

- [x] Lines 67:78 are [blatantly plagiarized from LLVM](https://github.com/llvm-mirror/clang/blob/dcd8d797b20291f1a6b3e0ddda085aa2bbb382a8/lib/Headers/avxintrin.h#L398)
- [x] _mm_sub_ps256 is failing its test
- [x] assert_instr is throwing for _mm_round_pd

I'll try to iron those out tomorrow.


alexcrichton:
Thanks for the PR! What do the test failures look like if you add `assert_instr`?

AdamNiederer:
```
LLVM ERROR: Cannot select: intrinsic %llvm.x86.avx.round.pd.256
```

My intrinsics reference says there is something for VROUNDPD, though

```
  def int_x86_avx_round_pd_256 : GCCBuiltin<"__builtin_ia32_roundpd256">,
        Intrinsic<[llvm_v4f64_ty], [llvm_v4f64_ty,
                  llvm_i32_ty], [IntrNoMem]>;
```

https://gist.githubusercontent.com/anonymous/a25d3e3b4c14ee68d63bd1dcb0e1223c/raw/fb798be53a7b29f9b28a0c7a9e2b9a16ae915f71/allintrinsics

I'm not 100% sure how I should proceed; I've tried some permutations on the name of the intrinsic.

alexcrichton:
I think this means that the argument must be an immediate? Right now it takes an i32, but should it take a byte instead?

AdamNiederer:
I've tried it with both an i8 and a u8, which both fail to compile. It looks like the icc intrinsic takes a c-int (http://www.felixcloutier.com/x86/ROUNDPD.html). Does the immediate have to be a constant expression?

alexcrichton:
Ah yeah you'll find a few macros (I think `constify*` style) in the rest of the codebase to turn variables into immediates

AdamNiederer:
@alexcrichton We should be good to go now; the broken appveyor build appears to be unrelated to this change.

alexcrichton:
Looks great, thanks @AdamNiederer!

PR #43, opened 2017-09-26
Added _mm_unpacklo_ps, _mm_movehl_ps, _mm_movelh_ps and _mm_shuffle_ps

mohanrajendran:
Squashing commits from #39 .

@alexcrichton The shuffle intrinsic involves a macro that expands for each shuffle mask (8 bits -> 256 cases). Thus I have removed the `probably_only_one_instruction` flag from the `assert-instr` macro. Let me know if this is fine or whether we should find another solution.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -154,11 +154,11 @@ sse
 * [ ] `_mm_storeu_ps`
 * [ ] `_mm_storer_ps`
 * [ ] `_mm_move_ss`
-* [ ] `_mm_shuffle_ps`
+* [x] `_mm_shuffle_ps`
 * [x] `_mm_unpackhi_ps`
-* [ ] `_mm_unpacklo_ps`
-* [ ] `_mm_movehl_ps`
-* [ ] `_mm_movelh_ps`
+* [x] `_mm_unpacklo_ps`
+* [x] `_mm_movehl_ps`
+* [x] `_mm_movelh_ps`
 * [x] `_mm_movemask_ps`
 * [ ] `_mm_undefined_ps`
 
--- a/assert-instr/src/lib.rs
+++ b/assert-instr/src/lib.rs
@@ -269,9 +269,7 @@ pub fn assert(fnptr: usize, fnname: &str, expected: &str) {
         }
     }
 
-    let probably_only_one_instruction = function.instrs.len() < 20;
-
-    if found && probably_only_one_instruction {
+    if found {
         return
     }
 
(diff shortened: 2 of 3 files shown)


alexcrichton:
@mohanrajendran shouldn't `_mm_shuffle_ps` be corresponding to just one instruction? I figured it may be either a bug in the code or our own codegen if it's generating a whole lot more

mohanrajendran:
@alexcrichton : Because the `simd_shuffle4` requires constant values as its selector parameter, I am using @BurntSushi 's macros `sse2.rs` to expand each possible input mask to its own call of `simd_shuffle4`. This is generating many repeating instructions at disassembly(8 bit => 256 cases). Maybe there is a better way to do this?

alexcrichton:
Maybe in this case it's too many layers of indirection for LLVM to see through? It may be the case that a different macro expansion would work here?

BurntSushi:
@mohanrajendran LLVM should optimize through that. Are you checking the disassembly produced when optimizations are enabled? But yeah, I certainly could have made a mistake. I'll try to look into this myself and double check that other uses of the macro are correct.

alexcrichton:
Oh so actually what I think is happening is that the instruction itself *must* take an immediate. When we take the function pointer where one of the arguments is still a variable then LLVM has no choice to but to generate a ton of instructions (a jump table). Typically though with inlining LLVM will eliminate all other branches and just take one of them.

In that sense @mohanrajendran I think you're right, removing the limit on instructions here sounds like the best way to go.

alexcrichton:
I think one of the failures on AppVeyor is legitimate:

```
---- x86::sse::assert_instr__mm_movehl_ps stdout ----
        disassembly for stdsimd::x86::sse::_mm_movehl_ps:
         0: push rbp
         1: mov rbp,rsp
         2: movapd xmm0,xmmword ptr [rdx]
         3: unpckhpd xmm0,xmmword ptr [rcx]
         4: pop rbp
         5: ret
         6: ��
thread 'x86::sse::assert_instr__mm_movehl_ps' panicked at 'failed to find instruction `movhlps` in the disassembly', assert-instr\src\lib.rs:288:8
```

as for the others, I can't reproduce locally :(

BurntSushi:
> When we take the function pointer where one of the arguments is still a variable then LLVM has no choice to but to generate a ton of instructions (a jump table). Typically though with inlining LLVM will eliminate all other branches and just take one of them.

Oh man, that's nasty... Is there a way to fix `assert-instr` to account for this?

alexcrichton:
@BurntSushi now that I actually think about this, how about...

```rust
pub fn the_intrinsic(...) -> R {
}

#[cfg(test)]
#[assert_instr(foo)]
fn the_test(a: T, b: T) -> R {
    the_intrinsic(a, b, 1)
}
```

@mohanrajendran does that make sense for testing this?

alexcrichton:
Alright @mohanrajendran looks like that last strategy worked for fixing up the test while keeping the length assertions, so I've merged. Thanks so much again!

PR #44, opened 2017-09-26
Add AVX2 instruction assertiosn

alexcrichton:
Also a few other assorted modules

### DIFF
--- a/src/x86/avx2.rs
+++ b/src/x86/avx2.rs
@@ -2,9 +2,13 @@ use v256::*;
 use v128::*;
 use x86::__m256i;
 
+#[cfg(test)]
+use assert_instr::assert_instr;
+
 /// Computes the absolute values of packed 32-bit integers in `a`.
 #[inline(always)]
 #[target_feature = "+avx2"]
+#[cfg_attr(test, assert_instr(vpabsd))]
 pub fn _mm256_abs_epi32(a: i32x8) -> i32x8 {
     unsafe { pabsd(a) }
 }
@@ -12,6 +16,7 @@ pub fn _mm256_abs_epi32(a: i32x8) -> i32x8 {
 /// Computes the absolute values of packed 16-bit integers in `a`.
 #[inline(always)]
 #[target_feature = "+avx2"]
+#[cfg_attr(test, assert_instr(vpabsw))]
 pub fn _mm256_abs_epi16(a: i16x16) -> i16x16 {
     unsafe { pabsw(a) }
 }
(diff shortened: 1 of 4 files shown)


src/x86/avx2.rs
```diff
@@ -621,6 +687,7 @@ pub fn _mm256_sll_epi64(a: i64x4, count: i64x2) -> i64x4 {
 /// shifting in zeros, return the results;
 #[inline(always)]
 #[target_feature = "+avx2"]
+#[cfg_attr(test, assert_instr(vpsllw))] // TODO: should this be pslli
```
alexcrichton: @BurntSushi do you know what's up with this? (and the two below)

src/x86/avx2.rs
```diff
@@ -699,6 +774,7 @@ pub fn _mm256_sra_epi32(a: i32x8, count: i32x4) -> i32x8 {
 /// shifting in sign bits.
 #[inline(always)]
 #[target_feature = "+avx2"]
+#[cfg_attr(test, assert_instr(vpsraw))] // TODO: notvpsraiw?
```
alexcrichton: Same with this, and the one below

src/x86/avx2.rs
```diff
@@ -756,6 +838,7 @@ pub fn _mm256_srl_epi64(a: i64x4, count: i64x2) -> i64x4 {
 /// zeros
 #[inline(always)]
 #[target_feature = "+avx2"]
+#[cfg_attr(test, assert_instr(vpsrlw))] // TODO not vpsrliw?
```
alexcrichton: And this

alexcrichton:
Sounds like the minor inconsistencies here have a reason but we've forgotten why. Sounds like a comment to add at a future date!

src/x86/avx2.rs
BurntSushi: It looks like Clang's codegen tests think that `pslli` should be emitted, at least in the llvm IR: https://github.com/llvm-mirror/clang/blob/6d55fdd9ee95cdb0e6fd34df57e044faff8fe5f0/test/CodeGen/avx2-builtins.c#L988-L989

BurntSushi: But it looks like the llvm codegen tests agree with what we're observing, and specifically that pslli in the ir gets lowered to vpsllw: https://github.com/llvm-mirror/llvm/blob/4c92030df73fdfac8b57d86241d26796f7f35a80/test/CodeGen/X86/avx2-intrinsics-fast-isel.ll#L2707-L2754

PR #45, opened 2017-09-26
Run travis with as many extensions as possible

AdamNiederer:


alexcrichton:
Thanks for the PR! I've opened up https://github.com/rust-lang-nursery/stdsimd/pull/46 which I think is the strategy we'll want to take for tests here. I think the test failures are also "legitimate" in the sense that if you've got avx2 enabled some of the sse2 operations are compiled a little more optimized than they usually are. I believe the strategy in https://github.com/rust-lang-nursery/stdsimd/pull/46 handles this and should enable vanilla `cargo test` to work most of the time.

PR #46, opened 2017-09-26
Always test intrinsics unconditionally

alexcrichton:
This commit alters the test suite to unconditionally compile and run all tests,
regardless of the ambient target features enabled. This then uses a new
convenience macro, `#[simd_test]`, to guard all tests with the appropriate
`cfg_feature_enabled!` and also enable the `#[target_feature]` appropriately.


alexcrichton:
@AdamNiederer and @gnzlbg this is what I was thinking we'd do for testing, always run all tests. We then only actually run tests your CPU supports.

PR #48, opened 2017-09-26
Add a test for x86 runtime support

alexcrichton:
Make sure we agree with the `cupid` crate


alexcrichton:
cc @gnzlbg @shepmaster

PR #51, opened 2017-09-27
add unsafe to every vendor intrinsic

BurntSushi:
... and also add missing assert_instr tests.

I'm somewhat expecting CI for i686 to fail. I think a few of the vendor intrinsics lower to different instructions on i686 vs x86_64.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -21,3 +21,7 @@ build: false
 test_script:
   - cargo test --target %TARGET%
   - cargo test --target %TARGET% --release
+
+branches:
+  only:
+    - master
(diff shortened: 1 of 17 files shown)


alexcrichton:
lgtm, but looks like three failures?

osx

```
---- x86::sse2::assert_instr__mm_set1_epi8 stdout ----

	disassembly for stdsimd::x86::sse2::_mm_set1_epi8: 

	 0: pushq %rbp 

	 1: movq %rsp, %rbp 

	 2: movd %edi, %xmm0 

	 3: pxor %xmm1, %xmm1 

	 4: pshufb %xmm1, %xmm0 

	 5: popq %rbp 

	 6: retq 

	 7: nopw %cs:(%rax,%rax) 

thread 'x86::sse2::assert_instr__mm_set1_epi8' panicked at 'failed to find instruction `pshufd` in the disassembly', stdsimd-test/src/lib.rs:292:8

```

windows

```
---- x86::sse2::assert_instr__mm_cvtsi128_si32 stdout ----
	disassembly for stdsimd::x86::sse2::_mm_cvtsi128_si32: 
	 0: push rbp 
	 1: mov rbp,rsp 
	 2: mov eax,dword ptr [rcx] 
	 3: pop rbp 
	 4: ret 
	 5: �������� 
thread 'x86::sse2::assert_instr__mm_cvtsi128_si32' panicked at 'failed to find instruction `movd` in the disassembly', stdsimd-test\src\lib.rs:292:8
stack backtrace:
---- x86::sse2::assert_instr__mm_cvtsi128_si64 stdout ----
	disassembly for stdsimd::x86::sse2::_mm_cvtsi128_si64: 
	 0: push rbp 
	 1: mov rbp,rsp 
	 2: mov rax,qword ptr [rcx] 
	 3: pop rbp 
	 4: ret 
	 5: ������� 
thread 'x86::sse2::assert_instr__mm_cvtsi128_si64' panicked at 'failed to find instruction `movq` in the disassembly', stdsimd-test\src\lib.rs:292:8
```

alexcrichton:
Oh one thing is that I think we'll have to make the `#[test]` functions unsafe eventually as well (due to their usage of `#[target_feature]`, I'm fine if you want to do that here or not.

BurntSushi:
All right, I've made all the tests unsafe as well, and I've fixed up the codegen tests mostly by just removing them. Some of the sse2 intrinsics are just normal loads/stores, and it seems like expecting a particular variant of `mov` is folly. Even if you look at Intel's vendor API, they don't seem to list any specific CPU instruction for them, and most of them probably only exist for completeness (users will likely want to use the platform independent vector API for manipulating vectors instead).

PR #53, opened 2017-09-27
Add vroundps, vceilps, vfloorps, vsqrtps, vsqrtpd

AdamNiederer:
These pass their tests and should be good to go

### DIFF
--- a/src/x86/avx.rs
+++ b/src/x86/avx.rs
@@ -187,6 +187,7 @@ extern "C" {
 // TODO: Remove once a macro is ipmlemented to automate these tests
 // https://github.com/rust-lang-nursery/stdsimd/issues/49
 #[cfg(test)]
+#[target_feature = "avx"]
 #[cfg_attr(test, assert_instr(vroundps))]
 fn test_mm256_round_ps(a: f32x8) -> f32x8 {
     _mm256_round_ps(a, 0x00)
@@ -195,6 +196,7 @@ fn test_mm256_round_ps(a: f32x8) -> f32x8 {
 // TODO: Remove once a macro is ipmlemented to automate these tests
 // https://github.com/rust-lang-nursery/stdsimd/issues/49
 #[cfg(test)]
+#[target_feature = "avx"]
 #[cfg_attr(test, assert_instr(vroundpd))]
 fn test_mm256_round_pd(a: f64x4) -> f64x4 {
     _mm256_round_pd(a, 0x00)
--- a/src/x86/avx.rs
+++ b/src/x86/avx.rs
@@ -117,6 +117,65 @@ pub fn _mm256_floor_pd(a: f64x4) -> f64x4 {
     unsafe { roundpd256(a, 0x01) }
 }
 
+/// Round packed single-precision (32-bit) floating point elements in `a`
+/// according to the flag `b`. The value of `b` may be as follows:
+/// 0x00: Round to the nearest whole number.
+/// 0x01: Round down, toward negative infinity.
+/// 0x02: Round up, toward positive infinity.
+/// 0x03: Truncate the values.
+/// For a few additional values options, check the LLVM docs:
+/// https://github.com/llvm-mirror/clang/blob/dcd8d797b20291f1a6b3e0ddda085aa2bbb382a8/lib/Headers/avxintrin.h#L382
+#[inline(always)]
+#[target_feature = "+avx"]
+// #[cfg_attr(test, assert_instr(vroundps))]
+// TODO: Replace with assert_expanded_instr https://github.com/rust-lang-nursery/stdsimd/issues/49
+pub fn _mm256_round_ps(a: f32x8, b: i32) -> f32x8 {
+    macro_rules! call {
+        ($imm8:expr) => {
+            unsafe { roundps256(a, $imm8) }
+        }
+    }
+    constify_imm8!(b, call)
+}
+
+/// Round packed single-precision (32-bit) floating point elements in `a` toward
+/// positive infinity.
+#[inline(always)]
+#[target_feature = "+avx"]
+// #[cfg_attr(test, assert_instr(vroundps))]
+// TODO: Replace with assert_expanded_instr https://github.com/rust-lang-nursery/stdsimd/issues/49
+pub fn _mm256_ceil_ps(a: f32x8) -> f32x8 {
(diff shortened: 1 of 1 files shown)


src/x86/avx.rs
```diff
@@ -85,13 +85,78 @@ pub fn _mm256_floor_pd(a: f64x4) -> f64x4 {
...
+    }
+    constify_imm8!(b, call)
+}
+
+/// Round packed single-precision (32-bit) floating point elements in `a` toward
+/// positive infinity.
+#[inline(always)]
+#[target_feature = "+avx"]
+// #[cfg_attr(test, assert_instr(vroundps))]
+// TODO: Replace with assert_expanded_instr https://github.com/rust-lang-nursery/stdsimd/issues/49
```
alexcrichton: I think this can be tested b/c it doesn't use a macro internally?

src/x86/avx.rs
```diff
@@ -85,13 +85,78 @@ pub fn _mm256_floor_pd(a: f64x4) -> f64x4 {
...
+pub fn _mm256_ceil_ps(a: f32x8) -> f32x8 {
+    unsafe { roundps256(a, 0x02) }
+}
+
+/// Round packed single-precision (32-bit) floating point elements in `a` toward
+/// negative infinity.
+#[inline(always)]
+#[target_feature = "+avx"]
+// #[cfg_attr(test, assert_instr(vroundps))]
+// TODO: Replace with assert_expanded_instr https://github.com/rust-lang-nursery/stdsimd/issues/49
```
alexcrichton: Similarly as above I think this can be tested as well?

src/x86/avx.rs
```diff
@@ -85,13 +85,78 @@ pub fn _mm256_floor_pd(a: f64x4) -> f64x4 {
...
+/// according to the flag `b`. The value of `b` may be as follows:
+/// 0x00: Round to the nearest whole number.
+/// 0x01: Round down, toward negative infinity.
+/// 0x02: Round up, toward positive infinity.
+/// 0x03: Truncate the values.
+/// For a few additional values options, check the LLVM docs:
+/// https://github.com/llvm-mirror/clang/blob/dcd8d797b20291f1a6b3e0ddda085aa2bbb382a8/lib/Headers/avxintrin.h#L382
+#[inline(always)]
+#[target_feature = "+avx"]
+// #[cfg_attr(test, assert_instr(vroundps))]
```
alexcrichton: I think this can be tested through the make-a-shim-that-calls-this-with-an-immediate strategy, right?

src/x86/avx.rs
```diff
@@ -180,4 +245,56 @@ mod tests {
         let expected_up = f64x4::new(2.0, 3.0, 4.0, -1.0);
         assert_eq!(result_up, expected_up);
     }
+
+    #[test]
+    #[target_feature = "+avx"]
```
alexcrichton: Could this use the new `#[simd_test]` attribute as well?

src/x86/avx.rs
```diff
@@ -180,4 +245,56 @@ mod tests {
...
+    pub fn _mm256_round_ps() {
```
alexcrichton: I think you can avoid the `pub` on the test functions

src/x86/avx.rs
AdamNiederer: I couldn't get it working reliably for round_pd, so I'm leaving it off of round_ps for now. I think the stuff in mod tests should cover us until we find a better way of doing it.

alexcrichton: Does this work?

```rust
#[cfg(test)]
#[cfg_attr(test, assert_instr(vroundps))]
fn test_mm256_round_ps(a: f32x8) -> f32x8 {
    _mm256_found_ps(a, 3)
}
```

AdamNiederer: That's the ticket. Just added it for round_*.

alexcrichton:
Looks like appveyor may be red?

AdamNiederer:
I think I was due for a rebase. Here's to not breaking everything!

AdamNiederer:
Appveyor is red again, but I know you can avoid that error by specifying `RUSTFLAGS="-C target-cpu=native"`.

src/x86/avx.rs
```diff
@@ -126,9 +176,31 @@ extern "C" {
...
+    fn sqrtpd256(a: f64x4) -> f64x4;
+    #[link_name = "llvm.x86.avx.sqrt.ps.256"]
+    fn sqrtps256(a: f32x8) -> f32x8;
 }
 
+// TODO: Remove once a macro is ipmlemented to automate these tests
+// https://github.com/rust-lang-nursery/stdsimd/issues/49
 #[cfg(test)]
+#[cfg_attr(test, assert_instr(vroundps))]
+fn test_mm256_round_ps(a: f32x8) -> f32x8 {
```
alexcrichton: I think this is missing `#[target_feature]`? 

(and the one below)

could these also move next to the functions they're testing? (hopefully temporarily). With https://github.com/rust-lang-nursery/stdsimd/pull/57 this shouldn't even be necessary to have a shim!

src/x86/avx.rs
```diff
@@ -92,13 +92,16 @@ pub fn _mm256_round_pd(a: f64x4, b: i32) -> f64x4 {
     constify_imm8!(b, call)
 }
 
+// TODO: Remove once a macro is ipmlemented to automate these tests
+// https://github.com/rust-lang-nursery/stdsimd/issues/49
 #[cfg(test)]
+#[target_feature = "avx"]
```
alexcrichton: +avx I think?

src/x86/avx.rs
```diff
@@ -117,6 +120,72 @@ pub fn _mm256_floor_pd(a: f64x4) -> f64x4 {
...
+            unsafe { roundps256(a, $imm8) }
+        }
+    }
+    constify_imm8!(b, call)
+}
+
+// TODO: Remove once a macro is ipmlemented to automate these tests
+// https://github.com/rust-lang-nursery/stdsimd/issues/49
+#[cfg(test)]
+#[target_feature = "avx"]
```
alexcrichton: +avx here too

AdamNiederer:
Rebased and fixed conflicts. I think we're good now.

alexcrichton:
Thanks!

PR #57, opened 2017-09-27
Enhance `#[assert_instr]` with constant arguments

### LINKED ISSUE #49: Add assert_instr equivalent for expanded forms
Currently, assert_instr() doesn't work well for functions with `constify_imm8!` in them, but there is a workaround. Adding a test-only helper function of the following form (in this example, for `_mm_cmpestri`) makes it work.

```rust
    #[cfg(test)]
    #[target_feature = "+sse4.2"]
    #[cfg_attr(test, assert_instr(pcmpestri))]
    fn _mm_cmpestri_test(
        a: __m128i,
        la: i32,
        b: __m128i,
        lb: i32,
        imm8: i8,
    ) -> i32
    {
        sse42::_mm_cmpestri(a, la, b, lb, sse42::_SIDD_CMP_EQUAL_ORDERED)
    }
```

Because all of the information we need to do something like this is in the function header, there should probably be a macro like `assert_expanded_instr` which allows us to autogenerate this test function.

alexcrichton:
Some intrinsics need to be invoked with constant arguments to get the right
instruction to get generated, so this commit enhances the `assert_instr` macro
to enable this ability. Namely you pass constant arguments like:

    #[assert_instr(foo, a = b)]

where this will assert that the intrinsic, when invoked with argument `a` equal
to the value `b` and all other arguments passed from the outside, will generate
the instruction `foo`.

Closes #49

### DIFF
--- a/stdsimd-test/assert-instr-macro/Cargo.toml
+++ b/stdsimd-test/assert-instr-macro/Cargo.toml
@@ -5,3 +5,9 @@ authors = ["Alex Crichton <alex@alexcrichton.com>"]
 
 [lib]
 proc-macro = true
+
+[dependencies]
+proc-macro2 = { version = "0.1", features = ["unstable"] }
+quote = { git = 'https://github.com/dtolnay/quote' }
+syn = { git = 'https://github.com/dtolnay/syn', features = ["full"] }
+synom = { git = 'https://github.com/dtolnay/syn' }
--- a/src/x86/avx.rs
+++ b/src/x86/avx.rs
@@ -84,18 +84,12 @@ pub unsafe fn _mm256_sub_ps(a: f32x8, b: f32x8) -> f32x8 {
 /// ```
 #[inline(always)]
 #[target_feature = "+avx"]
+#[cfg_attr(test, assert_instr(vroundpd, b = 0x3))]
 pub unsafe fn _mm256_round_pd(a: f64x4, b: i32) -> f64x4 {
     macro_rules! call {
         ($imm8:expr) => { roundpd256(a, $imm8) }
     }
     constify_imm8!(b, call)
-}
-
-#[cfg(test)]
-#[cfg_attr(test, assert_instr(vroundpd))]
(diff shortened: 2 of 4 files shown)


stdsimd-test/assert-instr-macro/Cargo.toml
```diff
@@ -5,3 +5,9 @@ authors = ["Alex Crichton <alex@alexcrichton.com>"]
 
 [lib]
 proc-macro = true
+
+[dependencies]
+proc-macro2 = { version = "0.1", features = ["unstable"] }
+quote = { git = 'https://github.com/dtolnay/quote' }
+syn = { git = 'https://github.com/dtolnay/syn', features = ["full"] }
+synom = { git = 'https://github.com/dtolnay/syn' }
```
BurntSushi: If/when this crate gets moved to `core`, will we still be able to carry these dependencies?

BurntSushi: Orthogonally, is there a way to use non-git dependencies? I actually kind of want it to be possible for `stdsimd` to be on crates.io. It would be nice to use it before it gets to `core`. :-)

(I could very well be misunderstanding. Do these need to be non-git deps to be able to put `stdsimd` on crates.io?)

alexcrichton: Yeah I think this'll work out just fine, even when this is moved into `core`. The `core` crate itself isn't actually tested, and my guess is we'll still have this repo *actually* doing all the testing and then in-tree we won't do any testing. These dependencies and such shouldn't harm the build in rust-alng/rust though.


Also, yeah, it's true that this'd hinder publishing on crates.io. What we'd just do though is comment out the `[dev-dependencies]` section of Cargo.toml in the `stdsimd` crate, and then we'd be good to go.

BurntSushi: Bummer, but all right, that's fine.

PR #60, opened 2017-09-28
[arm] fix aarch64 cls intrinsic

### LINKED ISSUE #30: x86::arm::v8::cls doesn't lower to cls
This is due to [LLVM bug 31802](https://bugs.llvm.org/show_bug.cgi?id=31802). We could use inline assembly to workaround this in the mean time.

gnzlbg:
Closes #30 .


BurntSushi:
Nice!

PR #61, opened 2017-09-28
[arm] fix unused unsafe warning

gnzlbg:


src/arm/v8.rs
```diff
@@ -31,7 +31,7 @@ extern "C" {
 #[inline(always)]
 #[cfg_attr(test, assert_instr(rbit))]
 pub unsafe fn _rbit_u64(x: u64) -> u64 {
```
BurntSushi: Should we actually be marking this function as unsafe even though it doesn't use `#[target_feature]`? Maybe it will one day use `#[target_feature]`?

gnzlbg: This could probably use `#[target_feature = "+v8"]`. Right now it just calls an llvm intrinsic that is portable (it also works on x86 btw).

 @alexcrichton changed it in https://github.com/rust-lang-nursery/stdsimd/commit/556b2e7bda90dbbac7856066461b835f9831a92b

BurntSushi: Oh I see. I hadn't made them unsafe because they didn't use `#[target_feature]`, but it does seem smarter to just do it now in case we add target_feature later.

BurntSushi:
Thanks!

PR #68, opened 2017-09-28
ssse3

gwenn:
- _mm_abs_epi16
- _mm_abs_epi32
- _mm_hadd_epi16
- _mm_hadds_epi16
- _mm_hadd_epi32
- _mm_hsub_epi16
- _mm_hsubs_epi16
- _mm_hsub_epi32
- _mm_maddubs_epi16
- _mm_mulhrs_epi16
- _mm_sign_epi8
- _mm_sign_epi16
- _mm_sign_epi32

Cannot implement all `_pi`s intrinsics due to `Intrinsic has incorrect return type!` error.
Try [this](https://gist.github.com/gwenn/01d84c1a7c7d7b33fa996a51949e342f) patch or [that](https://gist.github.com/gwenn/7a37f3146bee665a9ebb9523bee25793) one if you want to give it a try.

And there is no LLVM instrinsic associated to `_mm_alignr_epi8` (no `__builtin_ia32_palignr128`) so I don't know what to do.


alexcrichton:
Thanks!

BurntSushi:
> And there is no LLVM instrinsic associated to _mm_alignr_epi8 (no __builtin_ia32_palignr128) so I don't know what to do.

Yeah this is a tricky one. Clang uses `__builtin_ia32_palignr128`, which appears to be a Clang specific compiler intrinsic. You can see its codegen here: https://github.com/llvm-mirror/clang/blob/c8b57899133eb605927d729d08ee1c12ada540d2/lib/CodeGen/CGBuiltin.cpp#L7883-L7924

In particular, these lines tell us what we need to do:

```cpp
    Value *Align = Builder.CreateShuffleVector(Ops[1], Ops[0],
                                               makeArrayRef(Indices, NumElts),
                                               "palignr");
```

That means we need to call one of the `simd_shuffleN` routines ourselves.

PR #69, opened 2017-09-28
added support for _mm_cvtpd_epi32 / cvtpd2dq

pythoneer:
i am not a hundred percent sure if the last three test are what we really want here, just looks weird that it "defaults" to `i32::MIN` but maybe that's supposed to happen, can't really find docs about that behavior.

### DIFF
--- a/src/x86/sse2.rs
+++ b/src/x86/sse2.rs
@@ -1726,6 +1726,14 @@ pub unsafe fn _mm_cvtpd_ps(a: f64x2) -> f32x4 {
     cvtpd2ps(a)
 }
 
+/// Convert packed double-precision (64-bit) floating-point elements in `a` to packed 32-bit integers.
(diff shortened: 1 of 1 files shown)


BurntSushi:
Thanks!

I'm not sure where that behavior is documented either. The [Intel docs](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_cvtpd_epi32&expand=1605) say that each float is converted using `Convert_FP64_To_Int32`, but the behavior of `Convert_FP64_To_Int32` isn't documented anywhere. It does at least appear consistent with [Rust's rules for converting `f64` values to `i32` values](https://play.rust-lang.org/?gist=33b1aaf4dd6afeb2049dce8b72a4a7ee&version=stable), so I guess that's good?

Anyway, this looks great to me! Awesome work. :-)

PR #70, opened 2017-09-29
Implement the sse4.2 string comparison intrinsics

dlrobertson:
- Implement
   - _mm_cmpistrm
   - _mm_cmpistri
   - _mm_cmpistrz
   - _mm_cmpistrc
   - _mm_cmpistrs
   - _mm_cmpistro
   - _mm_cmpistra
   - _mm_cmpestrm
   - _mm_cmpestrz
   - _mm_cmpestrc
   - _mm_cmpestrs
   - _mm_cmpestro
   - _mm_cmpestra
 - Add documentation to _mm_cmpestri
 - Add missing constants
 - Fix typo in lib docs

I'm working on getting sse4.2 covered, but I figured this was a nice chunk to start with. Comments and critiques are welcomed!

### DIFF
--- a/src/x86/sse42.rs
+++ b/src/x86/sse42.rs
@@ -1,25 +1,182 @@
 #[cfg(test)]
 use stdsimd_test::assert_instr;
 
+use v128::*;
 use x86::__m128i;
 
+/// String contains unsigned 8-bit characters
 pub const _SIDD_UBYTE_OPS: i8 = 0b00000000;
+/// String contains unsigned 16-bit characters
 pub const _SIDD_UWORD_OPS: i8 = 0b00000001;
+/// String contains signed 8-bit characters
 pub const _SIDD_SBYTE_OPS: i8 = 0b00000010;
+/// String contains unsigned 16-bit characters
 pub const _SIDD_SWORD_OPS: i8 = 0b00000011;
 
+/// For each character in `a`, find if it is in `b`
 pub const _SIDD_CMP_EQUAL_ANY: i8 = 0b00000000;
+/// For each character in `a`, determine if `b[0] <= c <= b[1] or b[1] <= c <= b[2]...`
 pub const _SIDD_CMP_RANGES: i8 = 0b00000100;
+/// String equality
 pub const _SIDD_CMP_EQUAL_EACH: i8 = 0b00001000;
+/// Substring search
 pub const _SIDD_CMP_EQUAL_ORDERED: i8 = 0b00001100;
 
+/// Do not negate results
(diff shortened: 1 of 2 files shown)


alexcrichton:
Looks great to me, thanks! I think @BurntSushi may want to take a look at the docs though to help out?

src/x86/sse42.rs
```diff
@@ -2,6 +2,7 @@
 use stdsimd_test::assert_instr;
 
 use x86::__m128i;
+use v128::*;
```
BurntSushi: Please arrange imports in alphabetical order. :-)

src/x86/sse42.rs
```diff
@@ -15,11 +16,151 @@ pub const _SIDD_CMP_EQUAL_ORDERED: i8 = 0b00001100;
...
 pub const _SIDD_POSITIVE_POLARITY: i8 = 0b00000000;
 pub const _SIDD_NEGATIVE_POLARITY: i8 = 0b00010000;
+pub const _SIDD_MASKED_POSITIVE_POLARITY: i8 = 0b00100000;
 pub const _SIDD_MASKED_NEGATIVE_POLARITY: i8 = 0b00110000;
 
 pub const _SIDD_LEAST_SIGNIFICANT: i8 = 0b00000000;
 pub const _SIDD_MOST_SIGNIFICANT: i8 = 0b01000000;
 
+pub const _SIDD_BIT_MASK: i8 = 0b00000000;
+pub const _SIDD_UNIT_MASK: i8 = 0b01000000;
```
BurntSushi: We'll want to add docs to each of these constants, but I'm happy to punt on that for now!

dlrobertson: Added some short doc strings.

src/x86/sse42.rs
dlrobertson: Good point.

dlrobertson:
> I'm hoping we can improve the docs all around here since the packed string compare functions have a lot of interesting semantics and modes, and docs for them are rather sparse.

100% agree. I'd like to circle back at some point and add some examples etc.

> One thing I think would be nice here would be to improve the tests by using actual strings. I think the first test started out that way, just because it seems more intelligible to me... Would you want to try to do that in this PR?

I started by using strings, but I found it easier to debug when manually specifying the ASCII or UTF-16 bytes. Should be a simple change to use strings instead of the raw bytes. I can definitely include that in this PR.

BurntSushi:
@dlrobertson Awesome, thanks! (Maybe I'm wrong that strings are easier?)

BurntSushi:
@dlrobertson Nice work on the tests! They look much better. :-) I'll merge once CI passes.

Issue #71, opened 2017-09-29
LLVM emits single precision variation of double precision intrinsics


p32blo:
The affected intrinsics are:

AVX Intrinsic | Expected Instruction | Emitted Instruction
----------|----------------------------|-----------------------
`_mm256_and_pd` | `vandpd` | `vandps`
`_mm256_andnot_pd` | `vandnpd` | `vandnps`
`_mm256_or_pd` | `vorpd` | `vorps`
`_mm256_xor_pd` | `vxorpd` | `vxorps`

Checking in clang [here](https://godbolt.org/g/cmWBdu) confirms this behavior. Note how gcc emits the correct instructions while clang always uses the single precision variation.

The result does not seem to be affected, since these are bitwise operations on vectors of the same size. However, since the expected instructions are not generated,  the `assert_*` tests will fail if this behavior is not taken into account.

alexcrichton:
Ok I'm going to close this for now as I think in general we're not going to be able to fix this but we're ruling this as "it's ok". LLVM in general will performan lots of optimizations under the hood, so we can't really guarantee *precise* intructions unfortunately!

avl:
Don't vandpd and vandps do exactly the same thing anyway? Are they actually emitted as different opcodes by the assembler?

gnzlbg:
Yes and yes.

PR #72, opened 2017-09-29
Add vandpd, vandps, vorpd and vorps for AVX

p32blo:
In clang's  [avxintrin.h](https://github.com/llvm-mirror/clang/blob/master/lib/Headers/avxintrin.h#L528-L532) this intrinsics are implemented with a cast to an **unsigned** type but in [sse2.rs](https://github.com/rust-lang-nursery/stdsimd/blob/master/src/x86/sse2.rs#L1382), for example, similar instructions use a **signed** type. Does this make a difference? Should I change for the sake of consistency?

Also, llvm emits the wrong instruction for the double precision intrinsics (see Issue #71), so the `assert_instr` test is changed to the emitted instruction, as suggested on the gitter chat, to please the tests.

### DIFF
--- a/src/x86/avx.rs
+++ b/src/x86/avx.rs
@@ -20,6 +20,48 @@ pub unsafe fn _mm256_add_ps(a: f32x8, b: f32x8) -> f32x8 {
     a + b
 }
 
+/// Compute the bitwise AND of a packed double-precision (64-bit) floating-point elements
+/// in `a` and `b`.
+#[inline(always)]
+#[target_feature = "+avx"]
+#[cfg_attr(test, assert_instr(vandpd))]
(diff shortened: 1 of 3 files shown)


src/x86/avx.rs
```diff
@@ -1,6 +1,8 @@
 #[cfg(test)]
 use stdsimd_test::assert_instr;
 
+use std::mem;
```
BurntSushi: I try to group imports in three groups: std imports, external crate imports and then finally internal imports. So I'd write it like this:

```rust
use std::mem;

#[cfg(test)]
use stdsimd_test::assert_instr;

use v256::*;
```

BurntSushi:
Thanks so much! I'll bring this in once CI passes. :-)

PR #75, opened 2017-09-29
Added some more sse2 conversion intrinsics

krampenschiesser:
added some more conversions.
However I have a problem with ```cvtsd2si64``` which gets converted to ```cvtsd2si rax,mmword ptr [rcx] ```
and not ```cvtsd2si64``` . Maybe somebody has an idea?


krampenschiesser:
Will do again

BurntSushi:
@krampenschiesser Note that it's OK to just rebase and force push!

PR #78, opened 2017-09-30
Implement _mm_prefetch

nominolo:
This boils down to using LLVMs [`prefetch`][1] intrinsic. The regular [`_mm_prefetch`][2] intrinsic takes a constant int as a second parameter, so I used the same API.

[1]: https://llvm.org/docs/LangRef.html#llvm-prefetch-intrinsic
[2]: https://gist.github.com/anonymous/25d752fda8521d29699a826b980218fc#file-intel-intrinsics-3-3-15-xml-L2280

### DIFF
--- a/src/x86/sse.rs
+++ b/src/x86/sse.rs
@@ -1,5 +1,6 @@
 use simd_llvm::simd_shuffle4;
 use v128::*;
+use std::os::raw::c_void;
 
 #[cfg(test)]
 use stdsimd_test::assert_instr;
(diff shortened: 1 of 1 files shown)


src/x86/sse.rs
```diff
@@ -267,6 +268,80 @@ pub unsafe fn _mm_movemask_ps(a: f32x4) -> i32 {
...
+///
+/// * Prefetching may also fail if there are not enough memory-subsystem
+///   resources (e.g., request buffers).
+///
+#[inline(always)]
+#[target_feature = "+sse"]
+#[cfg_attr(test, assert_instr(prefetcht0, strategy = _MM_HINT_T0))]
+// #[cfg_attr(test, assert_instr(prefetcht1, strategy = _MM_HINT_T1))]
+// #[cfg_attr(test, assert_instr(prefetcht2, strategy = _MM_HINT_T2))]
+// #[cfg_attr(test, assert_instr(prefetchnta, strategy = _MM_HINT_NTA))]
```
BurntSushi: Oh this is interesting. It looks like this is something we'd actually want to test. @alexcrichton Is there a way to have multiple `assert_instr` tests for one function?

alexcrichton: An interesting idea! I suspect this'd require some small tweaks to the macro definition, notably [this line](https://github.com/rust-lang-nursery/stdsimd/blob/3a8bf80ba313214274324e5231a2326067f0a5ff/stdsimd-test/assert-instr-macro/src/lib.rs#L70), we only want to append `target_feature` attributes, not all I think.

nominolo: The other change would be to allow multiple strings in case the compiler is free to choose. See https://github.com/rust-lang-nursery/stdsimd/issues/71 and https://github.com/rust-lang-nursery/stdsimd/issues/50

alexcrichton:
This looks like it's good to go modulo more testing, which I'll work on in parallel

PR #79, opened 2017-09-30
sse3

gwenn:
Complete.

### DIFF
--- a/src/x86/sse2.rs
+++ b/src/x86/sse2.rs
@@ -1762,6 +1762,16 @@ pub unsafe fn _mm_store_pd(mem_addr: *mut f64, a: f64x2) {
     *(mem_addr as *mut f64x2) = a;
 }
 
+/// Load a double-precision (64-bit) floating-point element from memory
+/// into both elements of returned vector.
+#[inline(always)]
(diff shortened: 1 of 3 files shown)


src/x86/sse2.rs
```diff
@@ -1762,6 +1762,16 @@ pub unsafe fn _mm_store_pd(mem_addr: *mut f64, a: f64x2) {
     *(mem_addr as *mut f64x2) = a;
 }
 
+/// Load a double-precision (64-bit) floating-point element from memory
+/// into both elements of returned vector.
+#[inline(always)]
+#[target_feature = "+sse2"]
+#[cfg_attr(test, assert_instr(movddup))]
```
BurntSushi: I've found that testing specific instructions for intrinsics that are basically just moves isn't too reliable. The specific mov variants seems to differ across targets, so I've just been leaving off tests for those.

BurntSushi: (And yeah, it looks like this is causing CI to fail on Linux.)

gwenn: Fixed.

BurntSushi:
Sweet!

PR #82, opened 2017-09-30
Alignr

gwenn:
`_mm_alignr_epi8` is implemented.
 I would also like to implement `_mm256_alignr_epi8`. Is there something like simd_shuffle32 ?

### DIFF
--- a/src/x86/ssse3.rs
+++ b/src/x86/ssse3.rs
@@ -1,6 +1,7 @@
 #[cfg(test)]
 use stdsimd_test::assert_instr;
 
+use simd_llvm::simd_shuffle16;
 use v128::*;
 
 /// Compute the absolute value of packed 8-bit signed integers in `a` and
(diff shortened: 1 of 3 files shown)


src/x86/ssse3.rs
```diff
@@ -61,6 +62,54 @@ pub unsafe fn _mm_shuffle_epi8(a: u8x16, b: u8x16) -> u8x16 {
...
+    if n > 32 {
+        return i8x16::splat(0);
+    }
+    // If palignr is shifting the pair of input vectors more than one lane,
+    // but less than two lanes, convert to shifting in zeroes.
+    let (a, b, n) = if n > 16 {
+        (i8x16::splat(0), a, n - 16)
+    } else {
+        (a, b, n)
+    };
```
BurntSushi: Are you sure we need these if conditionals? Do they result in additional codegen with branches? If so, I believe that's wrong. Unfortunately we don't have the machinery to test something like that yet.

BurntSushi:
Thanks! I'm not sure about the implementation here though unfortunately.

> I would also like to implement _mm256_alignr_epi8. Is there something like simd_shuffle32 ?

If my [reading of rustc's codegen for simd shuffles is right](https://github.com/rust-lang/rust/blob/4491ea5a3f21b63e7d0f39d9294308673a6e9715/src/librustc_trans/intrinsic.rs#L1047), then it sounds like you can just [add `simd_shuffleXX`](https://github.com/rust-lang-nursery/stdsimd/blob/5503eff1f8a6310e4e581be8143f611feb87cee3/src/simd_llvm.rs#L9-L12)c and it should work fine as long as you get the types right in the declaration.

src/x86/ssse3.rs
gwenn: The corresponding clang code is [here](https://github.com/llvm-mirror/clang/blob/c8b57899133eb605927d729d08ee1c12ada540d2/lib/CodeGen/CGBuiltin.cpp#L7898-L7902).

BurntSushi: Yeah but that's in trans operating on compile time constants, so the codegen from that is very explicit. We're in a different boat here, where we need to make sure we aren't introducing extra branches, which kind of relies heavily on the optimizer. Maybe paste the codegen for this function somewhere? (An easy way to get it is to cause the `assert_instr` test to fail.)

gwenn: When `n` is 15:
```
	 0: pushq %rbp
	 1: movq %rsp, %rbp
	 2: palignr $0xf, %xmm1, %xmm0
	 3: popq %rbp
	 4: retq
	 5: nopl (%rax)
```
When `n` is 17:
```
	 0: pushq %rbp
	 1: movq %rsp, %rbp
	 2: psrldq $0x1, %xmm0
	 3: popq %rbp
	 4: retq
	 5: nopl (%rax,%rax)
```

BurntSushi: Good enough for me!

PR #88, opened 2017-09-30
Implement _mm_getcsr, _mm_setcsr, _mm_sfence

nominolo:
This PR implements the following `sse` intrinsics:

- `_mm_getcsr`
- `_mm_setcsr`
- `_MM_GET_EXCEPTION_STATE`
- `_MM_SET_EXCEPTION_STATE`
- `_MM_GET_EXCEPTION_MASK`
- `_MM_SET_EXCEPTION_MASK`
- `_MM_GET_ROUNDING_MODE`
- `_MM_SET_ROUNDING_MODE`
- `_MM_GET_FLUSH_ZERO_MODE`
- `_MM_SET_FLUSH_ZERO_MODE`
- `_mm_sfence`

However, all the `_MM_*` helpers are macros in C, but functions in Rust, so I turned them into lower-case names. See https://github.com/rust-lang-nursery/stdsimd/issues/84 for the open policy discussion on this.

Testing this turned out to be a bit tricky because almost all of the small unit tests seem to get constant folded (see https://github.com/rust-lang-nursery/stdsimd/issues/86).

The `_mm_sfence` intrinsic doesn't quite fit in thematically. I can put it into a separate PR if that's preferred.


nominolo:
Hm, the test failure is likely due to constant folding. Do Travis CI and appveyor use different LLVM versions? The `rustc` versions seem to be the same.

alexcrichton:
I wonder, if this is setting a CPU register, does it work in multithreaded programs? Will this need to be tested as a separate executable? Other than that using functions for macros seems fine, and maybe `test::black_box` could be used to inhibit constant folding?

nominolo:
`_mm_setcsr` must only affect the current thread (otherwise, that's a CPU or Kernel bug). I wrote tests in such a way that there are no asserts between modifying and restoring the state, so even if multiple tests are run in the same thread, it should be safe. I don't know if Rust uses a new thread per test.

alexcrichton:
Awesome, thanks @nominolo!

PR #89, opened 2017-10-01
added _mm_cvttps_epi32

pythoneer:


src/x86/sse2.rs
```diff
@@ -3606,6 +3618,19 @@ mod tests {
...
+    unsafe fn _mm_cvttps_epi32() {
+        use std::{f32, i32};
+
+        let a = f32x4::new(-1.1, 2.2, -3.3, 6.6);
+        let r = sse2::_mm_cvttps_epi32(a);
+        assert_eq!(r, i32x4::new(-1, 2, -3, 6));
+
+        let a = f32x4::new(f32::NEG_INFINITY, f32::INFINITY, f32::MIN, f32::MAX);
+        let r = sse2::_mm_cvttps_epi32(a);
+        assert_eq!(r, i32x4::new(i32::MIN, i32::MIN, i32::MIN, i32::MIN));
```
nominolo: Actually, this behavior is quite interesting. Shouldn't this be part of the docs? Or, at least mention the name of the instruction so people can google for the full spec?

From the spec: "When a conversion is inexact, a truncated (round toward zero) value is returned. If a converted result is larger than the maximum signed doubleword integer, the floating-point invalid exception is raised, and if this exception is masked, the indefinite integer value (80000000H) is returned."

alexcrichton:
Thanks @pythoneer!

src/x86/sse2.rs
alexcrichton: Sounds like a great idea to me! @nominolo would you be willing to file a follow-up PR?

nominolo: Yep, will do.

PR #90, opened 2017-10-01
Avx

gwenn:
I had trouble implementing `_mm256_dp_ps`, `_mm_cmp_pd`, `_mm_cmp_sd` (I will try again).
I submit this PR because I don't want to prevent you from implementing other avx intrinsics but I will probably continue to commit in this branch until it is merged.

PR #91, opened 2017-10-03
AVX: vandnpd, vandnps, vblendvpd and vblendvps

p32blo:
`_mm256_andnot_pd` also generates a single-precision instruction, so I added the intrinsic to #71

### DIFF
--- a/src/x86/avx.rs
+++ b/src/x86/avx.rs
@@ -45,6 +45,30 @@ pub unsafe fn _mm256_and_ps(a: f32x8, b: f32x8) -> f32x8 {
     mem::transmute(a & b)
 }
 
+/// Compute the bitwise NOT of a packed double-precision (64-bit) floating-point elements in `a` 
+/// and then AND with `b`.
+#[inline(always)]
+#[target_feature = "+avx"]
+// Should be 'vandnpd' instuction.
+// See https://github.com/rust-lang-nursery/stdsimd/issues/71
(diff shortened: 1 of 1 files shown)


p32blo:
So, apparently this intrinsics are already implemented at #90, sorry for that.

I will take this opportunity to ask about masks types. [Here](https://github.com/rust-lang-nursery/stdsimd/blob/aa77a7fbd2c8b1a2c60fc5fb413dbd2e41501865/src/x86/sse41.rs#L13) and [here](https://github.com/rust-lang-nursery/stdsimd/blob/3a8bf80ba313214274324e5231a2326067f0a5ff/src/x86/avx2.rs#L195) the masks use the `__mm*` types and I was assuming that for example this:

    fn blendvpd256(a: f64x4, b: f64x4, mask: f64x4) -> f64x4;

should not be written as this:

    fn blendvpd256(a: f64x4, b: f64x4, mask: __m256d) -> f64x4;
    
Is there a reason for the `__mm*` types in mask?

And again, sorry for not checking the previous PR

nominolo:
Regarding mask types, that's also mentioned in https://github.com/rust-lang-nursery/stdsimd/issues/6

alexcrichton:
Thanks for the PR @p32blo! As you've noticed though I think these were added in #90 as well so I'm going to close this in favor of that.

Let's continue discussion of the mask types on #6 I think?

PR #93, opened 2017-10-04
Improve _mm256_round_* docs

AdamNiederer:
Fix a grammatical error, use a list instead of using a code block or nothing,
and add the LLVM immediate reference.


alexcrichton:
Thanks!

PR #97, opened 2017-10-06
WIP - [x86] Implement avx2 broadcast intrinsics

dlrobertson:
Implement
 - `_mm_broadcastb_epi8`
 - `_mm256_broadcastb_epi8`
 - `_mm_broadcastd_epi32`
 - `_mm256_broadcastd_epi32`
 - `_mm_bradcastq_epi64`
 - `_mm256_broadcastq_epi64`
 - `_mm_broadcastsd_pd`
 - `_mm256_broadcastsd_pd`
 - `_mm256_broadcastsi128_si256`
 - `_mm_broadcastss_ps`
 - `_mm256_broadcastss_ps`
 - `_mm_broadcastw_epi16`
 - `_mm256_broadcast2_epi16`

### DIFF
--- a/src/x86/avx2.rs
+++ b/src/x86/avx2.rs
@@ -1,4 +1,5 @@
-use simd_llvm::{simd_shuffle8, simd_shuffle32};
+use simd_llvm::{simd_shuffle2, simd_shuffle4, simd_shuffle8};
+use simd_llvm::{simd_shuffle16, simd_shuffle32};
 use v256::*;
 use v128::*;
 use x86::__m256i;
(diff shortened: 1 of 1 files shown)


src/macros.rs
```diff
@@ -138,6 +138,29 @@ macro_rules! define_from {
     }
 }
 
+macro_rules! define_from_smaller {
```
dlrobertson: I needed to be able to convert a 128-bit value to its 256-bit equivalent (E.g. `i64x2 -> i64x4`) in the `_mm256_broadcast*` functions. I created this macro to implement `From<T>` for this.

src/macros.rs
```diff
@@ -138,6 +138,29 @@ macro_rules! define_from {
...
+                fn from(f: $from) -> $to {
+                    use std::mem::size_of;
+                    use std::ptr;
+
+                    let mut a = $to::splat(0 as $ty);
+                    unsafe {
+                        ptr::copy_nonoverlapping(
+                            &f as *const $from as *const u8,
+                            &mut a as *mut $to as *mut u8,
+                            size_of::<$from>());
```
dlrobertson: `size_of::<$from>()` should always be less than `size_of<$to>()`, but I still feel somewhat uneasy about this being unchecked.

src/x86/avx2.rs
```diff
@@ -196,20 +197,126 @@ pub unsafe fn _mm256_blendv_epi8(a:i8x32,b:i8x32,mask:__m256i) -> i8x32 {
...
+/// Broadcast the low double-precision (64-bit) floating-point element
+/// from `a` to all elements of the 256-bit returned value.
+#[inline(always)]
+#[target_feature = "+avx2"]
+#[cfg_attr(test, assert_instr(vbroadcastsd))]
+pub unsafe fn _mm256_broadcastsd_pd(a: f64x2) -> f64x4 {
+    simd_shuffle4(f64x4::from(a), f64x4::splat(0f64), [0u32; 4])
+}
+
 // TODO _mm_broadcastsi128_si256
```
dlrobertson: Not sure what to do with this. A quick `rg _mm_broadcastsi128_si256` in the clang source didn't yield any results.

dlrobertson:
How important is it that the instruction be compiled to the exact instruction? I've verified that `_mm256_broadcastq_epi64` emits the following LLVM IR using clang.

```
%1 = shufflevector <2 x i64> %0, <2 x i64> undef, <4 x i32> zeroinitializer
```

And the current implementation of `_mm256_broadcastq_epi64` here emits the following LLVM IR

```
%1 = shufflevector <2 x i64> %0, <2 x i64> undef, <4 x i32> zeroinitializer
```

Of course the variable numbers will vary based on the surrounding code, but they consistently compile to the same intermediate representation. Maybe the intrinsics that use llvms vector API should have something like a separate  `assert_llvm` test that ensures the correct LLVM IR is emitted?

alexcrichton:
It does indeed seem that LLVM is a little loose sometimes of what's actually being emitted. I think it's fine to just add assertions for what LLVM is currently emitting to ensure we don't regress, and probably also just make sure that it looks reasonable in terms of functionality

src/macros.rs
dlrobertson: Took a second look at this and realized I was misusing `simd_shuffle*`. No need for this now.

src/macros.rs
dlrobertson: Same

src/x86/avx2.rs
dlrobertson: After some discussion on gitter, following clang and not including `_mm_broadcastsi128_si256`

alexcrichton:
@dlrobertson looks like this may need a rebase now and tests are still failing? Anything I can help out with?

dlrobertson:
@alexcrichton thanks for the heads up! Not sure why the tests were failing. After a rebase and rustup update the tests are passing locally

alexcrichton:
Awesome, and thanks!

PR #98, opened 2017-10-06
sse4.1 instructions

p32blo:
I have some doubts over several aspects of the PR so I will add inline comments.


src/x86/sse41.rs
```diff
@@ -15,6 +18,174 @@ pub unsafe fn _mm_blendv_epi8(
     pblendvb(a, b, mask)
 }
 
+#[inline(always)]
+#[target_feature = "+sse4.1"]
+#[cfg_attr(test, assert_instr(pblendw, imm8=0xF0))]
+pub unsafe fn _mm_blend_epi16(a: i16x8, b: i16x8, imm8: u8) -> i16x8 {
```
p32blo: Should the `imm8` be `u8` or `i32`? The public interface in C is `const int`, but I followed [`_mm_dp_ps`](https://github.com/p32blo/stdsimd/blob/08574d8f02395f417cab2a179cf8fae326fa1126/src/x86/sse41.rs#L216) for consistency and used `u8`.

src/x86/sse41.rs
```diff
@@ -15,6 +18,174 @@ pub unsafe fn _mm_blendv_epi8(
...
+
+/// Blend packed double-precision (64-bit) floating-point elements from `a` and `b` using control mask `imm2`
+#[inline(always)]
+#[target_feature = "+sse4.1"]
+#[cfg_attr(test, assert_instr(blendpd, imm2=0b10))]
+pub unsafe fn _mm_blend_pd(a: f64x2, b: f64x2, imm2: u8) -> f64x2 {
+    macro_rules! call {
+        ($imm2:expr) => { blendpd(a, b, $imm2) }
+    }
+    constify_imm2!(imm2, call)
```
p32blo: I don't think this is the correct way, since I'm messing with macros, but it was the most obvious way to make it work for now.

src/x86/sse41.rs
```diff
@@ -15,6 +18,174 @@ pub unsafe fn _mm_blendv_epi8(
...
+        ($imm4:expr) => { blendps(a, b, $imm4) }
+    }
+    constify_imm4!(imm4, call)
+}
+
+/// Extract a single-precision (32-bit) floating-point element from `a`, selected with `imm8`
+#[inline(always)]
+#[target_feature = "+sse4.1"]
+#[cfg_attr(test, assert_instr(extractps, imm8=0))]
+pub unsafe fn _mm_extract_ps(a: f32x4, imm8: u8) -> i32 {
```
p32blo: This is again a question of how much should C be followed.

This intrinsic returns somewhat nonsensically an `i32`:
```
int _mm_extract_ps (__m128 a, const int imm8)
```
Should we just return an `f32` or continue to use the transmute here.

src/x86/sse41.rs
```diff
@@ -15,6 +18,174 @@ pub unsafe fn _mm_blendv_epi8(
...
+    mem::transmute(a.extract((imm8 & 0b11) as u32))
+}
+
+/// Extract an 8-bit integer from `a` selected with `imm8`
+#[inline(always)]
+#[target_feature = "+sse4.1"]
+#[cfg_attr(test, assert_instr(pextrb, imm8=0))]
+pub unsafe fn _mm_extract_epi8(a: i8x16, imm8: u8) -> i32 {
```
p32blo: Some as the previous one, but should it be `i8` instead of `i32`?

src/x86/sse41.rs
alexcrichton: I'd imagine that so long as more than 8 bits aren't read then passing `u8` here seems fine to me, but @BurntSushi do you think we should super strictly follow C?

src/x86/sse41.rs
alexcrichton: To me it seems ok to return `f32` here

alexcrichton:
Looks like CI may still be failing?

p32blo:
Some context:
- Travis is failing with some avx tests. They work on my machine so I don't know what to do to fix this. Also can't get why AppVeyor is failing.
- Couldn't change `_mm_extract_ps` to return `f32` because it does not generate the correct instruction with the change.
- Changed `_mm_extract_epi64` and `_mm_insert_epi64` to only work on `x86_64` since it failed to generate the instruction on 32-bit builds. See [here](http://www.felixcloutier.com/x86/PEXTRB:PEXTRD:PEXTRQ.html).
- used `i8x16` instead of `__m128i` on `_mm_blendv_epi8` for consistency.

nominolo:
Win64 uses a different calling convention. Unless you define a different calling convention, it will pass vector arguments in memory instead of via the XMM/YMM registers. This causes different code to be generated. See, e.g., https://msdn.microsoft.com/en-us/library/9z1stfyw.aspx

p32blo:
Thanks @nominolo !

Is there a way I can fix this?

nominolo:
@p32blo You probably want to just change the expected code. I.e., something like: https://github.com/rust-lang-nursery/stdsimd/blob/2dbe8d0b39a1b07e9cf4846d593908bed588ce9f/src/x86/sse.rs#L751

src/x86/sse41.rs
```diff
@@ -1,20 +1,191 @@
...
+#[cfg_attr(all(test, not(windows)), assert_instr(pextrd, imm8=1))]
+pub unsafe fn _mm_extract_epi32(a: i32x4, imm8: u8) -> i32 {
+    a.extract((imm8 & 0b11) as u32)
+}
+
+/// Extract an 64-bit integer from `a` selected with `imm8`
+#[cfg(target_arch = "x86_64")]
+#[inline(always)]
+#[target_feature = "+sse4.1"]
+#[cfg_attr(all(test, windows), assert_instr(mov, imm8=1))]
```
p32blo: Should windows be tested at all, since it does not generate the correct output? Or is testing for `mov` ok?

nominolo: I think testing for plain `mov` doesn't make sense, because it'll always be satisfied by `mov rbp,rsp` which is part of every function prelude.

src/x86/sse41.rs
nominolo: Actually, it really should be an `i32`. See answers at https://stackoverflow.com/questions/5526658/intel-sse-why-does-mm-extract-ps-return-int-instead-of-float

alexcrichton:
Ok looks great, thanks! We can always work on adding more tests later if necessary.

PR #99, opened 2017-10-06
Implement SSE _mm_load* instructions

nominolo:
This implements SSE instructions:

- `_mm_loadh_pi`
- `_mm_loadl_pi`
- `_mm_load_ss`
- `_mm_load1_ps`
- `_mm_load_ps1`
- `_mm_load_ps`
- `_mm_loadu_ps`
- `_mm_loadr_ps`

I also added some experimental doctests, but I can't get them to run. (That is, I can't get them to fail by adding `assert!(false)`. They are being compiled, though)

### DIFF
--- a/src/x86/sse.rs
+++ b/src/x86/sse.rs
@@ -316,6 +316,51 @@ pub unsafe fn _mm_loadh_pi(a: f32x4, p: *const f32) -> f32x4 {
     simd_shuffle4(a, bb, [0, 1, 4, 5])
 }
 
+/// Load two floats from `p` into the lower half of a `f32x4`. The upper half
+/// is copied from the upper half of `a`.
+///
+/// This corresponds to the `MOVLPS` / `MOVLDP` / `VMOVLDP` instructions.
+///
+/// ```rust
+/// # #![feature(cfg_target_feature)]
+/// # #![feature(target_feature)]
+/// #
+/// # #[macro_use] extern crate stdsimd;
+/// #
+/// # // The real main function
+/// # fn main() {
+/// #     if cfg_feature_enabled!("sse") {
+/// #         #[target_feature = "+sse"]
+/// #         fn worker() {
+/// #
+/// #   use stdsimd::simd::f32x4;
+/// #   use stdsimd::vendor::_mm_loadl_pi;
+/// #
+/// let a = f32x4::new(1.0, 2.0, 3.0, 4.0);
+/// let data: [f32; 4] = [5.0, 6.0, 7.0, 8.0];
+///
+/// let r = unsafe { _mm_loadl_pi(a, data[..].as_ptr()) };
+///
+/// assert_eq!(r, f32x4::new(5.0, 6.0, 3.0, 4.0));
+/// #
+/// #         }
+/// #         worker();
+/// #     }
+/// # }
+/// ```
+#[inline(always)]
+#[target_feature = "+sse"]
+// TODO: generates MOVLPD if the CPU supports SSE2.
+// #[cfg_attr(test, assert_instr(movlps))]
+#[cfg_attr(test, assert_instr(movlpd))]
+// TODO: Like _mm_loadh_pi, this also isn't limited to floats.
+pub unsafe fn _mm_loadl_pi(a: f32x4, p: *const f32) -> f32x4 {
+    let q = p as *const f32x2;
+    let b: f32x2 = *q;
+    let bb = simd_shuffle4(b, b, [0, 1, 0, 1]);
+    simd_shuffle4(a, bb, [4, 5, 2, 3])
+}
+
 /// Perform a serializing operation on all store-to-memory instructions that
 /// were issued prior to this instruction.
 ///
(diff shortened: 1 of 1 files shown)


src/x86/sse.rs
```diff
@@ -268,6 +271,190 @@ pub unsafe fn _mm_movemask_ps(a: f32x4) -> i32 {
...
+/// assert_eq!(r, f32x4::new(1.0, 2.0, 5.0, 6.0));
+/// #
+/// #         }
+/// #         worker();
+/// #     }
+/// # }
+/// ```
+#[inline(always)]
+#[target_feature = "+sse"]
+// TODO: generates MOVHPD if the CPU supports SSE2.
```
alexcrichton: So this actually got me thinking! I remembered that x86_64 is defined with SSE2 support by default, which may explain this function on that target. It turns out as well, however, that i686 *also* has SSE2 turned on by default!

I think to assert this instruction we'd have to use the i586 target (which has sse/sse2 disabled), but in the meantime this is probably ok.

src/x86/sse.rs
```diff
@@ -268,6 +271,190 @@ pub unsafe fn _mm_movemask_ps(a: f32x4) -> i32 {
...
+/// may be faster.
+///
+/// This corresponds to instructions `VMOVUPS` / `MOVUPS`.
+#[inline(always)]
+#[target_feature = "+sse"]
+#[cfg_attr(test, assert_instr(movups))]
+pub unsafe fn _mm_loadu_ps(p: *const f32) -> f32x4 {
+    // TODO: This also seems to generate the same code. Don't know which one
+    // will behave better when inlined into other code.
+    // f32x4::new(*p, *p.offset(1), *p.offset(2), *p.offset(3))
```
alexcrichton: I think that if there's no restrictions on alignment here that the implementation below is the one we'll have to go with, AFAIK a `*p` will require correct alignment

alexcrichton:
Awesome, thanks!

src/x86/sse.rs
nominolo: Ah, OK, then I'll remove this TODO.

alexcrichton:
Looks like the tests may be failing on i686?

nominolo:
Looks like it picks different instructions in 32-bit mode:

```
disassembly for stdsimd::x86::sse::_mm_loadh_pi: 
	 0: push %ebp 
	 1: mov %esp,%ebp 
	 2: mov 0x8(%ebp),%eax 
	 3: movsd (%eax),%xmm1        // 3&4 could be combined into
	 4: unpcklpd %xmm1,%xmm0  // movhps (%eax),%xmm0
	 5: pop %ebp 
	 6: ret 
```

```
disassembly for stdsimd::x86::sse::_mm_loadl_pi: 
	 0: push %ebp 
	 1: mov %esp,%ebp 
	 2: mov 0x8(%ebp),%eax 
	 3: movsd (%eax),%xmm1   // 3&4 could be written as
	 4: movsd %xmm1,%xmm0 // movlps (%eax),%xmm0
	 5: pop %ebp 
	 6: ret
```
Not sure what to do about this. The implementation is based on the one from `xmmintrin.h`

We have some other problems with `assert_instr` being too strict / fragile. We can make it less strict, or we could have a test based on the result matching the C intrinsic. But that would require invoking a C compiler somewhere. Perhaps, we can have a separate C file `reference_sse.c` and then have a macro `matches_c_body(<function_name>)`.

alexcrichton:
Hm maybe we could just relax it here for now? I'm not sure how to deal with this as I'm not sure why there's different codegen...

src/x86/sse.rs
```diff
@@ -269,6 +272,194 @@ pub unsafe fn _mm_movemask_ps(a: f32x4) -> i32 {
...
+/// may be faster.
+///
+/// This corresponds to instructions `VMOVUPS` / `MOVUPS`.
+#[inline(always)]
+#[target_feature = "+sse"]
+#[cfg_attr(test, assert_instr(movups))]
+pub unsafe fn _mm_loadu_ps(p: *const f32) -> f32x4 {
+    // Note: Using `*p` would require `f32` alignment, but `movups` has no
+    // alignment restrictions.
+    let mut dst = mem::uninitialized();
```
alexcrichton: I think this may be causing the faults seen on Travis, I had to use `f32x4::splat(uninitialized())` to fix these issues.

src/x86/sse.rs
```diff
@@ -831,6 +1022,88 @@ mod tests {
...
+
+        let r = sse::_mm_load_ps(p);
+        assert_eq!(r, f32x4::new(1.0, 2.0, 3.0, 4.0) + f32x4::splat(fixup));
+    }
+
+    #[simd_test = "sse"]
+    unsafe fn _mm_loadu_ps() {
+        let vals = &[1.0f32, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0];
+        let p = vals.as_ptr().offset(3);
+        let r = sse::_mm_loadu_ps(black_box(p));
```
alexcrichton: I think another case of a possible crash here?

nominolo: No, the failing tests are due to different instruction selection on i586 vs i686 for `loadh_ps` and `loadl_ps`. I don't see a convenient way to distinguish them, though. They're both `target_arch = "x86"` and I see no other way to distinguish them based on standard attributes: https://stackoverflow.com/questions/41742046/is-there-a-list-of-all-cfg-features

alexcrichton: Oops misread the logs!

You can conditionalize for i586/i686 with `#[cfg(target_feature = "sse2")]`, that's enabled on i686 and disabled on i586

alexcrichton:
Thanks again!

PR #100, opened 2017-10-06
Implement SSE _mm_set* intrinsics

nominolo:
This implements SSE intrinsics:

- `_mm_set_ss`
- `_mm_set1_ps`
- `_mm_set_ps1`
- `_mm_set_ps`
- `_mm_setr_ps`
- `_mm_setzero_ps`

### DIFF
--- a/src/x86/sse.rs
+++ b/src/x86/sse.rs
@@ -174,6 +174,22 @@ pub unsafe fn _mm_set_ss(a: f32) -> f32x4 {
     f32x4::new(a, 0.0, 0.0, 0.0)
 }
 
+/// Construct a `f32x4` with all element set to `a`.
+#[inline(always)]
+#[target_feature = "+sse"]
+#[cfg_attr(test, assert_instr(shufps))]
+pub unsafe fn _mm_set1_ps(a: f32) -> f32x4 {
+    f32x4::new(a, a, a, a)
+}
+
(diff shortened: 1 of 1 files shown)


alexcrichton:
Awesome thanks! Looks like the tests may be failing for i686 though?

alexcrichton:
Hm it looks like this may be segfaulting on the new i586 target? I'm not sure where though... If you run locally with `RUST_TEST_THREADS=1` (or push that to CI) it should say what test

src/x86/sse.rs
```diff
@@ -789,6 +863,40 @@ mod tests {
...
+    #[simd_test = "sse"]
+    unsafe fn _mm_setr_ps() {
+        let r = sse::_mm_setr_ps(
+            black_box(1.0), black_box(2.0), black_box(3.0), black_box(4.0));
+        assert_eq!(r, f32x4::new(1.0, 2.0, 3.0, 4.0));
+    }
+
+    #[simd_test = "sse"]
+    unsafe fn _mm_setzero_ps() {
+        let r = black_box(sse::_mm_setzero_ps());
```
alexcrichton: Oh actually this may be where the segfault is, unfortunately `black_box` isn't using the right ABI (wrong `target_feature`) so this may work instead:

```rust
let r = *black_box(&sse::_mm_setzero_ps());
```

alexcrichton:
Nice!

PR #103, opened 2017-10-07
Implement bitwise SSE ops & _mm_cmp*_ss 

nominolo:
Implement SSE bitwise AND, OR, AND-NOT, and XOR.
Implement all variants of `_mm_cmp*_ss`.

### DIFF
--- a/src/x86/sse.rs
+++ b/src/x86/sse.rs
@@ -232,6 +232,17 @@ pub unsafe fn _mm_cmplt_ss(a: f32x4, b: f32x4) -> f32x4 {
     cmpss(a, b, 1)
 }
 
+/// Compare the lowest `f32` of both inputs for less than or equal. The lowest
+/// 32 bits of the result will be `0xffffffff` if `a.extract(0)` is less than or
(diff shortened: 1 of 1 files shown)


src/x86/sse.rs
```diff
@@ -165,6 +166,188 @@ pub unsafe fn _mm_max_ps(a: f32x4, b: f32x4) -> f32x4 {
     maxps(a, b)
 }
 
+/// Bitwise AND of packed single-precision (32-bit) floating-point elements.
+#[inline(always)]
+#[target_feature = "+sse"]
+// i586 only seems to generate plain `and` instructions, so ignore it.
```
alexcrichton: This seems a little worrisome, I wonder if we don't have the right codegen for that then?

alexcrichton: Er well [apparently this is normal](https://github.com/llvm-mirror/llvm/blob/7e5d086d1e5ede0d73f51fca55dd1fdd6adfc796/test/CodeGen/X86/sse-intrinsics-fast-isel.ll#L38-L70)!

alexcrichton:
Thanks! Mind rebasing now as well?

alexcrichton:
👍

PR #104, opened 2017-10-08
Add _mm256_shuffle_epi8

AdamNiederer:
Need this for base100. I don't have an AVX2 capable CPU, so I might need to thrash travis a bit if this doesn't work off the bat.


AdamNiederer:
I've discovered that this may need to use the shuffle intrinsics. I'm going to close this for now so I don't discourage anybody from implementing it before I can grok them.

PR #105, opened 2017-10-08
Avx

gwenn:


gwenn:
I don't know what to do with the `i586` build failure !

src/x86/avx.rs
```diff
@@ -484,6 +484,152 @@ pub unsafe fn _mm256_xor_ps(a: f32x8, b: f32x8) -> f32x8 {
...
+    }
+    constify_imm6!(imm8, call)
+}
+
+/// Compare packed single-precision (32-bit) floating-point
+/// elements in `a` and `b` based on the comparison operand
+/// specified by `imm8`.
+#[inline(always)]
+#[target_feature = "+avx,+sse"]
+#[cfg_attr(test, assert_instr(vcmpeqps, imm8 = 0))] // TODO Validate vcmpps
```
alexcrichton: Does this mean that we just need more `assert_instr` attributes? Or should the intrustruction here be different?

src/x86/sse.rs
```diff
@@ -868,6 +868,13 @@ pub unsafe fn _mm_prefetch(p: *const c_void, strategy: i8) {
     pref!(strategy)
 }
 
+/// Return vector of type __m128 with undefined elements.
+#[inline(always)]
+#[target_feature = "+sse"]
+pub unsafe fn _mm_undefined_ps() -> f32x4 {
+    mem::uninitialized()
```
alexcrichton: I think this may be causing a fault on i586?

src/x86/avx.rs
gwenn: We need more `assert_instr`: the instruction depends on the comparison operand.

src/x86/sse.rs
gwenn: Thanks: when I saw #101, I should have fix `_mm_undefined_ps` but I did not.

alexcrichton:
Nice!

PR #106, opened 2017-10-10
Add _MM_TRANSPOSE4_PS pseudo-macro.

MaloJaffre:
This adds a strange macro, which I've replaced with a function, because it
seems there are not many better alternatives.

Also adds a test, and `#[allow(non_snake_case)]` to `#[simd_test]`.


alexcrichton:
Thanks!

PR #117, opened 2017-10-14
[example] nbody

gnzlbg:
Ports @huonw 's nbody example from the old simd library.

### DIFF
--- a/examples/nbody.rs
+++ b/examples/nbody.rs
@@ -1,0 +1,206 @@
+#![feature(cfg_target_feature)]
+#![feature(target_feature)]
+
+extern crate stdsimd;
+
+use self::stdsimd::simd;
+use self::stdsimd::vendor;
+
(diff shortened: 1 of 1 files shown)


alexcrichton:
A great idea! Perhaps this could use `cfg_feature_enabled!` as well to provide a runtime diagnostic when it would otherwise not work?

gnzlbg:
Yes we should definetely improve this:

- clean up the `cfg(target_feature)` macros
- implement the required ARM intrinsics (I am doing this right now, are you on IRC?)
- use run-time feature detection

alexcrichton:
I am indeed on IRC!

gnzlbg:
@alexcrichton why isn't travis running anymore?

gnzlbg:
does this mean I cannot squash? I can't change the past anymore!

alexcrichton:
👍

PR #119, opened 2017-10-14
[arm] v6/v7/v8 run-time tests

gnzlbg:


gnzlbg:
@alexcrichton I didn't manage to make the `#[simd_test = "v8"]` macros work, but wanted to comit this before I add one or two NEON intrinsics so that we at least have something to improve on.

alexcrichton:
Looks good! What was the error with the macro?

gnzlbg:
I think it did not recognize `+v7` and `+v8` as features.

alexcrichton:
Alas :(

PR #126, opened 2017-10-16
Avx

gwenn:


gwenn:
It seems that all `si128` or `si256` intrinsics that I've implemented will have to be fixed.
Because they do not use the `__m128i` or `__m256i` type.

alexcrichton:
Thanks!

PR #131, opened 2017-10-19
sse2

pythoneer:
added missing doc _mm_cvtps_pd
added missing doc & test _mm_load_pd
added missing doc & test _mm_store_pd
added _mm_store1_pd
added _mm_store_pd1
added _mm_storer_pd
added _mm_load_pd1
added _mm_loadr_pd
added _mm_loadu_pd

### DIFF
--- a/src/x86/sse2.rs
+++ b/src/x86/sse2.rs
@@ -1726,6 +1726,9 @@ pub unsafe fn _mm_cvtpd_ps(a: f64x2) -> f32x4 {
     cvtpd2ps(a)
 }
 
+
+/// Convert packed single-precision (32-bit) floating-point elements in `a` to packed
+/// double-precision (64-bit) floating-point elements.
 #[inline(always)]
 #[target_feature = "+sse2"]
 #[cfg_attr(test, assert_instr(cvtps2pd))]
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! Looks like tests may be segfaulting though?

pythoneer:
interesting that i don't experience those on my system (something related to the specific way its compiled on my system + my exact processor i guess) i try replicating this on another win and macOS machine on the weekend. 

i am just using `cargo test --release` (on nightly) should i also define the `target` like we do in the [run.sh](https://github.com/rust-lang-nursery/stdsimd/blob/59de334cfb0f53c3cb79bf59203b9828f7414deb/ci/run.sh#L21)? to replicate this better for the future?

src/x86/sse2.rs
```diff
@@ -3588,6 +3661,102 @@ mod tests {
...
 
     #[simd_test = "sse2"]
+    unsafe fn _mm_load_pd() {
+        let vals = &[1.0f64, 2.0, 3.0, 4.0];
+        let mut d = vals.as_ptr();
+
+        // Align d to 16-byte boundary
+        let mut offset = 0;
+        if (d as usize) & 0xf != 0 {
+            offset = (16 - (d as usize) & 0xf) >> 2;
```
alexcrichton: This is overflowing on travis, I wonder if the precedence here is making `(a - b) & 0xf` ?

alexcrichton:
Thanks!

Issue #139, opened 2017-10-22
Rename ARMv8 to AArch64


gnzlbg:
LLVM calls ARMv8 AArch64.  Some ARM documents call it A64 or v8. For consistency with LLVM we should probably rename it to `aarch64`. Thoughts? @parched @japaric

petrochenkov:
>LLVM calls ARMv8 AArch64. Some ARM documents call it A64

\<pedantry>

All the listed names mean different things.
- ARMv8 - ARM *architecture* version 8
    - ARMv8-A - *Architectural profile* inside ARMv8 
        - AArch64 - *Execution state* inside ARMv8-A (execution on one core can switch between different execution states, e.g. between AArch64 OS and AArch32 application).
            - A64 - *Instruction set*
        - AArch32 - *Execution state* inside ARMv8-A.
            - A32 - *Instruction set* (execution in one execution state can switch between different instruction sets using jumps).
            - T32 - *Instruction set*
    - ARMv8-M, ARMv8-R - ...

\</pedantry>

gnzlbg:
Currently, we have the arm intrinsics in the `v6`, `v7`, and `v8` modules +- `neon` (`v7_neon`, `v8_neon`), would maybe splitting `v8` into `A32` and `A64` make sense?

parched:
Why not just 2 modules, scalar and vector/simd/neon?

gnzlbg:
I should rephrase the question. First, the modules are private. Second, we use them only for conditional compilation, that is, for deciding which intrinsics to provide depending on the target triple.

parched:
Ok understood. It's probably not possible to use the target triple for that as they don't have to follow convention. Custom targets can have any name and even some of the builtin targets don't follow a convention. e.g., all the `arm-*` targets have no version and `armv7-linux-androideabi` generates Thumb not Arm code. I guess you could special case for every builtin target but custom targets would still be an issue.

gnzlbg:
AFAIK LLVM cannot emit code that contains `aarch64` instructions but is able to run on `v6`/`v7`.

This means that run-time feature detection cannot be used to generate an, e.g., `armv7` binary that uses `aarch64` instructions if they are available at run-time. 

I think that what I will do is move the `v8` and `v8_neon` modules out of the `arm` architecture into the `aarch64` architecture module to make this distinction more clear.

parched:
[...]

PR #140, opened 2017-10-22
[arm] vadd, vaddd, vaddq, vaddl

gnzlbg:


src/v64.rs
```diff
@@ -54,4 +54,12 @@ define_casts!(
...
     (i16x4, u16x4, as_u16x4),
     (u8x8, i8x8, as_i8x8),
-    (i8x8, u8x8, as_u8x8));
+    (i8x8, u8x8, as_u8x8),
+    (i8x8, i16x8, as_i16x8),
+    (i16x4, i32x4, as_i32x4),
+    (i32x2, i64x2, as_i64x2),
+    (u8x8, u16x8, as_u16x8),
+    (u16x4, u32x4, as_u32x4),
+    (u32x2, u64x2, as_u64x2)
```
alexcrichton: Oh I thought these were only for sem-width casts? What are the semantics of the different-width casts? Does each lane just get promoted?

gnzlbg: > Does each lane just get promoted?

Yes, that's what `simd_cast` does and what I intended. @BurntSushi how did you envision this? The number of `as_...` functions is going to explode if I keep doing this for the arm intrinsics. I would be fine with it, but maybe we should consider providing `simd_cast` as part of the library API.

gnzlbg:
@BurntSushi @alexcrichton I've filled #151 to discuss this. Right now this blocks the `aarch64` refactor of #139 .

BurntSushi:
@alexcrichton and I talked about this in person and we're good with this.

PR #147, opened 2017-10-24
avx2: add _mm256_unpack{hi,lo}_epi{8,16,32,64}

hdevalence:
Add AVX2 unpack intrinsics using LLVM shuffles.


alexcrichton:
Awesome, thanks! While you're at it, mind adding a few doc blocks for what these are doing, and if you're extra enterprising a test or two as well?

hdevalence:
Edited the commit to add doctests. I modified the test setup incantation from one of the SSE examples, hopefully it's done correctly.

alexcrichton:
Awesome, thanks!

PR #150, opened 2017-10-24
x86: fixed testing equality of floating point numbers

Mrowqa:
Hi,

I have run the tests and five of them failed. I checked the Intel Intrinsics Guide and it said that these specific instructions returns approximated values while the tests don't take the error into account. I have Ryzen 1700X. I guess all of test runs until now could be done on Intel CPUs where it could possibly always return those hardcoded specific values.

I also haven't known how to cleanly make the compare, so I have added an extra macro. It is basically copy-paste of [this](https://github.com/ashleygwilliams/assert_approx_eq/blob/master/src/lib.rs) since I didn't want to add new dependencies.

### DIFF
--- a/src/x86/avx.rs
+++ b/src/x86/avx.rs
@@ -3114,7 +3114,10 @@ mod tests {
         let r = avx::_mm256_rcp_ps(a);
         let e = f32x8::new(0.99975586, 0.49987793, 0.33325195, 0.24993896,
                            0.19995117, 0.16662598, 0.14282227, 0.12496948);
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks! Looks like CI is failing though?

Mrowqa:
It was still building when I went sleep. I haven't marked the new auxiliary macro as used only in tests, so normal build emitted unused macro warning. Now it should be fixed.

alexcrichton:
Awesome, thanks!

PR #154, opened 2017-10-26
[runtime-detection-x86] detect avx and avx2 only if osxsave is true

### LINKED ISSUE #153: AVX detection needs to check if OS supports save/restore
See https://hg.mozilla.org/mozilla-central/file/64bab5cbb9b6/mozglue/build/SSE.cpp#l190 for an example

gnzlbg:
Closes #153.

### DIFF
--- a/src/x86/runtime.rs
+++ b/src/x86/runtime.rs
@@ -162,78 +162,100 @@ fn inv_test_bit(v: usize, idx: u32) -> bool {
 /// [intel64_ref]: http://www.intel.de/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-instruction-set-reference-manual-325383.pdf
 /// [amd64_ref]: http://support.amd.com/TechDocs/24594.pdf
 fn detect_features() -> usize {
-    let ebx;
-    let ecx;
-    let edx;
+    let extended_features_ebx;
+    let proc_info_ecx;
+    let proc_info_edx;
 
     unsafe {
         /// To obtain all feature flags we need two CPUID queries:
 
         /// 1. EAX=1, ECX=0: Queries "Processor Info and Feature Bits"
         /// This gives us most of the CPU features in ECX and EDX (see
-        /// below),
+        /// below).
         asm!("cpuid"
-             : "={ecx}"(ecx), "={edx}"(edx)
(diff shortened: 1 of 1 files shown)


src/x86/runtime.rs
```diff
@@ -111,48 +111,56 @@ fn inv_test_bit(v: usize, idx: u32) -> bool {
...
+
+    // ECX[26] detects XSAVE and ECX[27] detects OSXSAVE, that is, whether the
+    // OS is AVX enabled and supports saving the state of the AVX/AVX2 vector
+    // registers on context-switches, see:
+    //
+    // - https://software.intel.com/en-us/blogs/2011/04/14/is-avx-enabled
+    // - https://hg.mozilla.org/mozilla-central/file/64bab5cbb9b6/mozglue/build/SSE.cpp#l190
+    //
+    if inv_test_bit(proc_info_ecx, 26) && inv_test_bit(proc_info_ecx, 27) {
+        if inv_test_bit(proc_info_ecx, 28) { value = set_bit(value, __Feature::avx as u32); }
```
jrmuizel: You need to actually use the xgetbv instruction here to confirm that the OS is going to do the right thing.

gnzlbg:
@jrmuizel indeed, I've updated the PR.

@BurntSushi we might want to expose `xgetbv` on x86 "somewhere".

petrochenkov:
>we might want to expose xgetbv on x86 "somewhere"

BTW, since this crate is supposed to cover only SIMD intrinsics (?), are there any plans to cover everything else?
(`xgetbv` in particular is [available](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=xgetbv&expand=5686) as an intrinsic in `immintrin.h`.)

gnzlbg:
@petrochenkov this crate implements `core::vendor` and aims to cover all vendor intrinsics. It is called `stdsimd` because its main current focus are SIMD intrinsics, but it already covers many non-SIMD-related intrinsics.

jrmuizel:
Looks good to me.

alexcrichton:
Looks great, thanks! I think CI is failing though?

gnzlbg:
Yes, I screwed up the assembly syntax, I'll fix it.

gnzlbg:
@alexcrichton this is ready

alexcrichton:
Awesome, thanks!

PR #158, opened 2017-10-27
x86: implemented roundings for SSE4.1

Mrowqa:
Hi,

I was basing mainly on [clang](https://clang.llvm.org/doxygen/smmintrin_8h_source.html) and, obviously, Intel Intrinsics Guide. Tests for `_mm_round_*` does not cover all possible roundings for all intrinsics. Furthermore, I haven't added docs, just left some TODOs.

### DIFF
--- a/src/x86/sse41.rs
+++ b/src/x86/sse41.rs
@@ -7,19 +7,31 @@ use stdsimd_test::assert_instr;
 use v128::*;
 
 // SSE4 rounding constans
-// TODO docs
+/// round to nearest
 pub const _MM_FROUND_TO_NEAREST_INT: i32 = 0x00;
+/// round down
 pub const _MM_FROUND_TO_NEG_INF: i32 = 0x01;
+/// round up
 pub const _MM_FROUND_TO_POS_INF: i32 = 0x02;
+/// truncate
 pub const _MM_FROUND_TO_ZERO: i32 = 0x03;
+/// use MXCSR.RC; see `vendor::_MM_SET_ROUNDING_MODE`
 pub const _MM_FROUND_CUR_DIRECTION: i32 = 0x04;
+/// do not suppress exceptions
 pub const _MM_FROUND_RAISE_EXC: i32 = 0x00;
+/// suppress exceptions
 pub const _MM_FROUND_NO_EXC: i32 = 0x08;
+/// round to nearest and do not suppress exceptions
 pub const _MM_FROUND_NINT: i32 = (_MM_FROUND_RAISE_EXC | _MM_FROUND_TO_NEAREST_INT);
+/// round down and do not suppress exceptions
 pub const _MM_FROUND_FLOOR: i32 = (_MM_FROUND_RAISE_EXC | _MM_FROUND_TO_NEG_INF);
+/// round up and do not suppress exceptions
 pub const _MM_FROUND_CEIL: i32 = (_MM_FROUND_RAISE_EXC | _MM_FROUND_TO_POS_INF);
+/// truncate and do not suppress exceptions
 pub const _MM_FROUND_TRUNC: i32 = (_MM_FROUND_RAISE_EXC | _MM_FROUND_TO_ZERO);
+/// use MXCSR.RC and do not suppress exceptions; see `vendor::_MM_SET_ROUNDING_MODE`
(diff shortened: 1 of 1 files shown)


Mrowqa:
It fails with 'too many instructions in the disassembly'. Found the check in `stdsimd-test/src/lib.rs`. Any ideas how to fix this in a clean and desired way? It looks like `constify_imm4` in `_mm_round_*` generates a lot of instructions (the macro is a giant match), but it is needed for intrinsic, otherwise LLVM fails a build.

src/x86/sse41.rs
```diff
@@ -233,6 +249,116 @@ pub unsafe fn _mm_dp_ps(a: f32x4, b: f32x4, imm8: u8) -> f32x4 {
...
+#[target_feature = "+sse4.1"]
+#[cfg_attr(test, assert_instr(roundss))]
+pub unsafe fn _mm_ceil_ss(a: f32x4, b: f32x4) -> f32x4 {
+    roundss(a, b, _MM_FROUND_CEIL)
+}
+
+// TODO docs
+#[inline(always)]
+#[target_feature = "+sse4.1"]
+#[cfg_attr(test, assert_instr(roundpd))]
```
BurntSushi: When testing intrinsics with const immediates, you need to generate code with a constant, otherwise you won't get the right codegen. So in this case I think something like `#[cfg_attr(test, assert_instr(roundpd, rounding = 1))]` should work. See an example here: https://github.com/rust-lang-nursery/stdsimd/blob/master/src/x86/sse2.rs#L316

src/x86/sse41.rs
```diff
@@ -6,6 +6,34 @@ use stdsimd_test::assert_instr;
...
+/// round down and do not suppress exceptions
+pub const _MM_FROUND_FLOOR: i32 = (_MM_FROUND_RAISE_EXC | _MM_FROUND_TO_NEG_INF);
+/// round up and do not suppress exceptions
+pub const _MM_FROUND_CEIL: i32 = (_MM_FROUND_RAISE_EXC | _MM_FROUND_TO_POS_INF);
+/// truncate and do not suppress exceptions
+pub const _MM_FROUND_TRUNC: i32 = (_MM_FROUND_RAISE_EXC | _MM_FROUND_TO_ZERO);
+/// use MXCSR.RC and do not suppress exceptions; see `vendor::_MM_SET_ROUNDING_MODE`
+pub const _MM_FROUND_RINT: i32 = (_MM_FROUND_RAISE_EXC | _MM_FROUND_CUR_DIRECTION);
+/// use MXCSR.RC and suppress exceptions; see `vendor::_MM_SET_ROUNDING_MODE`
+pub const _MM_FROUND_NEARBYINT: i32 = (_MM_FROUND_NO_EXC | _MM_FROUND_CUR_DIRECTION);
```
Mrowqa: This one and that one above it have a little bit confusing names. Not sure if I should define these some extra constants (not searchable in Intel Intrinsics Guide), but [clang does](https://clang.llvm.org/doxygen/smmintrin_8h_source.html), so I did.

src/x86/sse41.rs
```diff
@@ -233,6 +261,191 @@ pub unsafe fn _mm_dp_ps(a: f32x4, b: f32x4, imm8: u8) -> f32x4 {
...
+        ($imm4:expr) => { roundsd(a, b, $imm4) }
+    }
+    constify_imm4!(rounding, call)
+}
+
+/// Round the lower single-precision (32-bit) floating-point element in `b`
+/// using the `rounding` parameter, store the result as a single-precision
+/// floating-point element in the lower element of the intrinsic result,
+/// and copy the upper 3 packed elements from `a` to the upper elements
+/// of the instrinsic result.
```
Mrowqa: Some clever way how to call "the intrinsic result"? Now it is hard to read and understand. Intel's Guide calls it `dst`, but it also provides some pseudocode.

alexcrichton:
This looks great to me, thanks @Mrowqa!

src/x86/sse41.rs
alexcrichton: Yeah I think it's ok to match clang in this regard

src/x86/sse41.rs
alexcrichton: Eh I'm sort of fine with whatever, this plus intel's docs are likely sufficient regardless

PR #163, opened 2017-11-01
Avx

gwenn:


alexcrichton:
👍 Thanks!

gnzlbg:
@alexcrichton this PR broke the rustfmt build. We should either check that it doesn't break or... just allow the build to fail because of it. I mean, worst case we can't commit one day because rustfmt fails to build but... that is fixed quickly, and as soon as it is shipped with rustup this will never fail.

alexcrichton:
@gnzlbg oh I'm in "if it's green we merge it" mode, I'm not looking at the test results to see if the "allow fail" entries are failing.

gnzlbg:
I'm in the same mode :D

Issue #40, opened 2017-09-25
Implement all x86 vendor intrinsics


alexcrichton:
This is intended to be a tracking issue for implementing all vendor intrinsics in this repository.
This issue is also intended to be a guide for documenting the process of adding new vendor intrinsics to this crate.

If you decide to implement a set of vendor intrinsics, please check the list below to make sure somebody else isn't already working on them. If it's not checked off or has a name next to it, feel free to comment that you'd like to implement it!

At a high level, each vendor intrinsic should correspond to a single exported Rust function with an appropriate `target_feature` attribute. Here's an example for `_mm_adds_epi16`:

```rust
/// Add packed 16-bit integers in `a` and `b` using saturation.
#[inline]
#[target_feature(enable = "sse2")]
#[cfg_attr(test, assert_instr(paddsw))]
pub unsafe fn _mm_adds_epi16(a: __m128i, b: __m128i) -> __m128i {
    unsafe { paddsw(a, b) }
}
```

Let's break this down:

* The `#[inline]` is added because vendor intrinsic functions generally should always be inlined because the intent of a vendor intrinsic is to correspond to a single particular CPU instruction. A vendor intrinsic that is compiled into an actual function call could be quite disastrous for performance.
* The `#[target_feature(enable = "sse2")]` attribute intructs the compiler to generate code with the `sse2` target feature enabled, *regardless* of the target platform. That is, even if you're compiling for a platform that doesn't support `sse2`, the compiler will still generate code for `_mm_adds_epi16` *as if* `sse2` support existed. Without this attribute, the compiler might not generate the intended CPU instruction.
* The `#[cfg_attr(test, assert_instr(paddsw))]` attribute indicates that when we're testing the crate we'll assert that the `paddsw` instruction is generated inside this function, ensuring that the SIMD intrinsic truly is an intrinsic for the instruction!
* The types of the vectors given to the intrinsic should match exactly the types as provided in the vendor interface. (with things like `int64_t` translated to `i64` in Rust)
* The implementation of the vendor intrinsic is generally very simple. Remember, the goal is to compile a call to `_mm_adds_epi16` down to a single particular CPU instruction. As such, the implementation typically defers to a compiler intrinsic (in this case, `paddsw`) when one is available. More on this below as well.
* The intrinsic itself is `unsafe` due to the usage of `#[target_feature]`

Once a function has been added, you should also add at least one test for basic functionality. Here's an example for `_mm_adds_epi16`:

```rust
[...]

PR #165, opened 2017-11-02
Add SSE2 trivial aliases and conversions.

MaloJaffre:
`_mm_cvtsd_f64`, `_mm_cvtsd_si64x` and `_mm_cvttsd_si64x`.
See #40.

### DIFF
--- a/src/x86/sse2.rs
+++ b/src/x86/sse2.rs
@@ -1795,6 +1795,13 @@ pub unsafe fn _mm_cvtsd_si64(a: f64x2) -> i64 {
     cvtsd2si64(a)
 }
 
+/// Alias for [`_mm_cvtsd_si64`](fn._mm_cvtsd_si64_ss.html).
+#[cfg(target_arch = "x86_64")]
+#[inline(always)]
+#[target_feature = "+sse2"]
+#[cfg_attr(test, assert_instr(cvtsd2si))]
+pub unsafe fn _mm_cvtsd_si64x(a: f64x2) -> i64 { _mm_cvtsd_si64(a) }
+
 /// Convert the lower double-precision (64-bit) floating-point element in `b`
 /// to a single-precision (32-bit) floating-point element, store the result in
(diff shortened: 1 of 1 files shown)


src/x86/sse2.rs
```diff
@@ -1845,6 +1860,13 @@ pub unsafe fn _mm_cvttsd_si64(a: f64x2) -> i64 {
     cvttsd2si64(a)
 }
 
+/// Alias for [`_mm_cvttsd_si64`](fn._mm_cvttsd_si64_ss.html).
+#[cfg(target_arch = "x86_64")]
+#[inline(always)]
+#[target_feature = "+sse2"]
+#[cfg_attr(test, assert_instr(cvttsd2si))]
+pub unsafe fn _mm_cvttsd_si64x(a: f64x2) -> i64 { _mm_cvttsd_si64(a) }
```
BurntSushi: Can you add a test for this please? I know it's an alias, but we should try to cover the full public API.

src/x86/sse2.rs
```diff
@@ -1795,6 +1795,13 @@ pub unsafe fn _mm_cvtsd_si64(a: f64x2) -> i64 {
     cvtsd2si64(a)
 }
 
+/// Alias for [`_mm_cvtsd_si64`](fn._mm_cvtsd_si64_ss.html).
+#[cfg(target_arch = "x86_64")]
+#[inline(always)]
+#[target_feature = "+sse2"]
+#[cfg_attr(test, assert_instr(cvtsd2si))]
+pub unsafe fn _mm_cvtsd_si64x(a: f64x2) -> i64 { _mm_cvtsd_si64(a) }
```
BurntSushi: Can you add a test for this please? I know it's an alias, but we should try to cover the full public API.

MaloJaffre:
Thanks for your review!
I've split the tests between the aliases, to avoid repetition, and squashed my fixup commit.

alexcrichton:
👍

PR #166, opened 2017-11-02
[x86] expose cpuid, xgetbv, pushfd, popfd

gnzlbg:
This refactors the `cpuid` and `xgetby` intrinsics into its own functions and exposes them as part of the `x86` and `x86_64` architectures. 

The `cpuid` intrinsic is not available on all `x86` CPUs, so this also exposes a `has_cpuid() -> bool` intrinsic that detect this on non `x86_64` hosts, and the `pushfd` and `popfd` intrinsics required to implement it (GCC exposes all of this as intrinsics as well). 

When doing run-time feature detection for `x86`/`x86_64` we now check whether the `cpuid` instruction is available, and if not, then just "disable" all features.

One TODO:

- The `xgetbv` intrinsic requires the `xsave` target feature but this is not currently exposed by rustc, see #167 .

---

- `__readeflags` is equivalent to GCC's `__readeflags` exposed in [`/i386/ia32intrin.h`](https://github.com/gcc-mirror/gcc/blob/master/gcc/config/i386/ia32intrin.h#L260) and Clang's `__readeflags` exposed in [`ia32intrin.h`](https://github.com/llvm-mirror/clang/blob/master/lib/Headers/ia32intrin.h#L33)
- `__writeeflags` is equivalent to GCC's `__writeeflags` exposed in [`/i386/ia32intrin.h`](https://github.com/gcc-mirror/gcc/blob/master/gcc/config/i386/ia32intrin.h#L288) and Clang's `__writeeflags` exposed in [`ia32intrin.h`](https://github.com/llvm-mirror/clang/blob/master/lib/Headers/ia32intrin.h#L52)
- `_xgetbv` is equivalent to GCC's `_xgetbv` exposed in [`i386/xsaveintrin.h`](https://github.com/gcc-mirror/gcc/blob/154452f45aad4bfe7a1f719c65cd26837f416a48/gcc/config/i386/xsaveintrin.h#L60) and Clangs' `_xgetbv` exposed in [`intrin.h`](https://github.com/llvm-mirror/clang/blob/master/lib/Headers/intrin.h#L916)
- `_cpuid_count`is equivalent to GCC's `__cpuid_count` exposed in [`i386/cpuid.h`](https://github.com/gcc-mirror/gcc/blob/master/gcc/config/i386/cpuid.h) and Clang's `__cpuid_count` exposed in [`cpuid.h`](https://github.com/llvm-mirror/clang/blob/master/lib/Headers/cpuid.h#L214)

### DIFF
--- a/src/x86/cpuid.rs
+++ b/src/x86/cpuid.rs
@@ -1,0 +1,120 @@
+//! `cpuid` intrinsics
+
+#[cfg(test)]
+use stdsimd_test::assert_instr;
+
+/// Result of the `cpuid` instruction.
+#[derive(Copy, Clone, Eq, Ord, PartialEq, PartialOrd)]
+#[cfg_attr(feature = "cargo-clippy", allow(stutter))]
+pub struct CpuidResult {
+    /// EAX register.
+    pub eax: u32,
+    /// EBX register.
+    pub ebx: u32,
+    /// ECX register.
+    pub ecx: u32,
+    /// EDX register.
+    pub edx: u32,
+}
+
+/// `cpuid` instruction.
+///
+/// The [CPUID Wikipedia page][wiki_cpuid] contains how to query which
+/// information using the `eax` and `ecx` registers, and the format in
+/// which this information is returned in `eax...edx`.
+///
+/// The `has_cpuid()` intrinsics can be used to query whether the `cpuid`
+/// instruction is available.
+///
+/// The definitive references are:
+/// - [Intel 64 and IA-32 Architectures Software Developer's Manual Volume 2:
+///   Instruction Set Reference, A-Z][intel64_ref].
+/// - [AMD64 Architecture Programmer's Manual, Volume 3: General-Purpose and
+///   System Instructions][amd64_ref].
+///
+/// [wiki_cpuid]: https://en.wikipedia.org/wiki/CPUID
+/// [intel64_ref]: http://www.intel.de/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-instruction-set-reference-manual-325383.pdf
+/// [amd64_ref]: http://support.amd.com/TechDocs/24594.pdf
+#[inline(always)]
+#[cfg_attr(test, assert_instr(cpuid))]
+#[cfg_attr(feature = "cargo-clippy", allow(stutter))]
+pub unsafe fn __cpuid_count(eax: u32, ecx: u32) -> CpuidResult {
+    let mut r = ::std::mem::uninitialized::<CpuidResult>();
+    asm!("cpuid"
+         : "={eax}"(r.eax), "={ebx}"(r.ebx), "={ecx}"(r.ecx), "={edx}"(r.edx)
+         : "{eax}"(eax), "{ecx}"(ecx)
+         : :);
+    r
+}
+
+/// `cpuid` instruction.
+///
+/// See `__cpuid_count`.
+#[inline(always)]
+#[cfg_attr(test, assert_instr(cpuid))]
+#[cfg_attr(feature = "cargo-clippy", allow(stutter))]
+pub unsafe fn __cpuid(eax: u32) -> CpuidResult {
+    __cpuid_count(eax, 0)
+}
+
+/// Does the host support the `cpuid` instruction?
+#[inline(always)]
+#[cfg_attr(feature = "cargo-clippy", allow(stutter))]
+pub fn has_cpuid() -> bool {
+    #[cfg(target_arch = "x86_64")]
+    {
+        true
+    }
+    #[cfg(target_arch = "x86")]
+    {
+        use super::ia32::{__readeflags, __writeeflags};
+
+        // On `x86` the `cpuid` instruction is not always available.
+        // This follows the approach indicated in:
+        // http://wiki.osdev.org/CPUID#Checking_CPUID_availability
+        unsafe {
+            // Read EFLAGS:
+            let eflags: u32 = __readeflags();
+
+            // Invert the ID bit in EFLAGS:
+            let eflags_mod: u32 = eflags | 0x0020_0000;
+
+            // Store the modified EFLAGS (ID bit may or may not be inverted)
+            __writeeflags(eflags_mod);
+
+            // Read EFLAGS again:
+            let eflags_after: u32 = __readeflags();
+
+            // Check if the ID bit changed:
+            eflags_after != eflags
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_always_has_cpuid() {
+        // all currently-tested targets have the instruction
+        // FIXME: add targets without `cpuid` to CI
+        assert!(has_cpuid());
+    }
+
+    #[cfg(target_arch = "x86")]
+    #[test]
+    fn test_has_cpuid() {
+        use vendor::__readeflags;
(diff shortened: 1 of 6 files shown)


petrochenkov:
>AFAIK if there is an x86 processor for which it is not available, detecting this at run-time is impossible because... that's cpuids job.

Presence of `CPUID` can be checked by setting/clearing `EFLAGS.ID` (*if* you need to support really old CPUs).

gnzlbg:
@petrochenkov do you have an example of how that is done, or a link to some relevant material? I would then make cpuid unsafe and update the docs about how to check it (maybe offering a simpler way to check it).

petrochenkov:
(I wish I didn't leave that comment, 80486 is very old.)
Intel SDM / Vol.2 Instruction Set Reference / CPUID
http://wiki.osdev.org/CPUID

gnzlbg:
@petrochenkov don't worry, I am on it. 

I have added a `has_cpuid` function that returns `true` if the CPU has the `cpuid` intrinsic. On `x86_64` it always returns true. On x86 it does this:  https://github.com/gnzlbg/stdsimd/blob/dogfood_cpuid/src/x86/x86.rs#L52

gnzlbg:
So this is mostly fixed. For some reason, `i586` fails to compile, but I can't differentiate between i686 and i586 (`target_arch = "x86"` is the only tool I have available here). Is there a way to make this portable?

nominolo:
To distinguish between i586 and i686 you should check for `target_arch="x86"` and `target_feature="sse2"`. The latter is only available on i686.

src/x86/misc.rs
```diff
@@ -0,0 +1,150 @@
...
+            let eflags_mod: u32 = eflags | 0x0020_0000;
+
+            // Store the modified EFLAGS (ID bit may or may not be inverted)
+            popfd(eflags_mod);
+
+            // Read EFLAGS again:
+            let eflags_after: u32 = pushfd();
+
+            // Check if the ID bit changed:
+            eflags_after != eflags
```
petrochenkov: @gnzlbg 
Have you checked what exact instruction sequence is produced?
OSDev wiki warns: "Note: Implementing this routine in for example C can lead to issues, because the compiler may change EFLAGS at any time."
(Any intermediate arithmetic operations can change `EFLAGS`.)

gnzlbg: Yes I have, and the build should be green now :)

gnzlbg: It produces this in release:

```asm
ex:
	pushfl
	popl	%eax
	movl	%eax, %ecx
	orl	$2097152, %ecx
	pushl	%ecx
	popfl
	pushfl
	popl	%ecx
	cmpl	%eax, %ecx
	setne	%al
	retl
```

and this in debug:

```asm
	ex:
	subl	$12, %esp
	pushfl
	popl	%eax
	movl	%eax, 4(%esp)
	movl	4(%esp), %eax
	movl	%eax, %ecx
	orl	$2097152, %ecx
	pushl	%ecx
	popfl
	pushfl
	popl	%ecx
	movl	%ecx, 8(%esp)
	cmpl	%eax, 8(%esp)
	setne	%dl
	movb	%dl, 3(%esp)
	movb	3(%esp), %al
	andb	$1, %al
	movzbl	%al, %eax
	addl	$12, %esp
	retl
```

not "perfect" but good enough, plus I got to factor `pushfd` and `popf` out :)

gnzlbg:
So @alexcrichton @petrochenkov this is finished and green.

The only thing that worries me is that `pushfd` clobbers the stack but AFAIK it is impossible to indicate this via the `asm!` macro; it is not possible in clang and gcc either.

src/x86/misc.rs
```diff
@@ -0,0 +1,152 @@
...
+    let eflags: u32;
+    asm!("pushfd; popl $0" : "=r"(eflags) : : : "volatile");
+    eflags
+}
+
+/// Write EFLAGS.
+#[cfg(target_arch = "x86")]
+#[inline(always)]
+pub unsafe fn popfd(eflags: u32) {
+    asm!("pushl $0; popfd" : : "r"(eflags) : "cc", "flags" : "volatile");
```
gnzlbg: Also I am not 100% sure if the `cc` clobber is required here.

gnzlbg:
@BurntSushi general design question: I've put these intrinsics in a `misc` module. For users this doesn't matter much because our internal modules are private, but from the point-of-view of library organization I think `misc` is a pretty bad name. 

The only other idea I've had is to name the modules .../`i386`/`i486`/`i586`/`i686`/... instead. The main issue I have with this is that, for example, havint to use `cfg!(all(target_arch = "x86", target_feature = "-sse2"))` to detect `i686` is... messy.

gnzlbg:
@alexcrichton 

- `pushfd` is equivalent to GCC's `__readeflags` exposed in [`/i386/ia32intrin.h`](https://github.com/gcc-mirror/gcc/blob/master/gcc/config/i386/ia32intrin.h#L260) and Clang's `__readeflags` exposed in [`ia32intrin.h`](https://github.com/llvm-mirror/clang/blob/master/lib/Headers/ia32intrin.h#L33)
- `popfd` is equivalent to GCC's `__writeeflags` exposed in [`/i386/ia32intrin.h`](https://github.com/gcc-mirror/gcc/blob/master/gcc/config/i386/ia32intrin.h#L288) and Clang's `__writeeflags` exposed in [`ia32intrin.h`](https://github.com/llvm-mirror/clang/blob/master/lib/Headers/ia32intrin.h#L52)
- `xgetbv` is equivalent to GCC's `_xgetbv` exposed in [`i386/xsaveintrin.h`](https://github.com/gcc-mirror/gcc/blob/154452f45aad4bfe7a1f719c65cd26837f416a48/gcc/config/i386/xsaveintrin.h#L60) and Clangs' `_xgetbv` exposed in [`intrin.h`](https://github.com/llvm-mirror/clang/blob/master/lib/Headers/intrin.h#L916)
- `cpuid`is equivalent to GCC's `_get_cpuid` exposed in [`i386/cpuid.h`](https://github.com/gcc-mirror/gcc/blob/master/gcc/config/i386/cpuid.h) and Clang's `__cpuid` exposed in [`cpuid.h`](https://github.com/llvm-mirror/clang/blob/master/lib/Headers/cpuid.h#L214)

gnzlbg:
I could rename `pushfd` and `popfd` to `__readeflags` and `__writeeflags` and put them in a `ia32.rs` module, rename `xgetbv` to `_xgetbv` and move that to an `intrin.rs` module, and move `cpuid` into its own module (but I don't know whether I should rename it to `__cpuid` or `__get_cpuid`, clang and gcc differ here).

Note that the signature of `cpuid` in clang and gcc is identical and takes 4 int by reference but we can just return a struct or a tuple here (I prefer the struct with the named members but can switch that to a tuple).

gnzlbg:
@alexcrichton I've refactor this to agree naming-wise with GCC and Clang. 

The main difference remains that the `cpuid` instructions return the read values in a struct instead of modifying 4 ints via `&mut`. Also, AFAIK neither Clang nor GCC expose their intrinsics to detect `cpuid` support.

gnzlbg:
Blocked by https://github.com/rust-lang/rust/pull/45761 . I'll reopen this PR once that is merged upstream and I can properly use the `xsave` and `xsaveopt` target features. Turns out LLVM has a `llvm.x86.xgetbv` intrinsic that might allow me to remove some of the assembly.

PR #168, opened 2017-11-03
sse4.1: convert intrinsics 

p32blo:


src/x86/sse41.rs
```diff
@@ -238,11 +239,42 @@ pub unsafe fn _mm_max_epi32(a: i32x4, b: i32x4) -> i32x4 {
...
+#[cfg_attr(test, assert_instr(pcmpeqq))]
+pub unsafe fn _mm_cmpeq_epi64(a: i64x2, b: i64x2) -> i64x2 {
+    a.eq(b)
+}
+
+/// Sign extend packed 8-bit integers in `a` to packed 16-bit integers
+#[target_feature = "+sse4.1"]
+#[cfg_attr(test, assert_instr(pmovsxbw))]
+pub unsafe fn _mm_cvtepi8_epi16(a: i8x16) -> i16x8 {
+    simd_cast::<::v64::i8x8, _>(simd_shuffle8(a, a, [0, 1, 2, 3, 4, 5, 6, 7]))
```
BurntSushi: I think you can say `.as_i16x8()` here?

src/x86/sse41.rs
```diff
@@ -238,11 +239,42 @@ pub unsafe fn _mm_max_epi32(a: i32x4, b: i32x4) -> i32x4 {
...
+}
+
+/// Sign extend packed 8-bit integers in `a` to packed 32-bit integers
+#[target_feature = "+sse4.1"]
+#[cfg_attr(test, assert_instr(pmovsxbd))]
+pub unsafe fn _mm_cvtepi8_epi32(a: i8x16) -> i32x4 {
+    let cast = simd_cast::<_, ::v512::i32x16>(a);
```
BurntSushi: Why are you using a 512 bit vector here? What does Clang do?

src/x86/sse41.rs
p32blo: Yes, but the type still needs to be specified, for example:

```rust
simd_shuffle8::<_, ::v64::i8x8>(a, a, [0, 1, 2, 3, 4, 5, 6, 7]).as_i16x8()
```

BurntSushi: Yeah I'd prefer that. I'd like to use the wrappers around the `simd` intrinsics as much as possible.

src/x86/sse41.rs
p32blo: I can't do this:
```rust
simd_cast::<::v32::i8x4, _>(simd_shuffle4(a, a, [0, 1, 2, 3]))
```
Because there is no `i8x4` type (note that `v32` is an hypothetical module containing the type `i8x4`)

So I instead convert first `i8x16` to `i32x16` and then get the `i32x4` with `simd_suffle4`, instead of the other way around. 

I do realize that this might no be a good way to solve this even do it generated the right instruction. Will the usage of a `v512` type require 512 bit wide registers in a machine that might not have them?

src/x86/sse41.rs
```diff
@@ -724,6 +788,87 @@ mod tests {
...
+        let a = i16x8::splat(10);
+        let r = sse41::_mm_cvtepi16_epi64(a);
+        let e = i64x2::splat(10);
+        assert_eq!(r, e);
+        let a =  i16x8::splat(-10);
+        let r = sse41::_mm_cvtepi16_epi64(a);
+        let e = i64x2::splat(-10);
+        assert_eq!(r, e);
+    }
+*/
```
BurntSushi: Why is this test commented out?

src/x86/sse41.rs
BurntSushi: Can you do what Clang does? It doesn't use an `i8x4` or a 512 bit vector type: https://github.com/llvm-mirror/clang/blob/a38ba9770a70056f9fdd6c71f819e5db45a105e4/lib/Headers/smmintrin.h#L1283-L1289

p32blo: What clang does would be this line:
```rust
simd_cast::<::v32::i8x4, _>(simd_shuffle4(a, a, [0, 1, 2, 3]))
```
But rust's `simd_suffle4` has to return a `i8x4` which does not exist

BurntSushi: @p32blo Hmm sorry, I guess I'm not seeing it. Can you point out where the 32 bit vector is being used? I see `__v4si` and `__v16qs`, both of which I think are 128 bit vectors?

p32blo: Ok, we have a 128 bit vector with 8 bit elements, so there are 16 elements. First, with `simd_shuffle` we get the first 4 elements. This leaves us with 32 bits. Then with `simd_cast`, we sign extend this 4 elements to be 32 bits and this brings us back to the 128 bit total.

The problem were is when we take the first 4 elements we need a 32 bit type to hold only four 8bit elements. In clang we don't see this intermediate types but they are still there implicitly.

src/x86/sse41.rs
p32blo: Done ;)

src/x86/sse41.rs
p32blo: It was temporarily commented until the respective intrinsic could be implemented

p32blo:
As @nominolo suggested over at gitter I tried implementing the needed types. I defined the types directly in `lib.rs` but they can be moved to new files if needed. I also don't know if I should export them or if they should remain implementation details.

PR #173, opened 2017-11-06
Fix mm256_round_epi* return types

AdamNiederer:
From the Intel intrinsics manual (emphasis mine):

> Compute the absolute value of packed 16-bit integers in a, and store the
> *unsigned* results in dst.


alexcrichton:
Thanks!

Issue #174, opened 2017-11-06
Implement GCC's `__get_cpuid_max`


gnzlbg:
Once the `cpuid.rs` module is merged adding this should be trivial so we might want to mark this as an  easy task for beginners. I could mentor it.

The definition of `__get_cpuid_max` can be found here:

https://clang.llvm.org/doxygen/cpuid_8h.html#abe7c43218d2ca542413f9fc0f5063238

Issue #146, opened 2017-10-23
AVX-512 intrinsics


hdevalence:
I tried getting an AVX-512 intrinsic to work and ran into a bunch of difficulties. Some points:

- It looks like the combination of AVX512's masks and AVX512VL (which lets AVX512 instructions operate on 128/256bit vectors) means that for most instructions there's one C intrinsic for each of {no mask, write mask, zero mask} x {xmm, ymm, zmm}.

- These would probably be good to generate with a macro?

- Because AVX512 uses mask registers, the `constify!` macro hacks are probably not needed for mask instructions.

- The list of intrinsics [linked in the readme](https://gist.githubusercontent.com/anonymous/a25d3e3b4c14ee68d63bd1dcb0e1223c/raw/fb798be53a7b29f9b28a0c7a9e2b9a16ae915f71/allintrinsics) doesn't seem to have non-masked versions; I don't know if this is just an accident of how it was made.

- Trying to use the `int_x86_avx512_mask_pmul_dq_512` intrinsic from that list using 
```
#[link_name = "llvm.x86.avx512.mask.pmul.dq.512"]
fn mask_pmul_dq_512(a: i32x16, b: i32x16, src: i64x8, k: i8) -> i64x8;
```
didn't work, failing with 
```
rustc: /checkout/src/llvm/include/llvm/Support/Casting.h:236: typename llvm::cast_retty<X, Y*>::ret_type llvm::cast(Y*) [with X = llvm::VectorType; Y = llvm::Type; typename llvm::cast_retty<X, Y*>::ret_type = llvm::VectorType*]: Assertion `isa<X>(Val) && "cast<Ty>() argument of incompatible type!"' failed.
```
which I guess means I was linking to the intrinsic incorrectly?

@alexcrichton reduced to this minimal example for the `vpmuldq` instruction: https://godbolt.org/g/VMCtYy and found https://github.com/rust-lang/rust/blob/4c053db233d69519b548e5b8ed7192d0783e582a/src/librustc_trans/cabi_x86_64.rs#L30-L31 which hardcodes the biggest vector as 256 bits (the size of a ymm register).

alexcrichton:
I've opened up a PR for the some rustc features at f764eaf453162cd19ef484ece07cc21e14dfb2c1 https://github.com/rust-lang/rust/pull/45528

hdevalence:
[...]

PR #175, opened 2017-11-06
XSAVE, CPUID, Runtime AVX-512 and ARM support

### LINKED ISSUE #171: Implement missing `xsave` intrinsics
# `x86`

- [ ] `fn _xsave (p: *mut c_void, m: u64) -> ()`
- [ ] `fn _xrstor (p: *mut c_void, m: u64) -> ()`
- [ ] `fn _xsetbv (a: u32, v: u64) -> ()`
- [ ] `fn _xgetbv (a: u32) -> u64`

# `x86_64` 

- [ ] `fn _xsave64 (p: *mut c_void, m: u64) -> ()`
- [ ] `fn _xrstor64 (p: *mut c_void, m: u64) -> ()`

gnzlbg:
This PR implements:

- the `xsave` intrinsics,
- the `__readeflags`/`__writeeflags` intrinsics,
- the `cpuid` intrinsics (except for GCC's `__get_cpuid_max`, Issue #174 ), and
- run-time feature detection for `xsave` and `AVX-512`,

and fixes one bug:
- the `CPUID` instruction was used before testing whether the CPU supported it.

Note that the `has_cpuid` intrinsic does not exist in GCC nor Clang (it is a sub-set of GCC's `__get_cpuid_max`) but I've included it because we can make it a no-op on `x86_64` which is what most folks are targeting nowadays, which is useful for our own run-time feature detection support and for cpuid crates like @shepmaster's cupid.

This PR does arguably a lot of stuff. We need the `EFLAGS` intrinsics to properly refactor `cpuid` into its own intrinsic. We need `xsave` run-time support for properly refactoring the `xgetbv` intrinsic. And `xsave` run-time support is tightly coupled with AVX run-time feature detection, which is different for AVX/AVX2 and AVX-512 and since we want to support AVX-512 soon anyways I did not want to have to go through the spec twice. I also wanted to test the `xgetbv` intrinsic and ended implementing all of the `xsave` intrinsics for this. Arguably I just needed to implement `xsetbv` but, again, I did not wanted to have to go through the spec twice.

The run-time feature detection for x86 grows significantly with AVX-512 support, so I backported one of the refactorings I have in the run-time feature detection module for ARM (which is even bigger). There I refactor it further into its own top-level module to be able to share some code between both.

---

This is blocked on:

- ~https://github.com/rust-lang/rust/pull/45761 (whitelisting XSAVE in rustc)~
- https://github.com/rust-lang/rust/pull/45528 (whitelisting AVX-512 in rustc)

Closes #171 .
Helps with the run-time part of #146 .

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ opt-level = 3
 
 [dev-dependencies]
 stdsimd-test = { version = "0.*", path = "stdsimd-test" }
-cupid = "0.3"
+cupid = "0.4.0"
 
 [features]
 strict = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,3 +33,4 @@ cupid = "0.4.0"
 [features]
 strict = []
 std = []
+intel_sde = []
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ matrix:
     - env: TARGET=i586-unknown-linux-gnu
     - env: TARGET=i686-unknown-linux-gnu
     - env: TARGET=x86_64-unknown-linux-gnu NO_ADD=1
-    - env: TARGET=x86_64-unknown-linux-gnu-emulated NO_ADD=1 STDSIMD_TEST_EVERYTHING=1
+    - env: TARGET=x86_64-unknown-linux-gnu-emulated NO_ADD=1 STDSIMD_TEST_EVERYTHING=1 FEATURES="intel_sde"
     - env: TARGET=arm-unknown-linux-gnueabihf
     - env: TARGET=armv7-unknown-linux-gnueabihf
     - env: TARGET=aarch64-unknown-linux-gnu
@@ -33,7 +33,7 @@ install:
 
 script:
   - cargo generate-lockfile
-  - ci/run-docker.sh $TARGET
+  - ci/run-docker.sh $TARGET $FEATURES
 
 notifications:
   email:
--- a/ci/docker/aarch64-unknown-linux-gnu/Dockerfile
+++ b/ci/docker/aarch64-unknown-linux-gnu/Dockerfile
@@ -8,6 +8,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
   qemu-user \
   make \
   file
+
 ENV CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc \
     CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_RUNNER="qemu-aarch64 -L /usr/aarch64-linux-gnu" \
     OBJDUMP=aarch64-linux-gnu-objdump
--- a/ci/run-docker.sh
+++ b/ci/run-docker.sh
@@ -4,10 +4,11 @@
 set -ex
 
 run() {
-    echo $1
+    echo "Building docker container for TARGET=${1}"
     docker build -t stdsimd ci/docker/$1
     mkdir -p target
     target=$(echo $1 | sed 's/-emulated//')
+    echo "Running docker"
     docker run \
       --user `id -u`:`id -g` \
       --rm \
--- a/ci/run-docker.sh
+++ b/ci/run-docker.sh
@@ -16,6 +16,7 @@ run() {
       --env CARGO_HOME=/cargo \
       --volume `rustc --print sysroot`:/rust:ro \
       --env TARGET=$target \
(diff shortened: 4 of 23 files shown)


alexcrichton:
This all looks great to me, thanks so much for digging into all this! One day I'd love to dig into all the cpu detection but it looks quite thorough here and there's plenty of links to follow so I'm more than ok with the current implementation.

I'm ok personally landing this whenever CI is green (I don't think we need AVX-512 support in rustc to block this b/c it doesn't add any AVX-512 intrinsics, right?). It looks like CI currently has an error with `llvm.x86.xrstor` though?

alexcrichton:
I'll also note that personally I'm a little anxious about these intrinsics in the sense that there's not an Intel manual with the signatures (or maybe there is?) in the same way we have such a guide for all the SIMD intrinsics. That being said though I think this library will go through normal stabilization processes as other libraries in libstd, so it's ok for part of this library to be unstable (such as these intrinsics which I'm not currently sure about) and the other parts can be stable (like the Intel-defined signatures).

gnzlbg:
> I'm a little anxious about these intrinsics in the sense that there's not an Intel manual with the signatures (or maybe there is?)

AFAIK such a manual does not exist (maybe there is one in the Intel icc compiler?) but there is something much better, [Intel's online data-base](https://software.intel.com/sites/landingpage/IntrinsicsGuide/) !

The data-base gives you the C signatures, the assembly instruction the CPUID flag (or flags) that you need to check... Clang and GCC deviate slightly from the C signatures in the Intel guide though (mainly when it comes to signed vs unsigned integers), so it is useful to also check the:

- [Clang intrinsic headers](https://github.com/llvm-mirror/clang/tree/master/lib/Headers)
- GCC intrinsic headers for [i386](https://github.com/gcc-mirror/gcc/tree/master/gcc/config/i386) and [ia64](https://github.com/gcc-mirror/gcc/tree/master/gcc/config/ia64)(here without looking at the implementation!!!).

at least, when you read that Intel says some `int` should be signed, but the operation is clearly doing bit manipulations, in that case, if clang chooses an `unsigned` integer type I would suggest doing the same in Rust. 

---

> It looks like CI currently has an error with llvm.x86.xrstor though?

I got these in my computer as well but I thought I fixed them, will recheck these once the rust PR is merged. LLVM is picky with the pointer types that the intrinsics take.

gnzlbg:
Also technically `XSAVE` intrinsics are SIMD-related, because they allow you to save the state of the SIMD registers, so if a Rust library is going to, for whatever reasons (e.g. green threads) implement context switches correctly, it should probably use them.

gnzlbg:
> I don't think we need AVX-512 support in rustc to block this b/c it doesn't add any AVX-512 intrinsics, right?

Right. The only thing blocking this is the `XSAVE` whitelisting PR that you already accepted.

alexcrichton:
> AFAIK such a manual does not exist (maybe there is one in the Intel icc compiler?) but there is something much better, Intel's online data-base !

Nice! That's perfect.

This all sounds great to me, let's merge when CI is green

alexcrichton:
@gnzlbg could the 4000 line macro be left to a separate PR perhaps? I'm not sure I'm entirely convinced we should add it yet...

gnzlbg:
@alexcrichton yes, I can split that to a different PR before merging this one. But there was some runtime detection code there that I wanted to have here. Is that ok if I do that when the `xsave` PR is merged into rustc (it seems to be taking a while longer than I expected).

gnzlbg:
@BurntSushi @alexcrichton this one should be ready to go as well. The ARM ci still doesn't use run-time feature detection because there is a PR for `cross` that adds support for `qemu-system` and I think it makes more sense to use `cross` instead of trying to roll our own solution.

alexcrichton:
@gnzlbg yeah want to split out the bextri fix? Otherwise seems fine by me!

gnzlbg:
@alexcrichton done :)

Cargo.toml
```diff
@@ -28,8 +28,9 @@ opt-level = 3
...
 stdsimd-test = { version = "0.*", path = "stdsimd-test" }
-cupid = "0.3"
+cupid = "0.4.0"
 
 [features]
 strict = []
-std = []
\ No newline at end of file
+std = []
+intel_sde = []
```
alexcrichton: When you get a chance, could you add documentation here for what `intel_sde` is?

gnzlbg: Sure, I'll add a comment to the Cargo.toml.

PR #177, opened 2017-11-06
Implement missing std::ops

gnzlbg:
Implements missing `std::ops` and adds test for all operators of all vector types.

Blocked on https://github.com/rust-lang/rust/pull/45804

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -166,13 +166,7 @@ macro_rules! define_common_ops {
                     unsafe { simd_mul(self, other) }
                 }
             }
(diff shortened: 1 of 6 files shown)


BurntSushi:
This LGTM, but it looks like `simd_rem` isn't a thing?

gnzlbg:
@BurntSushi this PR is blocked on rust-lang/rust#45804 which implements `simd_rem` and the missing `simd_div`s.

gnzlbg:
This is ready.

PR #187, opened 2017-11-09
[x86][sse4.1] Add phminposuw & pmul* instructions

crypto-universe:
Unfortunately, pmulld is not supported by llvm.

This is a PR for https://github.com/rust-lang-nursery/stdsimd/issues/40 tracking issue.

### DIFF
--- a/src/x86/sse41.rs
+++ b/src/x86/sse41.rs
@@ -580,6 +580,39 @@ pub unsafe fn _mm_round_ss(a: f32x4, b: f32x4, rounding: i32) -> f32x4 {
     constify_imm4!(rounding, call)
 }
 
+/// Find minimal u16 element in vector.
+/// Place it in the first element of resulting vector and it's index
+/// in second element (formally bits [16..18] inclusive).
+/// All other elements are set to zero.
+#[inline(always)]
+#[target_feature = "+sse4.1"]
+#[cfg_attr(test, assert_instr(phminposuw))]
+pub unsafe fn _mm_minpos_epu16(a: u16x8) -> u16x8 {
+    phminposuw(a)
+}
+
+/// Multiply the low 32-bit integers from each packed 64-bit element
+/// in a and b, and store the signed 64-bit results in dst.
+#[inline(always)]
+#[target_feature = "+sse4.1"]
+#[cfg_attr(test, assert_instr(pmuldq))]
+pub unsafe fn _mm_mul_epi32(a: i32x4, b: i32x4) -> i64x2 {
+    pmuldq(a, b)
+}
+
+/// Multiply the packed 32-bit integers in a and b, producing intermediate
+/// 64-bit integers, and  returns the lowest 32-bit, whatever they might be,
+/// reinterpreted as a signed integer.
+/// While pmulld i32x4::splat(2), i32x4::splat(2) returns the obvious
+/// i32x4::splat(4), pmulld i32x4::splat(i32::MAX), i32x4::splat(2)
+/// would return a negative number.
+#[inline(always)]
+#[target_feature = "+sse4.1"]
+#[cfg_attr(test, assert_instr(pmulld))]
+pub unsafe fn _mm_mullo_epi32(a: i32x4, b: i32x4) -> i32x4 {
+    a * b
+}
+
 
 #[allow(improper_ctypes)]
 extern "C" {
@@ -627,6 +660,10 @@ extern "C" {
     fn roundsd(a: f64x2, b: f64x2, rounding: i32) -> f64x2;
     #[link_name = "llvm.x86.sse41.round.ss"]
     fn roundss(a: f32x4, b: f32x4, rounding: i32) -> f32x4;
+    #[link_name = "llvm.x86.sse41.phminposuw"]
+    fn phminposuw(a: u16x8) -> u16x8;
+    #[link_name = "llvm.x86.sse41.pmuldq"]
+    fn pmuldq(a: i32x4, b: i32x4) -> i64x2;
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


src/x86/sse41.rs
```diff
@@ -627,6 +647,13 @@ extern "C" {
     fn roundsd(a: f64x2, b: f64x2, rounding: i32) -> f64x2;
     #[link_name = "llvm.x86.sse41.round.ss"]
     fn roundss(a: f32x4, b: f32x4, rounding: i32) -> f32x4;
+    #[link_name = "llvm.x86.sse41.phminposuw"]
+    fn phminposuw(a: u16x8) -> u16x8;
+    #[link_name = "llvm.x86.sse41.pmuldq"]
+    fn pmuldq(a: i32x4, b: i32x4) -> i64x2;
+    // pmulld is not supported by llvm so far
+    // #[link_name = "llvm.x86.sse41.pmulld"]
+    // fn pmulld(a: i32x4, b: i32x4) -> i32x4;
```
BurntSushi: This is intentionally not supported by LLVM because we're supposed to implement it using their portable vector API. As #40 says, do what Clang does: https://github.com/llvm-mirror/clang/blob/a38ba9770a70056f9fdd6c71f819e5db45a105e4/lib/Headers/smmintrin.h#L546-L564

crypto-universe:
@BurntSushi `pmulld` gives a misleading result when a sign bit is set during multiplication of two positives.
Can we do smth with that?

BurntSushi:
@crypto-universe Sorry, I don't understand your question. Can you unpack? Remember the goal: *match the vendor interface.*

crypto-universe:
Sorry, what do you mean by unpack? It is done similar to `_mm_add_epi16` in sse2 in this repo.
Interface fits an [Intel Intrinsics Guide](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#techs=SSE4_1&expand=3744).

BurntSushi:
@crypto-universe It means I'd like you to explain more. I don't understand your question. It might help to provide code examples.

crypto-universe:
I was wondering if it good to leave it as it is, when 2 positive values after multiplication gives a negative value. Of course, SIMD instruction works like this and doesn't care about signed/unsigned data types, but maybe we should care on rust level somehow?
```
    #[simd_test = "sse4.1"]
    unsafe fn _mm_mullo_epi32() {
        let a = i32x4::new(15, -2, **1234567**, 99999);
        let b = i32x4::new(-20, -256, **666666**, -99999);
        let r = sse41::_mm_mullo_epi32(a, b);
        // Pay attention, most significant bit in e[2] is treated
        // as a sign bit!
        // **1234567 * 666666 = -1589877210**
        let e = i32x4::new(-300, 512, **-1589877210**, -1409865409);
        assert_eq!(r, e);
    }
```

gnzlbg:
> I was wondering if it good to leave it as it is, when 2 positive values after multiplication gives a negative value.

`_mm_mullo_epi32`'s is specified by Intel to work on signed integers. It multiplies the pairs of 32bit integers as 64bit integers [*], and the lowest 32 to bits of that 64bit signed integer are returned. 

Intel declares the intrinsic to return a signed integer, but specifies the semantics of the intrinsic as returning "just some bits". 

I'd say that we should stick to returning a `i32x4` for consistency with Intel intrinsics. While the behavior can be surprising, returning an `u32x4` would just make things surprising in a different way. 

Maybe you could improve the docs a bit to state that this instructions don't return the result of the multiplication of two signed integers, but emphasize how they work (that is, that they perform a lossless 64-bit multiply, and then return the lowest 32-bit, whatever they might be). And then give as an example that while `pmulld i32x4::splat(2), i32x4::splat(2)` behaves as an user would expect, `pmulld i32x4::splat(i32::MAX) i32x4::splat(2)` would return a negative number. 

[*](avoiding signed integer overflow in C and C++ which is undefined behavior) semantics are to wrap around on overflow, so I think that is fine.

crypto-universe:
@gnzlbg if you don't mind, I'll use sentences from your comment in documentation :)

gnzlbg:
@crypto-universe np, don't forget to run `rustfmt` :)

crypto-universe:
@gnzlbg `rustfmt` has no complains about sse41.rs content (but fixed another 22 files :P)
@BurntSushi GitHub claims that "Changes requested" by you. Changes are done, but with `--amend` and `git push --force`, so GitHub didn't notice changes. Make a review again, please.

gnzlbg:
> rustfmt has no complains about sse41.rs content

The build bot is broken, are you using rustfmt-nightly?

> cargo install rustfmt-nightly --force
> cargo fmt --all

The diff is:

```shell
-        let a = i32x4::new(15, 2/*ignored*/, 1234567, 4/*ignored*/);⏎
-        let b = i32x4::new(-20, -256/*ignored*/, 666666, 666666/*ignored*/);⏎
+        let a =⏎
+            i32x4::new(15, 2 /* ignored */, 1234567, 4 /* ignored */);⏎
+        let b = i32x4::new(⏎
+            -20,⏎
+            -256, /* ignored */⏎
+            666666,⏎
+            666666, /* ignored */⏎
+        );⏎
```

crypto-universe:
Previously I did it with `cargo install rustfmt`. Now should be good.

crypto-universe:
@BurntSushi , @gnzlbg any comments to this PR?

gnzlbg:
@crypto-universe I'm just waiting on @BurntSushi to check it out again, but this looks good to me.

I want to thank you for working on this, and please don't take the long delay here personally. Sometimes we all have a lot of time and the timezone god blesses us and things get merged very quickly, but it is normal for pull-requests to take ~1 week to get merged. 

If you need this changes in a release to be able to use them it always helps to let us know about it and we try to speed things up.

I know it is hard but what I try to do is pick a different item in the mean time and have multiple parallel pull-requests. I am pretty sure your next pull-request will be merged quicker :)

PR #189, opened 2017-11-10
syn API breaking change

gnzlbg:
The syn API changed in the following commit breaking our builds: https://github.com/dtolnay/syn/commit/c6b55bcbe4627a9293b9354b3d76a7db8bf7d6dd 

This commit fixes it by using the new API.

### DIFF
--- a/stdsimd-test/assert-instr-macro/src/lib.rs
+++ b/stdsimd-test/assert-instr-macro/src/lib.rs
@@ -28,8 +28,8 @@ pub fn assert_instr(
         .expect("expected #[assert_instr(instr, a = b, ...)]");
     let item =
         syn::parse::<syn::Item>(item).expect("must be attached to an item");
-    let func = match item.node {
-        syn::ItemKind::Fn(ref f) => f,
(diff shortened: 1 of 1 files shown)


stdsimd-test/assert-instr-macro/src/lib.rs
```diff
@@ -70,8 +70,11 @@ pub fn assert_instr(
                 }
             };
         }
-        let attrs = item.attrs
-            .iter()
+
+        let attrs = match item {
+            syn::Item::Fn(ref item) => &item.attrs,
+            _ => panic!("item must be a function!"),
+        }.iter()
```
p32blo: Can this be done like this?

```rust
let attrs = func.attrs.iter()
```
I really don't know a lot about `syn` but it looks like the match with `item` is done previously and you can already access the attrs from `func`.

gnzlbg:
I can't update this before tomorrow so feel free to pick it up if you want
this fixed sooner :)

alexcrichton:
👍

PR #192, opened 2017-11-16
[avx2] add _mm_256_cvtepu{8,16,32}_epi{16,32,64}

governortarkin:
## [avx2]
- _mm256_cvtepu16_epi32
- _mm256_cvtepu16_epi64
- _mm256_cvtepu32_epi64
- _mm256_cvtepu8_epi16
- _mm256_cvtepu8_epi32
- _mm256_cvtepu8_epi64

An intermediate `u8x4` type is needed for the cast in `_mm256_cvtepu8_epi64`, so I defined one inline just like it's signed counterpart. I see there might be some overlap with other open PR's (regarding the `u8x4`), but this seemed like the best route to take for now given the context and the desire to define these just like they are in `avx2intrin.h`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -173,6 +173,9 @@ mod v32 {
     define_ty! { i8x4, i8, i8, i8, i8 }
     define_impl! { i8x4, i8, 4, i8x4, x0, x1, x2, x3 }
 
+    define_ty! { u8x4, u8, u8, u8, u8 }
+    define_impl! { u8x4, u8, 4, i8x4, x0, x1, x2, x3 }
+
     define_casts!((i8x4, i32x4, as_i32x4), (i16x2, i64x2, as_i64x2));
 }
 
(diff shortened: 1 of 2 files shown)


gnzlbg:
@governortarkin thank you for the pull-request! 

> An intermediate u8x4 type is needed for the cast in _mm256_cvtepu8_epi64, so I defined one inline just like it's signed counterpart.

The `i8x4` type is already available in the [`v32` module](https://github.com/rust-lang-nursery/stdsimd/blob/master/src/lib.rs#L167), I think I would prefer for the `u8x4` type to be there as well.

governortarkin:
@gnzlbg I must have overlooked the `v32` module. In that case, I can also prune the existing [duplicate](https://github.com/rust-lang-nursery/stdsimd/blob/9b81ab7a79810a08034fa697b68f5cb657ee5e8c/src/x86/avx2.rs#L590) `i8x4` definition.

In the mean time, it looks like #183 adds the `u8x4` that we'll need [here](https://github.com/rust-lang-nursery/stdsimd/pull/183/commits/49c19ee88829dc3dbff18f44933fa2aaf2249a45#diff-b4aea3e418ccdb71239b96952d9cddb6R175), so I guess that concern is addressed once #183 is merged.

gnzlbg:
I think that this PR is going to end up being merged sooner than #183 (the rest of the PR LGTM) so I would suggest fixing it here, we can amend #183 afterwards.

gnzlbg:
Thanks.

PR #195, opened 2017-11-19
refactor the x86 module

gnzlbg:
This PR refactors the x86 module into `i386`, `i586`, `i686` and `x86_64`. Most of the intrinsics that were previously `x86_64` only can now be used on `i686` as well.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,10 @@ stdsimd-test = { version = "0.*", path = "stdsimd-test" }
 cupid = "0.4.0"
 
 [features]
-strict = []
 std = []
+
+# Internal-only: denies all warnings.
+strict = []
+# Internal-only: enables only those intrinsics supported by Intel's
+# Software Development Environment (SDE).
 intel_sde = []
--- a/src/x86/i586/sse41.rs
+++ b/src/x86/i586/sse41.rs
@@ -6,7 +6,6 @@ use std::mem;
 use stdsimd_test::assert_instr;
 use simd_llvm::{simd_shuffle2, simd_shuffle4, simd_shuffle8};
 
-use x86::__m128i;
 use v128::*;
 
 // SSE4 rounding constans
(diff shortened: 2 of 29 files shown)


src/x86/i586/sse41.rs
```diff
@@ -198,16 +187,6 @@ pub unsafe fn _mm_insert_epi32(a: i32x4, i: i32, imm8: u8) -> i32x4 {
     a.replace((imm8 & 0b11) as u32, i)
 }
 
-/// Return a copy of `a` with the 64-bit integer from `i` inserted at a
-/// location specified by `imm8`.
-#[cfg(all(target_arch = "x86_64", not(target_feature = "sse2")))] // i586
```
gnzlbg: These checks were wrong but should be right now.

src/x86/mod.rs
```diff
@@ -1,24 +1,26 @@
...
+// x86 w/o sse2
+mod i586;
+pub use self::i586::*;
+
+#[cfg(any(all(target_arch = "x86", target_feature = "sse2"),
+          target_arch = "x86_64"))]
+mod i686;
+#[cfg(any(all(target_arch = "x86", target_feature = "sse2"),
+          target_arch = "x86_64"))]
+pub use self::i686::*;
```
gnzlbg: Everything in the `i686` modules was previously only available in `x86_64`. The check for `i686` is `x86 + sse2`.

src/x86/mod.rs
```diff
@@ -1,24 +1,26 @@
...
+mod i386;
+pub use self::i386::*;
+
+// x86 w/o sse2
+mod i586;
+pub use self::i586::*;
+
+#[cfg(any(all(target_arch = "x86", target_feature = "sse2"),
+          target_arch = "x86_64"))]
+mod i686;
```
alexcrichton: Could you add a coment for this module indicating that this will be compiled in for all `i686-*` targets by default, and it's just omitted for `i586`?

gnzlbg: Done. I've documented the `intel_sde` feature as well.

Cargo.toml
```diff
@@ -33,4 +33,6 @@ cupid = "0.4.0"
 [features]
 strict = []
 std = []
+# This feature enables only those intrinsics supported by
+# Intel's Software Development Environment (SDE):
```
alexcrichton: Thanks! Could this extend the documentation to indicate that this is an internal-only feature not intended for consumers?

Additionally is the `strict` feature like that as well, where it's' internal-only? If so mind noting that as well in addition to adding some docs to `strict`? (this can happen in a future PR though)

gnzlbg: Yes both `intel_sde` and `strict` are internal features, is there a way to add development-only features? 

Or is there a standard convention for documenting these? 

I would otherwise just add `# internal only: explanation`.

alexcrichton: Right now the only way of having a development-only feature is to specify it through `RUSTFLAGS` but that's not really the best solution. A comment I think should work just fine!

gnzlbg: done :)

PR #196, opened 2017-11-19
Unconditionally flag as `#![no_std]`

alexcrichton:
This is more idiomatic for no-std-compatible crates where imports are
unconditionally rewritten to `core` and then only when necessary `std` is pulled
in explicitly.


gnzlbg:
LGTM. The `extern crate core as std` was me just being lazy.

PR #197, opened 2017-11-20
[coresimd] extracts the no_std components into the coresimd crate

gnzlbg:
- Extracts the `no_std` components into the `coresimd` crate, which is unconditionally `no_std` (most of the library).
- Re-exports the `coresimd` crate as `stdsimd` + `std` components (e.g. run-time feature detection for ARM)
- Removes the dependency on the `std::os::raw::c_void` type by making those intrinsics take a `*mut u8` instead. 

The crates `coresimd` and `stdsimd` currently are not part of the same `workspace` because I got linker errors. We can fix this either in this PR or a subsequent one. Also, this PR broke Windows tests. PRs to this PR that fix this are welcome.


.gitignore
```diff
@@ -2,3 +2,7 @@ Cargo.lock
 .*.swp
 /target
 tags
+/coresimd/target
+/stdsimd-test/target
+/stdsimd-test/assert-instr-macro/target
+/stdsimd-test/simd-test-macro/target
```
gnzlbg: is there a convention to ignore all paths containing target? `*/target/*` ?

.travis.yml
```diff
@@ -25,14 +25,15 @@ matrix:
       script: |
         cargo install clippy
         cargo clippy --all -- -D clippy-pedantic
+        cd coresimd
+        cargo clippy --all -- -D clippy-pedantic
```
gnzlbg: cargo clippy --all` only works for the crates in a workspace, but `coresimd` and `stdsimd` are not in the same workspace yet: https://github.com/rust-lang-nursery/rust-clippy/issues/2238

.travis.yml
```diff
@@ -25,14 +25,15 @@ matrix:
...
   allow_failures:
     - env: RUSTFMT=On TARGET=x86_64-unknown-linux-gnu NO_ADD=1
     - env: CLIPPY=On TARGET=x86_64-unknown-linux-gnu NO_ADD=1
 install:
   - if [ "$NO_ADD" == "" ]; then rustup target add $TARGET; fi
 
 script:
-  - cargo generate-lockfile
```
gnzlbg: `cargo generate-lockfile` only really works if all the crates are in a workspace, so I have added write permissions to the docker runs.

Cargo.toml
```diff
@@ -18,6 +18,9 @@ is-it-maintained-issue-resolution = { repository = "BurntSushi/stdsimd" }
 is-it-maintained-open-issues = { repository = "BurntSushi/stdsimd" }
 maintenance = { status = "experimental" }
 
+[dependencies]
+coresimd = { version = "0.0.3", path = "coresimd/" }
```
gnzlbg: I've set the first version of `coresimd` to `0.0.3` to match that of `stdsimd`, we should release these together.

ci/dox.sh
```diff
@@ -22,7 +22,10 @@ dox() {
   rm -rf target/doc/$arch
   mkdir target/doc/$arch
 
-  rustdoc --target $target -o target/doc/$arch src/lib.rs --crate-name stdsimd
+  cargo clean
+  cargo build --target $target
+
+  rustdoc --target $target -o target/doc/$arch src/lib.rs --crate-name stdsimd --library-path target/$target/debug/deps
```
gnzlbg: Now `rustdoc` needs a build of the `coresimd` to generate the docs for `stdsimd`. I haven't taken a detailed look at the docs, and haven't looked at the output of this at all. `cargo doc` seems to work just fine for `stdsimd`, wrapping the docs of both crates nicely, although some links on `coresimd` are still pointing to `stdsimd`.

ci/run-docker.sh
```diff
@@ -19,7 +19,7 @@ run() {
       --env TARGET=$target \
       --env FEATURES=$2 \
       --env STDSIMD_TEST_EVERYTHING \
-      --volume `pwd`:/checkout:ro \
+      --volume `pwd`:/checkout \
```
gnzlbg: I did not managed to get this to work any other way :( 

The `checkout` directory is not readonly anymore because now building `stdsimd` requires building `coresimd` and the lock files are not generated a priori anymore.

ci/run.sh
```diff
@@ -16,14 +16,14 @@ case ${TARGET} in
...
-cargo test --release --target $TARGET --features $FEATURES --verbose -- --nocapture
-
-cargo test --target $TARGET --features $FEATURES_STD --verbose -- --nocapture
-cargo test --release --target $TARGET --features $FEATURES_STD --verbose -- --nocapture
+cd coresimd
+cargo test --all --target $TARGET --features $FEATURES --verbose -- --nocapture
+cargo test --all --release --target $TARGET --features $FEATURES --verbose -- --nocapture
+cd ..
+cargo test --all --target $TARGET --features $FEATURES --verbose -- --nocapture
+cargo test --all --release --target $TARGET --features $FEATURES --verbose -- --nocapture
```
gnzlbg: The `std` crate feature is gone, and so are the two `$FEATURES_STD` runs. 

However, `cargo test --all` only works for `workspaces`, so we need to run the tests in `coresimd` manually...

coresimd/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+//! * [Overal tracking issue for SIMD support][simd_tracking_issue]
+//! * [`cfg_target_feature` tracking issue][cfg_target_feature_issue]
+//! * [SIMD types currently not sound][simd_soundness_bug]
+//! * [`#[target_feature]` improvements][target_feature_impr]
+//!
+//! [vendor]: https://github.com/rust-lang-nursery/stdsimd/issues/40
+//! [simd_tracking_issue]: https://github.com/rust-lang/rust/issues/27731
+//! [cfg_target_feature_issue]: https://github.com/rust-lang/rust/issues/29717
+//! [simd_soundness_bug]: https://github.com/rust-lang/rust/issues/44367
+//! [target_feature_impr]: https://github.com/rust-lang/rust/issues/44839
```
gnzlbg: These docs still need to be updated to reflect `coresimd` instead of `stdsimd`.

coresimd/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+    #[cfg(target_arch = "aarch64")]
+    pub use aarch64::*;
+
+    pub use nvptx::*;
+
+    #[cfg(
+        // x86/x86_64:
+        any(target_arch = "x86", target_arch = "x86_64")
+    )]
+    pub use runtime::{__unstable_detect_feature, __Feature};
```
gnzlbg: In `coresimd` only `x86`'s run-time is available, `stdsimd` just re-exports it.

coresimd/src/runtime/bit.rs
```diff
@@ -0,0 +1,11 @@
...
+
+/// Sets the `bit` of `x`.
+pub const fn set(x: usize, bit: u32) -> usize {
+    x | 1 << bit
+}
+
+/// Tests the `bit` of `x`.
+pub const fn test(x: usize, bit: u32) -> bool {
+    x & (1 << bit) != 0
+}
```
gnzlbg: This file is duplicated in `coresimd` and `stdsimd`...

coresimd/src/runtime/cache.rs
```diff
@@ -0,0 +1,31 @@
...
+/// PLEASE: do not use this, it is an implementation detail subject to change.
+pub fn test<F>(bit: u32, f: F) -> bool
+where
+    F: FnOnce() -> usize,
+{
+    if CACHE.load(Ordering::Relaxed) == usize::MAX {
+        CACHE.store(f(), Ordering::Relaxed);
+    }
+    bit::test(CACHE.load(Ordering::Relaxed), bit)
+}
```
gnzlbg: This file is duplicated in `coresimd` and `stdsimd`...

coresimd/src/runtime/macros.rs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    #[cfg(target_arch = "x86_64")]
+    #[test]
+    fn test_macros() {
+        assert!(cfg_feature_enabled!("sse"));
+    }
+}
```
gnzlbg: This file is duplicated in `coresimd` and `stdsimd`...

coresimd/src/runtime/x86.rs
```diff
@@ -424,7 +423,6 @@ pub fn detect_features() -> usize {
 
 #[cfg(test)]
 mod tests {
-    #[cfg(feature = "std")]
```
gnzlbg: `std` is always available when running `coresimd`'s tests.

coresimd/src/x86/i586/avx2.rs
```diff
@@ -1257,7 +1257,7 @@ pub unsafe fn _mm256_shuffle_epi8(a: u8x32, b: u8x32) -> u8x32 {
 /// # #![feature(cfg_target_feature)]
 /// # #![feature(target_feature)]
 /// #
-/// # #[macro_use] extern crate stdsimd;
+/// # #[macro_use] extern crate coresimd as stdsimd;
```
gnzlbg: I was a bit lazy here because I don't know the best way to fix the examples so that they show `coresimd` in the docs of `coresimd` and `stdsimd` in the docs of `stdsimd`. So here I just changed as little as possible for now.

coresimd/src/x86/i586/sse.rs
```diff
@@ -1572,7 +1571,7 @@ pub const _MM_HINT_NTA: i8 = 0;
 #[cfg_attr(test, assert_instr(prefetcht1, strategy = _MM_HINT_T1))]
 #[cfg_attr(test, assert_instr(prefetcht2, strategy = _MM_HINT_T2))]
 #[cfg_attr(test, assert_instr(prefetchnta, strategy = _MM_HINT_NTA))]
-pub unsafe fn _mm_prefetch(p: *const c_void, strategy: i8) {
+pub unsafe fn _mm_prefetch(p: *const u8, strategy: i8) {
```
gnzlbg: I don't know why this intrinsic used `c_void`, but [Intel specifies it as taking a `const char*`]( https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=prefetch&expand=4101). 

The intrinsic is also untested, I've filled: https://github.com/rust-lang-nursery/stdsimd/issues/198

coresimd/src/x86/i586/sse2.rs
```diff
@@ -29,7 +28,7 @@ pub unsafe fn _mm_pause() {
 #[inline(always)]
 #[target_feature = "+sse2"]
 #[cfg_attr(test, assert_instr(clflush))]
-pub unsafe fn _mm_clflush(p: *mut c_void) {
+pub unsafe fn _mm_clflush(p: *mut u8) {
```
gnzlbg: Intel specifies that the [`_mm_clflush` intrinsic](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_clflush&expand=4101,657) takes a `const void*`. I don't know why it takes a `*mut c_void`, and am slightly uncomfortable with it taking a `*mut u8` instead. 

The tracking issue to fix this is: https://github.com/rust-lang/rust/issues/36193

I've opened this issue to track the intrinsics in which we break the API because of the dependency on `c_void`. This should block the stabilization of these intrinsics.

https://github.com/rust-lang-nursery/stdsimd/issues/199

.gitignore
alexcrichton: Oh you can just remove the leading `/` on the `/target` line near the top of the file

alexcrichton:
Looks like some appveyor tests may be failing? (I'd recommend removing `--nocapture` from CI to get some better output)

gnzlbg:
@alexcrichton done (btw why does `--nocapture` produce worse output on windows?)

gnzlbg:
So I've removed `--nocapture` but the output looks the same.

alexcrichton:
Oh `--nocapture` isn't windows specific, it, when passed, just means that the otuput of each test is garbled with all the other outputs, but now that it's not passed the output for each test is in its own delineated section.

gnzlbg:
I think that we are testing with only one thread so that shouldn’t matter
that much, but anyways I have no clue why the windows test are failing and
no window machine to test :/

alexcrichton:
So FWIW I just added a `[workspace]` directive to the root `Cargo.toml` and it worked? I wonder if the link error you got was maybe due to a stale target directory? Or maybe I'm running tests differently?

I'll test on Windows

gnzlbg:
The linking errors happened on 32 bit targets on Travis :/

Maybe I just did it wrong.

alexcrichton:
Aha, I've reproduced the 32-bit failure as well

alexcrichton:
Ok I think I've pushed a fix for workspaces, it was I believe just related to disabling tests for the procedural macros

alexcrichton:
That seems to have... magically fixed windows?

So I can't actually run the tests locally on Windows due to a faulting instruction. The test is failing on the `xsaves` test and instruction. The disassembly in visual studio when it faults leads with:

```
00007FF7EBEF990F  mov         edx,0FFFFFFFFh  
00007FF7EBEF9914  mov         eax,0FFFFFFFFh  
00007FF7EBEF9919  xsaves      [rsp+80h]  
```

where the faulting instruction is `xsaves` and RSP is `000000696F3FB580`. So I think that's 64-byte aligned? The reported exception is:

> Unhandled exception at 0x00007FF7EBEF9919 in coresimd-80b6104084dfdbe5.exe: 0xC0000005: Access violation reading location 0xFFFFFFFFFFFFFFFF. occurred

I'm not really sure what's going on there... I wonder if maybe xsaves detection isn't working? I apparently have an i7-7700k, although I'm not sure if it actually supports `xsaves` or not. Do you know what's going on?

gnzlbg:
> So I think that's 64-byte aligned?

Yes.

> I'm not really sure what's going on there... I wonder if maybe xsaves detection isn't working?

I don't think so, we currently only enable `xsaves` if the CPU supports AVX and XSAVES, and if the OS supports it (OSXSAVE). Also, [Appveyor](https://ci.appveyor.com/project/rust-lang-libs/stdsimd/build/1.0.623#L149) says that `xsaves` is not enabled, while on your machine it says that it is enabled, so I think that looks fine.

> Do you know what's going on?

This probably means that the `xsaves`/`xsaves64` tests were always broken. 

Two things point in this direction: 

- The [`--nocapture` results from a previous Appveyor build](https://ci.appveyor.com/project/rust-lang-libs/stdsimd/build/1.0.623#L149) show that the test for `xsaves` returns false on `Appveyor`, and `true` on your machine, so these were not tested before on Windows. 
- I disabled `xsaves` tests on Intel SDE because... I am an idiot, so these are probably not tested on Linux either.

It is weird that you get an error saying that you are trying to read from `0xFFFF...`. That looks surprisingly similar to the `mask` that `XSAVES` uses, but this mask is not a memory address; it is used to specify which registers should be saved, where `0xFFFF...` just means save all registers.

But IMO the simplest explanation is that they were broken already, and that the bug surfaced only on your machine, so can you comment them out and test if the `xsavec` tests work? 

If so, I would move on with this and I'll try to fix the tests this week (I have some machines with `xsaves` enabled).

gnzlbg:
@alexcrichton can you post a `cat /proc/cpuinfo` of that cpu? (not sure how easy it is to do so from windows but maybe you happen to have dual boot?)

gnzlbg:
rebased

alexcrichton:
r=me, but looks like it needs a rebase (feel free to merge after rebasing)

PR #217, opened 2017-11-26
update docs

gnzlbg:
The `cargo clean` command was removing the previously generated docs in `ci/dox.sh` so hopefully after this the links to the architecture specific docs should work again. 

I've also removed duplicated docs between `stdsimd` and `coresimd`, remove all uses of `stdsimd` in `coresimd` examples, and unified the examples to better show how to use the library.

### DIFF
--- a/ci/dox.sh
+++ b/ci/dox.sh
@@ -22,7 +22,6 @@ dox() {
   rm -rf target/doc/$arch
   mkdir target/doc/$arch
 
-  cargo clean
   cargo build --target $target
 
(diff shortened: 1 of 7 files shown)


alexcrichton:
Hm I don't think we want to use `coresimd` all over the place in examples? Sort of like how libcore examples still use libstd. (it's the idiomatic one to use and users of `coresimd` know how to do the translation)

gnzlbg:
@alexcrichton So I think I've fixed that (please check). I've added `stdsimd` as a dev dependency of `coresimd`, and now it seems that we can just use `stdsimd` in the documentation comments. 

I wonder why that wasn't necessary for the other tests.

alexcrichton:
Looks great!

PR #221, opened 2017-11-28
[ci] test i686-apple-darwin

gnzlbg:
`i686-apple-darwin` is a tier 1 target. This PRs adds the target to CI and makes the tests pass.

### DIFF
--- a/coresimd/src/x86/i586/sse2.rs
+++ b/coresimd/src/x86/i586/sse2.rs
@@ -3686,38 +3686,33 @@ mod tests {
         assert_eq!(r, 0b11);
     }
 
+    #[repr(align(16))]
+    struct Memory {
+        data: [f64; 4],
+    }
+
     #[simd_test = "sse2"]
     unsafe fn _mm_load_pd() {
-        let vals = &[1.0f64, 2.0, 3.0, 4.0];
-        let mut d = vals.as_ptr();
-
-        // Align d to 16-byte boundary
-        let mut offset = 0;
-        while (d as usize) & 0xf != 0 {
-            d = d.offset(1 as isize);
-            offset += 1;
-        }
+        let mem = Memory {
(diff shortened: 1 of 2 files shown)


alexcrichton:
Were the test changes here needed to fix i686-apple-darwin for something that didn't show up on i686-unknown-linux-gnu? Unfortunately Travis's OSX offering is *super* slow nowadays so it's best of we can avoid it and I figured that the two 32-bit targets would be basically the same, but if test fixes were required...

coresimd/src/x86/i586/sse2.rs
```diff
@@ -3686,38 +3686,33 @@ mod tests {
...
     #[simd_test = "sse2"]
     unsafe fn _mm_load_pd() {
-        let vals = &[1.0f64, 2.0, 3.0, 4.0];
-        let mut d = vals.as_ptr();
-
-        // Align d to 16-byte boundary
-        let mut offset = 0;
-        while (d as usize) & 0xf != 0 {
-            d = d.offset(1 as isize);
-            offset += 1;
```
gnzlbg: @alexcrichton this makes `offset` overflow on `i686-apple-darwin` but not on `i686-linux-gnu` :/

gnzlbg: librustc_back targets [i686_unknown_linux_gnu](https://github.com/rust-lang/rust/blob/0defa208dcb5a0f3c1ca5bc4df7acf52759ece44/src/librustc_back/target/i686_unknown_linux_gnu.rs) and [i686_apple_darwin](https://github.com/rust-lang/rust/blob/0defa208dcb5a0f3c1ca5bc4df7acf52759ece44/src/librustc_back/target/i686_apple_darwin.rs) look pretty similar. 

The only difference is the data layout.

alexcrichton:
Ah well, if it finds failures it finds failures!

Issue #222, opened 2017-11-28
Adding an `extract_unchecked` method to `define_impl`?


governortarkin:
Is there any opposition to adding an `extract_unchecked` method to the `define_impl` macro? Currently, `extract` is defined like this:
```rust
pub fn extract(self, idx: u32) -> $elemty {
    assert!(idx < $nelems);
    unsafe { simd_extract(self, idx) }
}
```
Here's an [example](https://github.com/rust-lang-nursery/stdsimd/blob/db8c6ac6f4ed734406f6aaf65ec9b4588e0501f3/coresimd/src/x86/i586/sse2.rs#L1422) of the current `extract` method in use.

I'm currently working on a pull request for some incorrect return types in the _mm_extract_epi* intrinsics (which also use `extract`), and everything seems to hold together fine with this addition.

BurntSushi:
Seems fine to me. Might as well add `replace_unchecked` as well.

gnzlbg:
Fine for me as well as long as you make them `unsafe`. 

---

@BurntSushi we should probably add a `::len()` method to the simd types to query their number of lanes as well. 

The `extract`/`replace`/etc. functions are all `#[inline(always)]` so if you are accessing the values with an `idx` known at compile-time, or in a `for i in 0..f32x4::len()` loop (where the result of `len()` is a compile-time constant) the compiler should remove the asserts completely in all cases when enabling optimizations.

governortarkin:
@gnzlbg thanks for that, makes sense. It didn't quite click until I just looked at the C headers again (and saw that `const` keyword) that what you are saying should also apply in the case of #223

gnzlbg:
Closed by #223 .

PR #223, opened 2017-11-29
Add `unchecked` methods, fix _mm_extract_epi* return types

governortarkin:
- Adds `extract_unchecked` + `replace_unchecked` + `len` (#222 )
- Fixes the return types + uses `extract_unchecked` for:
    - _mm_extract_epi8
    - _mm_extract_epi16
    - _mm256_extract_epi8
    - _mm256_extract_epi16
- Minor changes to the other `extract_epi*` intrinsics for style consistency

These should now zero-extend the extracted int and behave appropriately. An old typo makes these a bit confusing, See [this](https://reviews.llvm.org/D20468) llvm issue.

### DIFF
--- a/coresimd/src/macros.rs
+++ b/coresimd/src/macros.rs
@@ -31,6 +31,11 @@ macro_rules! define_impl {
             }
 
             #[inline(always)]
+            pub fn len() -> i32 {
+                $nelems
+            }
+
+            #[inline(always)]
             pub fn splat(value: $elemty) -> $name {
                 $name($({
                     #[allow(non_camel_case_types, dead_code)]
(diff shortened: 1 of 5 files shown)


gnzlbg:
@governortarkin sorry for taking this long, I thought the test were broken but it seems that those failures are spurious! Feel free to ping us directly if this ever happen again!

@governortarkin could you remove the `::len` associated function? I feel that its addition hasn't been discussed enough (e.g. do we really need this, if so, should it be an associated function, associated `const`, etc.). Or how do you feel about that @BurntSushi ? If you are ok with it we can always merge this and change it later.

gnzlbg:
@governortarkin thanks for doing this! 

@BurntSushi I've marked the len issue as a release blocker, we can remove/rename the `len` associated function or change it to an associated const later, but we should do so before the next release.

PR #224, opened 2017-12-02
ssse3

gwenn:
`_mm_alignr_pi8` is still missing due to:
```
Intrinsic has incorrect argument type!
x86_mmx (x86_mmx, x86_mmx, i32)* @llvm.x86.mmx.palignr.b
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,9 @@ matrix:
     - env: TARGET=armv7-unknown-linux-gnueabihf
     - env: TARGET=aarch64-unknown-linux-gnu
     - os: osx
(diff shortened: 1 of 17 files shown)


alexcrichton:
Nice! I wonder if `_mm_abs_pi8` is broken on i586 though?

gwenn:
Does it mean that they are only available on `i686` ?
Like [here](https://github.com/gnzlbg/stdsimd/commit/4ba0256b7a284ef9635e5842950dda435aa0e31f).

alexcrichton:
Thanks!

PR #229, opened 2017-12-10
non-x86 runtime detection updates

lu-zero:
The arm32 variant is not tested due a small hardware failure.

### DIFF
--- a/src/runtime/aarch64.rs
+++ b/src/runtime/aarch64.rs
@@ -32,11 +32,11 @@ pub enum __Feature {
 }
 
 pub fn detect_features<T: linux::FeatureQuery>(mut x: T) -> usize {
-    let value: usize = 0;
+    let mut value: usize = 0;
     {
         let mut enable_feature = |f| {
             if x.has_feature(&f) {
-                bit::set(value, f as u32);
+                value = bit::set(value, f as u32);
             }
         };
         enable_feature(__Feature::asimd);
--- a/src/runtime/aarch64.rs
+++ b/src/runtime/aarch64.rs
@@ -45,6 +45,25 @@ pub fn detect_features<T: linux::FeatureQuery>(mut x: T) -> usize {
     value
 }
 
+/// Probe the ELF Auxiliary vector for hardware capabilities
+///
+/// The values are part of the platform-specific [asm/hwcap.h][hwcap]
+///
+/// [hwcap]: https://github.com/torvalds/linux/blob/master/arch/arm64/include/uapi/asm/hwcap.h
+impl linux::FeatureQuery for linux::AuxVec {
+    fn has_feature(&mut self, x: &__Feature) -> bool {
+        use self::__Feature::*;
+        if let Some(caps) = self.lookup(linux::AT::HWCAP) {
+            match *x {
+                asimd => caps & (1 << 1) != 0,
+                pmull => caps & (1 << 4) != 0,
+            }
+        } else {
+            false
+        }
+    }
+}
+
 impl linux::FeatureQuery for linux::CpuInfo {
     fn has_feature(&mut self, x: &__Feature) -> bool {
         use self::__Feature::*;
--- a/src/runtime/linux/auxvec.rs
+++ b/src/runtime/linux/auxvec.rs
@@ -1,0 +1,89 @@
+//! Reads /proc/self/auxv on Linux systems
+
+use std::prelude::v1::*;
+use std::slice;
+use std::mem;
+
+/// Simple abstraction for the ELF Auxiliary Vector
+///
+/// the elf.h provide the layout of the single entry as auxv_t.
+/// The desugared version is a usize tag followed by a union with
+/// the same storage size.
+///
+/// Cache only the HWCAP and HWCAP2 entries.
+#[derive(Debug)]
+pub struct AuxVec {
+    hwcap: Option<usize>,
+    hwcap2: Option<usize>,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+#[allow(dead_code)]
+/// ELF Auxiliary vector entry types
+///
+/// The entry types are specified in  [linux/auxvec.h][auxvec_h].
+///
+/// [auxvec_h]: https://github.com/torvalds/linux/blob/master/include/uapi/linux/auxvec.h
+pub enum AT {
+    /// CPU Hardware capabilities, it is a bitfield.
+    HWCAP = 16,
+    /// CPU Hardware capabilities, additional bitfield.
+    HWCAP2 = 26,
+}
+
+impl AuxVec {
+    /// Reads the ELF Auxiliary Vector
+    ///
+    /// Try to read `/proc/self/auxv`.
+    // TODO: Make use of getauxval once it is available in a
+    // reliable way.
+    pub fn new() -> Result<Self, ::std::io::Error> {
+        use std::io::Read;
+        let mut file = ::std::fs::File::open("/proc/self/auxv")?;
+        let mut buf = [0usize; 64];
+        let mut raw = unsafe {
+            slice::from_raw_parts_mut(
+                buf.as_mut_ptr() as *mut u8,
+                buf.len() * mem::size_of::<usize>(),
+            )
+        };
+
+        let _ = file.read(&mut raw)?;
+
+        mem::forget(raw);
+
+        let mut auxv = AuxVec { hwcap: None, hwcap2: None };
+
+        for el in buf.chunks(2) {
+            if el[0] == AT::HWCAP as usize {
+                auxv.hwcap = Some(el[1]);
+            }
+            if el[0] == AT::HWCAP2 as usize {
+                auxv.hwcap2 = Some(el[1]);
+            }
+        }
+
+        Ok(auxv)
+    }
+
+    /// Returns the value for the AT key
+    pub fn lookup(&self, key: AT) -> Option<usize> {
+        match key {
+            AT::HWCAP => self.hwcap,
+            AT::HWCAP2 => self.hwcap2,
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[cfg(target_os = "linux")]
+    #[test]
+    fn test_auxvec_linux() {
+        let auxvec = AuxVec::new().unwrap();
+        println!("{:?}", auxvec.lookup(AT::HWCAP));
+        println!("{:?}", auxvec);
+    }
+}
--- a/src/runtime/linux/cpuinfo.rs
+++ b/src/runtime/linux/cpuinfo.rs
@@ -214,4 +214,89 @@ CPU revision	: 1";
         assert!(cpuinfo.field("Features").has("neon"));
         assert!(cpuinfo.field("Features").has("asimd"));
     }
+
+    const POWER8E_POWERKVM: &str = r"processor       : 0
+cpu             : POWER8E (raw), altivec supported
+clock           : 3425.000000MHz
+revision        : 2.1 (pvr 004b 0201)
+
+processor       : 1
+cpu             : POWER8E (raw), altivec supported
+clock           : 3425.000000MHz
+revision        : 2.1 (pvr 004b 0201)
+
+processor       : 2
+cpu             : POWER8E (raw), altivec supported
+clock           : 3425.000000MHz
+revision        : 2.1 (pvr 004b 0201)
+
+processor       : 3
+cpu             : POWER8E (raw), altivec supported
+clock           : 3425.000000MHz
+revision        : 2.1 (pvr 004b 0201)
+
+timebase        : 512000000
+platform        : pSeries
+model           : IBM pSeries (emulated by qemu)
+machine         : CHRP IBM pSeries (emulated by qemu)";
+
+    #[test]
+    fn test_cpuinfo_linux_power8_powerkvm() {
+        let cpuinfo = CpuInfo::from_str(POWER8E_POWERKVM).unwrap();
+        assert_eq!(cpuinfo.field("cpu"), "POWER8E (raw), altivec supported");
+
+        assert!(cpuinfo.field("cpu").has("altivec"));
+    }
+
+    const POWER5P: &str = r"processor       : 0
+cpu             : POWER5+ (gs)
+clock           : 1900.098000MHz
+revision        : 2.1 (pvr 003b 0201)
+
+processor       : 1
+cpu             : POWER5+ (gs)
+clock           : 1900.098000MHz
+revision        : 2.1 (pvr 003b 0201)
+
+processor       : 2
+cpu             : POWER5+ (gs)
+clock           : 1900.098000MHz
+revision        : 2.1 (pvr 003b 0201)
+
+processor       : 3
+cpu             : POWER5+ (gs)
+clock           : 1900.098000MHz
+revision        : 2.1 (pvr 003b 0201)
+
+processor       : 4
+cpu             : POWER5+ (gs)
+clock           : 1900.098000MHz
+revision        : 2.1 (pvr 003b 0201)
+
+processor       : 5
+cpu             : POWER5+ (gs)
+clock           : 1900.098000MHz
+revision        : 2.1 (pvr 003b 0201)
+
+processor       : 6
+cpu             : POWER5+ (gs)
+clock           : 1900.098000MHz
+revision        : 2.1 (pvr 003b 0201)
+
+processor       : 7
+cpu             : POWER5+ (gs)
+clock           : 1900.098000MHz
+revision        : 2.1 (pvr 003b 0201)
+
+timebase        : 237331000
+platform        : pSeries
+machine         : CHRP IBM,9133-55A";
+
+    #[test]
+    fn test_cpuinfo_linux_power5p() {
+        let cpuinfo = CpuInfo::from_str(POWER5P).unwrap();
+        assert_eq!(cpuinfo.field("cpu"), "POWER5+ (gs)");
+
+        assert!(!cpuinfo.field("cpu").has("altivec"));
+    }
 }
--- a/src/runtime/linux/mod.rs
+++ b/src/runtime/linux/mod.rs
@@ -2,6 +2,9 @@
 mod cpuinfo;
 pub use self::cpuinfo::CpuInfo;
 
+mod auxvec;
+pub use self::auxvec::*;
+
 use super::__Feature;
 
 pub trait FeatureQuery {
@@ -19,9 +22,12 @@ fn detect_features_impl<T: FeatureQuery>(x: T) -> usize {
     }
 }
 
-/// Detects ARM features:
+/// Detects CPU features:
 pub fn detect_features() -> usize {
-    // FIXME: use libc::getauxval, and if that fails /proc/auxv
+    // Try to read the ELF Auxiliary Vector
+    if let Ok(v) = auxvec::AuxVec::new() {
+        return detect_features_impl(v);
+    }
     // Try to read /proc/cpuinfo
     if let Ok(v) = cpuinfo::CpuInfo::new() {
         return detect_features_impl(v);
--- a/src/runtime/linux/mod.rs
+++ b/src/runtime/linux/mod.rs
@@ -20,6 +20,10 @@ fn detect_features_impl<T: FeatureQuery>(x: T) -> usize {
     {
         super::aarch64::detect_features(x)
     }
+    #[cfg(target_arch = "powerpc64")]
+    {
+        super::powerpc64::detect_features(x)
+    }
 }
 
 /// Detects CPU features:
--- a/src/runtime/powerpc64.rs
+++ b/src/runtime/powerpc64.rs
@@ -1,0 +1,94 @@
+//! Run-time feature detection on PowerPC64.
+use super::{bit, linux};
+
+#[macro_export]
+#[doc(hidden)]
+macro_rules! __unstable_detect_feature {
+    ("altivec") => {
+        $crate::vendor::__unstable_detect_feature($crate::vendor::__Feature::altivec{})
+    };
+    ("vsx") => {
+        $crate::vendor::__unstable_detect_feature($crate::vendor::__Feature::vsx{})
+    };
+    ("power8") => {
+        $crate::vendor::__unstable_detect_feature($crate::vendor::__Feature::power8{})
+    };
+    ($t:tt) => { compile_error!(concat!("unknown PowerPC target feature: ", $t)) };
+}
+
+/// PowerPC CPU Feature enum. Each variant denotes a position in a bitset
+/// for a particular feature.
+///
+/// PLEASE: do not use this, it is an implementation detail subject to change.
+#[doc(hidden)]
+#[allow(non_camel_case_types)]
+#[repr(u8)]
+pub enum __Feature {
+    /// Altivec
+    altivec,
+    /// VSX
+    vsx,
+    /// Power8
+    power8,
+}
+
+pub fn detect_features<T: linux::FeatureQuery>(mut x: T) -> usize {
+    let mut value: usize = 0;
+    {
+        let mut enable_feature = |f| {
+            if x.has_feature(&f) {
+                value = bit::set(value, f as u32);
+            }
+        };
+        enable_feature(__Feature::altivec);
+        enable_feature(__Feature::vsx);
+        enable_feature(__Feature::power8);
+    }
+    value
+}
+
+/// Probe the ELF Auxiliary vector for hardware capabilities
+///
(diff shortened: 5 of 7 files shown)


alexcrichton:
This looks great to me, thanks!

Mind adding some more documentation as well? For example for a newbie like me, I'm not even sure what `auxv` is :(. It'd be great if that file explained a little bit about the purpose of the file, the format, and then maybe provided a link to more info?

Additionally for the bits defined in powerpc64, could documentation be referenced for where the values come from?

src/runtime/aarch64.rs
```diff
@@ -32,11 +32,11 @@ pub enum __Feature {
 }
 
 pub fn detect_features<T: linux::FeatureQuery>(mut x: T) -> usize {
-    let value: usize = 0;
+    let mut value: usize = 0;
     {
         let mut enable_feature = |f| {
             if x.has_feature(&f) {
-                bit::set(value, f as u32);
+                value = bit::set(value, f as u32);
```
gnzlbg: thanks for fixing this! The x86 run-time did not have this issue, I wish there was an easy way to unify this code. https://github.com/rust-lang-nursery/stdsimd/blob/master/coresimd/src/runtime/x86.rs#L329

src/runtime/linux/auxvec.rs
```diff
@@ -0,0 +1,86 @@
...
+            .finish()
+    }
+}
+
+impl AuxVec {
+    /// Reads /proc/self/auxv into AuxVec.
+    // TODO: make it less ugly if possible
+    pub fn new() -> Result<Self, ::std::io::Error> {
+        use std::io::Read;
+        let mut file = ::std::fs::File::open("/proc/self/auxv")?;
```
gnzlbg: This is a good first step, but we should only do this if libc's `getauxval` does not exist. I think we can add `getauxval` as a `weak_symbol`, and check whether the value of a function pointer to it is null or not at run-time. If it is not null we should just call that, and if it is null, then the binary does not have a linked extern C `getauxval`, and we can proceed to read `proc/self/auxv`.

I wish someone (@alexcrichton , @BurntSushi ) would review #211 since that would allow us to add some more ARM targets for testing and use qemu-system instead of qemu-user to test this on linux and android targets.

lu-zero: I avoided `getauxval` since it is nonstandard, I wouldn't rely on weak_symbol to check for its presence (static linking and non-gnu libcs aren't uncommon:))

src/runtime/aarch64.rs
```diff
@@ -45,6 +45,23 @@ pub fn detect_features<T: linux::FeatureQuery>(mut x: T) -> usize {
     value
 }
 
+/// Probe the ELF Auxiliary vector for hardware capabilities
+///
+/// The values are part of the platform-specific `asm/hwcap.h`
```
gnzlbg: Could you maybe add a link to the linux header where hwcap is defined? That is, here: https://github.com/torvalds/linux/blob/master/arch/arm64/include/uapi/asm/hwcap.h

src/runtime/linux/auxvec.rs
alexcrichton: Typically we don't use weak symbols as they're not very portable, but a solution is to do what libstd does which is `dlsym` at runtime.

src/runtime/linux/auxvec.rs
```diff
@@ -0,0 +1,86 @@
...
+use std::mem;
+use std::fmt;
+
+/// Simple abstraction for the ELF Auxiliary Vector
+///
+/// the elf.h provide the layout of the single entry as auxv_t.
+/// The desugared version is a usize tag followed by a union with
+/// the same storage size.
+pub struct AuxVec {
+    raw: Vec<usize>,
```
gnzlbg: I think we should avoid the `Vec` here. Can't we just use a `[usize; 64]` + a length?

src/runtime/linux/mod.rs
```diff
@@ -17,11 +20,18 @@ fn detect_features_impl<T: FeatureQuery>(x: T) -> usize {
...
     }
+    #[cfg(target_arch = "powerpc64")]
+    {
+        super::powerpc64::detect_features(x)
+    }
 }
 
 /// Detects ARM features:
 pub fn detect_features() -> usize {
-    // FIXME: use libc::getauxval, and if that fails /proc/auxv
```
gnzlbg: We should keep this fixme because we are still not calling `libc::getauxval`. We should try to do that first, and if that fails, then read `/proc/self/auxv`, and if that fails, then try to read `/proc/cpuinfo`.

src/runtime/powerpc64.rs
```diff
@@ -0,0 +1,92 @@
...
+            altivec => self.field("Features").has("altivec"),
+            vsx => self.field("Features").has("vsx"),
+            power8 => self.field("Features").has("arch_2_07"),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
```
gnzlbg: Can you add a test that parses one or two dumps of `/proc/cpuinfo` for different CPUs?

src/runtime/linux/auxvec.rs
```diff
@@ -0,0 +1,86 @@
...
+            }
+        }
+        None
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
```
gnzlbg: I don't know if its easily possible but we should try to refactor this a bit so that we can test it by adding one or two dumps of auxv for arm cpus.

gnzlbg:
Looks very good, thanks! I've left some comments in the code, but they are all minor issues.

src/runtime/linux/auxvec.rs
gnzlbg: IIRC the problem was that `/proc/self/auxv` is not guaranteed to exist on all Android targets, but `libc::getauxval` is (and also, `libc::getauxval` is always faster than reading `/proc/self/auxv` on all platforms that have it). 

Anyways, we don't need to add `libc::getauxval` on this PR, that was more me thinking out loud. This PR is a good step forward as it is. Good job!

src/runtime/powerpc64.rs
lu-zero: No, since the cpuinfo is w/out information at least for now.

src/runtime/aarch64.rs
lu-zero: Sure.

src/runtime/linux/auxvec.rs
lu-zero: I could just copy the hwcaps when present maybe?

src/runtime/powerpc64.rs
gnzlbg: Check this out: https://github.com/randombit/cpuinfo/tree/master/ppc

That's a collection of `cpuinfo` files for a lot of powerpc cpus. Maybe one could c&p one of those files here and add it as a test?

src/runtime/aarch64.rs
gnzlbg: Thanks!

src/runtime/linux/auxvec.rs
lu-zero: I do not have an arm system handy right now, could somebody else do that?

src/runtime/linux/auxvec.rs
gnzlbg: I think that would be better. 

I know we are already doing a lot of syscals by reading the file, and that allocating the vector should not matter much, but I was hoping that if we managed to add support for `libc::getauxval` in the future we could reuse at least some of this for that.

src/runtime/linux/auxvec.rs
gnzlbg: Yeah just open an issue and tag it here. Otherwise you can also try googling for dumps of auxv, maybe you get lucky :D

src/runtime/powerpc64.rs
lu-zero: I'm afraid the comment is not clear in the function: `/proc/cpuinfo` on powerpc does not provide useful information, I left that code there since I plan to ask the linux kernel developers to provide the same information as arm does. I can change the function with a simple return false so it is less misleading for not.

gnzlbg: The comment is clear, I just skimmed too fast through it. 

What is the value of adding the cpuinfo implementation? We can always add it later once a kernel is released that "supports" it right?

lu-zero: I just prefer to leave the placeholder since it is sort of future proof.

src/runtime/linux/auxvec.rs
lu-zero: Actually it won't work since the auxv entry size is system dependent so the synthetic 32bit tests would fail on 64bit platforms.

gnzlbg: We could use `#[cfg(target_pointer_width = "32")]` to enable the tests on 32-bit platforms or to add different tests for 32 and 64 bit platforms.

src/runtime/powerpc64.rs
gnzlbg: Does the `linux` mod require this implementation to exist? If that is the case I think we should fix that instead. 

I can do that later if you prefer, but I think that leaving it with the expectation that a future Linux kernel version implements this is a long shot.

lu-zero: The `linux` mod requires the implementation and I wouldn't spend more time on that. I updated it so it leverages the small bit of information to get to know if at least 1 feature is enabled.

src/runtime/linux/auxvec.rs
lu-zero: I would not go this direction.

src/runtime/powerpc64.rs
gnzlbg: Now that the examples of cpuinfo for powerpc are there everything has become clear (sorry that it took so long). Just to check that I understand correctly, the only thing one can be checked from `/proc/cpuinfo` for PowerPC is whether `altivec` is supported or not, is that right?

src/runtime/linux/cpuinfo.rs
```diff
@@ -214,4 +214,37 @@ CPU revision	: 1";
...
+model           : IBM pSeries (emulated by qemu)
+machine         : CHRP IBM pSeries (emulated by qemu)";
+
+    #[test]
+    fn test_cpuinfo_linux_power8_powerkvm() {
+        let cpuinfo = CpuInfo::from_str(POWER8E_POWERKVM).unwrap();
+        assert_eq!(cpuinfo.field("cpu"), "POWER8E (raw), altivec supported");
+
+        assert!(cpuinfo.field("cpu").has("altivec"));
+    }
```
gnzlbg: This looks great, thanks! Could you add a second test for a CPU without altivec, like for example this one: https://github.com/randombit/cpuinfo/blob/master/ppc/power5 ?

gnzlbg: @lu-zero it looks like the Power5 test got lost in the merge?

gnzlbg:
I've opened #234 to track testing auxv. Modulo the Power5 tests that got lost in the merge this LGTM.

lu-zero:
I updated patch to add it, I hope it doesn't cause additional disruption

gnzlbg:
@lu-zero thank you for working on this! great job!

src/runtime/linux/auxvec.rs
marshallpierce: IIRC 64 bit solaris still uses 32 bit auxv, but how many solaris users are out there anymore...

marshallpierce: BTW, since I don't run into a whole lot of people who know/care about auxval, is there anything I could have done in https://crates.io/crates/auxv to make it more likely that you would use that instead of re-rolling your own?

src/runtime/linux/auxvec.rs
```diff
@@ -0,0 +1,89 @@
...
+impl AuxVec {
+    /// Reads the ELF Auxiliary Vector
+    ///
+    /// Try to read `/proc/self/auxv`.
+    // TODO: Make use of getauxval once it is available in a
+    // reliable way.
+    pub fn new() -> Result<Self, ::std::io::Error> {
+        use std::io::Read;
+        let mut file = ::std::fs::File::open("/proc/self/auxv")?;
```
marshallpierce: Apparently it is common on Android for this file to not be readable (permissions wise), but `getauxval()` is widely available because bionic (their libc) has implemented it for some years now.

src/runtime/linux/auxvec.rs
lu-zero: I wasn't aware of that crate otherwise I'd use it assuming stdsimd is fine with external crates.

Does it survive static linking when the libc does not provide the function? From a cursory look I guess I'd add some more cfg guards and/or ask for a cfg item for the binary layout.

the stack crawler is brilliant btw :)

marshallpierce: Well, to be fair, who would expect someone to make a crate for such an esoteric thing... ;) The stack crawler is pretty heinous, but it _does work_, mostly... 

I'm not sure what [`__attribute__((weak))`](https://bitbucket.org/marshallpierce/rust-auxv/src/d52f19ab9d73bf379aa0eb439c842574b11f71d8/c/getauxval-wrapper.c?at=master&fileviewer=file-view-default#getauxval-wrapper.c-6) does when static linking. [This person](https://stackoverflow.com/questions/2627839/weak-linking-with-static-libraries) seems to have not had a great time with it. Is there a specific test case you had in mind? I'd be curious to see what happens in practice.

lu-zero: You might try to see what happens when you build a native static library with musl (the use-case I have in mind is building VLC, we want to support more rust and the build-glue is... peculiar).

The stack crawler would be the best solution beside the "tiny" problem of being quite unsafe.

gnzlbg: @marshallpierce

> IIRC 64 bit solaris still uses 32 bit auxv, but how many solaris users are out there anymore...

64 bit solaris is a tier 2 platform, but even though the rest of core/std is not tested against it, external core/std components are in a position to test beyond tier 1 if they want to. 

Please fill an issue in this repo so that we don't forget this. 

---

@marshallpierce 

> BTW, since I don't run into a whole lot of people who know/care about auxval, is there anything I could have done in https://crates.io/crates/auxv to make it more likely that you would use that instead of re-rolling your own?

I honestly don't think there is anything you could have done. This is just my opinion, but in a nutshell:

- `coresimd`/`stdsimd` are intended to be part of `core`/`std` so right now AFAIK they cannot technically have any external dependencies (if these libraries weren't intended to be part of the `std` library, I think we would have used `cupid` and `auxv` directly, at least initially)
- no other language run-time that I know of aims to offer as comprehensive and safe "feature/simd" support as Rust, which means we don't have any place to draw inspiration from which makes it hard to tell a priori whether a dependency fulfills our needs. 

For example, @shepmaster also wishes that we would reuse the `cupid` crate, but back then we did not know exactly whether it would fulfill our needs (and we couldn't directly use it even if we wanted). In hindsight, now that the run-time feature detection support for `x86/x86_64` is finished, it is much easier to discuss whether we should be using `cupid` internally or not, but back then it was a big unknown. The same applies to the `auxv` crate right now. Maybe when the whole `auxv` implementation is finished we realize "actually, we could just use the `auxv` crate for this" and try to somehow position doing that somewhere in the roadmap, but right now this is a question that we don't have enough data to answer. 

This doesn't mean that there isn't anything here for the `auxv` crate though. For example, we use `cupid` as a dev dependency, which has resulted in some bug reports, bug fixes, etc. landing on `cupid` upstream. I might not be the easiest person to work with, but one of the next things we should do is add the `auxv` crate as a dev dependency and check that our results match those of the `auxv` crate.

gnzlbg: @lu-zero @marshallpierce I still think that `getauxval`, when available, should be our preferred way of testing features on ARM at least because of performance, availability on Android, and... because IIUC `getauxval()` does not require a `std` dependency. 

That is, if we manage to implement `getauxval()` on `coresimd`, at least on Linux, all `#[no_std]` crates would have as good run-time feature support as `x86/x86_64` on ARM/PowerPC/...

marshallpierce: That all makes sense. If I can find any documentation (or reproduce it in a VM) for Solaris's odd behavior, I'll file an issue. I might just be mis-remembering other Solaris oddities.

marshallpierce: another nice thing about `getauxval()` is that there are no syscalls; it's purely userspace. (The libc implementation is basically the same as my stack-walking code, except that they can skip the environment vars reliably and I cannot.)

gnzlbg: Are linux binaries guaranteed to have an `auxv` ? Or would we still need the fall back to try to read `/proc/self/auxv` and `/proc/cpuinfo` ? 

If by using your stack-walking code we could reliably read `auxv` on Linux without a libc dependency we might be able to drop the `std` component completely which would simplify many things.

marshallpierce: ELF binaries have an `auxv`, but I'm not sure about non-ELF things on Linux. http://phrack.org/issues/58/5.html has more info on the ELF runtime layout.

I don't think the stack walking code is suitable for a library because it couldn't necessarily guarantee that nobody has changed an environment variable, for instance. However, calling `getauxval()` (when available) is safe, and I don't see why it would require anything from the stdlib. 

`getauxval()` is available in glibc 2.16 (2012), bionic since Android 4.3 (also 2012), and musl 1.1.0 (2014). Even RHEL, champion of ancient versions, has glibc 2.17 in RHEL7. Of the target market (people using rust with simd on linux), that probably covers just about everybody, right?

lu-zero: As long it is happening in the main thread and nothing plays with that memory it is quite reliable. Just it requires that nobody is changing the environment while you are walking the stack from what I recall.

marshallpierce: @lu-zero is there a good way to get started with rust on musl? rustup-init links against glibc, which makes it hard to get going on, say, alpine linux.

lu-zero: you could unpack somewhere a Gentoo stage3 and you should have everything needed more or less.

lu-zero: [this](https://github.com/korhalio/stasis) and [this](https://github.com/japaric/steed) could be the most contrived corner cases we should support decently IMHO.

gnzlbg: I have a branch that implements getauxval in core (https://github.com/gnzlbg/stdsimd/tree/core_auxv) and should support those use cases, the C file is only compiled and linked if the target is linux. 

@marshallpierce I've stolen the C header file from the auxv crate (the one I had before was a naive implementation without all the glibc workarounds and once I saw yours I could not unsee it) but that might introduce some licensing issues. Any chances you could also license the crate under the MIT and Apache licenses? All the code in Rust stdlib must be dual licensed under MIT/Apache so that users can choose what suits them best. Otherwise I'll switch it back to the naive weak linking implementation and we can figure out how to improve it in a license-compatible way in the future.

BTW reviews welcome.

gnzlbg: I've submitted a PR to make review easier #242.

marshallpierce: Any sufficiently good hack should be slightly bothersome, right? ;) I have no problem dual (triple) licensing. The existing license is already less restrictive than ASL2 and basically equivalent to MIT. If I put a note in the LICENSE.txt, will that appease the licensing rules? (License haggles are probably my least favorite part of open source work.)

BTW, stack crawling is broken even if nothing is concurrently modifying the environment. See the [elf_stack_broken_modified_environ](https://bitbucket.org/marshallpierce/rust-auxv/src/d52f19ab9d73bf379aa0eb439c842574b11f71d8/examples/elf_stack_broken_modified_environ.rs?at=master&fileviewer=file-view-default) example, which produces output like this for me on glibc 2.25:

```
Auxv before modifying environ:
33      140725387726848
16      3219913727
6       4096
17      100
3       94512514588736
4       56
5       10
7       140372545884160
8       0
9       94512514617760
11      1000
12      1000
13      1000
14      1000
23      0
25      140725386234953
31      140725386244032
15      140725386234969
Setting new env var, which sets environ to new array
Auxv after
```

IIRC, `setenv` changes the stack in ways we can't navigate around after the fact, but of course the libc implementation is free to hold on to the pointer straight to aux for use in `getauxval()` so that will still work.

gnzlbg: > If I put a note in the LICENSE.txt, will that appease the licensing rules? (License haggles are probably my least favorite part of open source work.)

IANAL but I think that adding [LICENSE_MIT](https://github.com/rust-lang-nursery/stdsimd/blob/master/LICENSE-MIT) and [LICENSE-APACHE](https://github.com/rust-lang-nursery/stdsimd/blob/master/LICENSE-APACHE) files to your repo should and mentioning that the library is triple licensed somewhere should be enough.

> BTW, stack crawling is broken even if nothing is concurrently modifying the environment. 

I don't understand how this works in general, so can't really follow why this breaks at all.

In my mental model, the arguments to the binary, environment of the shell, and auxiliary vector sections are all setup on binary initialization, and are contiguously allocated next to each other at the top of the stack. 

`argc` gives you the number of arguments, so one is able to directly jump to the environment variables section. Something like `argc` for environment variables must exist, so that one can jump to the auxiliary vector sections. 

How does adding/removing environment variables work? In my mental model adding environment variables must add them somewhere else, because otherwise if you add too many then at some point adding more would need to push all the stack down, right? Or in other words, in my mental model the section were the auxiliary vector starts cannot change.

Since my mental model doesn't fit the issue you are describing, I assume that it is probably wrong. Do you have a link that explains this?

marshallpierce: You're right about the stack layout: nothing will ever change that the stack has args, env, and auxv one after the other. However, once you call `setenv` with a new environment var, glibc detects that it doesn't have an entry in the stack-resident table that it can update, so it reallocs the whole env var table. Once it does that, `environ` now points somewhere far away, not into the stack, and now we can't navigate from it down the stack to auxv. However, libc conveniently keeps a pointer to the stack auxv, which is how `getauxval()` continues to work even after monkeying with env vars. 

For the gory details, you can check `stdlib/setenv.c` in a [glibc tarball](https://mirrors.syringanetworks.net/gnu/libc/).

marshallpierce: I added the other two licenses to `rust-auxv` so that should be good to go.

lu-zero: I guess we could "just" save the auxv correct pointer on the `init()` call that is used to save the argv/argc.

At that time you are 100% sure that the layout is stable and the pointer to argv is always valid.
then you walk your way twice and you have the auxv for our purposes.

basically my idea boils down to merge an extended version of your walker in `sys::args`.

gnzlbg: Note that then for this to work in `core` we would still be relying on `getauxval` existing, but I guess that at that point we could just weakly-link `std` from core and use that instead.

lu-zero: My idea is to avoid `getauxval` completely and just walk from argv in `imp::init()` and save a `GLOBAL_AUXV_PTR` there. then wrap the whole thing as it is done for args.

marshallpierce: Before we go to the extent of bypassing `getauxval()` and doing our own `unsafe` version, I think it would make sense to evaluate whether or not there is actually a problem with `getauxval()` availability in practice. Are people going to be writing Rust code that uses SIMD that gets deployed on very old Android devices? I suspect that that is likely to be a marginal use case at most.

lu-zero: I assume that the `getauxval()` support is sorted out already and I looked at how hard would be avoiding it  to support the fringe cases I mentioned.

Apparently it would be within the 50-lines range.

gnzlbg: @lu-zero I thought that the `init` call was only available in `std`, is it also usable from `core`?

---

@marshallpierce once `getauxval` lands things will be in my opinion very good for Linux users and my time is better spent on improving things for `macos` users (macos x86 users are fine already, but `iOS` users on ARM chips are not). Having said this, if someone wants to spearhead support for libc's without `getauxval` on `coresimd` I'd completely support that (for `steed` users it might be better to just implement `getauxval` on `steed` though).

gnzlbg: @lu-zero if you want to pursue this a good way to get initial feedback is to fill an issue with your thoughts on the matter, we are kind of mixing many different discussions in this PR, and while many are subscribed to the issues in this repo, most of them are probably not being notified about any of this.

PR #232, opened 2017-12-11
make __m64/__m128i/__m256i types

### LINKED ISSUE #214: x86 __m128, __m256, __m512 rename / retype ?
Currently the `_m128i` and `_m256i` are [just type aliases for `i8x16` and `i8x32`](https://github.com/rust-lang-nursery/stdsimd/blob/f3ee983f44088064ca55ec9c00eb5be21da645ae/coresimd/src/x86/mod.rs#L29):

```rust
/// 128-bit wide signed integer vector type
#[allow(non_camel_case_types)]
pub type __m128i = ::v128::i8x16;
/// 256-bit wide signed integer vector type
#[allow(non_camel_case_types)]
pub type __m256i = ::v256::i8x32;
```

These are the places that are currently using them (unless my grep-fu failed, please correct me if I am missing some):

- Comparing strings where the input is reinterpreted as 1byte.8bytes, ... depending on a flag:
  - [`_mm_cmpistrm`](https://github.com/rust-lang-nursery/stdsimd/blob/f3ee983f44088064ca55ec9c00eb5be21da645ae/coresimd/src/x86/i586/sse42.rs#L54): Compare packed strings
  - `_mm_cmpestri`: Compare packed strings
  - ... most of the intrinsics in the sse42 module.

- Shift bytes:
  - [`_mm_slli_si128`](https://github.com/rust-lang-nursery/stdsimd/blob/f3ee983f44088064ca55ec9c00eb5be21da645ae/coresimd/src/x86/i586/sse2.rs#L318): Shift `a` left by `imm8` bytes while shifting in zeros.
  - `_mm_bslli_si128`: Shift `a` left by `imm8` bytes while shifting in zeros.
  - `_mm_bsrli_si128`:  Shift `a` right by `imm8` bytes while shifting in zeros.
  - `_mm_srli_si128`: Shift `a` right by `imm8` bytes while shifting in zeros.

- As raw bits `i1x128` / `i1x256`:
  - [`_mm_testz_si128`](https://github.com/rust-lang-nursery/stdsimd/blob/f3ee983f44088064ca55ec9c00eb5be21da645ae/coresimd/src/x86/i686/sse41.rs#L59): Tests whether the specified bits in a 128-bit integer vector are all zeros.
  - [`_mm256_and_si256`](https://github.com/rust-lang-nursery/stdsimd/blob/f3ee983f44088064ca55ec9c00eb5be21da645ae/coresimd/src/x86/i586/avx2.rs#L188): Compute the bitwise AND of 256 bits (representing integer data) in `a` and `b`.
  - [`_mm256_extractf128_si256`](https://github.com/rust-lang-nursery/stdsimd/blob/f3ee983f44088064ca55ec9c00eb5be21da645ae/coresimd/src/x86/i586/avx.rs#L859): Extract 128 bits (composed of integer data) from `a`, selected with `imm8`.
  - [`_mm_lddqu_si128`](https://github.com/rust-lang-nursery/stdsimd/blob/f3ee983f44088064ca55ec9c00eb5be21da645ae/coresimd/src/x86/i586/sse3.rs#L70): Load 128-bits of integer data from unaligned memory.

~~Switching `__mm_cmpistrm` to use bytes explicitly, and rename `__m128i`/`__m256i` to `i1x128`/`i1x256` _might_ make the library more consistent. I am not sure. I also haven't used `__mm_cmpistrm` before so maybe I am wrong to think that bytes there make sense.~~ It makes sense for these functions to operate on generic `__m128i` types because how these are interpreted by the function depends on the function arguments passed. The shift byte functions should, however, operate on bytes.  

Whether we can implement an `i1x128` `repr(simd)` type in current Rust is a different topic (LLVM supports this, but we might need to "tweak" rustc to make it work).

Thoughts?

--- 

This discussion spawned in #212 .

gnzlbg:
This PR makes `__m{64, 128i, 256i}` types instead of aliases and allows us to change the `repr` of these types without breaking client code. It also means that the only things users can do with these types is convert from/into them from other types (no `splat`, no operators, no `extract`, `replace`, `store`, etc.).

It also changes two places where the `__m`/`i_x_` types were previously used but the other types make more sense:
-  The `_mm_slli_si128`,`_mm_bslli_si128`,`_mm_bsrli_si128`, `_mm_srli_si128` intrinsics shift bytes (so it makes sense for them to take `i8x16` as an argument).
- whether the result of `_mm_cmpestrm` works on bits, bytes, `u16`, ... depends on the `mode` argument, so returning a `__m128i` makes more sense than an `u8x16` (also for consistency with the other string manipulation intrinsics which already do this). 

This partially fixes #214 . The only part of that issue that remains open is whether we should rename these types to something else (e.g. `i1x64`,`i1x128`, `i1x256`), but there is no consensus about this yet and that should be done in a different PR anyways.

---

This PR also:
- documents intrinsics that do not correspond to any instruction or should not generate any instructions: when going through the code some intrinsics were missing the `assert_instr` macro and I did not know if that was on purpose or an oversight, so I had to check them all. We were already doing this for some intrinsics but should probably try to be more disciplined about it.
- remove unnecessary `mem::uninitialized`/`mem::transmute`, probably from the times where we did not have the intrinsics to create uninitialized vectors or `from/into`.


gnzlbg:
`rustfmt` is broken-ish (https://github.com/rust-lang-nursery/rustfmt/issues/2260 fix hasn't landed yet?) and the latest upstream changes introduce a lot of tiny differences all over the place so I think it would be better to reformat the library after this and #229 are merged to avoid any conflicts.

BurntSushi:
Looks great to me! Thanks for doing this!

Issue #215, opened 2017-11-24
`.into()` is very slow


quininer:
* rustc 1.23.0-nightly (246a6d19c 2017-11-23)
* CPU: Intel i7-6700HQ (8)

I am trying to [use `coresimd` in `gimli-permutation`](https://github.com/quininer/gimli/blob/5f512d1cbb5c38a915ee8758be436754f3f19be3/gimli-permutation/src/sse.rs#L19), but it very slow.

![screenshot_20171124_153814](https://user-images.githubusercontent.com/6286761/33199956-85b5dfe8-d12d-11e7-8fcc-f64db3dacc96.png)

I found a lot of `Into` is not optimized.

```rust
            x = _mm_shuffle_epi8(
                x.into(),
                _mm_set_epi8(12, 15, 14, 13, 8, 11, 10, 9, 4, 7, 6, 5, 0, 3, 2, 1).into()
            ).into();
            y = (shift(y.into(), 9) | _mm_srli_epi32(y.into(), 32 - 9)).into();
            let newz = x.as_i32x4() ^ shift(z.into(), 1)    ^ shift((y & z).into(), 2);
            let newy = y            ^ x                     ^ shift((x | z).into(), 1).into();
            x = z ^ y ^ shift((x & y).into(), 3).into();
```

![index](https://user-images.githubusercontent.com/6286761/33199998-bab0456c-d12d-11e7-8c60-84bc915268b9.png)

I don't know if this should be a rust issue, so I report it here.

BurntSushi:
Please provide a reproducible example that we can easily compile and test. Please also tell us how you're compiling your code. For example, are you using codegen units or incremental compilation? Are you compiling in release mode?

gnzlbg:
Since those are `cargo bench` results I assume that is `release` mode but it would be good to know if you have a custom `bench` profile in your `Cargo.toml`. 

@BurntSushi I will try to send a PR with the bench to `stdsimd` so that we can either add it or easily look at it, but I am fully booked till sunday.

@quininer how urgent is this?

quininer:
@BurntSushi Yes, I use incremental compilation.

@gnzlbg It is less urgent, I currently use `transmute` instead of `Into`.

gnzlbg:
@quininer have you tested with the `.as_` methods?

alexcrichton:
[...]

PR #233, opened 2017-12-11
add test for into

gnzlbg:
This test using `.into()` fails on ~~32 bits~~ `i586-unknown-linux-gnu` (it segfaults), but it should pass. 

Using `from` instead of `.into()` does not segfault. 

It might be related to #215, @alexcrichton ? Is `into` getting inlined a pure performance issue or is it required for correctness?


alexcrichton:
Yeah unfortunately I think this is just https://github.com/rust-lang-nursery/stdsimd/issues/215#issuecomment-346856620,  a method like this won't work until upstream rustc has fixed `#[target_feature]` support to insert "shims" when necessary.

gnzlbg:
I don't know why this PR was still open  😅

PR #242, opened 2017-12-15
More run-time detection improvements

gnzlbg:
* Implement run-time detection in Linux targets in `coresimd` using `getauxval` for non-`x86` targets.
* The run-time detection on Linux in `stdsimd` for non-x86 targets now tries `getauxval`, and if that fails tries to read `/proc/self/auxv`, and if that fails tries to read `/proc/cpuinfo`. This approach is the best we can do on Linux / Android for all architectures.
* Reuse the `coresimd` run-time detection implementation in `stdsimd` without duplicating code (e.g. the `cache` is reused). The runtime components are now in `__vendor_runtime` instead of in `vendor` because of this. 
* Use the `auxv` crate to validate the results of the `auxv` implementations (the `getauxval` in `core`, and the `/proc/self/auxv` in `std`)
* Adds synthetic tests for `/proc/self/auxv` 
* Use the `simd_test` macro for ARM and Aarch64 tests.


gnzlbg:
The `DOCUMENTATION` build fails because we need more clever cross-compilation to generate the docs or work around the lack of it.

gnzlbg:
@alexcrichton do you have a suggestion about how to fix the `DOCUMENTATION` build bot?

---

@lu-zero I'd like your feedback on this as well.

Cargo.toml
```diff
@@ -23,6 +23,9 @@ maintenance = { status = "experimental" }
 [dependencies]
 coresimd = { version = "0.0.3", path = "coresimd/" }
 
+[dev-dependencies]
+auxv = "0.3.3"
```
gnzlbg: We use the `auxv` crate in `coresimd` to test `getauxval` and in `stdsimd` to test reading from `/proc/self/auxv`.

coresimd/Cargo.toml
```diff
@@ -18,7 +18,11 @@ is-it-maintained-issue-resolution = { repository = "BurntSushi/stdsimd" }
 is-it-maintained-open-issues = { repository = "BurntSushi/stdsimd" }
 maintenance = { status = "experimental" }
 
+[build-dependencies]
+cc = "1.0"
```
gnzlbg: We now need the `cc` crate to compile the `getauxval` wrapper on linux systems.

coresimd/src/aarch64/neon.rs
```diff
@@ -39,41 +41,43 @@ pub unsafe fn vaddd_u64(a: u64, b: u64) -> u64 {
...
 #[cfg(test)]
 mod tests {
-    use super::*;
+    use super::f64x2;
+    use aarch64::neon;
+    use stdsimd_test::simd_test;
 
-    #[test]
-    fn vadd_f64_() {
+    #[simd_test = "neon"]
```
gnzlbg: The `simd_test` macro now works for `arm`-like targets and it should also work for powerpc and mips as well.

coresimd/src/lib.rs
```diff
@@ -56,18 +56,34 @@ pub mod vendor {
...
+    // the `coresimd` one, but it is not exactly the same. These
+    // re-exports are only intended to be used by the `stdsimd`
+    // run-time but not re-exported by it.
+    #[cfg(all(target_os = "linux",
+              any(target_arch = "arm", target_arch = "aarch64",
+                  target_arch = "powerpc64")))]
+    #[doc(hidden)]
+    pub mod __runtime {
+        pub use runtime::*;
+    }
```
gnzlbg: I don't know if this is a good way to reuse code. It makes some items public for `core` users that wouldn't be public otherwise :/

coresimd/src/runtime/linux/auxv.rs
```diff
@@ -0,0 +1,140 @@
...
+
+/// Cache HWCAP bitfields of the ELF Auxiliary Vector.
+///
+/// If an entry cannot be read all the bits in the bitfield
+/// are set to zero.
+#[cfg(target_arch = "aarch64")]
+#[derive(Debug, Copy, Clone)]
+pub struct AuxVec {
+    pub hwcap: usize,
+}
```
gnzlbg: I've chosen to have different `AuxVec` structs depending on the architecture to only read what's really needed and avoid the `Option`s.

coresimd/src/runtime/linux/auxv.rs
```diff
@@ -0,0 +1,140 @@
...
+    pub fn auxv() -> Result<AuxVec, ()> {
+        if let Ok(hwcap) = getauxval(AT_HWCAP) {
+            #[cfg(target_arch = "aarch64")]
+            {
+                return Ok(AuxVec { hwcap });
+            }
+            #[cfg(any(target_arch = "arm", target_arch = "powerpc64"))]
+            {
+                if let Ok(hwcap2) = getauxval(AT_HWCAP2) {
+                    return Ok(AuxVec { hwcap, hwcap2 });
```
gnzlbg: This is a bit different from what @lu-zero . Now, either reading both fields succeeds, or the whole read fails. This allows that, in case reading one of them succeeds but reading the other fails, we can still try to read `/proc/self/auxv` or `/proc/cpuinfo` to obtain the rest of the information.

Basically either `getauxval` works 100% reliably, or we fall back to doing something slower but reliable.

src/lib.rs
```diff
@@ -123,34 +123,38 @@
 //! [simd_soundness_bug]: https://github.com/rust-lang/rust/issues/44367
 //! [target_feature_impr]: https://github.com/rust-lang/rust/issues/44839
 
-#![feature(macro_reexport, const_fn, const_atomic_usize_new)]
+#![feature(const_fn, const_size_of, use_extern_macros, cfg_target_feature)]
```
gnzlbg: I moved the crate from `macro_reexport` to `use_extern_macros` here because `macro_reexport` was behaving weirdly when trying to reuse the macros in `stdsimd`; `use_extern_macros` seems to work way better.

src/runtime/linux/auxv.rs
```diff
@@ -0,0 +1,142 @@
...
+        let mut buf = [0usize; 64];
+        {
+            let raw: &mut [u8; 64 * mem::size_of::<usize>()] =
+                unsafe { mem::transmute(&mut buf) };
+            file.read(raw).or_else(|_| Err(()))?;
+        }
+        auxv_from_buf(&buf)
+    }
+
+    fn auxv_from_buf(buf: &[usize; 64]) -> Result<AuxVec, ()> {
```
gnzlbg: I've refactor `auxv` into these three functions for testing purposes (see below). It allows testing the `auxv` functionality by reading a file, using `include_bytes`, or just adding some `auxv` arrays to the test module.

src/runtime/linux/auxv.rs
```diff
@@ -0,0 +1,142 @@
...
+        #[cfg(target_arch = "aarch64")]
+        {
+            for el in buf.chunks(2) {
+                match el[0] {
+                    AT_HWCAP => return Ok(AuxVec { hwcap: el[1] }),
+                    _ => (),
+                }
+            }
+            return Ok(AuxVec { hwcap: 0 });
```
gnzlbg: Here and below I return `Ok` even if the values are not read property, I think this is a bug and that this should return `Err` so that the fallback to reading from `/proc/cpuinfo` triggers.

src/runtime/mod.rs
```diff
@@ -1,40 +1,51 @@
...
+    use coresimd::*;
+
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    #[test]
+    fn detect_feature() {
+        println!("avx {}", cfg_feature_enabled!("avx"));
+    }
+
+    #[cfg(target_arch = "arm")]
+    #[test]
```
gnzlbg: FIXME: This test and the ones below should be executed on `Linux` only.

coresimd/src/runtime/linux/auxv.rs
lu-zero: I guess when we'd want to support what's provided in hwcap2 will be easy enough to extend/collate the structures :)

gnzlbg:
Yes, I just kept this with the minimal stuff we currently need. But if we
want to offer sha2 and whatnot in the future (which we probably will) we
might be able to collate them in some arch’s at least.

alexcrichton:
In general crates like libcore can't pick up a dependency on `cc`, so I think we should avoid libc or C code in coresimd. Could that support stay in stdsimd where libc may actually be linked? I also continue to think it's ok to not bother sharing something like `cache`, isn't it just setting bits?

gnzlbg:
@alexcrichton 

> In general crates like libcore can't pick up a dependency on cc,

Dependency removed. This issue is unfortunate, though (there are some things that are more easily done in C than Rust).

> so I think we should avoid libc or C code in coresimd.
>
>  Could that support stay in stdsimd where libc may actually be linked?

I've switched from using C to a Rust only implementation using `linkage = "extern_weak"`. 

I can implement "better" support for `getauxval` in `stdsimd` later, but since `libc` might or might not be linked, we would need to use weak symbols there anyways (I'd just re-export them from core).

> I also continue to think it's ok to not bother sharing something like cache, isn't it just setting bits?

There are other things that are shared (the features supported in each architecture, the way to detect these features, the `bits` utilities, the struct layouts for auxiliary vectors, the `getauxval` functionality, ...). While we can reimplement all of that in `stdsimd` (with the cost of keeping it in sync), the `cache` is something that we shouldn't reimplement because:

- every binary would then have two caches, one in `coresimd` and one in `stdsimd` (it's just a single `usize` right now, so this isn't a big deal), however,
- these two caches will need to be initialized, so we need to run initialization twice (slightly bigger deal, but not tragic), and
- these two caches will store different values (e.g. `stdsimd` support is more accurate) which might lead to weird situations: this worries me the most.

So I would really prefer to have one single cache that produces consistent results. This PR doesn't get us there, but there are a couple of ways to achieve it. I wanted to open an issue about this but the easiest way would be for `coresimd` to use `stdsimd` to initialize the cache when `stdsimd` is linked, which is something we can do with weak symbols.

src/runtime/linux/auxv.rs
gnzlbg: This is fixed now.

src/runtime/mod.rs
gnzlbg: This is also fixed.

alexcrichton:
Hm yeah let's still avoid weak symbols in `core`, I think it's best to have this support in stdsimd. Weak linkage is a funny thing that I barely understand how it works, but in general I've found things like `dlsym` to be far more reliable, but that's of course only available in libstd. In general shouldn't coresimd (without stdsimd) only be used in a context *without* something like getauxval?

In terms of sharing I'm sort of assuming the auxval bits aren't actually needed to be shared (they'd live in stdsimd) and then the only sharing would basically be the `|` or `<<` operators, which I don't think really need sharing.

gnzlbg:
>  In general shouldn't coresimd (without stdsimd) only be used in a context without something like getauxval?

I don't know, is it? IIUC `coresimd` will be used from `#[no_std]` crates, which orthogonal to whether the final binary links to `libc`, `std`, or not. 

That is,  if I have a generic `#[no_std]` crate A that uses `coresimd`, and I use it from my other crate B, that uses `std` and `libc`. When `B` calls `A` and `A` does run-time feature detection, I want the better run-time feature detection from `std` trigger.

I think achieving that is a problem worth solving. But anyways, even if we solve it, just because `std` and `libc` are linked does not mean that `getauxval` is available, because some `libc`'s do not have a `getauxval` function. 

I haven't looked into `dlsym`, but this is a problem that we would need to solve in `stdsimd`'s `getauxval` implementation anyways.

> Hm yeah let's still avoid weak symbols in core, I think it's best to have this support in stdsimd. Weak linkage is a funny thing that I barely understand how it works, but in general I've found things like dlsym to be far more reliable, but that's of course only available in libstd. 

I'll move `getauxval` to `stdsimd`, but the auxval bits need to remain in `coresimd` because we still want to read the auxiliary vector from `coresimd` and we need the general bits for that. 

Every Linux ELF binary (independently of whether it links `std` or not) has an auxiliary vector that is set up on binary initialization. I think that every binary that has it should be able to read it, so the next step after this PR is to implement something like the stack crawler of the `auxv` crate but in a reliable way in the `coresimd` run-time.

> In terms of sharing I'm sort of assuming the auxval bits aren't actually needed to be shared (they'd live in stdsimd) and then the only sharing would basically be the | or << operators, which I don't think really need sharing.

Given the file structure in this PR, after we move `getauxval` to `stdsimd`, the only thing in `coresimd/runtime` that doesn't have to be shared/duplicated/re-exported is `coresimd/runtime/linux` (which, once we move `getauxval` to `stdsimd`, becomes empty). Everything else in `coresimd/runtime` must be either re-exported (for `x86`) or shared/duplicated (for other archs). This includes:

- the feature macros for each target (in `coresimd/runtime/${ARCH}.rs`), 
- how to detect features for each target (the `detect_feature` in `coresimd/runtime/${ARCH}.rs`),
- the cache
- the auxval types and auxval parsing functionality (we need this for stack-crawling in `coresimd`)

The `bit` setting/testing utilities are two one liners that don't change over time and we don't need to share between the two libraries. But the rest will evolve over time, needs to be kept consistent between both libraries, etc.

gnzlbg:
So I won't have much time to work on this until January probably. If somebody wants to pick it up please feel free to do so.

alexcrichton:
I'd be ok with leaving some bits in coresimd if it helps, but I don't think we'll want to make stack crawlers in coresimd? I don't really understand what's going on with auxval bits though, something like 'stack crawler' just sounds very error prone...

I'm all for code deduplication but I don't think we should bend over backwards for it. For example there's a crazy amount of duplication in the libc crate but it's all thoroughly verified. The portability here I feel should come before deduplication, in that we should get things working and then look for opportunities to share code later, but all in all I don't feel like anything here is complicated enough to warrant we-can't-land-this-without-sharing-code

gnzlbg:
> I don't really understand what's going on with auxval bits though, something like 'stack crawler' just sounds very error prone...

I'ts not really that hard once you get it. The only thing you need to know to understand how it works is the layout of the stack of a Linux process of an ELF binary after initialization. From the top of the start to the bottom of the stack: 

- stack pointer (on Linux the stack grows up, the stack frame of main/start will be at a higher address than this); the following addresses are bellow the stack pointer
- how many cmd line arguments: argc (fixed width)
- the cmd line arguments: argv[0...argc] (each argv has the same fixed width)
- the environment variables: env[0...last environment variable]; these are actually pointers to the environment strings, so each pointer has a fixed width, the last entry is a null pointer and indicates that there are no more environment variables,
- the auxiliary vector: auxv[0..last auxv] (the last auxiliary vector has a key equal to `AT_NULL`)
- some padding, other stuff, end marker, and finally, the bottom of the stack.

So stack crawler is just a scary term for the code that searches the auxiliary vector within the stack frame. The stack crawler in the `auxv` crate finds the auxiliary vector by getting a pointer to the first environment variable (which is a pointer to a string), and then just walking the stack down till it finds NULL, and incrementing the pointer once more gives you the beginning of the auxiliary vector. That's it, once you are there, you can iterate the vector using generic code that doesn't care where the vector comes from. In particular, the location of the auxiliary vector never changes, so once you have a pointer to its beginning, you can just store it and never have to look for it again.

Easy right? Well, I omitted a crucial bit of information: how to obtain the first environment variable at the bottom of a process' stack: you can't, or more precisely, you can only doing in `start` before `main`, which is what libc does (and btw, is also what libc does for `getauxval`). Why? Because the process can add more variables to the environment but there is no place for these in the bottom of the stack, so they are then relocated somewhere else, and if you try to read past the last one you are accessing heap memory out of bounds (which is undefined behavior).

So the way to implement `getauxval` in `coresimd` would be to find the beginning of the auxiliary vector in `start` and save that somewhere so that the `coresimd` run-time doesn't need to do any stack crawling at all.

gnzlbg:
@alexcrichton care to re-review? I finally managed to fix this this week. I moved `getauxval` to `stdsimd` (from `coresimd`), and also ended up fixing some error messages with the `cfg_enable_feature!` macro on platforms where we don't support run-time feature detection yet:
 previously it complained that `__Feature` or `__unstable_detect_feature` were missing but now it should fail nicely with `compile_fail!`.

Maybe we could set up some compile fail tests for this.

gnzlbg:
What I haven't done yet is using `dlsym`, will look into that next.

alexcrichton:
@gnzlbg was getauxval and business gonna move to stdsimd? It looks like the files are still in coresimd?

gnzlbg:
I forgot to remove the `getauxval-wrapper.c` file from `coresimd` (was unused) and the `auxv` crate dev dependency (was unused as well).

All the auxval detection is done right now in `stdsimd`, but the `AuxVec` type is still part of `coresimd` so that we can at least try to implement the "stack crawler" approach there. If that turns out to be impossible / harder than expected, we can always move the `AuxVec` bits to `stdsimd` later.

alexcrichton:
Looks good to me, thanks for the updates!

PR #244, opened 2017-12-17
sse2: implements last remaining intrinsics

gwenn:
sse2 is complete.

### DIFF
--- a/coresimd/src/x86/i686/sse2.rs
+++ b/coresimd/src/x86/i686/sse2.rs
@@ -196,6 +196,14 @@ mod tests {
     }
 
     #[simd_test = "sse2"]
+    unsafe fn _mm_cvtpi32_pd() {
+        let a = i32x2::new(1, 2);
+        let expected = f64x2::new(1., 2.);
+        let r = sse2::_mm_cvtpi32_pd(a);
+        assert_eq!(r, expected);
+    }
+
+    #[simd_test = "sse2"]
     unsafe fn _mm_cvtsi64_si128() {
         let r = sse2::_mm_cvtsi64_si128(5);
         assert_eq!(r, i64x2::new(5, 0));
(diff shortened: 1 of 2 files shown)


gwenn:
I don't know why `_mm_loadr_pd` is failing on Travis. It is not impacted by this PR !
I am going to fix `_mm_store_pd` on Windows.

gnzlbg:
@gwenn this is failing on 32 bit builds: 

```shell
test x86::i586::sse2::tests::_mm_loadr_pd ... thread 'x86::i586::sse2::tests::_mm_loadr_pd' panicked at 'attempt to add with overflow', coresimd/src/x86/i586/sse2.rs:4055:13
FAILED
```

Very similar code used to break the `i686-apple-darwin` (incrementing the `offset` in the `while` loops overflowed). IIRC the way I fixed it was by using `repr(align)` (see [here](https://github.com/gwenn/stdsimd/blob/ad12c27023e523144ec6910faf27dff203e5364d/coresimd/src/x86/i586/sse2.rs#L3881) ).

We should probably remove all similar looking code from the test suite.

gwenn:
I hope I will understand when I will see your patch how I made `_mm_loadr_pd` failed without modifying it.

gnzlbg:
@gwenn You did not break it, it was already broken, but the issue was not surfacing before. Either a new version of the compiler broke it (there was a change related to `simd` and `repr(transparent)` merged this week), or adding new tests/functions to the module did :/

gwenn:
@gnzlbg Thanks for your explanation.

gnzlbg:
@gwenn I've submitted a PR (#245) that _might_ fix this.

gnzlbg:
@gwenn I've merged the test fixes, maybe the fix the issues here (I've restarted the build but it might be picking the old master commit so you might have to rebase).

gnzlbg:
Now things look way better, the `i586`/`i686-unknown-linux-gnu` builds fail with:

```shell
x86::sse2::assert__mm_shuffle_pd_shufpd' panicked at 'too many instructions in the disassembly', stdsimd-test/src/lib.rs:322:8
```

You probably want to enable the `assert_instr` of `__mm_shuffle_pd` only for `x86_64` platforms but the tests should be green afterwards.

gwenn:
Sorry,
I don't have any windows machine so I can't investigate.

coresimd/src/x86/i686/sse2.rs
```diff
@@ -111,17 +184,53 @@ extern "C" {
...
     use stdsimd_test::simd_test;
 
+    use core::mem;
     use v128::*;
-    use v64::i32x2;
+    use v64::{__m64, i32x2, u32x2};
     use x86::i686::sse2;
 
+    #[simd_test = "sse2"]
+    unsafe fn _mm_add_si64() {
```
alexcrichton: It looks like this and the two tests below it are causing the failures on Windows, could a `#[cfg]` be added with a FIXME perhaps?

gnzlbg:
This is all green :)

PR #251, opened 2017-12-22
Verify Intel intrinsics against upstream definitions

### LINKED ISSUE #240: Find inconsistencies between the intel intrinsics XML file and the Rust code
XSLT (data.xsl):
```xml
<?xml version="1.0"?>
<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
    <xsl:output method="text" omit-xml-declaration="yes"/>
 
    <xsl:variable name="line">
        <xsl:text>&#10;</xsl:text>
    </xsl:variable>
    <xsl:variable name="sep">
        <xsl:text>;</xsl:text>
    </xsl:variable>
 
    <xsl:template match="/">
        <xsl:for-each select="intrinsics_list/intrinsic">
            <xsl:variable name="name" select="@name"/>
            <xsl:variable name="cpuid" select="CPUID"/>
            <xsl:variable name="instruction" select="instruction/@name"/>
            <xsl:value-of select="concat($name, $sep, $cpuid, $sep, $instruction, $line)"/>
        </xsl:for-each>
    </xsl:template>
</xsl:stylesheet>
```
And a script:
```sh
# Extract instrinsics from XML file
xsltproc data.xsl intel-intrinsics-3.3.15.xml > intel-intrinsics-3.3.15.csv
# Load intel intrinsics into an SQLite db
rm -f intrinsics.db
sqlite3 -batch intrinsics.db 2> /dev/null <<EOF
CREATE TABLE intrinsics(name TEXT NOT NULL UNIQUE, cpuid TEXT, instruction TEXT);
.separator ';'
.import intel-intrinsics-3.3.15.csv intrinsics
EOF

# Extract implemented intrinsics from Rust files
for cpuid in abm avx avx2 bmi bmi2 sse sse2 sse3 sse41 sse42 ssse3; do

rm -f $cpuid.csv
find coresimd/ -type f -name $cpuid.rs -print0 | xargs -0 grep -h '^#\[cfg_attr(.*assert_instr(\|pub unsafe fn' | sed -e 's/^#\[cfg_attr(.*assert_instr(\([^),]*\)[,)].*$/\1/' | sed -e 's/^pub unsafe fn \([^(]*\)(.*$/\1/' | awk '/^_/{print $0";"p; p=""}!/^_/{p=$0}' > $cpuid.csv
# Load them the SQLite db
sqlite3 -batch intrinsics.db <<EOF
CREATE TABLE $cpuid(name TEXT NOT NULL UNIQUE, instruction TEXT);
.separator ';'
.import $cpuid.csv $cpuid
-- find inconsistencies (or missing implementation)
select i.name, i.cpuid, i.instruction, s.instruction
from intrinsics i
left outer join $cpuid s on s.name = i.name
where i.cpuid like '$cpuid'
and i.instruction <> s.instruction
order by 1;
EOF

done
```

alexcrichton:
This commit adds a new crate for testing that the intrinsics listed in this
crate do indeed match the upstream definition of each intrinsic. A
pre-downloaded XML description of all Intel intrinsics is checked in which is
then parsed in the `stdsimd-verify` crate to verify that everything we write
down is matched against the upstream definitions.

Currently the checks are pretty loose to get this compiling but a few intrinsics
were fixed as a result of this. For example:

* `_mm256_extract_epi8` - AVX2 intrinsic erroneously listed under AVX
* `_mm256_extract_epi16` - AVX2 intrinsic erroneously listed under AVX
* `_mm256_extract_epi32` - AVX2 intrinsic erroneously listed under AVX
* `_mm256_extract_epi64` - AVX2 intrinsic erroneously listed under AVX
* `_mm_tzcnt_32` - erroneously had `u32` in the name
* `_mm_tzcnt_64` - erroneously had `u64` in the name
* `_mm_cvtsi64_si128` - erroneously available on 32-bit platforms
* `_mm_cvtsi64x_si128` - erroneously available on 32-bit platforms
* `_mm_cvtsi128_si64` - erroneously available on 32-bit platforms
* `_mm_cvtsi128_si64x` - erroneously available on 32-bit platforms
* `_mm_extract_epi64` - erroneously available on 32-bit platforms
* `_mm_insert_epi64` - erroneously available on 32-bit platforms
* `_mm256_extract_epi16` - erroneously returned i32 instead of i16
* `_mm256_extract_epi8` - erroneously returned i32 instead of i8
* `_mm_shuffle_ps` - the mask argument was erroneously i32 instead of u32
* `_popcnt32` - the signededness of the argument and return were flipped
* `_popcnt64` - the signededness of the argument was flipped and the argument
  was too large bit-wise
* `_mm_tzcnt_32` - the return value's sign was flipped
* `_mm_tzcnt_64` - the return value's sign was flipped
* A good number of intrinsics used `imm8: i8` or `imm8: u8` instead of `imm8:
  i32` which Intel was using. (we were also internally inconsistent)
* A number of intrinsics working with `__m64` were instead working with i64/u64,
  so they're now corrected to operate with the vector types instead.

Currently the verifications performed are:

* Each name in Rust is defined in the XML document
* The arguments/return values all agree.
* The CPUID features listed in the XML document are all enabled in Rust as well.

The type matching right now is pretty loose and has a lot of questionable
changes. Future commits will touch these up to be more strict and require closer
adherence with Intel's own types. Otherwise types like `i32x8` (or any integers
with 256 bits) all match up to `__m256i` right now, althoguh this may want to
change in the future.

Finally we're also not testing the instruction listed in the XML right now.
There's a huge number of discrepancies between the instruction listed in the XML
and the instruction listed in `assert_instr`, and those'll need to be taken care
of in a future commit.

Closes #240

### DIFF
--- a/coresimd/src/x86/i586/abm.rs
+++ b/coresimd/src/x86/i586/abm.rs
@@ -44,16 +44,16 @@ pub unsafe fn _lzcnt_u64(x: u64) -> u64 {
 #[inline(always)]
 #[target_feature = "+popcnt"]
 #[cfg_attr(test, assert_instr(popcnt))]
-pub unsafe fn _popcnt32(x: u32) -> u32 {
-    x.count_ones()
+pub unsafe fn _popcnt32(x: i32) -> i32 {
+    x.count_ones() as i32
 }
 
 /// Counts the bits that are set.
 #[inline(always)]
 #[target_feature = "+popcnt"]
 #[cfg_attr(test, assert_instr(popcnt))]
-pub unsafe fn _popcnt64(x: u64) -> u64 {
-    x.count_ones() as u64
+pub unsafe fn _popcnt64(x: i64) -> i32 {
+    x.count_ones() as i32
 }
 
 #[cfg(test)]
@@ -64,21 +64,21 @@ mod tests {
 
     #[simd_test = "lzcnt"]
     unsafe fn _lzcnt_u32() {
-        assert_eq!(abm::_lzcnt_u32(0b0101_1010u32), 25u32);
+        assert_eq!(abm::_lzcnt_u32(0b0101_1010), 25);
     }
 
     #[simd_test = "lzcnt"]
     unsafe fn _lzcnt_u64() {
-        assert_eq!(abm::_lzcnt_u64(0b0101_1010u64), 57u64);
+        assert_eq!(abm::_lzcnt_u64(0b0101_1010), 57);
     }
 
     #[simd_test = "popcnt"]
     unsafe fn _popcnt32() {
-        assert_eq!(abm::_popcnt32(0b0101_1010u32), 4);
+        assert_eq!(abm::_popcnt32(0b0101_1010), 4);
     }
 
     #[simd_test = "popcnt"]
     unsafe fn _popcnt64() {
-        assert_eq!(abm::_popcnt64(0b0101_1010u64), 4);
+        assert_eq!(abm::_popcnt64(0b0101_1010), 4);
     }
 }
--- a/coresimd/src/x86/i586/avx.rs
+++ b/coresimd/src/x86/i586/avx.rs
@@ -607,69 +607,69 @@ pub unsafe fn _mm256_xor_ps(a: f32x8, b: f32x8) -> f32x8 {
 }
 
 /// Equal (ordered, non-signaling)
-pub const _CMP_EQ_OQ: u8 = 0x00;
+pub const _CMP_EQ_OQ: i32 = 0x00;
 /// Less-than (ordered, signaling)
-pub const _CMP_LT_OS: u8 = 0x01;
+pub const _CMP_LT_OS: i32 = 0x01;
 /// Less-than-or-equal (ordered, signaling)
-pub const _CMP_LE_OS: u8 = 0x02;
+pub const _CMP_LE_OS: i32 = 0x02;
 /// Unordered (non-signaling)
-pub const _CMP_UNORD_Q: u8 = 0x03;
+pub const _CMP_UNORD_Q: i32 = 0x03;
 /// Not-equal (unordered, non-signaling)
-pub const _CMP_NEQ_UQ: u8 = 0x04;
+pub const _CMP_NEQ_UQ: i32 = 0x04;
 /// Not-less-than (unordered, signaling)
-pub const _CMP_NLT_US: u8 = 0x05;
+pub const _CMP_NLT_US: i32 = 0x05;
 /// Not-less-than-or-equal (unordered, signaling)
-pub const _CMP_NLE_US: u8 = 0x06;
+pub const _CMP_NLE_US: i32 = 0x06;
 /// Ordered (non-signaling)
-pub const _CMP_ORD_Q: u8 = 0x07;
+pub const _CMP_ORD_Q: i32 = 0x07;
 /// Equal (unordered, non-signaling)
-pub const _CMP_EQ_UQ: u8 = 0x08;
+pub const _CMP_EQ_UQ: i32 = 0x08;
 /// Not-greater-than-or-equal (unordered, signaling)
-pub const _CMP_NGE_US: u8 = 0x09;
+pub const _CMP_NGE_US: i32 = 0x09;
 /// Not-greater-than (unordered, signaling)
-pub const _CMP_NGT_US: u8 = 0x0a;
+pub const _CMP_NGT_US: i32 = 0x0a;
 /// False (ordered, non-signaling)
-pub const _CMP_FALSE_OQ: u8 = 0x0b;
+pub const _CMP_FALSE_OQ: i32 = 0x0b;
 /// Not-equal (ordered, non-signaling)
-pub const _CMP_NEQ_OQ: u8 = 0x0c;
+pub const _CMP_NEQ_OQ: i32 = 0x0c;
 /// Greater-than-or-equal (ordered, signaling)
-pub const _CMP_GE_OS: u8 = 0x0d;
+pub const _CMP_GE_OS: i32 = 0x0d;
 /// Greater-than (ordered, signaling)
-pub const _CMP_GT_OS: u8 = 0x0e;
+pub const _CMP_GT_OS: i32 = 0x0e;
 /// True (unordered, non-signaling)
-pub const _CMP_TRUE_UQ: u8 = 0x0f;
+pub const _CMP_TRUE_UQ: i32 = 0x0f;
 /// Equal (ordered, signaling)
-pub const _CMP_EQ_OS: u8 = 0x10;
+pub const _CMP_EQ_OS: i32 = 0x10;
 /// Less-than (ordered, non-signaling)
-pub const _CMP_LT_OQ: u8 = 0x11;
+pub const _CMP_LT_OQ: i32 = 0x11;
 /// Less-than-or-equal (ordered, non-signaling)
-pub const _CMP_LE_OQ: u8 = 0x12;
+pub const _CMP_LE_OQ: i32 = 0x12;
 /// Unordered (signaling)
-pub const _CMP_UNORD_S: u8 = 0x13;
+pub const _CMP_UNORD_S: i32 = 0x13;
 /// Not-equal (unordered, signaling)
-pub const _CMP_NEQ_US: u8 = 0x14;
+pub const _CMP_NEQ_US: i32 = 0x14;
 /// Not-less-than (unordered, non-signaling)
-pub const _CMP_NLT_UQ: u8 = 0x15;
+pub const _CMP_NLT_UQ: i32 = 0x15;
 /// Not-less-than-or-equal (unordered, non-signaling)
-pub const _CMP_NLE_UQ: u8 = 0x16;
+pub const _CMP_NLE_UQ: i32 = 0x16;
 /// Ordered (signaling)
-pub const _CMP_ORD_S: u8 = 0x17;
+pub const _CMP_ORD_S: i32 = 0x17;
 /// Equal (unordered, signaling)
-pub const _CMP_EQ_US: u8 = 0x18;
+pub const _CMP_EQ_US: i32 = 0x18;
 /// Not-greater-than-or-equal (unordered, non-signaling)
-pub const _CMP_NGE_UQ: u8 = 0x19;
+pub const _CMP_NGE_UQ: i32 = 0x19;
 /// Not-greater-than (unordered, non-signaling)
-pub const _CMP_NGT_UQ: u8 = 0x1a;
+pub const _CMP_NGT_UQ: i32 = 0x1a;
 /// False (ordered, signaling)
-pub const _CMP_FALSE_OS: u8 = 0x1b;
+pub const _CMP_FALSE_OS: i32 = 0x1b;
 /// Not-equal (ordered, signaling)
-pub const _CMP_NEQ_OS: u8 = 0x1c;
+pub const _CMP_NEQ_OS: i32 = 0x1c;
 /// Greater-than-or-equal (ordered, non-signaling)
-pub const _CMP_GE_OQ: u8 = 0x1d;
+pub const _CMP_GE_OQ: i32 = 0x1d;
 /// Greater-than (ordered, non-signaling)
-pub const _CMP_GT_OQ: u8 = 0x1e;
+pub const _CMP_GT_OQ: i32 = 0x1e;
 /// True (unordered, signaling)
-pub const _CMP_TRUE_US: u8 = 0x1f;
+pub const _CMP_TRUE_US: i32 = 0x1f;
 
 /// Compare packed double-precision (64-bit) floating-point
 /// elements in `a` and `b` based on the comparison operand
@@ -677,7 +677,7 @@ pub const _CMP_TRUE_US: u8 = 0x1f;
 #[inline(always)]
 #[target_feature = "+avx,+sse2"]
 #[cfg_attr(test, assert_instr(vcmpeqpd, imm8 = 0))] // TODO Validate vcmppd
-pub unsafe fn _mm_cmp_pd(a: f64x2, b: f64x2, imm8: u8) -> f64x2 {
+pub unsafe fn _mm_cmp_pd(a: f64x2, b: f64x2, imm8: i32) -> f64x2 {
     macro_rules! call {
         ($imm8:expr) => { vcmppd(a, b, $imm8) }
     }
@@ -690,7 +690,7 @@ pub unsafe fn _mm_cmp_pd(a: f64x2, b: f64x2, imm8: u8) -> f64x2 {
 #[inline(always)]
 #[target_feature = "+avx"]
 #[cfg_attr(test, assert_instr(vcmpeqpd, imm8 = 0))] // TODO Validate vcmppd
-pub unsafe fn _mm256_cmp_pd(a: f64x4, b: f64x4, imm8: u8) -> f64x4 {
+pub unsafe fn _mm256_cmp_pd(a: f64x4, b: f64x4, imm8: i32) -> f64x4 {
     macro_rules! call {
         ($imm8:expr) => { vcmppd256(a, b, $imm8) }
     }
@@ -703,7 +703,7 @@ pub unsafe fn _mm256_cmp_pd(a: f64x4, b: f64x4, imm8: u8) -> f64x4 {
 #[inline(always)]
 #[target_feature = "+avx,+sse"]
 #[cfg_attr(test, assert_instr(vcmpeqps, imm8 = 0))] // TODO Validate vcmpps
-pub unsafe fn _mm_cmp_ps(a: f32x4, b: f32x4, imm8: u8) -> f32x4 {
+pub unsafe fn _mm_cmp_ps(a: f32x4, b: f32x4, imm8: i32) -> f32x4 {
     macro_rules! call {
         ($imm8:expr) => { vcmpps(a, b, $imm8) }
     }
@@ -716,7 +716,7 @@ pub unsafe fn _mm_cmp_ps(a: f32x4, b: f32x4, imm8: u8) -> f32x4 {
 #[inline(always)]
 #[target_feature = "+avx"]
 #[cfg_attr(test, assert_instr(vcmpeqps, imm8 = 0))] // TODO Validate vcmpps
-pub unsafe fn _mm256_cmp_ps(a: f32x8, b: f32x8, imm8: u8) -> f32x8 {
+pub unsafe fn _mm256_cmp_ps(a: f32x8, b: f32x8, imm8: i32) -> f32x8 {
     macro_rules! call {
         ($imm8:expr) => { vcmpps256(a, b, $imm8) }
     }
@@ -731,7 +731,7 @@ pub unsafe fn _mm256_cmp_ps(a: f32x8, b: f32x8, imm8: u8) -> f32x8 {
 #[inline(always)]
 #[target_feature = "+avx,+sse2"]
 #[cfg_attr(test, assert_instr(vcmpeqsd, imm8 = 0))] // TODO Validate vcmpsd
-pub unsafe fn _mm_cmp_sd(a: f64x2, b: f64x2, imm8: u8) -> f64x2 {
+pub unsafe fn _mm_cmp_sd(a: f64x2, b: f64x2, imm8: i32) -> f64x2 {
     macro_rules! call {
         ($imm8:expr) => { vcmpsd(a, b, $imm8) }
     }
@@ -746,7 +746,7 @@ pub unsafe fn _mm_cmp_sd(a: f64x2, b: f64x2, imm8: u8) -> f64x2 {
 #[inline(always)]
 #[target_feature = "+avx,+sse"]
 #[cfg_attr(test, assert_instr(vcmpeqss, imm8 = 0))] // TODO Validate vcmpss
-pub unsafe fn _mm_cmp_ss(a: f32x4, b: f32x4, imm8: u8) -> f32x4 {
+pub unsafe fn _mm_cmp_ss(a: f32x4, b: f32x4, imm8: i32) -> f32x4 {
     macro_rules! call {
         ($imm8:expr) => { vcmpss(a, b, $imm8) }
     }
@@ -860,48 +860,6 @@ pub unsafe fn _mm256_extractf128_si256(a: __m256i, imm8: i32) -> __m128i {
         _ => simd_shuffle2(i64x4::from(a), b, [2, 3]),
     };
     __m128i::from(dst)
-}
-
-/// Extract an 8-bit integer from `a`, selected with `imm8`. Returns a 32-bit
-/// integer containing the zero-extended integer data.
-///
-/// See [LLVM commit D20468][https://reviews.llvm.org/D20468].
-#[inline(always)]
-#[target_feature = "+avx"]
-// This intrinsic has no corresponding instruction.
-pub unsafe fn _mm256_extract_epi8(a: i8x32, imm8: i32) -> i32 {
-    let imm8 = (imm8 & 31) as u32;
-    (a.extract_unchecked(imm8) as i32) & 0xFF
-}
-
-/// Extract a 16-bit integer from `a`, selected with `imm8`. Returns a 32-bit
-/// integer containing the zero-extended integer data.
-///
-/// See [LLVM commit D20468][https://reviews.llvm.org/D20468].
-#[inline(always)]
-#[target_feature = "+avx"]
-// This intrinsic has no corresponding instruction.
-pub unsafe fn _mm256_extract_epi16(a: i16x16, imm8: i32) -> i32 {
-    let imm8 = (imm8 & 15) as u32;
-    (a.extract_unchecked(imm8) as i32) & 0xFFFF
-}
-
-/// Extract a 32-bit integer from `a`, selected with `imm8`.
-#[inline(always)]
-#[target_feature = "+avx"]
-// This intrinsic has no corresponding instruction.
-pub unsafe fn _mm256_extract_epi32(a: i32x8, imm8: i32) -> i32 {
-    let imm8 = (imm8 & 7) as u32;
-    a.extract_unchecked(imm8)
-}
-
-/// Extract a 64-bit integer from `a`, selected with `imm8`.
-#[inline(always)]
-#[target_feature = "+avx"]
-// This intrinsic has no corresponding instruction.
-pub unsafe fn _mm256_extract_epi64(a: i64x4, imm8: i32) -> i64 {
-    let imm8 = (imm8 & 3) as u32;
-    a.extract_unchecked(imm8)
 }
 
 /// Zero the contents of all XMM or YMM registers.
@@ -1138,7 +1096,7 @@ pub unsafe fn _mm_permute_pd(a: f64x2, imm8: i32) -> f64x2 {
 #[inline(always)]
 #[target_feature = "+avx"]
 #[cfg_attr(test, assert_instr(vperm2f128, imm8 = 0x5))]
-pub unsafe fn _mm256_permute2f128_ps(a: f32x8, b: f32x8, imm8: i8) -> f32x8 {
+pub unsafe fn _mm256_permute2f128_ps(a: f32x8, b: f32x8, imm8: i32) -> f32x8 {
     macro_rules! call {
         ($imm8:expr) => { vperm2f128ps256(a, b, $imm8) }
     }
@@ -1150,7 +1108,7 @@ pub unsafe fn _mm256_permute2f128_ps(a: f32x8, b: f32x8, imm8: i8) -> f32x8 {
 #[inline(always)]
 #[target_feature = "+avx"]
 #[cfg_attr(test, assert_instr(vperm2f128, imm8 = 0x31))]
-pub unsafe fn _mm256_permute2f128_pd(a: f64x4, b: f64x4, imm8: i8) -> f64x4 {
+pub unsafe fn _mm256_permute2f128_pd(a: f64x4, b: f64x4, imm8: i32) -> f64x4 {
     macro_rules! call {
         ($imm8:expr) => { vperm2f128pd256(a, b, $imm8) }
     }
@@ -1163,7 +1121,7 @@ pub unsafe fn _mm256_permute2f128_pd(a: f64x4, b: f64x4, imm8: i8) -> f64x4 {
 #[target_feature = "+avx"]
 #[cfg_attr(test, assert_instr(vperm2f128, imm8 = 0x31))]
 pub unsafe fn _mm256_permute2f128_si256(
-    a: i32x8, b: i32x8, imm8: i8
+    a: i32x8, b: i32x8, imm8: i32
 ) -> i32x8 {
     macro_rules! call {
         ($imm8:expr) => { vperm2f128si256(a, b, $imm8) }
@@ -3144,47 +3102,6 @@ mod tests {
         let r = avx::_mm256_extractf128_si256(__m256i::from(a), 0);
         let e = i64x2::new(4, 3);
         assert_eq!(r, __m128i::from(e));
-    }
-
-    #[simd_test = "avx"]
-    unsafe fn _mm256_extract_epi8() {
-        #[cfg_attr(rustfmt, rustfmt_skip)]
-        let a = i8x32::new(
-            -1, 1, 2, 3, 4, 5, 6, 7,
-            8, 9, 10, 11, 12, 13, 14, 15,
-            16, 17, 18, 19, 20, 21, 22, 23,
-            24, 25, 26, 27, 28, 29, 30, 31
-        );
-        let r1 = avx::_mm256_extract_epi8(a, 0);
-        let r2 = avx::_mm256_extract_epi8(a, 35);
-        assert_eq!(r1, 0xFF);
-        assert_eq!(r2, 3);
-    }
-
-    #[simd_test = "avx"]
-    unsafe fn _mm256_extract_epi16() {
-        let a =
-            i16x16::new(-1, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15);
-        let r1 = avx::_mm256_extract_epi16(a, 0);
-        let r2 = avx::_mm256_extract_epi16(a, 19);
-        assert_eq!(r1, 0xFFFF);
-        assert_eq!(r2, 3);
-    }
-
-    #[simd_test = "avx"]
-    unsafe fn _mm256_extract_epi32() {
-        let a = i32x8::new(-1, 1, 2, 3, 4, 5, 6, 7);
-        let r1 = avx::_mm256_extract_epi32(a, 0);
-        let r2 = avx::_mm256_extract_epi32(a, 11);
-        assert_eq!(r1, -1);
-        assert_eq!(r2, 3);
-    }
-
-    #[simd_test = "avx"]
-    unsafe fn _mm256_extract_epi64() {
-        let a = i64x4::new(0, 1, 2, 3);
-        let r = avx::_mm256_extract_epi64(a, 3);
-        assert_eq!(r, 3);
     }
 
     #[simd_test = "avx"]
--- a/coresimd/src/x86/i586/bmi.rs
+++ b/coresimd/src/x86/i586/bmi.rs
@@ -27,8 +27,8 @@ pub unsafe fn _bextr_u32(a: u32, start: u32, len: u32) -> u32 {
 #[target_feature = "+bmi"]
 #[cfg_attr(test, assert_instr(bextr))]
 #[cfg(not(target_arch = "x86"))]
-pub unsafe fn _bextr_u64(a: u64, start: u64, len: u64) -> u64 {
-    _bextr2_u64(a, (start & 0xff_u64) | ((len & 0xff_u64) << 8_u64))
+pub unsafe fn _bextr_u64(a: u64, start: u32, len: u32) -> u64 {
+    _bextr2_u64(a, ((start & 0xff) | ((len & 0xff) << 8)) as u64)
 }
 
 /// Extracts bits of `a` specified by `control` into
@@ -133,16 +133,6 @@ pub unsafe fn _blsr_u64(x: u64) -> u64 {
 #[inline(always)]
 #[target_feature = "+bmi"]
 #[cfg_attr(test, assert_instr(tzcnt))]
-pub unsafe fn _tzcnt_u16(x: u16) -> u16 {
-    x.trailing_zeros() as u16
-}
-
-/// Counts the number of trailing least significant zero bits.
-///
-/// When the source operand is 0, it returns its size in bits.
-#[inline(always)]
-#[target_feature = "+bmi"]
-#[cfg_attr(test, assert_instr(tzcnt))]
 pub unsafe fn _tzcnt_u32(x: u32) -> u32 {
     x.trailing_zeros()
 }
@@ -163,8 +153,8 @@ pub unsafe fn _tzcnt_u64(x: u64) -> u64 {
 #[inline(always)]
 #[target_feature = "+bmi"]
 #[cfg_attr(test, assert_instr(tzcnt))]
-pub unsafe fn _mm_tzcnt_u32(x: u32) -> u32 {
-    x.trailing_zeros()
+pub unsafe fn _mm_tzcnt_32(x: u32) -> i32 {
+    x.trailing_zeros() as i32
 }
 
 /// Counts the number of trailing least significant zero bits.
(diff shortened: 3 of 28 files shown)


gnzlbg:
I’ll take a look at this tomorrow

coresimd/src/x86/i586/abm.rs
```diff
@@ -44,16 +44,16 @@ pub unsafe fn _lzcnt_u64(x: u64) -> u64 {
 #[inline(always)]
 #[target_feature = "+popcnt"]
 #[cfg_attr(test, assert_instr(popcnt))]
-pub unsafe fn _popcnt32(x: u32) -> u32 {
-    x.count_ones()
```
gnzlbg: This was on purpose, bit manipulation intrinsics prefer unsigned integers like the ones in `std::` do. Whatever we decide to do, we should choose a guideline, and follow it consistently (many other intrinsics that take `__m128i` take for example `u32x4` instead of `i32x4`).

We should open an issue for this.

coresimd/src/x86/i586/avx.rs
```diff
@@ -607,77 +607,77 @@ pub unsafe fn _mm256_xor_ps(a: f32x8, b: f32x8) -> f32x8 {
 }
 
 /// Equal (ordered, non-signaling)
-pub const _CMP_EQ_OQ: u8 = 0x00;
```
gnzlbg: The unsigned part was on purpose (bit flags consistently used unsigned integers), why we use `8` instead of `32` bits here, I don't know. 

Whatever we do, we should do it consistently, and have a guideline for it.

coresimd/src/x86/i586/avx.rs
```diff
@@ -1138,7 +1096,7 @@ pub unsafe fn _mm_permute_pd(a: f64x2, imm8: i32) -> f64x2 {
 #[inline(always)]
 #[target_feature = "+avx"]
 #[cfg_attr(test, assert_instr(vperm2f128, imm8 = 0x5))]
-pub unsafe fn _mm256_permute2f128_ps(a: f32x8, b: f32x8, imm8: i8) -> f32x8 {
```
gnzlbg: Arguments that can only be `8` bit long like this `imm8` were make `i8` or `u8` on purpose.

coresimd/src/x86/i586/avx.rs
gnzlbg: So IIRC as mentioned below, arguments that can only be `8`-bit long, `16` bit long, etc. were made `_8` (e.g. `i8`) on purpose. This happens on most intrinsics taking an `imm8` or `imm16`.

The Intel intrinsics use here `int`/`long`/... very loosely, so this was part of offering better types by default.

coresimd/src/x86/i586/bmi.rs
```diff
@@ -27,8 +27,8 @@ pub unsafe fn _bextr_u32(a: u32, start: u32, len: u32) -> u32 {
 #[target_feature = "+bmi"]
 #[cfg_attr(test, assert_instr(bextr))]
 #[cfg(not(target_arch = "x86"))]
-pub unsafe fn _bextr_u64(a: u64, start: u64, len: u64) -> u64 {
-    _bextr2_u64(a, (start & 0xff_u64) | ((len & 0xff_u64) << 8_u64))
+pub unsafe fn _bextr_u64(a: u64, start: u32, len: u32) -> u64 {
```
gnzlbg: This should probably be `u8` instead of `u32` since they can only take values in the `[0, 64)` range.

coresimd/src/x86/i586/bmi.rs
```diff
@@ -127,16 +127,6 @@ pub unsafe fn _blsr_u64(x: u64) -> u64 {
     x & (x.wrapping_sub(1))
 }
 
-/// Counts the number of trailing least significant zero bits.
-///
-/// When the source operand is 0, it returns its size in bits.
-#[inline(always)]
-#[target_feature = "+bmi"]
-#[cfg_attr(test, assert_instr(tzcnt))]
-pub unsafe fn _tzcnt_u16(x: u16) -> u16 {
```
gnzlbg: GCC and clang define these:

```c++
 #define _tzcnt_u16(a)     (__tzcnt_u16((a)))
```

E.g. see here: https://clang.llvm.org/doxygen/bmiintrin_8h_source.html

coresimd/src/x86/i586/bmi.rs
```diff
@@ -163,8 +153,8 @@ pub unsafe fn _tzcnt_u64(x: u64) -> u64 {
 #[inline(always)]
 #[target_feature = "+bmi"]
 #[cfg_attr(test, assert_instr(tzcnt))]
-pub unsafe fn _mm_tzcnt_u32(x: u32) -> u32 {
-    x.trailing_zeros()
+pub unsafe fn _mm_tzcnt_32(x: u32) -> i32 {
```
gnzlbg: The renaming is correct, but these should return an unsigned integer (it returns a size). This integer is in range `[0,32)`, so per what we were currently doing somewhere else it should be an `u8`, but I have opened an issue to clarify that. 

Also, clang and GCC also define

```c++
 #define _tzcnt_u32(a)     (__tzcnt_u32((a)))
```

and in this module the non-`_mm` prefixed versions are missing (so the typo might have come from here).

coresimd/src/x86/i586/bmi.rs
```diff
@@ -173,8 +163,8 @@ pub unsafe fn _mm_tzcnt_u32(x: u32) -> u32 {
 #[inline(always)]
 #[target_feature = "+bmi"]
 #[cfg_attr(test, assert_instr(tzcnt))]
-pub unsafe fn _mm_tzcnt_u64(x: u64) -> u64 {
-    x.trailing_zeros() as u64
+pub unsafe fn _mm_tzcnt_64(x: u64) -> i64 {
```
gnzlbg: Same as above, non-`_mm`-version missing, size and signedness of the result do not match what we were doing everywhere else.

coresimd/src/x86/i586/bmi2.rs
```diff
@@ -55,8 +55,8 @@ pub unsafe fn _bzhi_u32(a: u32, index: u32) -> u32 {
 #[target_feature = "+bmi2"]
 #[cfg_attr(test, assert_instr(bzhi))]
 #[cfg(not(target_arch = "x86"))]
-pub unsafe fn _bzhi_u64(a: u64, index: u64) -> u64 {
-    x86_bmi2_bzhi_64(a, index)
+pub unsafe fn _bzhi_u64(a: u64, index: u32) -> u64 {
```
gnzlbg: index is in range `[0, 64)`, so size of the index might be wrong

coresimd/src/x86/i586/sse.rs
```diff
@@ -764,7 +764,7 @@ pub unsafe fn _mm_setzero_ps() -> f32x4 {
 #[inline(always)]
 #[target_feature = "+sse"]
 #[cfg_attr(test, assert_instr(shufps, mask = 3))]
-pub unsafe fn _mm_shuffle_ps(a: f32x4, b: f32x4, mask: i32) -> f32x4 {
+pub unsafe fn _mm_shuffle_ps(a: f32x4, b: f32x4, mask: u32) -> f32x4 {
```
gnzlbg: It seems that in general the Intel intrinsics "try" to use `unsigned` integers for bit manipulations because that is what makes sense in C++ due to `unsigned` integers having modulo arithmetic (but `signed` integers having undefined behavior on overflow).

The Intel intrinsics don't do this consistently (probably because some of them are 20 years old and this guideline wasn't clear then) but we should probably follow what the new intrinsics do.

coresimd/src/x86/i586/sse.rs
```diff
@@ -906,7 +906,7 @@ pub unsafe fn _mm_movemask_ps(a: f32x4) -> i32 {
            assert_instr(unpcklps))]
 // TODO: This function is actually not limited to floats, but that's what
 // what matches the C type most closely: (__m128, *const __m64) -> __m128
-pub unsafe fn _mm_loadh_pi(a: f32x4, p: *const f32) -> f32x4 {
+pub unsafe fn _mm_loadh_pi(a: f32x4, p: *const u8x8) -> f32x4 {
```
gnzlbg: I think this should be `p: *const f32x2` because this loads two floats from `p` into the upper two floats of `a`.

coresimd/src/x86/i586/sse.rs
```diff
@@ -957,7 +957,7 @@ pub unsafe fn _mm_loadh_pi(a: f32x4, p: *const f32) -> f32x4 {
 #[cfg_attr(all(test, target_arch = "x86", not(target_feature = "sse2")),
            assert_instr(movss))]
 // TODO: Like _mm_loadh_pi, this also isn't limited to floats.
-pub unsafe fn _mm_loadl_pi(a: f32x4, p: *const f32) -> f32x4 {
+pub unsafe fn _mm_loadl_pi(a: f32x4, p: *const u8x8) -> f32x4 {
```
gnzlbg: Same here: `p: *const f32x2`.

coresimd/src/x86/i586/sse.rs
```diff
@@ -1070,14 +1070,14 @@ pub unsafe fn _mm_loadr_ps(p: *const f32) -> f32x4 {
 // On i586 (no SSE2) it just generates plain MOV instructions.
 #[cfg_attr(all(test, any(target_arch = "x86_64", target_feature = "sse2")),
            assert_instr(movhpd))]
-pub unsafe fn _mm_storeh_pi(p: *mut u64, a: f32x4) {
+pub unsafe fn _mm_storeh_pi(p: *mut u8x8, a: f32x4) {
```
gnzlbg: Same here, we should make the `storeh` functions consistent with the `loadh`. That is: `p: *mut f32x2`.

coresimd/src/x86/i586/sse.rs
gnzlbg: Also, in case we can't agree on that, then I think `_m64` would have been better than `u8x8`.

coresimd/src/x86/i586/sse.rs
```diff
@@ -1688,18 +1688,15 @@ pub unsafe fn _mm_stream_ps(mem_addr: *mut f32, a: f32x4) {
...
 #[inline(always)]
 #[target_feature = "+sse"]
-// generates movnti on i686 and x86_64 but just a mov on i586
-#[cfg_attr(all(test,
-               any(target_arch = "x86_64",
-                   all(target_arch = "x86", target_feature = "sse2"))),
-           assert_instr(movnti))]
-pub unsafe fn _mm_stream_pi(mem_addr: *mut i64, a: i64) {
+// #[cfg_attr(test, assert_instr(movntq))] // FIXME: llvm codegens wrong instr
+pub unsafe fn _mm_stream_pi(mem_addr: *mut i8x8, a: i8x8) {
```
gnzlbg: We should be using `_m64` here (which might probably fix the FIXME since that is the magic `x86_mmx` type).

coresimd/src/x86/i586/sse41.rs
```diff
@@ -61,7 +61,7 @@ pub unsafe fn _mm_blendv_epi8(a: i8x16, b: i8x16, mask: i8x16) -> i8x16 {
 #[inline(always)]
 #[target_feature = "+sse4.1"]
 #[cfg_attr(test, assert_instr(pblendw, imm8 = 0xF0))]
-pub unsafe fn _mm_blend_epi16(a: i16x8, b: i16x8, imm8: u8) -> i16x8 {
+pub unsafe fn _mm_blend_epi16(a: i16x8, b: i16x8, imm8: i32) -> i16x8 {
```
gnzlbg: Since this is an `imm8` from a bitmask I think it is more consistent with what we were previously doing to use an `u8` (because of bitmask, and because of its value range).

coresimd/src/x86/i586/sse41.rs
```diff
@@ -91,7 +91,7 @@ pub unsafe fn _mm_blendv_ps(a: f32x4, b: f32x4, mask: f32x4) -> f32x4 {
 #[inline(always)]
 #[target_feature = "+sse4.1"]
 #[cfg_attr(test, assert_instr(blendpd, imm2 = 0b10))]
-pub unsafe fn _mm_blend_pd(a: f64x2, b: f64x2, imm2: u8) -> f64x2 {
+pub unsafe fn _mm_blend_pd(a: f64x2, b: f64x2, imm2: i32) -> f64x2 {
```
gnzlbg: The Intel Intrinsics guide calls the last argument `imm8` (so `u8`).

coresimd/src/x86/i586/sse42.rs
```diff
@@ -8,49 +8,49 @@ use stdsimd_test::assert_instr;
 use v128::*;
 
 /// String contains unsigned 8-bit characters *(Default)*
-pub const _SIDD_UBYTE_OPS: i8 = 0b0000_0000;
+pub const _SIDD_UBYTE_OPS: i32 = 0b0000_0000;
```
gnzlbg: These should have been `u8` instead of `i8`.

coresimd/src/x86/i586/xsave.rs
```diff
@@ -36,7 +36,7 @@ extern "C" {
 #[inline(always)]
 #[target_feature = "+xsave"]
 #[cfg_attr(test, assert_instr(xsave))]
-pub unsafe fn _xsave(mem_addr: *mut u8, save_mask: u64) -> () {
+pub unsafe fn _xsave(mem_addr: *mut u8, save_mask: u64) {
```
gnzlbg: clippy should have complained about these, would be nice to know why it didn't and fill an issue upstream.

coresimd/src/x86/i686/sse2.rs
```diff
@@ -25,56 +25,22 @@ pub unsafe fn _mm_cvtsi64x_sd(a: f64x2, b: i64) -> f64x2 {
...
-}
-
 /// Returns the lower 64 bits of a 128-bit integer vector as a 64-bit
 /// integer.
 #[inline(always)]
 #[target_feature = "+sse2"]
-// no particular instruction to test
-pub unsafe fn _mm_movepi64_pi64(a: i64x2) -> i64 {
-    a.extract(0)
+// #[cfg_attr(test, assert_instr(movdq2q))] // FIXME: llvm codegens wrong instr?
```
gnzlbg: We should go through all FIXMEs and TODOs of the x86 intrinsics before stabilization. There was one before that say "add test on windows" with little explanation about exactly what was broken there.

coresimd/src/x86/i686/sse2.rs
```diff
@@ -25,56 +25,22 @@ pub unsafe fn _mm_cvtsi64x_sd(a: f64x2, b: i64) -> f64x2 {
...
+pub unsafe fn _mm_movepi64_pi64(a: i64x2) -> i8x8 {
```
gnzlbg: This is wrong, this function returns a single 64-bit integer. It should have been `__m64` instead of `i64` though (probably `__m64` wasn't implemented back then).

We should go over all intrinsics that return `__m64` and make sure that we don't make them worse by returning something else. In particular, chaining one intrinsic that returns `__m64` with one that takes `__m64` as an argument probably generates assembly that operates directly on that MMX register. However, `u8x8` is not an `x86_mmx` LLVM type, so by using a different type that does not translates to `x86_mmx` we might be in-curing a cost that the C compilers don't have.

coresimd/src/x86/i686/sse2.rs
```diff
@@ -25,56 +25,22 @@ pub unsafe fn _mm_cvtsi64x_sd(a: f64x2, b: i64) -> f64x2 {
...
 
 /// Moves the 64-bit operand to a 128-bit integer vector, zeroing the
 /// upper bits.
 #[inline(always)]
 #[target_feature = "+sse2"]
-// #[cfg_attr(test, assert_instr(movq2dq))] FIXME
-pub unsafe fn _mm_movpi64_epi64(a: i64) -> i64x2 {
-    i64x2::new(a, 0)
+// #[cfg_attr(test, assert_instr(movq2dq))] // FIXME: llvm codegens wrong instr?
+pub unsafe fn _mm_movpi64_epi64(a: i8x8) -> i64x2 {
```
gnzlbg: This function should take a `__m64` instead of `i64` or `i8x8` type. This would probably fix the FIXME.

coresimd/src/x86/i686/ssse3.rs
```diff
@@ -47,7 +47,7 @@ pub unsafe fn _mm_shuffle_pi8(a: u8x8, b: u8x8) -> u8x8 {
 #[inline(always)]
 #[target_feature = "+ssse3"]
 #[cfg_attr(test, assert_instr(palignr, n = 15))]
-pub unsafe fn _mm_alignr_pi8(a: u8x8, b: u8x8, n: u8) -> u8x8 {
+pub unsafe fn _mm_alignr_pi8(a: u8x8, b: u8x8, n: i32) -> u8x8 {
```
gnzlbg: `n: u8` was by design here as well.

coresimd/src/x86/x86_64/sse2.rs
```diff
@@ -58,6 +58,40 @@ pub unsafe fn _mm_stream_si64(mem_addr: *mut i64, a: i64) {
...
+#[target_feature = "+sse2"]
+#[cfg_attr(all(test, not(windows)), assert_instr(movq))]
+pub unsafe fn _mm_cvtsi128_si64(a: i64x2) -> i64 {
+    a.extract(0)
+}
+
+/// Return the lowest element of `a`.
+#[inline(always)]
+#[target_feature = "+sse2"]
+#[cfg_attr(all(test, not(windows)), assert_instr(movq))]
```
gnzlbg: We should add a FIXME to fix these tests on windows.

coresimd/src/x86/x86_64/sse41.rs
```diff
@@ -0,0 +1,49 @@
...
+
+#[cfg(test)]
+use stdsimd_test::assert_instr;
+
+/// Extract an 64-bit integer from `a` selected with `imm8`
+#[inline(always)]
+#[target_feature = "+sse4.1"]
+// TODO: Add test for Windows
+#[cfg_attr(all(test, not(windows)), assert_instr(pextrq, imm8 = 1))]
+pub unsafe fn _mm_extract_epi64(a: i64x2, imm8: i32) -> i64 {
```
gnzlbg: `imm8: u8`

coresimd/src/x86/x86_64/sse41.rs
```diff
@@ -0,0 +1,49 @@
...
+    let imm8 = (imm8 & 1) as u32;
+    a.extract_unchecked(imm8)
+}
+
+/// Return a copy of `a` with the 64-bit integer from `i` inserted at a
+/// location specified by `imm8`.
+#[inline(always)]
+#[target_feature = "+sse4.1"]
+#[cfg_attr(all(test, target_arch = "x86_64"), assert_instr(pinsrq, imm8 = 0))]
+pub unsafe fn _mm_insert_epi64(a: i64x2, i: i64, imm8: i32) -> i64x2 {
```
gnzlbg: - `imm8: u8`
- the check for `x86_64` in the test is not necessary (this module is only included for `x86_64`)

stdsimd-verify/build.rs
```diff
@@ -0,0 +1,24 @@
...
+        if file.file_type().unwrap().is_dir() {
+            walk(&file.path());
+            continue
+        }
+        let path = file.path();
+        if path.extension().and_then(|s| s.to_str()) != Some("rs") {
+            continue
+        }
+
+        println!("cargo:rerun-if-changed={}", path.display());
```
gnzlbg: Does this only redo the build.rs if any of the selected paths changed? What happens if I add a new path (e.g. by creating a new file in the `i686`/`x86_64` directories? Does it get automatically checked?

stdsimd-verify/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    functions.retain(|f| {
+        match f.vis {
+            syn::Visibility::Public(_) => {}
+            _ => return false,
+        }
+        match f.unsafety {
+            syn::Unsafety::Unsafe(_) => {}
+            _ => return false,
```
gnzlbg: Are there `pub` functions in the `x86` module that are not `unsafe`? What are those?

stdsimd-verify/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+                    _ => false,
+                }
+            })
+    });
+
+    let input = proc_macro2::TokenStream::from(input);
+
+    let functions = functions.iter()
+        .filter(|f| {
+            !f.ident.sym.as_str().starts_with("_mulx")
```
gnzlbg: What's the problem with `_mulx`? Is it that it doesn't generate the MULX instruction?

stdsimd-verify/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+        })
+        .map(|f| {
+            let name = f.ident;
+            // println!("{}", name);
```
gnzlbg: debug code

stdsimd-verify/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+                    ret: #ret,
+                    target_feature: #target_feature,
+                    instrs: &[#(stringify!(#instrs)),*],
+                }
+            }
+        })
+        .collect::<Vec<_>>();
+
+    let ret = my_quote! { #input: &[Function] = &[#(#functions),*]; };
+    // println!("{}", ret);
```
gnzlbg: debug code

stdsimd-verify/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+    ret.into()
+}
+
+fn to_type(t: &syn::Type) -> Tokens {
+    match *t {
+        syn::Type::Path(ref p) => {
+            match extract_path_ident(&p.path).sym.as_str() {
+                "__m128i" => my_quote! { &I8x16 },
+                "__m256i" => my_quote! { &I8x32 },
+                "__m64" => my_quote! { &I8x8 },
```
gnzlbg: This is a bit dubious, because `i8x8` does not currently need to map to an MMX register, but `__m64` does. 

Maybe the current approach of hardcoding `#[repr(simd)] struct __m64(i64)` to map to `x86_mmx` is wrong and we should be mapping all 64-bit `#[repr(simd)]` to `x86_mmx` instead. Some intrinsics take `__m64` but we would like to strongly type them as `i16x4` but currently can't (if we want proper code gen) due to that mismatch, so we need to use `__m64` instead.

stdsimd-verify/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+}
+
+fn find_instrs(attrs: &[syn::Attribute]) -> Vec<syn::Ident> {
+    attrs.iter()
+        .filter_map(|a| a.meta_item())
+        .filter_map(|a| {
+            match a {
+                syn::MetaItem::List(i) => {
+                    if i.ident == "cfg_attr" {
+                        Some(i.nested.into_vec())
```
gnzlbg: Would it be possible to map this to `None` if the `cfg_attr` is disabled. For example, `#[cfg_attr(test, ...` is always enabled on test builds, but `#[cfg_attr(all(test, not(windows)), ...]` should fail on windows builds, so we would like the windows CI to detect when an intrinsic is missing the `assert_instr` check for windows builds.

stdsimd-verify/tests/x86-intel.rs
```diff
@@ -0,0 +1,299 @@
...
+}
+
+#[derive(Deserialize)]
+struct Instruction {
+    name: String,
+}
+
+#[test]
+fn verify_all_signatures() {
+    let xml = include_bytes!("../x86-intel.xml");
```
gnzlbg: It would be nice to have a comment stating where does this file come from, how to update it, etc? Ideally, we would always be running tests against the most up-to-date xml file.

stdsimd-verify/tests/x86-intel.rs
```diff
@@ -0,0 +1,299 @@
...
+    let xml = &xml[..];
+    let data: Data = serde_xml_rs::deserialize(xml).expect("failed to deserialize xml");
+    let mut map = HashMap::new();
+    for intrinsic in data.intrinsics.iter() {
+        // This intrinsic has multiple definitions in the XML, so just ignore it.
+        if intrinsic.name == "_mm_prefetch" {
+            continue
```
gnzlbg: The `_mm_prefetch` intrinsics does something different if the target has AVX-512, SSE, or not even SSE. I think it is ok to ignore these for now (and AVX-512 overloads some intrinsics to have slightly different behavior).

stdsimd-verify/tests/x86-intel.rs
```diff
@@ -0,0 +1,299 @@
...
+        }
+
+        let intel = match map.get(rust.name) {
+            Some(i) => i,
+            None => panic!("missing intel definition for {}", rust.name),
+        };
+
+        // Verify that all `#[target_feature]` annotations are correct, ensuring
+        // that we've actually enabled the right instruction set for this
+        // intrinsic.
```
gnzlbg: could we check here that we have put the intrinsic in the right module (e.g. by saving the name of the module in which the intrinsics are implemented)? (there were some avx2 intrinsics in the avx module)

stdsimd-verify/tests/x86-intel.rs
```diff
@@ -0,0 +1,299 @@
...
+                continue
+            }
+
+            let cpuid = cpuid
+                .chars()
+                .flat_map(|c| c.to_lowercase())
+                .collect::<String>();
+
+            // Normalize `bmi1` to `bmi` as apparently that's what we're calling
+            // it.
```
gnzlbg: We might want to change the name of the target feature here to match intel's name before stabilization.

stdsimd-verify/tests/x86-intel.rs
```diff
@@ -0,0 +1,299 @@
...
+        (&Type::Bool, "int")
+            if intrinsic.starts_with("_mm_comi") && intrinsic.ends_with("_sd")
+            => {}
+        (&Type::Bool, "int")
+            if intrinsic.starts_with("_mm_ucomi") && intrinsic.ends_with("_sd")
+                => {}
+
+        _ => panic!("failed to equate: `{}` and {:?} for {}", intel, t, intrinsic),
+    }
+}
```
gnzlbg: Nice work here. 

The two main points that worry me a bit is that:

- this equates `int` to 32-bit integer, but an `int` in C can be a 64-bit integer (and actually are in some platforms). I think assuming `int` to mean 32-bit integer is fine, but
- Intel often writes that the argument is an 8-bit integer type, even though the argument type is `int`, or the specification mentions that the value is at most N, meaning that a 2-bit, 4-bit, 8-bit, 16-bit integer types might make more sense to represent these,
- this assumes that `int` maps to a signed integer, but in many intrinsics intel uses `int` to denote a count or a bit index, which can't never be negative.

gnzlbg:
So this looks awesome and it has surfaced lots of bugs and inconsistencies in the conventions that we were trying to follow. 

I think we should clarify whether we want to keep following these conventions and improve this a bit to enforce them: 

- we were use unsigned integer types when bit manipulation was intended (including bitflags, bitmasks, etc.): why? I don't really know. If I had to guess I'd say that this is because `unsigned` integers in C and C++ have modulo arithmetic, and the newer intrinsics tend to follow this convention even though some of the older ones don't. In Rust, however, arithmetic and signedness are orthogonal to each other.
- we were using appropriately sized types (as long as it was reasonable): why? for increased type safety (e.g. trying to pass a `0xFFFF` to a flag taking an `u8` raises a warning). Intel specifies most of these arguments as having type `int` (which can be any integer type from 8-bit to 64 bits, depending on the platform), but then the specification text adds extra semantic information like it always being positive, being in some range `[0, {2,4,8,16,...255, 16386,...})`. Many intrinsics use these information to translate `int` not only to `i32`, but every combination of `{i,u}{8,16,32}`.
- we did not have a properly working `__m64` for a long time some we have a lot of intrinsics with broken code gen that just use `i64`.  With these intrinsics we should be extra careful because `__m64` means an MMX register and LLVM calls this particular type `x86_mmx`. That is, `i8x8` has 64 bits but it is not necessarily an `x86_mmx` type. The easiest way to strongly type these is to just transmute internally to `__m64` but we should check that chaining these intrinsics doesn't break codegen.

coresimd/src/x86/i586/bmi.rs
alexcrichton: Perhaps, but this isn't on Intel's intrinsic list, which is what this crate is for.

stdsimd-verify/build.rs
alexcrichton: Yes, this only reruns if those are changed. If you add a new module it doesn't matter, what typically happens is you add `mod foo` to an existing file.

stdsimd-verify/src/lib.rs
alexcrichton: Presumably not because we said that all intrinsics should be unsafe.

stdsimd-verify/src/lib.rs
alexcrichton: That I'd like to leave in because it's nicest to just re-uncomment-it.

stdsimd-verify/src/lib.rs
alexcrichton: It doesn't matter, this is just mapping to a 64-bit vector type.

stdsimd-verify/src/lib.rs
alexcrichton: Not easily, no.

stdsimd-verify/tests/x86-intel.rs
alexcrichton: Perhaps, but I'll leave that to a future commit.

stdsimd-verify/tests/x86-intel.rs
alexcrichton: > but an int in C can be a 64-bit integer (and actually are in some platforms). 

Not on Intel AFAIK.

alexcrichton:
In general as the libs team has discussed SIMD over time it's becoming more and more apparent that we're *not* going to deviate at all from Intel's upstream definitions. I expect the `uNxN` types to all go away on Intel eventually in favor of `__m256` and such natively as we stabilize intrinsics. As such I've switched all arguments to exactly what intel says (i32 where intel says `int`). In general I don't think it really matters about all the constant arguments, they're all constant anyway so the type shouldn't propagate anywhere else in the program.

I'd personally prefer to not have special cases when reading Intel's own definitions and trying to bend over backwards to map those to our own "possibly more correct" types. If we can come up with a strong technical argument for why we'd make such a transformation I thin kthat makes sense, but "we happened to do it yesterday" I don't think is sufficient.

For `__m64` related things I believe `i8x8` (or any 64-bit vector) is all defined the exact same way as `__m64` (or at least that's what I thought it was done as). In that sense we'll likely eventually move to just `__m64` so that'll just get updated naturally. The main change here was switching from a literal `i64` to an mmx type.

coresimd/src/x86/i586/bmi.rs
gnzlbg: Yep, the Intel intrinsic list only defines `_tzcnt_u{32,64}`. 

MSVC does not define it here: https://msdn.microsoft.com/en-us/library/hh977023.aspx

and the Intel C++ compiler doesn't seem to define it either. The GCC compiler defines it The GCC definition is here: https://github.com/gcc-mirror/gcc/blob/master/gcc/config/i386/bmiintrin.h#L38M

> Perhaps, but this isn't on Intel's intrinsic list, which is what this crate is for

Intel Intrinsics list doesn't define many x86 intrinsics either (e.g. all of the AMD ones).

Anyways, I am fine with removing it, just saying that two widely used C and C++ compilers offer it, which is why it was offered here.

stdsimd-verify/build.rs
gnzlbg: Cool! TIL that `build.rs` could do this. Every crate I have with a `build.rs` file gets rebuild all the time, maybe one should add something like this to the jemalloc-sys crate as well.

stdsimd-verify/src/lib.rs
gnzlbg: Yes I think so too, maybe we should emit a warning here or something, but that can be done in a future PR.

stdsimd-verify/src/lib.rs
gnzlbg: So IIUC it is possible to change all the intrinsics that now take an `i8x8` to use an `i16x4` or `__m64` or any type with 64-bit wide argument right?

stdsimd-verify/src/lib.rs
gnzlbg: I see, well we can always add this in the future.

stdsimd-verify/tests/x86-intel.rs
gnzlbg: It doesn't depend on the CPU vendor, but on the platform (which must conform to the C standard). For example: 

- on `x86_64-unknown-linux-gnu` `int`s are 32-bit, and long are `64-bit`, 
- on `x86_64-pc-windows-msvc` `int`s and `long`s are 32-bit and `long long`s are 64-bit, and
- on `x86_64-cray-linux-gnu` `int`s are also 64-bit. 

All of these conform to the C standard.

gnzlbg:
I'll be ok with:

- 1. offering `__m{64. 128, 256}{_,i,d}`, types without any operations (like we do today)
- 2. specifying all intrinsics to use those just like C does
- 3. (optional as an after-step): offering the `{i,u,f}{8,16,32,64}x{2,4,8,16,32}` types, with conversions from/to `__m{64. 128, 256}{_,i,d}` (to avoid stabilizing `repr(simd)` and the LLVM generic operations like `simd_cast`).

That is, we have the same intrinsic API that C does, but higher-level libraries and users can use portable `{i,u,f}{8,16,32,64}x{2,4,8,16,32}` types in their APIs, but internally they will need to convert those to the appropriate `__m{64. 128, 256}{_,i,d}`.

The points 1. and 2. are non-controversial because that's what C does and this is a low-level API, so we can deliver those "soon". 

The point 3 would be a convenience for libraries like `faster`. They can build their own portable types on top of `__m{64. 128, 256}{_,i,d}` but that is really painful for library authors and we really should be using the generic LLVM intrinsics here anyways.

All in all, I'll agree to all of this, but its a huge change API wise, and it hasn't been discussed much in the open, I think. The quickest thing would be to create an issue here to discuss it, but this is already at a point where we (or I) can write an RFC about this.

---

> (i32 where intel says int) 

We probably should just add an `int` type alias somewhere in the library so that, if platform compatibility ends up giving us trouble in the future, we can at least hack around that in a single place. 

> I'd personally prefer to not have special cases when reading Intel's own definitions and trying to bend over backwards to map those to our own "possibly more correct" types. If we can come up with a strong technical argument for why we'd make such a transformation I thin kthat makes sense, but "we happened to do it yesterday" I don't think is sufficient.

Basically we will need to review every single intrinsic, and motivate why it uses one type instead of another. That's an absurd ton of work. From the pov of "we want to stabilize this someday", trying to deviate from the Intel intrinsics in any way makes this whole thing "un-rewieable/un-RFC'able".

> For __m64 related things I believe i8x8 (or any 64-bit vector) is all defined the exact same way as __m64 (or at least that's what I thought it was done as). 

IIRC when you added `x86_mmx` to the compiler the logic was that it had to be a 64-bit vector containing a single element, see: https://github.com/rust-lang/rust/blob/3cc68bac7c89a81ec83cbd8f0aff9db001425c50/src/librustc_trans/type_of.rs#L36 .

gnzlbg:
@BurntSushi help

AdamNiederer:
> (optional as an after-step): offering the {i,u,f}{8,16,32,64}x{2,4,8,16,32} types, with conversions from/to __m{64. 128, 256}{_,i,d} (to avoid stabilizing repr(simd) and the LLVM generic operations like simd_cast).

I'd be willing to steward a package containing the more stringently-typed definitions of these intrinsics, if that helps with the decision at all. Perhaps we could have a solution similar to the `http` crate with the "strong" vector types and perhaps portable typedefs of them?

> In the mean time I don't know what to do with this PR, I think we should merge it, but changing all those APIs is a huge breaking change.

I maintain 2 of the 5 packages which depend on stdsimd at the moment, and I'd recommend breaking the API as much as possible while we're still figuring this stuff out and few people actually use the crate.

stdsimd-verify/tests/x86-intel.rs
alexcrichton: Er what I mean is this is that this doesn't matter for us. Rust doesn't have any platforms where i32 isn't `int`, so we can just cross this bridge when we get there. There's no real need to add hundreds of lines of "portability code" that aren't actually buying us anything.

alexcrichton:
@gnzlbg 

> That is, we have the same intrinsic API that C does, but higher-level libraries and users can use portable {i,u,f}{8,16,32,64}x{2,4,8,16,32} types in their APIs, but 

Yeah that's what I'd imagine as well, the "more portable" types would probably be much more laggard in stabilization as well.

> All in all, I'll agree to all of this, but it's a huge change API wise ... but changing all those APIs is a huge breaking change.

Agreed it's a big change! We'd want to be somewhat careful about it I think and probably yes have a bit of discussion first. FWIW though I don't consider this library having any stability guarantees right now though.

> We probably should just add an int type alias somewhere in the library so that, if platform compatibility ends up giving us trouble in the future, we can at least hack around that in a single place.

I'd rather not, we have yet to even see from a distance a realistic platform for Rust where this is true. This has lots of complications when moving into libcore as well, let's just not.

> Basically we will need to review every single intrinsic, and motivate why it uses one type instead of another. That's an absurd ton of work. From the pov of "we want to stabilize this someday", trying to deviate from the Intel intrinsics in any way makes this whole thing "un-rewieable/un-RFC'able".

Correct, this is why "hey it's the same as intel" is so crucial to be able to rubber-stamp APIs and be confident in such a stamp

---

@AdamNiederer 

> It's worth noting, however, that the ambiguity of intel's vector types isn't necessarily duplicated across all platforms

Certainly! Something like `__m128` would not be portable, it'd only be available on x86. Platforms like AArch64 would have their own whole suite of types.

> I'd be willing to steward a package containing the more stringently-typed definitions of these intrinsics, if that helps with the decision at all. Perhaps we could have a solution similar to the http crate with the "strong" vector types and intrinsic definitions?

That'd be great! It's certainly something we always wanted to see on top of a library like this!

alexcrichton:
> In the mean time I don't know what to do with this PR

I plan on addressing some of the feedback and then merging. I don't think we'll break much usage by switching the types from things like `u8` to `i32` as they're all constants anyway, and otherwise it's mostly just mmx stuff that got fixed which we've seen claims about anyway that they shouldn't be exposed anyway!

BurntSushi:
I'm in favor of this. I basically don't think "breakage" at this point is something we should be concerning ourselves with. This crate is still very clearly in an experimental phase while we iron out the API. Folks using it today should be doing it with that knowledge, and with the knowledge that their experience should help inform us of future API decisions. We should semver bump and move on. :)

I don't have too many thoughts. I am slowly coming around to @alexcrichton's view that we should revert our efforts to provide more stringent types on the vector intrinsics. I am pretty disappointed myself since we put so much work into it, but I suppose it's just a sunk cost. I tend to think that having these be checked in an automated fashion is a pretty huge advantage. With that said, I do think we should have a contingency. i.e., The stabilization effort should lay out plans for what to do in the event that we did indeed get some intrinsics wrong.

stdsimd-verify/src/lib.rs
alexcrichton: Yes just like with `__m256i` you can have any `iAxB` so long as `A * B = 256`. In general I don't expect this to be a problem as well probaly hardcode everything to `__m256i` anyway.

gnzlbg:
@alexcrichton  So all in all I'd say this is good to once it's green again.

The only thing that it is a bit sketchy is switching so many APIs to use `i8x8` when we already know we want to switch them to `__m64` in the near future but that doesn't need to be done in this PR.

---

@AdamNiederer 

> Perhaps we could have a solution similar to the http crate with the "strong" vector types and intrinsic definitions?

I think we should just provide all vector types in `coresimd` for all platforms.

The `x86` intrinsics are only specified in terms of the `"weak" __mXYZw` types, while most of the non-`x86` intrinsics (ARM, MIPS, PowerPC, ...) are specified in terms of the "strong" `txWxL` vector types (with different notation, e.g., `v4i32`, but equal semantics). 

Implementing these types needs unstable intrinsics that we would prefer never to stabilize, and we need to implement both in `coresimd` anyways.

Given that they are already implemented, and that they currently work with each other (this is not hard), I don't currently see what we would gain from splitting the types into a different crate.

In particular, because `coresimd` would then need to depend on this crate, and this crate would need to be part of `core`.

Maybe my views on this would change as we implement more ARM, MIPS, and PowerPC intrinsics in `coresimd` and new problems arise, but I expect the differently-named "strong" vector types in these platforms to be just type-aliases. 

Also, for stabilizing the intrinsics of a particular platform we only need to stabilize the vector types that the platform uses. That is, to stabilize the `x86` intrinsics we only need to stabilize the "weak" vector types, and the "strong" ones can remain unstable. 

---

@BurntSushi 

> I am pretty disappointed myself since we put so much work into it, but I suppose it's just a sunk cost. 

Well its not 100% sunk: we learned something. Once the RFC for this arrives somebody is going to ask "why aren't the Intel intrinsics more strongly typed?" and now we have a pretty good answer for that ("they were more strongly typed for a while; it did not work out").

> I tend to think that having these be checked in an automated fashion is a pretty huge advantage. 

Agreed.

alexcrichton:
@gnzlbg 

> The only thing that is a bit sketchy is switching so many APIs to use i8x8 when we already know we want to switch them to __m64 in the near future but that doesn't need to be done in this PR.

Oh I think I moved them over to `__m64` where I could.

PR #257, opened 2017-12-31
Add bswap

AdamNiederer:
This uses LLVM's architecture-independent bswap intrinsic to mimic Intel's _bswap and _bswap64 intrinsics.

`bswap` was added to the 486, so I've added it to the i586 module.


alexcrichton:
Thanks!

PR #258, opened 2018-01-02
Last missing avx and avx2 intrinsics

gwenn:
avx and avx2 is complete.

For _mm256_slli_si256 and _mm256_srli_si256, I don't know why Clang does not use the builtin instructions ?


alexcrichton:
Awesome, thanks!

PR #261, opened 2018-01-03
Assert intrinsic implementations are inlined properly

alexcrichton:
This adds an assertion with the `assert_instr` macro that intrinsics aren't literally calling one another but rather everything is nice and inlined. This then touches up various intrinsics that accidentally had a `call` instruction in their implementation.

### DIFF
--- a/stdsimd-test/src/lib.rs
+++ b/stdsimd-test/src/lib.rs
@@ -293,9 +293,25 @@ pub fn assert(fnptr: usize, fnname: &str, expected: &str) {
         }
     }
 
-    let probably_only_one_instruction = function.instrs.len() < 30;
+    // Look for `call` instructions in the disassembly to detect whether
+    // inlining failed: all intrinsics are `#[inline(always)]`, so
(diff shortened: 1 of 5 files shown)


stdsimd-test/src/lib.rs
```diff
@@ -293,9 +293,41 @@ pub fn assert(fnptr: usize, fnname: &str, expected: &str) {
...
+        };
+        if !part.contains("call") {
+            continue
+        }
+
+        // On 32-bit x86 position independent code will call itself and be
+        // immediately followed by a `pop` to learn about the current address.
+        // Let's not take that into account when considering whether a function
+        // failed inlining something.
+        let followed_by_pop = function.instrs.get(i + 1)
```
gnzlbg: how did you figure this out?

alexcrichton: Oh I look at the disassembly and recognized the instruction sequence. Additionally compiling with `-C relocation-model=static` made it go away.

PR #264, opened 2018-01-05
Add RDTSC and RDTSCP intrinsics

shepmaster:


alexcrichton:
Thanks!

gnzlbg:
Is there a reason why there aren't in i386 ? That's where clang puts them: 

https://github.com/llvm-mirror/clang/blob/a38ba9770a70056f9fdd6c71f819e5db45a105e4/lib/Headers/ia32intrin.h#L64

There doesn't seem to be an llvm feature to enable these conditionally, so I'd guess they are available on all x86 LLVM targets without exceptions.

shepmaster:
TBQH, I just followed the BSWAP intrinsic, although I thought I looked at the Intel manual to see when it was supported. 

Either way, you can assume I didn't know what I was doing ;-)

gnzlbg:
So `_bswap` also belongs in `i386` :D 

https://github.com/gcc-mirror/gcc/blob/master/gcc/config/i386/ia32intrin.h

I'll move them there.

PR #265, opened 2018-01-05
Migrate `_mm_add_ss` to `__m128`

alexcrichton:
This commit starts the migration towards Intel's types one intrinsic at a time,
starting with `_mm_add_ss`. This is mostly just to get a feel for what the tests
will start to look like.


alexcrichton:
r? @gnzlbg 

I'm curious what you think of this strategy. My thinking is that we'll start moving towards *stripped down* versions of the `__mXXX` types. For now I'd like to start with *only* the `Copy`, `Clone`, and `Debug` traits implemented (not even `PartialEq`) and see how it goes. Ideally we should be able to define all tests and such purely in terms of the exposed vendor intrinsics (as, after all, that's all we're gonna give users). 

That being said, I'd like your thoughts on this approach!

alexcrichton:
Totally! If @gnzlbg is on board with this approach I'll publish what we've got, merge this, and then start translating more (hopefully figuring out a way to do it mechanically along the way)

gnzlbg:
@alexcrichton what is the advantage of this over not reusing `v128::__m128` ?

The main difference I see is that the new type doesn't implement `PartialEq` (and the conversions), but if this is what we want we can just make `v128::__m128` not implement `PartialEq` either.

Also, I think we should reuse the strongly typed vector types in the tests (they are just tests after all). In C one "interfaces" with `__m128` by using a portable vector type (e.g. `f32x4`) and then just converting to it (the conversions are implicit but they are there).

alexcrichton:
@gnzlbg oh I was going to reuse that typedef but it doesn't exist, so I created it. Additionally I was thinking we'd want to move the type definitions into the x86 folders as they shouldn't be available on other paltforms (in theory). This does raise questions about the eventual structure of the crate in libcore/libstd, but I figured for now the exact location it was imported from didn't matter too mch.

I was thinking though that it'd be good to *not* use the unstable features to tests all these features. In that we're hoping to stabilize the intrinsics before the portable types, so I think it'd be good to have a hefty body of code of how to actually use them in a "purely stable" way. I think we'll definitely want tests for the portable types for sure, but I'd also like to get experience (locally) with stable types as well.

I do think though that we'll want conversions between the portable types and these x86 types (which don't currently exist for `__m128` itself), but I'm thinking we may want to untangle a bit the stable/unstable surfaces of this library (if possible).

gnzlbg:
@alexcrichton 

> oh I was going to reuse that typedef but it doesn't exist, so I created it.

Indeed, i was thinking of [`__m128i`](https://github.com/rust-lang-nursery/stdsimd/blob/master/coresimd/src/v128.rs#L46) but this type is different.

I would prefer to reuse the macros to define these types, or maybe add some new macros here for `x86` specific stuff (or in the `x86` directory), so that we keep all types with the same API.

> Additionally I was thinking we'd want to move the type definitions into the x86 folders as they shouldn't be available on other paltforms (in theory).

I think that's ok.

> This does raise questions about the eventual structure of the crate in libcore/libstd, but I figured for now the exact location it was imported from didn't matter too mch.

Maybe we just put the platform-specific vendor types in `core::vendor::` along the intrinsics, and keep the `core::simd::` for the strongly-typed vector types? That way we can just keep them unstable by not stabilizing the `core::simd` module.

> I was thinking though that it'd be good to not use the unstable features to tests all these features. In that we're hoping to stabilize the intrinsics before the portable types, so I think it'd be good to have a hefty body of code of how to actually use them in a "purely stable" way. I think we'll definitely want tests for the portable types for sure, but I'd also like to get experience (locally) with stable types as well.

I think that we should definitely have code that show how to use the intrinsics in a purely stable way. Maybe a compromise could be to start porting the intrinsics required for the examples, and make the examples show "purely stable" usage of the library? 

If we keep the strongly-typed vector types in the tests, porting the intrinsics reduces to sparkling `.into()` here and there, which is a lot of work. We can remove them later if we want, but removing them is way more work and it is unclear to me whether it is worth doing so right now. For all we know we might want to stabilize the strongly-typed vector types soonish after the intrinsics anyways. 

Arguably, we don't have many examples, but I think that porting some more examples from the `simd` crate is something we should do anyways and time would be time better spent there.

alexcrichton:
@gnzlbg 

> I would prefer to reuse the macros to define these types, or maybe add some new macros here for x86 specific stuff (or in the x86 directory), so that we keep all types with the same API.

Sure! I just wanted to do one first to see what the macro would generated. Just hoping to get some experience before we generalize everything. I do think though that we'll want to avoid the same API as the strongly-typed types (e.g. no `extract` and such). We may eventually want some constructors as I think you can do that in C without the intrinsics, but we can get to that down the road.

> Maybe we just put the platform-specific vendor types in core::vendor:: along the intrinsics, and keep the core::simd:: for the strongly-typed vector types? That way we can just keep them unstable by not stabilizing the core::simd module.

My thoughts exactly!

> We can remove them later if we want, but removing them is way more work and it is unclear to me whether it is worth doing so right now. For all we know we might want to stabilize the strongly-typed vector types soonish after the intrinsics anyways.

Heh well I'm sort of volunteering to get the work started :)

It's true that we don't *know* the strongly-typed vectors won't be stabilized, though.

In general how do you feel about this PR? Do you think we *shouldn't* take this strategy when updating the tests?

gnzlbg:
>  I do think though that we'll want to avoid the same API as the strongly-typed types (e.g. no extract and such).

Yeah the `__m128i` type does not have `extract` and friends because that leaks the internal representation of the type (`(i64,i64)` vs `(i32,i32,i32,i32)`).

> We may eventually want some constructors as I think you can do that in C without the intrinsics, but we can get to that down the road.

I did not managed to add any meaningful constructors to `__m128i`. If we go as planned and don't stabilize the strongly-typed vector types together with the `x86` intrinsics we need to find a solution for this, but maybe we'll get some new ideas as we port the examples to the new API (pain points will become obvious).

> In general how do you feel about this PR? Do you think we shouldn't take this strategy when updating the tests?

The approach is what we previously discussed so I feel that this PR is in the direction that we want to go. It is progress, so we can merge it as is. 

I would however prefer if subsequent PRs would reuse the macros for  `__m128i` for `__m128` or make the APIs of these two types at least consistent (we probably want to do the same for all other `__m` types). Also I would be fine if subsequent PRs just replace the types in the intrinsics to use the `_m` types and add enoughs `.into` in the tests to make them compile. What is done in this PR is not wrong (and if you prefer to do it like this and volunteer to do it I won't mention it again) but because I don't think doing it like this is worth the effort right now.

alexcrichton:
Ok sounds good to me! In that case I'll merge...

coresimd/src/x86/mod.rs
```diff
@@ -3,6 +3,11 @@
 #[macro_use]
 mod macros;
 
+#[repr(simd)]
+#[derive(Clone, Copy, Debug)]
+#[allow(non_camel_case_types)]
+pub struct __m128(f32, f32, f32, f32);
```
gnzlbg: Docs missing on this type.

alexcrichton: Oops yeah I'll get around to that.

PR #267, opened 2018-01-06
Fix `_mm_setzero_si64`

TSPMP:
The previous FIXME's do not seem to be correct since the author
previously expected the resulting instruction to be `xorps` but
according to the Intel intrinsics document
(https://gist.github.com/anonymous/25d752fda8521d29699a826b980218fc)
`xorps` is a SSE instruction and the expected instruction for
`_mm_setzero_si64` is `pxor`. Looking at the clang header `mmintrin.h`,
however, the comments say they expect the `vxorps`/`xorps` instructions,
too. Using the `pxor` instruction results in
```
pxor %mm0,%mm0
```
assembly, though, which seems correct since it uses the `%mmX` type
registers.

### DIFF
--- a/coresimd/src/x86/i686/mmx.rs
+++ b/coresimd/src/x86/i686/mmx.rs
@@ -14,14 +14,13 @@ use core::mem;
 #[cfg(test)]
 use stdsimd_test::assert_instr;
 
-/// Constructs a 64-bit integer vector initialized to zero.
+/// Return vector of type `__m64` with all elements set to zero.
 #[inline(always)]
 #[target_feature = "+mmx"]
-// FIXME: this produces a movl instead of xorps on x86
-// FIXME: this produces a xor intrinsic instead of xorps on x86_64
-#[cfg_attr(all(test, target_arch = "x86_64"), assert_instr(xor))]
+#[cfg_attr(test, assert_instr(pxor))]
 pub unsafe fn _mm_setzero_si64() -> __m64 {
-    mem::transmute(0_i64)
+    let a: __m64 = mem::uninitialized();
+    pxor(a, a)
 }
 
 /// Convert packed 16-bit integers from `a` and `b` to packed 8-bit integers
@@ -131,6 +130,8 @@ pub unsafe fn _mm_unpacklo_pi32(a: __m64, b: __m64) -> __m64 {
 
 #[allow(improper_ctypes)]
 extern "C" {
+    #[link_name = "llvm.x86.mmx.pxor"]
+    fn pxor(a: __m64, b: __m64) -> __m64;
     #[link_name = "llvm.x86.mmx.packsswb"]
     fn packsswb(a: __m64, b: __m64) -> __m64;
     #[link_name = "llvm.x86.mmx.packssdw"]


alexcrichton:
Thanks! Loks like CI may be failing though? Unfortunately I'm not sure this use of uninitialized memory will work, I think in release mode it's undefined behavior.

TSPMP:
It looks like you're right :( In theory, it shouldn't be undefined behaviour as any value of type `__m64` will be zeroed by `pxor(__m64, __m64)`. However, it looks like the compiler cannot see this and for `_mm_cvtps_pi8`, which uses `_mm_setzero_si64`,
```rust
pub unsafe fn _mm_setzero_si64() -> __m64 {
    let a: __m64 = mem::uninitialized();
    pxor(a, a)
}
```
results in
```
---- x86::i686::sse::assert__mm_cvtps_pi8_foo stdout ----
        disassembly for coresimd::x86::i686::sse::_mm_cvtps_pi8:
         0: push %rbp
         1: mov %rsp,%rbp
         2: cvtps2pi %xmm0,%mm0
         3: vpermilpd $0x3,%xmm0,%xmm0
         4: cvtps2pi %xmm0,%mm1
         5: packssdw %mm1,%mm0
         6: pxor %mm0,%mm1
         7: packsswb %mm1,%mm0
         8: movq2dq %mm0,%xmm0
         9: pop %rbp
        10: retq
        11: nop
```
note the `pxor %mm0,%mm1`. :(((( Any idea how to fix this?

alexcrichton:
Nah yeah unfortunately I'm not sure how to fix it, but it's also trivial-ish to verify by hand so I'm not too worried about getting this wrong :)

gnzlbg:
> Unfortunately I'm not sure this use of uninitialized memory will work, I think in release mode it's undefined behavior.

If `xor`ing uninitialized memory with itself is undefined behavior that is something that might be worth defining or fixing.

gnzlbg:
> note the `pxor %mm0,%mm1`. :(((( Any idea how to fix this?

You can also try with the `mem::transmute(0_i64)` and then doing the `pxor`, maybe? That way you at least avoid undefined behavior.

gnzlbg:
I asked about whether the undefined behavior was ok (it isn't) and whether there was a way to make it defined (it can't) so actually what we are already doing is what we should be doing and if LLVM emits the wrong code then that might be an LLVM bug. I've opened: https://bugs.llvm.org/show_bug.cgi?id=35869

We should definitely fix the FIXME to say that per the spec this should generate `pxor` instead of `xorps` and link to the LLVM bug report. It might well be that `xorps` is better than `pxor` in general for "reasons" and that the current behavior is the intended one.

gnzlbg:
So the LLVM dev said that LLVM chooses `pxor`/`xorps` depending on how the intrinsic is used. If it is used with an integer in the same context `pxor` will be choosen because it is faster, otherwise `xorps` is generally preferred because it has a shorter encoding. 

Also the x86_64-ABI turns `__m64` in function return types from MMX into SSE registers. That is, using `pxor` would mean that the return value would need to be turned into an SSE register afterwards, which is slower than just using `xorps` in an SSE register in the first place (this is why `assert_instr` detects `xorps`).

gnzlbg:
LLVM 6 should generate better code once this patch lands: https://reviews.llvm.org/D41908

TSPMP:
Wow! Thanks @gnzlbg for taking this to the LLVM devs. Ah, I was not aware that MMX function return types were moved to SSE registers according to the x86_64 ABI. This explains why the assembly code actually showed moving the `%mm0` to the `%xmm0` register after the `pxor %mm0, %mm0` which I left out in the PR. I guess we'll have to wait for the patch to land then. I will close this PR. @gnzlbg will you update the FIXME to link to the LLVM patch or do you want me to do it?

gnzlbg:
> I will close this PR. @gnzlbg will you update the FIXME to link to the LLVM patch or do you want me to do it?

I tried to updated to:

```rust
// FIXME: this produces a movl instead of xorps on x86
// FIXME: this produces a xor instead of xorps on x86_64
// NOTE (x86_64): This intrinsic should produce a pxor
// but because mmx registers are passed via SSE registers 
// in the x86_64 ABI LLVM will generate a xorps instruction
// here. 
// 
// This will work better once Rust migrates to LLVM 6:
//  * https://bugs.llvm.org/show_bug.cgi?id=35869
//  * https://reviews.llvm.org/D41908
```

but I don't recall why we test for `xor` and not `xorps`. That is, why does rustc produce a `xor` in `x86_64` instead of `xorps` ? I think there is still a bug here?

PR #273, opened 2018-01-10
Migrate a bunch of i586::sse2 to native types

alexcrichton:


coresimd/src/x86/i586/sse3.rs
```diff
@@ -131,57 +131,58 @@ mod tests {
 
     use v128::*;
     use x86::i586::sse3;
+    use x86::*;
 
     #[simd_test = "sse3"]
-    unsafe fn _mm_addsub_ps() {
+    unsafe fn tset_mm_addsub_ps() {
```
gnzlbg: typo: tset instead of test on this file: `sed s/tset/test/g` this file

coresimd/src/x86/i586/sse2.rs
```diff
@@ -582,149 +587,150 @@ pub unsafe fn _mm_or_si128(a: __m128i, b: __m128i) -> __m128i {
...
 }
 
 /// Compare packed 8-bit integers in `a` and `b` for equality.
 #[inline(always)]
 #[target_feature = "+sse2"]
 #[cfg_attr(test, assert_instr(pcmpeqb))]
-pub unsafe fn _mm_cmpeq_epi8(a: i8x16, b: i8x16) -> i8x16 {
-    a.eq(b)
+pub unsafe fn _mm_cmpeq_epi8(a: __m128i, b: __m128i) -> __m128i {
+    mem::transmute::<i8x16, _>(simd_eq(a.as_i8x16(), b.as_i8x16()))
```
gnzlbg: Is the `::< >` necessary here? All types are fixed, same below.

coresimd/src/x86/i586/sse2.rs
```diff
@@ -744,51 +750,51 @@ pub unsafe fn _mm_set_epi16(
 pub unsafe fn _mm_set_epi8(
     e15: i8, e14: i8, e13: i8, e12: i8, e11: i8, e10: i8, e9: i8, e8: i8,
     e7: i8, e6: i8, e5: i8, e4: i8, e3: i8, e2: i8, e1: i8, e0: i8,
-) -> i8x16 {
+) -> __m128i {
     #[cfg_attr(rustfmt, rustfmt_skip)]
-    i8x16::new(
+    mem::transmute(i8x16::new(
```
gnzlbg: I guess we still need the portable vector types to implement some of the intrinsics, at least for the constructors.

coresimd/src/x86/mod.rs
```diff
@@ -8,6 +10,70 @@ mod macros;
...
+pub use v64::__m64;
+
+#[cfg(test)]
+mod test;
+#[cfg(test)]
+pub use self::test::*;
+
+#[doc(hidden)]
+#[allow(non_camel_case_types)]
+trait m128iExt: Sized {
```
gnzlbg: This is somewhat smaller than what the macros were doing, but we'll how it evolves as the avx intrinsics get ported.

alexcrichton:
@gnzlbg 

> I honestly find that removing the portable vector types from the implementation of the intrinsics makes the code a bit harder to read

It's true yeah I agree too. I believe though the set of types provided in, for example, the `v128` module, are critical to actually implementing the intrinsics though in terms of that's how we get LLVM to recognize them. AFAIK there's no way to implement the intrinsics without *some* notion of `i8x16` for example to define various intrinsics with.

At that point it's sort of just a question of then how do we actually implement the intrinsics? Do we try to make use of the full API on `i8x16` or not? I figured that for now it's probably best to stick to raw LLVM things if we can to give us maximal flexibility to tweak and change `i8x16` without having to change thousands of intrinsics later on.

In general though I don't think changing our minds later on will cost too much. These are just internal implementations so they can change at any time, meaning that if we decide to change them we just also decide to bring on a lot of debt :)

coresimd/src/x86/mod.rs
alexcrichton: Yeah this was just what I felt was needed to define the various intrinsics, although it may be a better idea to separate out what these are converting too from the actual `::simd` portable types (they don't need to be the same)

coresimd/src/x86/i586/sse2.rs
alexcrichton: Unfortunately yeah, the [`simd_eq` intrinsic](https://github.com/rust-lang-nursery/stdsimd/blob/6ef0ae27aaa087b97b9482f22eaf8c9ed6cf9853/coresimd/src/simd_llvm.rs#L6) has a generic return type, but I believe that generic return type needs to at least have the same number of lanes as the comparison. That means that we just have to constrain `U` there to say what it needs to be.

PR #276, opened 2018-01-11
Migrate i586::sse41 to vendor types

alexcrichton:


alexcrichton:
@gnzlbg I also don't mean to overwhelm you at all, if you don't want to be tagged as reviewer on all these just let me know!

PR #277, opened 2018-01-12
Make rustc's job a little esaier in sse42

alexcrichton:
Move all the casts from `__m128i` to `i8x16` outside the macro invocations so
rustc only has to resolve a few function calls, not thousands!


TheIronBorn:
Ah this might help my build speed. Thanks!

PR #279, opened 2018-01-12
Migrate the i686 module to vendor types

alexcrichton:


coresimd/src/x86/i686/mmx.rs
```diff
@@ -185,6 +185,69 @@ pub unsafe fn _mm_unpacklo_pi32(a: __m64, b: __m64) -> __m64 {
...
+    mem::transmute(i32x2::new(e0, e1))
+}
+
+/// Set packed 8-bit integers in dst with the supplied values in reverse order.
+#[inline(always)]
+#[target_feature = "+mmx"]
+pub unsafe fn _mm_setr_pi8(e0: i8, e1: i8, e2: i8, e3: i8, e4: i8, e5: i8, e6: i8, e7: i8) -> __m64 {
+    mem::transmute(i8x8::new(e0, e1, e2, e3, e4, e5, e6, e7))
+}
+
```
gnzlbg: where do these come from?

gnzlbg: ah you need them for the tests :)

coresimd/src/x86/i686/sse42.rs
```diff
@@ -10,22 +12,21 @@ use stdsimd_test::assert_instr;
...
     use stdsimd_test::simd_test;
 
     #[simd_test = "sse4.2"]
-    unsafe fn _mm_cmpgt_epi64() {
-        let a = i64x2::splat(0x00).replace(1, 0x2a);
-        let b = i64x2::splat(0x00);
-        let i = sse42::_mm_cmpgt_epi64(a, b);
-        assert_eq!(i, i64x2::new(0x00, 0xffffffffffffffffu64 as i64));
+    unsafe fn test_mm_cmpgt_epi64() {
+        let a = _mm_setr_epi64x(0, 0x2a);
```
gnzlbg: which intrinsic is this?

coresimd/src/x86/i686/mmx.rs
alexcrichton: ah yeah these are defined in the mmx suite, just weren't bound yet!

coresimd/src/x86/i686/sse42.rs
alexcrichton: Heh a good question! All types so far have had an `_mm_set_XX` constructor which sets all the lanes, and *most* have a `setr` which is the opposite order of specifying the lanes. The existing methods like `i64x2::new` are the same order as the `setr` intrinsics, so I've been mechanically changing `foo::new` to `_mm_setr_foo`. Unfortunately though `i64x2`'s equivalent *only* has `_mm_set_epi64x` and doesn't have the `setr` equivalent, so I just [vendored it](https://github.com/rust-lang-nursery/stdsimd/blob/master/coresimd/src/x86/test.rs#L34-L39) to allow easy translation.

It should maybe have a name that indicates it's not an intrinsic though...

alexcrichton: (note that the intrinsic here is only defined for tests, not exported from the library)

PR #283, opened 2018-01-16
Update `target_feature` syntax

alexcrichton:
This commit updates to the latest nightly's syntax where `#[target_feature =
"+foo"]` is now deprecated in favor of `#[target_feature(enable = "foo")]`.
Additionally `#[target_feature]` can only be applied to `unsafe` functions for
now.

Along the way this removes a few exampels that were just left around and also
disables the `fxsr` modules as that target feature will need to land in upstream
rust-lang/rust first as it's currently unknown to the compiler.


alexcrichton:
This was like 99% a sed script btw, and if desired I can try to split it up more based on the script

gnzlbg:
Is there a PR for enabling `fxsr` upstream or should I do that?

coresimd/src/x86/i586/sse42.rs
```diff
@@ -186,8 +186,8 @@ pub unsafe fn _mm_cmpistrm(a: __m128i, b: __m128i, imm8: i32) -> __m128i {
 /// #
 /// # fn main() {
 /// #     if cfg_feature_enabled!("sse4.2") {
-/// #         #[target_feature = "+sse4.2"]
-/// #         fn worker() {
+/// #         #[target_feature(enable = "sse4.2")]
+/// #         unsafe fn worker() {
```
gnzlbg: Is `unsafe` required now? Cool!

examples/types.rs
```diff
@@ -1,37 +0,0 @@
-#![cfg_attr(feature = "strict", deny(warnings))]
-#![feature(target_feature)]
```
gnzlbg: What happened to `examples/types.rs` and `examples/wat.rs` ?

coresimd/src/x86/i586/sse42.rs
alexcrichton: It is indeed!

examples/types.rs
alexcrichton: Ah these just looked like things that were playing around and not actually legitimate examples (in the sense of things we wanted to maintain over time) so I just deleted them.

gnzlbg: I also thought this, but never really looked into them.

gnzlbg:
So LGTM, merge once tests pass!

alexcrichton:
Ah looks like rustc also doesn't know about `v7` on arm!

PR #305, opened 2018-01-29
Verify intrinsics don't leak to i686

### LINKED ISSUE #304: Some x86_64 only intrinsics are available on x86
Intrinsics like `_mm_cvttsd_si64` are currently only defined on x86_64. Looks like [clang agrees](https://github.com/llvm-mirror/clang/blob/09aa26c22dcad5968beaafbe44d017a6780e13c5/lib/Headers/emmintrin.h#L3354-L3413) that these should be x86_64 only!

However we also have intrinsics like `_mm256_insert_epi64` which for us is available on x86/x86_64 targets but on clang is [only available on x86_64](https://github.com/llvm-mirror/clang/blob/09aa26c22dcad5968beaafbe44d017a6780e13c5/lib/Headers/avxintrin.h#L2172-L2198).

We should figure out a way to verify whether an intrinsic should only be defined on x86_64 and then add that to stdsimd-verify!

alexcrichton:
Some intrinsics take `i64` or `u64` arguments which typically means that they're
using 64-bit registers and aren't actually available on x86. This commit adds a
check to stdsimd-verify to assert this and moves around some intrinsics that I
believe should only be available on x86_64.

This commit was checked in many places against gcc/clang/MSVC using godbolt.org
to ensure that we're agreeing with what other compilers are doing.

Closes #304

### DIFF
--- a/coresimd/src/lib.rs
+++ b/coresimd/src/lib.rs
@@ -15,7 +15,7 @@
            simd_ffi, target_feature, cfg_target_feature, i128_type, asm,
            const_atomic_usize_new, stmt_expr_attributes, core_intrinsics,
            crate_in_paths)]
-#![cfg_attr(test, feature(proc_macro, test, repr_align, attr_literals))]
+#![cfg_attr(test, feature(proc_macro, test, attr_literals))]
 #![cfg_attr(feature = "cargo-clippy",
             allow(inline_always, too_many_arguments, cast_sign_loss,
                   cast_lossless, cast_possible_wrap,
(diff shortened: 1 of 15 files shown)


alexcrichton:
Note that the "verification" here is purely "look for intrinsics that take i64 or u64 as an argument", but it looks like even that has a [few exceptions](https://github.com/alexcrichton/stdsimd/blob/c0976e8a7f4d7205c0e339cd6d58c87ae4dd1348/stdsimd-verify/tests/x86-intel.rs#L236-L255).

PR #311, opened 2018-01-30
Add x86 AES-NI vendor intrinsics

### LINKED ISSUE #295: Implement all x86 AES-NI vendor intrinsics
#40 makes no mention of the [AES-NI](https://en.wikipedia.org/wiki/AES_instruction_set) intrinsics, so I am opening an issue to track them here. See also the [Intel intrinsics guide](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#othertechs=AES).

  * [ ] [`_mm_aesdec_si128`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_aesdec_si128&expand=5236)
  * [ ] [`_mm_aeskeygenassist_si128`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_aeskeygenassist_si128&expand=5236)
  * [ ] [`_mm_aesenc_si128`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_aesenc_si128&expand=5236)
  * [ ] [`_mm_aesenclast_si128`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_aesenclast_si128&expand=5236)
  * [ ] [`_mm_aesimc_si128`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_aesimc_si128&expand=5236)
  * [ ] [`_mm_aesdeclast_si128`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_aesdeclast_si128&expand=5236)

ruuda:
Fixes #295.

I have not tested this yet, waiting for a nightly that includes https://github.com/rust-lang/rust/pull/47822/commits/2497d10ff34bf4d03283e4562143762426492789 to be released.

Also, the [MSDN documentation](https://msdn.microsoft.com/en-us/library/cc714206(v=vs.100).aspx) is a bit clearer and more comprehensive than the Intel intrinsics guide (thanks @gnzlbg for pointing this out), but I am not sure if we could copy it into the docs license-wise. In the worst case we could add links to the MSDN docs.


coresimd/src/x86/x86_64/aes.rs
```diff
@@ -0,0 +1,139 @@
...
+    aeskeygenassist(a, imm8)
+}
+
+#[cfg(test)]
+mod tests {
+    use stdsimd_test::simd_test;
+
+    use x86::*;
+
+    #[simd_test = "aex"]
```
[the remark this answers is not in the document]
gnzlbg: Same issue below.

alexcrichton:
Thanks! These are currently in `x86_64` but that means they're only compiled in the `x86_64-*` targets, but I think these can move into `i686` to be on 32-bit as well!

coresimd/src/x86/x86_64/aes.rs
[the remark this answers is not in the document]
ruuda: Fixed.

coresimd/src/x86/x86_64/aes.rs
```diff
@@ -0,0 +1,139 @@
...
+
+/// Assist in expanding the AES cipher key.
+///
+/// Assist in expanding the AES cipher key by computing steps towards generating
+/// a round key for encryption cipher using data from `a` and an 8-bit round constant.
+#[inline]
+#[target_feature(enable = "aes")]
+#[cfg_attr(test, assert_instr(aeskeygenassist))]
+pub unsafe fn _mm_aeskeygenassist_si128(a: __m128i, imm8: u8) -> __m128i {
+    aeskeygenassist(a, imm8)
```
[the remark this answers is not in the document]
ruuda: With constify it does work, indeed.

ruuda:
The tests now pass on `rustc 1.25.0-nightly (def3269a7 2018-01-30)`.

> Thanks! These are currently in x86_64 but that means they're only compiled in the x86_64-* targets, but I think these can move into i686 to be on 32-bit as well!

Can they? [Intel does not list any 32-bit CPUs with AES-NI](https://ark.intel.com/Search/FeatureFilter?productType=processors&AESTech=true&InstructionSet=32-bit), unless I am missing something.

> I would just use that as "inspiration" for the tests. For the docs you can check the clang-docs. IIRC @alexcrichton suggested that we could base our docs on those and then just add attribution on the readme.

The constants in the tests should not be an issue. The documentation comments in Clang’s [`__wmmintrin_aes.h`](https://clang.llvm.org/doxygen/____wmmintrin__aes_8h_source.html) do not seem much more useful that what we have now. What I was referring to is the [remarks section](https://msdn.microsoft.com/en-us/library/cc664949(v=vs.100).aspx#Anchor_3) of MSDN docs, which contains a bit more information on how these instructions are supposed to be used together.

alexcrichton:
@ruuda ah that's true, but making making code compatible across x86/x86_64 tends to be easier if it's available everywhere. The `cfg_feature_enabled!` would probably just always return false on i686 which means they'd never be called anyway.

I think it also matches what clang/gcc do, making the intrinsics available on both platforms.

coresimd/src/x86/x86_64/aes.rs
```diff
@@ -0,0 +1,144 @@
...
+        let e = __m128i(0xc66c82284ee40aa0, 0x6633441122770055);
+        let r = _mm_aesimc_si128(a);
+        assert_eq_m128i(r, e);
+    }
+
+    #[simd_test = "aes"]
+    unsafe fn test_mm_aeskeygenassist_si128() {
+        // Constants taken from https://msdn.microsoft.com/en-us/library/cc714138.aspx.
+        let a = __m128i(0x0123456789abcdef, 0x8899aabbccddeeff);
+        let e = __m128i(0x857c266b7c266e85, 0xeac4eea9c4eeacea);
```
alexcrichton: Could this use `_mm_set_epi64x` to avoid making an assumption about the internal layout of `__m128i`?

ruuda: Done.

ruuda:
> making making code compatible across x86/x86_64 tends to be easier if it's available everywhere

Ah, I see, that makes sense. Moved them into i686.

gnzlbg:
@ruuda 

> Can they? Intel does not list any 32-bit CPUs with AES-NI, unless I am missing something.

64-bit CPUs can run 32-bit binaries. AFAIK 32-bit binaries can call 64-bit CPU intrinsics without issues, but they obviously won't be able to run on CPUs that don't have these intrinsics.

alexcrichton:
Looks like stdsimd's runtime detection may need to be taught the aes feature?

gnzlbg:
@alexcrichton yes, I am on it.

alexcrichton:
@ruuda want to rebase on https://github.com/rust-lang-nursery/stdsimd/pull/312 and see if the tests are fixed/

ruuda:
>     thread 'x86::i686::aes::assert__mm_aeskeygenassist_si128_aeskeygenassist' panicked at 'instruction found, but the disassembly contains too many instructions: #instructions = 781 >= 20 (limit)', stdsimd-test/src/lib.rs:375:9

Looks like that constify macro expanded to an enormous function ... is there anything we can do about that?

coresimd/src/x86/i686/aes.rs
```diff
@@ -0,0 +1,144 @@
...
+    aesimc(a)
+}
+
+/// Assist in expanding the AES cipher key.
+///
+/// Assist in expanding the AES cipher key by computing steps towards generating
+/// a round key for encryption cipher using data from `a` and an 8-bit round constant.
+#[inline]
+#[target_feature(enable = "aes")]
+#[cfg_attr(test, assert_instr(aeskeygenassist))]
```
gnzlbg: Change this

```rust
assert_instr(aeskeygenassist)
```

to (from the top of my head):

```rust
assert_instr(aeskeygenassist, imm8 = 0)
```

That should fix that.

gnzlbg: Just to explain what's going on, `imm8` must be a compile-time argument, and while testing the disassembly, we need to pass it a value. If one doesn't specify it in the `assert_instr` macro, it will pass it a run-time value, and that won't work for this intrinsic. 

We don't have a way of expressing that function arguments must be const yet.

ruuda: I see, thanks for the explanation. Should be fixed now.

I was a bit worried that this meant that the entire match statement would end up in the final binary, so I also verified the disassembly of `test_mm_aeskeygenassist_si128` in release mode, and fortunately it does get const-folded:

    0000000000041720 <coresimd::x86::i686::aes::tests::test_mm_aeskeygenassist_si128::test_mm_aeskeygenassist_si128>:
       41720:       55                      push   %rbp
       41721:       48 89 e5                mov    %rsp,%rbp
       41724:       48 81 ec 80 00 00 00    sub    $0x80,%rsp
       4172b:       66 0f 3a df 05 3b c7    aeskeygenassist $0x5,0x11c73b(%rip),%xmm0        # 15de70 <str.4y+0x20>
       41732:       11 00 05 
       41735:       66 0f 7f 45 90          movdqa %xmm0,-0x70(%rbp)
       4173a:       0f 28 05 9f c7 11 00    movaps 0x11c79f(%rip),%xmm0        # 15dee0 <str.4y+0x90>
       41741:       0f 29 45 80             movaps %xmm0,-0x80(%rbp)

In debug mode it doesn’t though, instead there is a call to `_mm_aeskeygenassist_si128` which contains the entire match statement. Something along the lines of rust-lang/rust#47980 would be good to have indeed.

ruuda:
The only thing failing at this point, is that the Intel xml file claims that the `aeskeygenassist` argument has type `const int`. The description clearly specifies that it is an “8-bit round constant” though. Other intrinsics with an `imm8` argument specify it as `i32`. Do we want to blindly copy this idiosyncrasy from the C API? Or shall I update the test to allow `const int imm8` arguments to be `u8`? The right instruction is emitted either way.

alexcrichton:
@ruuda for now I'd stick exactly to what Intel says, and Intel says `int` which we translate to `i32`, so I'd switch to `i32`.

alexcrichton:
Bah it looks like that yield an LLVM error. Maybe the LLVM intrinsic needs a `u8`?

ruuda:
It compiles :tada:

gnzlbg:
Nice! Now you just need to fix the formatting and the clippy issues and this is good to go! (see the last two build bots!)

alexcrichton:
Thanks @ruuda!

PR #312, opened 2018-02-01
Run-time feature detection for AES-NI and TSC

gnzlbg:
TSC doesn't have an LLVM feature, so IIUC we still can't use `simd_test` for it (its enabled for all x86 targets). Anyways, offering run-time feature detection for it should do no harm.


gnzlbg:
I wish we would make the clippy and and rustfmt builds fail hardly. Otherwise every time I do a commit and want to format things appropriately I get a lot of noise in other parts of the library. The same applies to clippy: when I want to check that I haven't written something unnecessarily complex: `.map(...).unwrap_or` vs `.map_or`, `foo.iter()` vs `&foo`, `xxx` vs `xxx_with` (I always forget that the `_with` methods exist...), unnecessary `&foo`s, unnecessary `transmutes`, etc., etc., etc. I get a lot of noise from everywhere in the library as well.

gnzlbg:
So... it seem the feature cache was full on 32-bit because this PR just added the 32th run-time feature test and I was using an AtomicUsize for the cache...

So... I have added a `debug_assert!` for this, which although not perfect, should catch this in the future as long as we test the run-time feature detection system a bit (which we do, against `cupid`). 

I have also increased the size of the cache on 32-bit by making it an `AtomicU64`. The underlying type of the cache was previously being leaked everywhere, so I've compartmentalized it a bit more, so that in the future increasing the cache again to, for example, be implemented using a pair of `AtomicU64`s can be done locally.

alexcrichton:
Hm I think `AtomicU64` may not be quite as portable as we might like, but on 32-bit could we use a pair of two `AtomicUsize`?

alexcrichton:
(otherwise though this looks good to me!)

gnzlbg:
@alexcrichton Is there a reason to prefer AtomicUsize over AtomicU32 on 32-bit ?

alexcrichton:
Oh I just tend to prefer stable things to unstable things, but either should work fine!

gnzlbg:
@alexcrichton we would still the `new` methods to be `const fn` for `AtomicUsize` to work here without unstable features because of the `static` :/

PR #319, opened 2018-02-10
Add #[rustc_args_required_const] annotations

alexcrichton:
Support isn't quite in nightly to make this work yet, but using a local build
this gets everything passing again! This also implements native verification
that we have the attribute in the right place

### DIFF
--- a/coresimd/src/x86/i586/avx.rs
+++ b/coresimd/src/x86/i586/avx.rs
@@ -96,6 +96,7 @@ pub unsafe fn _mm256_or_ps(a: __m256, b: __m256) -> __m256 {
 #[inline]
 #[target_feature(enable = "avx")]
 #[cfg_attr(test, assert_instr(vshufpd, imm8 = 0x1))]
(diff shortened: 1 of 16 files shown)


alexcrichton:
FWIW this is blocked on https://github.com/rust-lang/rust/pull/48078 reaching nightlies

coresimd/src/x86/i586/avx.rs
```diff
@@ -96,6 +96,7 @@ pub unsafe fn _mm256_or_ps(a: __m256, b: __m256) -> __m256 {
 #[inline]
 #[target_feature(enable = "avx")]
 #[cfg_attr(test, assert_instr(vshufpd, imm8 = 0x1))]
+#[rustc_args_required_const(2)]
```
petrochenkov: It looks like supporting attributes attached directly to function parameters would help with cases like this.
*\*Adds it to the list of things to implement some time\**

alexcrichton: Indeed!

PR #320, opened 2018-02-10
CLMUL instruction set

### LINKED ISSUE #318: Implement CLMUL instruction set
Carry-less Multiplication (CLMUL) allows to improve the speed of applications doing block cipher encryption in Galois/Counter Mode, which depends on finite field GF(2^k) multiplication. Another application is the fast calculation of CRC values, including those used to implement the LZ77 sliding window DEFLATE algorithm in zlib and pngcrush. (wiki)

newpavlov:
Fixes #318

Depends on rust-lang/rust#48126.

### DIFF
--- a/coresimd/src/x86/i686/clmul.rs
+++ b/coresimd/src/x86/i686/clmul.rs
@@ -1,0 +1,58 @@
+use x86::__m128i;
+
+#[cfg(test)]
+use stdsimd_test::assert_instr;
+
+#[allow(improper_ctypes)]
+extern "C" {
+    #[link_name = "llvm.x86.pclmulqdq"]
+    fn pclmulqdq(a: __m128i, round_key: __m128i, imm8: u8) -> __m128i;
+}
+
+/// Perform a carry-less multiplication of two 64-bit polynomials over the
+/// finite field GF(2^k).
+///
+/// The immediate byte is used for determining which halves of `a` and `b`
+/// should be used. Immediate bits other than 0 and 4 are ignored.
+#[inline]
+#[target_feature(enable = "pclmul")]
+#[cfg_attr(test, assert_instr(pclmulqdq, imm8 = 0))]
+pub unsafe fn _mm_clmulepi64_si128(a: __m128i, b: __m128i, imm8: u8) -> __m128i {
+    macro_rules! call {
+        ($imm8:expr) => (pclmulqdq(a, b, $imm8))
+    }
+    constify_imm8!(imm8, call)
+}
+
+
+#[cfg(test)]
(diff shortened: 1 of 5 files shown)


alexcrichton:
Thanks! The verification program is reporting 

  * intel cpuid `pclmulqdq` not in `pclmul` for _mm_clmulepi64_si128

Perhaps the feature in Rust itself could be called `PCLMULQDQ`?

newpavlov:
We have 3 options for feature name:
- `clmul` - shortest one, coherent with intrinsic name `_mm_clmulepi64_si128`, matches with instruction set name
- `pclmul` - used by LLVM
- `pclmulqdq` - longest one, matches with CPUID and instruction mnemonic.

I've chose `pclmul` because it required simple whitelisting, while other options will require more involved changes, which I am currently don't know how to implement.

alexcrichton:
So far I think our intention is to provide a direct mapping from what vendors (Intel in this case) says. So far everything (except bmi/bmi1) has aligned with LLVM but I think for new ones we'll want to align with Intel rather than LLVM.

newpavlov:
Can you please list necessary changes for adding `pclmulqdq` feature which will use `pclmul` LLVM feature under the hood? (I think it's better to discuss in the relevant Rust PR)

alexcrichton:
Sure! I'll comment on the Rust PR

alexcrichton:
@newpavlov I think these should be in nightly now, right? If so, want to rebase to kick CI?

newpavlov:
I will change `imm8: u8` to `imm8: i32` to be coherent with other intrinsics.

Is it worth to open an issue for discussing potential change of `i32` to `u8` or `i8` for immediate 8 bits arguments? I understand `i32` was chosen to be coherent with headers, but I personally don't think we should follow them in this case.

newpavlov:
Hm, not sure how to fix this. In [some cases](https://travis-ci.org/rust-lang-nursery/stdsimd/jobs/342902348) `_mm_clmulepi64_si128(a, b, 0x00)` gets compiled to `pclmullqlqdq %xmm1,%xmm0 ` and in [others](https://travis-ci.org/rust-lang-nursery/stdsimd/jobs/342902356) to `pclmulqdq $0x0, %xmm1, %xmm0 `. IIUC both mnemonics are equivalent.

alexcrichton:
Ah no worries, wanna only run that test on Linux and leave a fixme?

newpavlov:
Ah, indeed I could've run OS specific tests. I think last commit should fix it.

alexcrichton:
Awesome, thanks!

PR #321, opened 2018-02-10
Fix instruction assertions on LLVM 6

alexcrichton:
Looks like some instructions changed here and there, so this updates the
assertions (no behavior appears to have changed though)


alexcrichton:
Note that to land this should wait for a new nightly to guarantee it's all green

Issue #325, opened 2018-02-18
Imlement rdrand and rdseed intrinsics


newpavlov:
- `_rdrand16_step`
- `_rdrand32_step`
- `_rdrand64_step`
- `_rdseed16_step`
- `_rdseed32_step`
- `_rdseed64_step`

It will require addition of two target features `rdrand` and `rdseed` to the language.

cc @nagisa

newpavlov:
I don't quite get how to link to the LLVM intrinsic `llvm.x86.rdrand.64`, IUUC correctly it has two return values `i64` and `i32`. All options which I've tried result in "Intrinsic has incorrect return type!". Can someone help with it?

nagisa:
You cannot link to this intrinsic this way, because it returns {i1,
i<width>}, which is not expressible in a stable manner (there's no
guarantee that any rust type will lower down into that particular LLVM
type).

nagisa:
I misremembered. The type is {I<width>, i32} but the general point about it
not being expressible stands.

alexcrichton:
@newpavlov I think it's ok to use a `asm!` block temporarily for this, but yeah as @nagisa mentioned we'll need to add a dedicated intrinsic to rustc for this otherwise to get the ABI to work.

nagisa:
rustc has the dedicated intrinsic. It is used by the rdrand crate.

newpavlov:
@alexcrichton 
Rust already supports `rdrnd` and `rdseed` features. But `rdrnd` is a LLVM feature name, while with `pclmulqdq` it was decided to use vendor name. What do you think about renaming this feature to `rdrand` for consistency? Will this change be backwards compatible? This questions was earlier [raised](https://github.com/rust-lang/rust/pull/38561#issue-99176874) by @nagisa, but it was left unanswered.

nagisa:
The change wouldn’t be backwards compatible, but since anything related to it is unstable in the first place, I do not see any reason to do the epoch dance.

newpavlov:
@alexcrichton 
Do we have to follow intel intrinsics 1-to-1? It feels a bit weird to have signature `_rdrand64_step(val: &mut u64) -> i32` more natural for Rust will be `_rdrand64_step() -> (u64, bool)` or even `_rdrand64_step() -> Option<u64>`, plus it's closer to how instruction works.

gnzlbg:
> Do we have to follow Intel intrinsics 1-to-1? 

Yep :/

[...]

PR #326, opened 2018-02-20
Implemented rdrand and rdseed intrinsics

newpavlov:
Fixes: #325 
Depends on rust-lang/rust#48369.


coresimd/x86/i686/rdrand.rs
```diff
@@ -0,0 +1,80 @@
...
+//! on-chip hardware random number generator which has been seeded by an on-chip
+//! entropy source.
+extern "platform-intrinsic" {
+    fn x86_rdrand16_step() -> (u16, i32);
+    fn x86_rdrand32_step() -> (u32, i32);
+    fn x86_rdrand64_step() -> (u64, i32);
+    fn x86_rdseed16_step() -> (u16, i32);
+    fn x86_rdseed32_step() -> (u32, i32);
+    fn x86_rdseed64_step() -> (u64, i32);
+}
```
BurntSushi: We don't use `platform-intrinsic` in this crate. Can you link to the LLVM intrinsics directly?

newpavlov: No, as was discussed in the issue LLVM  type {I<width>, i32} can not be currently represented by Rust.

BurntSushi: I'm not sure that means we want to start using platform-intrinsics here.

We should probably wait for @alexcrichton to review this and help decide how to proceed.

newpavlov: He [suggested](https://github.com/rust-lang-nursery/stdsimd/issues/325#issuecomment-366839315) to use `asm!` block as a temporary solution, so I think platform-intrinsics will be an ok temporary solution as well.

alexcrichton: Ah yeah the platform intrinsic ABI here is the way to go for now. This may change in the future, but we can always just update the crate!

In any case this PR looks good to me, just waiting on the rust-lang/rust PR!

alexcrichton:
@newpavlov there's a [failure on CI](https://travis-ci.org/rust-lang-nursery/stdsimd/jobs/343838992) related to the verification of intrinsics, but I think it's benign and mostly just means there's a [missing case here](https://github.com/rust-lang-nursery/stdsimd/blob/0df4359c17779aab542e665d549546f932dac53b/crates/stdsimd-verify/tests/x86-intel.rs#L364-L433). Mind fixing that in this PR as well?

newpavlov:
@alexcrichton
Done!

alexcrichton:
Ok! I reran the builds with the new nightly, but it looks like there may be some errors?

newpavlov:
Not sure how to debug this error:
```
LLVM ERROR: Cannot select: 0x13893f8f0: ch = store<ST4[%9]> 0x13893fb60, 0x13893f208:1, 0x13893f068, undef:i32
  0x13893f068: i32 = add nuw FrameIndex:i32<0>, Constant:i32<8>
    0x13893f958: i32 = FrameIndex<0>
    0x13893f3a8: i32 = Constant<8>
  0x13893f680: i32 = undef
In function: _ZN8coresimd8coresimd3x864i6866rdrand14_rdrand64_step17hd58a2bbfbca06dcdE
```
Maybe it's a bug in the current nigthly? Can you please try to restart test for `master` and see if it will pass tests?

alexcrichton:
Looks like rdrand64 is only available on x86_64 targets

alexcrichton:
Oh I think it's just rdrand64, the other functions should be available on x86

alexcrichton:
Thanks!

PR #332, opened 2018-02-27
Remove `_xgetbv` for now

alexcrichton:
Looks like LLVM 6 may have removed the intrinsic? We can always figure out how
to re-include later with inline asm if necessary


alexcrichton:
Aha looks like we use it, so let's implement it...

gnzlbg:
We use it for run-time feature detection. I had an implementation that used inline assembly, might be able to find it. This is what clang does now: https://github.com/llvm-mirror/clang/blob/f9eb14961ddd9a48da63cdb792d06386287a5dfd/lib/Headers/intrin.h#L916

alexcrichton:
Oh actually I believe I misdiagnosed the error in rust-lang/rust. Rather the intrinsic wasn't removed but added after LLVM 3.9 I think (which is what Travis tests). In any case I think we'll want to switch to inline asm.

PR #338, opened 2018-03-02
Prepare portable packed vector types for RFCs

gnzlbg:
This commit cleans up the implementation of the Portable Packed SIMD Vectors
(PPSV), adds some new features, and makes some breaking changes.

The implementation is moved to `coresimd/ppsv` (they are
still exposed via `coresimd::simd`).

As before, the vector types of a certain width are implemented in the `v{width}`
submodules. The `macros.rs` file has been rewritten as an `api` module that
exposes the macros to implement each API.

It should now hopefully be really clear where each API is implemented, and which types
implement these APIs. It should also now be really clear which APIs are tested and how.

# Additions

- boolean vectors of the form `b{element_size}x{number_of_lanes}`.
- reductions: arithmetic (sum, product), bitwise (and, or, xor), min/max, and boolean (all, any, none) - mainly implemented via `llvm.experimental.vector.reduction.{...}`  modulo bugs.
- FromBits trait analogous to `{f32,f64}::from_bits` that perform "safe" transmutes.
  Instead of writing `From::from`/`x.into()` (see below for breaking changes) now one writes
  `FromBits::from_bits(x)`/`x.into_bits()`.
- portable vector types implement `Default` and `Hash`
- tests for all portable vector types and all portable operations (~2000 new tests; this does hurts compile-time of `cargo test` a lot, and increases the memory requirements...).
- (hopefully) comprehensive implementation of bitwise transmutes and lane-wise
  casts (before `From` and the `.as_...` methods where implemented "when they were needed").
- documentation for PPTV (not great yet, but better than nothing)
- conversions/transmutes from/to x86 architecture specific vector types

# Breaking changes

- `{store,load}{,_unchecked}` API has been replaced with `{store,load}_{aligned,unaligned}{,_unchecked}`
- `eq,ne,lt,le,gt,ge` APIs now return boolean vectors
- The `.as_{...}` methods have been removed. Lane-wise casts are now performed via `From`/`Into`.
- `From/Into` traits now perform lane-wise casts (see above). Previously they used to perform bitwise transmutes.
- `simd` vectors' `replace` method's result is now `#[must_use]`; executing `replace` and dropping the result is an easy to make error.


stdsimd/mod.rs
```diff
@@ -240,101 +241,6 @@
 /// we'll be using SSE4.1 features to implement hex encoding.
 ///
 /// ```
-/// #![feature(cfg_target_feature, target_feature, stdsimd)]
```
gnzlbg: I have no idea about what happened here. Did rustfmt delete all of this?

gnzlbg: Yep, it look like `rustfmt` just deleted all of this. @nrc does this look familiar?

coresimd/ppvt/api/fmt.rs
```diff
@@ -0,0 +1,53 @@
...
+            fn fmt(&self, f: &mut fmt::Formatter)
+                   -> fmt::Result {
+                write!(f, "{}(", stringify!($id))?;
+                let n = mem::size_of_val(self)
+                    / mem::size_of::<$elem_ty>();
+                for i in 0..n {
+                    if i > 0 {
+                        write!(f, ", ")?;
+                    }
+                    write!(f, "{:#x}", self.extract(i))?;
```
alexcrichton: I think here if you do `self.extract(i).fmt(f)` it'll automatically forward formatting flags like `#` which means we may not need to hardcode the `#`

coresimd/ppvt/api/hash.rs
```diff
@@ -0,0 +1,50 @@
...
+        impl hash::Hash for $id {
+            #[inline(always)]
+            fn hash<H: hash::Hasher>(&self, state: &mut H) {
+                union A {
+                    data: [$elem_ty; $id::lanes()],
+                    vec: $id
+                }
+                unsafe {
+                    let mut bytes: A = mem::uninitialized();
+                    self.store_aligned_unchecked(&mut bytes.data);
```
alexcrichton: Could this function be simplified to:

```rust
A { vec: *self }.data.hash(state)
```

stdsimd/mod.rs
```diff
@@ -1,343 +1,345 @@
...
+//!         static TABLE: &[u8] = b"0123456789abcdef";
+//!         TABLE[byte as usize]
+//!     }
+//!
+//!     for (byte, slots) in src.iter().zip(dst.chunks_mut(2)) {
+//!         slots[0] = hex((*byte >> 4) & 0xf);
+//!         slots[1] = hex(*byte & 0xf);
+//!     }
+//! }
+//! ```
```
alexcrichton: This was all actually intended to be the rustdoc documentation for the `arch` module (which will show up in libstd's docs soon), mind leaving it on the `arch` module instead of the `stdsimd` module?

gnzlbg: Ah! I got a complaint about the `stdsimd` module not having any documentation and thought this was the other way around, I'll change this.

coresimd/ppvt/api/bitwise_reductions.rs
```diff
@@ -0,0 +1,83 @@
+//! Implements portable bitwise vector reductions.
+
+macro_rules! impl_bitwise_reductions {
+    ($id:ident, $elem_ty:ident) => {
+        impl $id {
+            /// Lane-wise bitwise `and` of the vector elements.
+            pub fn and(self) -> $elem_ty {
```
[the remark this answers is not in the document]
gnzlbg: So all these 3 are provided by llvm. They are necessary to implement the reductions of boolean vectors (`all`,`any`,`none`), but maybe we shouldn't provide them for the integer types (llvm supports that though).

coresimd/ppvt/api/arithmetic_reductions.rs
```diff
@@ -0,0 +1,136 @@
+//! Implements portable arithmetic vector reductions.
+
+macro_rules! impl_arithmetic_reductions {
+    ($id:ident, $elem_ty:ident) => {
+        impl $id {
+            /// Lane-wise addition of the vector elements.
+            #[inline(always)]
+            pub fn add(self) -> $elem_ty {
```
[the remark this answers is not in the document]
gnzlbg: So llvm only provides ‘llvm.experimental.vector.reduce.{add, fadd, mul, fmul, and, or, xor, smax, smin, umax, umin}`, which means `sub`, `div`, and `rem` cannot really be implemented here any better than in a third party crate. I provided them for completeness, but I think it makes sense to leave them out . Writing tests for them felt weird.

coresimd/ppvt/api/bool_vectors.rs
```diff
@@ -0,0 +1,152 @@
+//! Minimal boolean vector implementation
+
+/// Minimal interface: all packed SIMD boolean vector types implement this.
+macro_rules! impl_bool_minimal {
+    ($id:ident, $elem_ty:ident, $elem_count:expr, $($elem_name:ident),+) => {
+
+        #[cfg_attr(feature = "cargo-clippy", allow(expl_impl_clone_on_copy))]
+        impl Clone for $id {
+            #[inline] // currently needed for correctness
```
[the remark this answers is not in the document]
gnzlbg: The tests passed without this, I just saw that the types in the `x86` module were still doing this and decided to add it for consistency. I can remove them there as well.

coresimd/ppvt/api/arithmetic_ops.rs
```diff
@@ -0,0 +1,142 @@
...
+        impl ops::Rem for $id {
+            type Output = Self;
+            #[inline(always)]
+            fn rem(self, other: Self) -> Self {
+                unsafe { simd_rem(self, other) }
+            }
+        }
+
+        impl ops::AddAssign for $id {
+            #[inline(always)]
```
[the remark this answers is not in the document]
gnzlbg: I thought about this and recalled that with ThinLTO `#[inline(always)]` should be necessary anymore, but then looked at what the types in the `x86` module were doing (and the vector types before that), and they were all still using `#[inline(always)]`. 

I think we should do a pass through the library and see if we can replace most of the `#[inline(always)]` attributes with just `#[inline]`.

coresimd/ppvt/api/hash.rs
gnzlbg: I'll give it a try.

I should fill an issue to discuss the semantics of this, it basically just has the vector as a slice of bytes. This is a bit differently of how `Hash` works for arrays, where first the length is hashes, and then each element is hashed.

coresimd/ppvt/api/arithmetic_reductions.rs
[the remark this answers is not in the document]
gnzlbg: Done, I've renamed add/mul to sum/product (just like the Iterator methods) and removed the sub/div/rem reductions.

coresimd/ppvt/api/fmt.rs
gnzlbg: Done.

coresimd/ppvt/api/hash.rs
gnzlbg: Done.

stdsimd/mod.rs
gnzlbg: Done.

coresimd/ppvt/api/arithmetic_ops.rs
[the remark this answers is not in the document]
gnzlbg: I've filled #340 for this.

coresimd/ppvt/api/bitwise_reductions.rs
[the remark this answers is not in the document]
gnzlbg: I am leaving these here for now since I can just map them directly to the llvm intrinsics. We might just decide to never stabilize these and expose them only for boolean vectors via `all`,`any`,`none`.

gnzlbg:
@alexcrichton so I was able to implement most of the reductions on top of `llvm.experimental.vector.reduction.{...}`. However, for floating-point vector types, `sum` and `product` produce code-gen errors with everything I've tried (passing `0. as $elem_ty` and `mem::uninitialized()` as an accumulator).

coresimd/ppvt/api/bool_vectors.rs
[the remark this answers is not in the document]
alexcrichton: Oh yeah I think at this point in time they can all move to `#[derive(Clone)]`

alexcrichton:
Hm I wonder if the added tests are stressing out rustc a bit much? Travis looks like it's timing out quite a lot :(

alexcrichton:
Looks good to me to merge modulo CI

gnzlbg:
@alexcrichton yes build times have doubled. I could split the simd types into their own crate, and import it from coresimd.

alexcrichton:
Hm splitting crates I don't think will be feasible due to integration into libstd, do you know why this takes so long to compile?

gnzlbg:
I think it's because of the tests.

gnzlbg:
I am also unable to do re-compilations of the crate without doing a `cargo clean` first. Doing a `cargo test`, one line change, `cargo test` again, makes the second `cargo test` require huge amounts of memory and HDD space..

rkruppe:
Why are boolean vectors represented as integer vectors containing 0 or -1? I know that some popular platforms don't have native vectors with one bit elements but why does that impact the portable types?

gnzlbg:
@rkruppe Those are the values that the simd comparison instructions return on at least `x86` to represent `true` (`0xFFF...`) and `false` (`0x000...`), and those are the values that the portable vector comparisons (and the reductions) of LLVM also use.

The API of boolean vector doesn't expose these values, and it also doesn't expose any conversions nor bitwise transmutes from other vector types.

In any case, we should document this, since this is relevant for those explicitly calling `mem::transmute`.

rkruppe:
> Those are the values that the simd comparison instructions return on at least x86

I know, that's what I was alluding to, but again, why does that have to impact the *portable* types? This is like saying i64 arithmetic doesn't exist on 32 bit targets so the portable `i64xN` types should be implemented in terms of `i32xM` for M = 2 N. When a target doesn't support some portable vector type, the type should be legalized by the backend. (Sometimes that doesn't work in practice, like for i128 on some targets, but I know that `i1` vectors can be legalized on x86 at least.)

> and those are the values that the portable vector comparisons (and the reductions) of LLVM also return. 

That is not true. `icmp eq <N x i32>` returns `<N x i1>`, for example. I haven't checked reductions but since they are overloaded, in principle they should also work with `i1`.

Clang does lower C-language vector compares to `icmp` (returning an `i1` vector) + `sext` so that the end result is an integer vector containing 0 and -1, but this is a front end decision (inherited from GCC), not anything inherent about LLVM IR.

gnzlbg:
> That is not true. icmp eq <N x i32> returns <N x i1>, for example. I haven't checked reductions but since they are overloaded, in principle they should also work with i1.

I'll give this a try.

gnzlbg:
@rkruppe so I tried to use bools, but got a "SIMD vector element type should be machine type" error: https://play.rust-lang.org/?gist=a0331d2eb68fec6c5e32b8a49356cd8d&version=nightly

rkruppe:
Come to think of it, I actually doubt there is a portable way to expose `<N x i1>` to Rust -- because Rust types can be stored in memory, but `<N x i1>` is stored either like `<N x i8>` (one byte per element) or `iN` (packed, individual bits not addressable) depending on the target. And indeed rustc won't let you do `#[repr(simd)] struct BoolVec(bool, bool, ...)` currently.

Probably best to use a memory layout compatible with `[bool; N]` or `(bool, bool, ...)`: u8 elements, 0 or 1.

Cargo.toml
```diff
@@ -11,3 +11,10 @@ opt-level = 3
 [profile.bench]
 debug = 1
 opt-level = 3
+
+[profile.test]
+opt-level = 0
+lto = false
+debug-assertions = true
+codegen-units = 1
+panic = 'unwind'
```
GabrielMajeri: Minor nitpick: should be a newline at the end of this file.

gnzlbg:
@rkruppe the question is, should we only expose `b8x{2,4,6,8,32,74}` or should we also expose wider boolean vector types? 

From the POV of portable operations exposing `b8xN` should be enough, because the actual size of the boolean vector type is irrelevant. 

From the POV of the architecture specific intrinsics, some of them return "boolean vectors" of a larger width stored in either integer or floating point registers. I don't know if `b8xN` is enough to provide type-safe wrappers around these intrinsics that have zero-runtime cost. The original `simd` crate had types like `b32fx8` probably for this purpose.

Cargo.toml
gnzlbg: I will revert these changes before merging. These profiles were added only to test if setting codegen-units to 1 would improve either compiletimes or remove some issues of incremental compilation.

gnzlbg:
> Come to think of it, I actually doubt there is a portable way to expose <N x i1> to Rust -- because Rust types can be stored in memory, but <N x i1> is stored either like <N x i8> (one byte per element) or iN (packed, individual bits not addressable) depending on the target. 

Could you elaborate on this? (or are you on IRC?). My question is: why does this matter?

gnzlbg:
@alexcrichton so this should be good to go modulo compile-times.

alexcrichton:
Ok thanks! I'd like to dig into this a bit first and investigate compile times a bit, I'll do that now.

alexcrichton:
Hm ok there may have been a recent rustc regression that was fixed, in any case looks like it's not *too* slow now, and yeah it's all almost entirely tests which we can of course move around later if need be. Thanks again @gnzlbg!

PR #339, opened 2018-03-03
Run-time feature detection for new AArch64 features

### LINKED ISSUE #337: Implement run-time feature detection for new AArch64 features
The following features were whitelisted here: https://github.com/rust-lang/rust/pull/48570/files#diff-f32b852c683c99fc96cfbda848664debR84

* [ ] `fp`: maps to `HWCAP_FP`
* [ ] `simd` (or `asimd`): 
   * [ ] maps to `HWCAP_ASIMD` (modulo known broken CPUs)
   * [ ] should probably require `fp` 
* [ ] `neon`: should be renamed to `simd` or `asimd` here an in rustc #125 
* [ ] `sve`: 
  * [ ] maps to `HWCAP_SVE` 
  * [ ] should probably require `fp`, `fp16`, and `simd`
* [ ] `crc`: maps to `HWCAP_CRC32`
* [ ] `crypto`: 
  - [ ] maps to `HWCAP_AES`, `HWCAP_PMULL`, HWCAP_SHA1`, `HWCAP_SHA2`
  - [ ] should probably require `simd` and transitively `fp` (this is how it works on GCC)
* [ ] `ras`: is this detectable at run-time? (Reliability, Availability, and Serviceability) If not, should use `compile_error!`.
* [ ] `lse`: maps to `HWCAP_ATOMICS`
* [ ] `rdm`: 
  * [ ] maps to `HWCAP_ASIMDRDM`
  * [ ] should also require `simd`
* [ ] `fp16`: 
  * [ ] maps to `HWCAP_FPHP`
  * [ ] should also require/enable `fp`
  * [ ] if `simd` is enabled, might also need to require `HWCAP_ASIMDHP`.
* [ ] `rcpc`: maps to `HWCAP_LRCPC`
* [ ] `dotprod`: 
  * [ ] maps to `HWCAP_ASIMDDP`
  * [ ] should also require `simd`
* [ ] `v8.1a`, `v8.2a`, `v8.3a`: these features are not detectable at run-time AFAICT, we should probably just produce a nice `compile_error!` when the user tries to query run-time feature detection for them.

---

Linux kernel HWCAP src for ARM64: https://elixir.bootlin.com/linux/latest/source/arch/arm64/include/uapi/asm/hwcap.h
GCC src for ARM64: https://github.com/gcc-mirror/gcc/blob/master/gcc/config/aarch64/aarch64-option-extensions.def
LLVM src for ARM64: https://github.com/llvm-mirror/llvm/blob/e8664b0b333ae9862978db563db8cf58dca41377/lib/Support/Host.cpp#L1273

--

cc @Amanieu

gnzlbg:
Closes #337 .

ping: @Amanieu

### DIFF
--- a/crates/stdsimd/tests/cpu-detection.rs
+++ b/crates/stdsimd/tests/cpu-detection.rs
@@ -19,9 +19,17 @@ fn arm_linux() {
 #[test]
 #[cfg(all(target_arch = "aarch64", target_os = "linux"))]
 fn aarch64_linux() {
+    println!("fp: {}", is_target_feature_detected!("fp"));
+    println!("fp16: {}", is_target_feature_detected!("fp16"));
     println!("neon: {}", is_target_feature_detected!("neon"));
     println!("asimd: {}", is_target_feature_detected!("asimd"));
-    println!("pmull: {}", is_target_feature_detected!("pmull"));
+    println!("sve: {}", is_target_feature_detected!("sve"));
+    println!("crc: {}", is_target_feature_detected!("crc"));
+    println!("crypto: {}", is_target_feature_detected!("crypto"));
+    println!("lse: {}", is_target_feature_detected!("lse"));
+    println!("rdm: {}", is_target_feature_detected!("rdm"));
+    println!("rcpc: {}", is_target_feature_detected!("rcpc"));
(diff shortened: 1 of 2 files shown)


alexcrichton:
Nice!

Issue #316, opened 2018-02-05
Flavorful return types like "uint8x16x2_t"


alexcrichton:
Some intrinsics on NEON like `vtrnq_u8` have a nonstandard return type like `uint8x16x2_t` defined as:

```c
 typedef struct uint8x16x2_t {
  uint8x16_t val[2];
} uint8x16x2_t;
```

How should we best represent this in Rust? A few possible options are:

1. `unsafe fn vtrnq_u8(a: uint8x16_t, b: uint8x16_t) -> uint8x16x2_t;` - aka we define the struct w/ pub fields in Rust like with the `libc` crate
2. `unsafe fn vtrnq_u8(a: uint8x16_t, b: uint8x16_t) -> [uint8x16_t; 2]` 
3. `unsafe fn vtrnq_u8(a: uint8x16_t, b: uint8x16_t) -> (uint8x16_t; uint8x16_6)`

I'd sort of lean towards the last one, but it also depends on how "flavorful" the return types get!

hsivonen:
I think the last option is the ideal one.

A quick look at clang's arm_neon.h suggests that

1. The multiple vectors wrapped by a returned C `struct` are always mutually of the same type. (Always declared as a C array within the `struct`.)
2. The possible array lengths are 2, 3 and 4.

alexcrichton:
Ok yeah that makes sense to me! In that case I think it's ok for us to take a bit of liberty here and return tuples instead of strictly the structs specified. Although I'd also be ok with fixed size arrays as well!

gnzlbg:
I would do what the spec does, that is, option 1.

gnzlbg:
Unless there is a good reason not to. Is it? Like if we can make a tuple, we can make a struct right? Or does this tuple need to be magical?

alexcrichton:
We'd sort of be defining a new type either way and adding logic to ensure that it matches with the spec no matter what we choose. In that sense I think it's an equal amount of work/verification logic for any choice here, so I figured we'd go with the most ergonomic to consume one.

hsivonen:
A tuple is nicer for Rust destructuring. (Also, since the ARM stuff is already strongly-typed, `uint8x16_t` could be `u8x16` from portable SIMD.)

gnzlbg:
> A tuple is nicer for Rust destructuring.

[With `#[feature(slice_patterns)]`](https://play.rust-lang.org/?version=nightly&mode=debug) you get:

```rust
[...]

PR #345, opened 2018-03-06
adds AArch64's {s,u,f}{min,max}{v,p} and ARM's {vmov}{n,l}

### LINKED ISSUE #314: encoding_rs aarch64 priority wishlist
In the SIMD meeting yesterday, @gnzlbg encouraged me to file an issue suggesting what aarch64 intrinsics to do first in the interest of letting encoding_rs's SIMD functionality migrate to stable Rust.

Assuming that LLVM portable shuffles become available on stable together with the rest of portable 128-bit SIMD, my aarch64 wishlist has just two items:

```rust
extern "platform-intrinsic" {
            fn aarch64_vmaxvq_u8(x: u8x16) -> u8;
            fn aarch64_vmaxvq_u16(x: u16x8) -> u16;
}
```

If we don't get portable shuffles on stable for a while, then my wishlist has two additional items:

* The intrinsic(s) that enables the operation of expanding a single `u8x16` into two `u16x8`s such that each `u8` lane is zero-extended to a `u16` lane. I believe this would be zipping the `u8x16` with an all-zero vector.
* The intrinsic(s) that takes two `u16x8`s and in the situation where the high half of each lane is zeroed, produces a `u8x16` consisting of the low half of each lane. (If the high halves aren't zero, I don't care what garbage ends up in the result `u8x16` as long as it doesn't trap. For example, on SSE2, I use `x86_mm_packus_epi16`, which does signed saturation instead of just discarding the high halves.) I believe on aarch64 this would be some kind of unzip operation with the other output vector (the one with the high halves) ignored afterwards.

gnzlbg:
* adds {s,u,f}{min,max}{v,p} AArch64 intrinsics
* adds {vmov}{n,l} ARM intrinsics

Closes #314 (the missing part there are portable shuffles, but I'll open a new issue just for that). 

cc @hsivonen.

### DIFF
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -5,8 +5,8 @@ set -ex
 # Tests are all super fast anyway, and they fault often enough on travis that
 # having only one thread increases debuggability to be worth it.
 export RUST_TEST_THREADS=1
-export RUST_BACKTRACE=1
-export RUST_TEST_NOCAPTURE=1
+#export RUST_BACKTRACE=1
+#export RUST_TEST_NOCAPTURE=1
 
 # FIXME(rust-lang-nursery/stdsimd#120) run-time feature detection for ARM Neon
 case ${TARGET} in
--- a/coresimd/aarch64/neon.rs
+++ b/coresimd/aarch64/neon.rs
@@ -39,9 +39,319 @@ pub unsafe fn vaddd_u64(a: u64, b: u64) -> u64 {
     a + b
 }
 
+#[allow(improper_ctypes)]
+extern "C" {
+    #[link_name = "llvm.aarch64.neon.smaxv.i8.v8i8"]
+    fn vmaxv_s8_(a: i8x8) -> i8;
+    #[link_name = "llvm.aarch64.neon.smaxv.i8.6i8"]
+    fn vmaxvq_s8_(a: i8x16) -> i8;
+    #[link_name = "llvm.aarch64.neon.smaxv.i16.v4i16"]
+    fn vmaxv_s16_(a: i16x4) -> i16;
+    #[link_name = "llvm.aarch64.neon.smaxv.i16.v8i16"]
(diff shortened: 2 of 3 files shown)


alexcrichton:
Nice! r=me once CI is green

Before we go too hog wild with intrinsics though we may want to sort out the types first (aka add the dedicated types for arm/aarch64 and conversions from the portable ones)

gnzlbg:
@alexcrichton ci was green for me, the failure is spurious "no input received for the last 10 minutes". 

> Before we go too hog wild with intrinsics though we may want to sort out the types first (aka add the dedicated types for arm/aarch64 and conversions from the portable ones)

I tried to choose intrinsics that exercise types that we don't have yet. These happened to be what @hsivonen needed and also have some variants using `f16` that I wasn't able to implement. I'd implement a couple of the intrinsics that use "polynomial" vector types next, and the ones in issue #316 . Once we have a small corpus of intrinsics that exercise the weirdest things in ARM/AARCH64, I'll start adding the architecture specific vector types.

alexcrichton:
Ok! If CI continues to be flaky we probably want to start splitting up test suites into separate crates to ease the burden on rustc

gnzlbg:
Maybe we could use `travis_wait` somehow to increase the allowed time without output from 10 minutes to 20 minutes.

alexcrichton:
Yeah I'd also be fine w/ that, but taking forever to build locally can also be a pain :(

gnzlbg:
Could we:

* keep the tests in `coresimd` but behind a feature flag, e.g., `ppsv_tests`.
* add a third crate to the workspace that just includes the `coresimd` module, but always has this feature flag enabled

?

That way we can develop `coresimd`/`stdsimd` without compiling/running these tests, but still do run them as part of `cargo test --all` when executed on the root.

alexcrichton:
Seems plausible to me!

PR #346, opened 2018-03-06
Rename `is_target_feature_detected!`

alexcrichton:
This commit renames the `is_target_feature_detected!` macro to have different
names depending on the platform. For example:

* `is_x86_feature_detected!`
* `is_arm_feature_detected!`
* `is_aarch64_feature_detected!`
* `is_powerpc64_feature_detected!`

Each macro already has a platform-specific albeit similar interface. Currently,
though, each macro takes a different set of strings so the hope is that like
with the name of the architecture in the module we can signal the dangers of
using the macro in a platform-agnostic context.

One liberty taken with the macro currently though is to on both the x86 and
x86_64 architectures name the macro `is_x86_feature_detected` rather than also
having an `is_x86_64_feature_detected`. This mirrors, however, how all the
intrinsics are named the same on x86/x86_64.

### DIFF
--- a/coresimd/x86/avx2.rs
+++ b/coresimd/x86/avx2.rs
@@ -2127,7 +2127,7 @@ pub unsafe fn _mm256_shuffle_epi8(a: __m256i, b: __m256i) -> __m256i {
 /// use std::arch::x86_64::*;
 ///
 /// # fn main() {
-/// #     if is_target_feature_detected!("avx2") {
+/// #     if is_x86_feature_detected!("avx2") {
 /// #         #[target_feature(enable = "avx2")]
 /// #         unsafe fn worker() {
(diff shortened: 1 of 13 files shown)


BurntSushi:
I like it!

gnzlbg:
> so the hope is that like with the name of the architecture in the module we can signal the dangers of
using the macro in a platform-agnostic context.

I don't like to hope here. Could you add a `compile_error!` that triggers if the macro is used in the wrong arch ?

alexcrichton:
@gnzlbg oh we can do that at any time I think, it's already an error to invoke it on other platforms and this'd just be about making it a *nicer* error when invoked on other platforms. That seems like a great idea though!

alexcrichton:
Ok! I'm gonna merge this

PR #351, opened 2018-03-07
Add `cfg!` clauses to detection macro

### LINKED ISSUE #349: is_target_feature_detected not being statically removed with newer target-cpu
I believe is_target_feature_detected! is supposed to be statically removed if the build settings guarantee the target_features exist. 

This is not the case here,
Linux, rustc 1.26.0-nightly (c9334404f 2018-03-05)

Running the following with
RUSTFLAGS="-C target-cpu=haswell" cargo run --release
And checking perf top, shows
std::stdsimd::arch::detect::check_for
Taking the majority of cpu time, uncommenting out the cfg check removes that. 

The optimizer skips the relevent sse4.2 code, but the commented out random check can be used, and check_for will show up in perf (though it may need to be searched for, since it's less of the cpu time) 

If I look at the assembly it appears that with target-cpu=haswell, the sse4.2 function is inlined into the function with the check, and without it's a separate function.

```rust
#![feature(target_feature)]
#![feature(cfg_target_feature)]
#![feature(stdsimd)]

extern crate rand;
use rand::{thread_rng, Rng};

#[cfg(target_arch = "x86_64")]
use std::arch::x86_64::*;

#[cfg(target_arch = "x86")]
use std::arch::x86::*;

#[target_feature(enable = "sse4.2,popcnt")]
unsafe fn subset_len_sse42(a: [u8; 16], b: [u8; 16]) -> u32 {
    let a = _mm_loadu_si128(a.as_ptr() as *const __m128i);
    let b = _mm_loadu_si128(b.as_ptr() as *const __m128i);

    _mm_extract_epi32(_mm_cmpistrm(a, b, 0), 0).count_ones()
}

fn subset_len_portable(a: [u8; 16], b: [u8; 16]) -> u32 {
    a.iter().map(|x| if b.contains(x) { 1 } else { 0 }).sum()
}

fn subset_len(a: [u8; 16], b: [u8; 16]) -> u32 {
    /*
    //Uncomment this out to make it work
    #[cfg(all(any(target_arch = "x86", target_arch = "x86_64"), target_feature = "sse4.2",
              target_feature = "popcnt"))]
    {
        return unsafe { subset_len_sse42(a, b) };
    }
    */
    {
        //This check is supposed to compile out when a new enough arch is
        //enabled at build time, but isn't working
        if is_target_feature_detected!("sse4.2") && is_target_feature_detected!("popcnt") {
            return unsafe { subset_len_sse42(a, b) };
        }
    }
    subset_len_portable(a, b)
}

fn main() {
    // Zero marks end of string in _mm_cmpistrm
    let mut vals: Vec<_> = (1..256u16).map(|x| x as u8).collect();

    let mut rng = thread_rng();

    let mut a = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16];
    let mut b = [
        9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24
    ];

    loop {
        assert_eq!(subset_len(a, b), 8);
        /*
        rng.shuffle(&mut vals);
        a.iter_mut().zip(vals.iter()).for_each(|(a, v)| *a = *v);
        rng.shuffle(&mut vals);
        b.iter_mut().zip(vals.iter()).for_each(|(b, v)| *b = *v);
        assert_eq!(subset_len(a, b), subset_len_portable(a, b));
        */
    }
}
```

alexcrichton:
This way if the feature is statically detected then it'll be expanded to `true`

Closes #349

### DIFF
--- a/crates/stdsimd/src/lib.rs
+++ b/crates/stdsimd/src/lib.rs
@@ -8,6 +8,7 @@
 //! [stdsimd]: https://rust-lang-nursery.github.io/stdsimd/x86_64/stdsimd/
 
 #![feature(const_fn, integer_atomics, staged_api, stdsimd)]
+#![feature(cfg_target_feature)]
 #![cfg_attr(feature = "cargo-clippy", allow(shadow_reuse))]
 #![cfg_attr(target_os = "linux", feature(linkage))]
 #![no_std]
--- a/stdsimd/arch/detect/aarch64.rs
+++ b/stdsimd/arch/detect/aarch64.rs
@@ -9,40 +9,52 @@ use super::linux;
 macro_rules! is_aarch64_feature_detected {
     ("neon") => {
         // FIXME: this should be removed once we rename Aarch64 neon to asimd
-        $crate::arch::detect::check_for($crate::arch::detect::Feature::asimd)
+        cfg!(target_feature = "neon") ||
+            $crate::arch::detect::check_for($crate::arch::detect::Feature::asimd)
     };
     ("asimd") => {
-        $crate::arch::detect::check_for($crate::arch::detect::Feature::asimd)
+        cfg!(target_feature = "neon") ||
+            $crate::arch::detect::check_for($crate::arch::detect::Feature::asimd)
     };
     ("pmull") => {
-        $crate::arch::detect::check_for($crate::arch::detect::Feature::pmull)
+        cfg!(target_feature = "pmull") ||
+            $crate::arch::detect::check_for($crate::arch::detect::Feature::pmull)
     };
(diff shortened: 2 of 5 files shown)


gnzlbg:
LGTM

gnzlbg:
Is there a way that we can add some tests for this?

alexcrichton:
Hm we could probably do it in a roundabout way through inspecting optimized disassembly and asserting there are no `call` instructions?

gnzlbg:
Sounds good.

PR #361, opened 2018-03-08
Add initial wasm memory grow/current intrinsics

alexcrichton:
This exposes access to the `grow_memory` and `current_memory` instructions
provided by wasm in what will hopefully be a stable interface (the stable part
being x86 first in theory).

### DIFF
--- a/stdsimd/arch/detect/cache.rs
+++ b/stdsimd/arch/detect/cache.rs
@@ -1,6 +1,8 @@
 //! Caches run-time feature detection so that it only needs to be computed
 //! once.
 
+#![allow(dead_code)] // not used on all platforms
+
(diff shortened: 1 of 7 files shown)


stdsimd/arch/detect/cache.rs
```diff
@@ -1,6 +1,8 @@
 //! Caches run-time feature detection so that it only needs to be computed
 //! once.
 
+#![allow(dead_code)] // not used on all platforms
+
```
gnzlbg: I would prefer if we would just disable the `detect` module in the platforms in which it is not used. I can do this in the MIPS PR.

gnzlbg:
LGTM!

Issue #364, opened 2018-03-09
CI for MIPS/MIPS64


gnzlbg:
To test the MIPS SIMD Architecture (MSA) intrinsics we need to setup CI for MIPS/MIP64 targets. This is a summary of the problems I've encountered so far. 

* using MSA requires enabling floating-point features (`fp64`/`fpxx`). The current 32-bit MIPS targets (`{mips,mipsel}-unknown-linux-gnu`) do not have these enabled. Passing `-C target-feature=+msa,+fp64` does not help:

    ```
    LLVM ERROR: MSA requires a 64-bit FPU register file (FR=1 mode). See -mattr=+fp64.
    ```
* enabling floating point probably will require a std library compiled with it enabled, this should help: https://github.com/rust-lang/rust/pull/48874
* qemu only supports two CPUs with MSA: p5600 (32-bit) and i6400 (64-bit). 
* qemu has a bug in which the auxiliary vector bit for MSA is not set: https://bugs.launchpad.net/qemu/+bug/1754372 ; until that is fixed we can't really test run-time feature detection under qemu and must override it to test the intrinsics
* cross-compiling for `i6400` is challenging. We need to set `-C target-cpu=mips64r6` when compiling Rust. This is going to require the `std` lib to be compiled for this. Also, we need to set `-march=mips64r6` when compiling C (e.g. for backtrace, auxv, etc.). The cross-compilation toolchains for this are not available in Ubuntu, only on Debian testing/unstable: `libc6-dev-mips64r-cross`. I haven't managed to get them to work (they seem buggy: missing headers, linking failurs, etc. they do not work out-of-the-box). One also must be careful of not cross-compiling against code that was compiled for `mips64r2` since that will fail to link. All in all we might need to just use a different toolchain here, like the Codescape SDK. This explains how to set it up: https://github.com/v8mips/v8mips/wiki/Building-v8-for-MIPS
* cross-compiling for `p5600` requires targeting `-C target-cpu=mips64r5`/`-march=mips64r5`. The toolchains for it look mature, they might even work out-of-the-box.  We will probably still need to recompile the std library for these targets though.  

I would be surprised if these were the only issues. 

In my opinion, the most promising route towards CI for MS would be to focus on `mips/mipsel` and the `p5600`.

---

cc @jcowgill @malbarbo @japaric

gnzlbg:
QEMU 2.12 has auxv support for MIPS, which should enable us to run our tests there as well :)

[...]

Issue #355, opened 2018-03-08
Add /proc/cpuinfo support for mips64


gnzlbg:
Currently the mips64 feature detection run-time only uses auxiliary vectors. It would be nice to add parsing `/proc/cpuinfo` as a fallback.

PR #365, opened 2018-03-09
[mips/mips64: msa] add add_a_b intrinsic

gnzlbg:
This PR adds a single `mips::msa` intrinsic (`add_a_b`: vector add absolute values). 

It adds all the scaffolding for adding more mips/mips64 intrinsics "easily" in the future.

* adds an `arch::mips64` module
* generates documentation for `mips`/`mips64`
* adds run-time feature detection for `mips`/`mips64` (auxiliary vectors only)
* adds `simd_test` support for `mips`/`mips64` (requires parsing instructions as expressions instead of identifiers because `add_a.b` is a valid MIPS64 instruction name)

The following issues are left for future PRs:

* adding CI for MIPS: #364 
* adding synthetic tests using the auxval crate: #355 
* adding cpuinfo support: #354

---

@alexcrichton I cleaned up the run-time feature detection on linux a bit because it was getting out-of-hand. I have no idea whether this might break rustc builds or not. Is there a way for me to test this? If these changes break some build bots there, we should add these build bots to our CI as well.

cc @jcowgill


.travis.yml
```diff
@@ -47,7 +51,14 @@ install:
 
 script:
   - cargo generate-lockfile
-  - ci/run-docker.sh $TARGET $FEATURES
+  - |
+    if [ "$NORUN" == "1" ]; then
+      cd crates/stdsimd
```
alexcrichton: Could this use the `-p` argument to avoid changing the cwd?

stdsimd/arch/detect/linux/auxvec.rs
```diff
@@ -0,0 +1,290 @@
+//! Parses ELF auxiliary vectors.
+
+use core::mem;
+use _std::prelude::v1::*;
+use _std::fs::File;
+use _std::io::Read;
```
alexcrichton: This unfortunately will cause some problems when building in libstd itself as this doens't exist, but could this continue to use the same strategy as before?

stdsimd/arch/detect/arm.rs
```diff
@@ -35,9 +36,13 @@ pub enum Feature {
 pub fn detect_features() -> cache::Initializer {
     let mut value = cache::Initializer::default();
     fill_features(&mut value);
-    return value
+    value
 }
 
+#[cfg(not(target_os = "linux"))]
```
alexcrichton: I'm personally not a huge fan of `#[cfg]` as it can make already complicated code even more complicated, so could this continue to compile on all platforms and just statically return an error on the "wrong" platforms?

stdsimd/arch/detect/linux/auxvec.rs
```diff
@@ -0,0 +1,290 @@
...
+
+#[cfg(not(all(target_os = "linux",
```
alexcrichton: Perhaps `cfg_if!` could be used to simplify this? That could be added as a dependency and it's available in libstd as well

stdsimd/arch/detect/linux/auxvec.rs
```diff
@@ -0,0 +1,290 @@
...
+/// If an entry cannot be read all the bits in the bitfield are set to zero.
+/// This should be interpreted as all the features being disabled.
+#[derive(Debug, Copy, Clone)]
+pub struct AuxVec {
+    #[cfg(any(target_arch = "aarch64", target_arch = "arm",
+              target_arch = "powerpc64", target_arch = "mips",
+              target_arch = "mips64"))]
+    pub hwcap: usize,
+    #[cfg(any(target_arch = "arm", target_arch = "powerpc64"))]
+    pub hwcap2: usize,
```
alexcrichton: Could both fields be perhaps unconditionally included to avoid the extra `#[cfg]` traffic?

stdsimd/arch/detect/linux/mod.rs
```diff
@@ -0,0 +1,9 @@
+//! Run-time feature detection on Linux
+
+#[cfg(not(any(target_arch = "mips", target_arch = "mips64")))]
+pub mod cpuinfo;
+
+#[cfg(any(target_arch = "aarch64", target_arch = "arm",
+          target_arch = "powerpc64", target_arch = "mips",
+          target_arch = "mips64"))]
```
alexcrichton: Ideally I'd prefer to avoid the `#[cfg]` here, it becomes a pretty big burden to maintain over time :(

.travis.yml
gnzlbg: Sure thing.

stdsimd/arch/detect/arm.rs
gnzlbg: IIRC the problem here was that the auxvec/cpuinfo modules are only available on some oses (currently only linux), but these functions should be available on all OSes and just return nothing on linux. 

So IMO the entanglement here is that we have target_os modules (e.g. linux) and arch modules (e.g. aarch64) and a mixture of archs and target_oses on both. I'll try to think of something better here.

.travis.yml
```diff
@@ -47,7 +51,14 @@ install:
 
 script:
   - cargo generate-lockfile
-  - ci/run-docker.sh $TARGET $FEATURES
+  - |
+    if [ "$NORUN" == "1" ]; then
+      cd -p crates/stdsimd
```
alexcrichton: Oh sorry I meant `-p` arguments to the `cargo` commands :)

gnzlbg:
@alexcrichton So I think I managed to clean up `cfg`-hell at least a bit. The gist in a nutshell is:

`stdsimd/detect/` now has two sub-directories:

* stdsimd/detect/arch/{target_arch}.rs: define architecture features, that is, the `Feature` enum and the `is_{target_arch}_feature_detected!` macros, nothing more.
* stdsimd/detect/os/{target_os}.rs: defines the `check_for` function, that is, how features are actually detected on the different platforms

I don't think this is great, but I think it is better than what we had. Now, `cpuinfo`, `auxvec`, and the `cache`, are only defined on the platforms that actually need them. For example, on non-x86 non-linux platforms, we don't need the cache, because we can't detect any features anyways.

gnzlbg:
Yes we will need to do so to add windows on arm support.

PR #367, opened 2018-03-09
document all arches when part of std

QuietMisdreavus:
This PR adds the `doc_cfg` feature to coresimd and stdsimd to tell rustdoc about which arch is required for each module. It also uses the `#[cfg(dox)]` configuration flag that the rest of libstd uses to ensure that all four major arch modules are included when rustdoc builds the library. This worked great for coresimd, which now proudly shows all the arch intrinsics. However, when trying to re-export these modules in stdsimd, the cfg(dox) trick doesn't work any more, so instead we mask out these modules and link to libcore. Rather than show the full docs for whatever the native arch of the build environment was, i masked out all of them and stubbed them out.

For a preview, i have [a rendering on my server](https://tonberry.quietmisdreavus.net/simd-std/core/arch/index.html). Here's what the modules listing looks like in `std::arch` when this branch is used as part of the std docs:

every arch? in this part of the docs? localized entirely on my server? may i see it?

### DIFF
--- a/coresimd/mod.rs
+++ b/coresimd/mod.rs
@@ -39,7 +39,8 @@ pub mod arch {
     /// Platform-specific intrinsics for the `x86` platform.
     ///
     /// See the [module documentation](../index.html) for more details.
-    #[cfg(target_arch = "x86")]
+    #[cfg(any(target_arch = "x86", dox))]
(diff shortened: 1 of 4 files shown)


gnzlbg:
Thanks!

PR #369, opened 2018-03-10
Add CI for android

gnzlbg:
This PR adds CI for android targets:

* x86_64-linux-android
* arm-linux-androideabi 
* aarch64-linux-android

The `assert_instr` tests are disabled on these via the `STDSIMD_ASSERT_INSTR_IGNORE` environment variable. 

cc @rubdos

---

@alexcrichton  IIUC it is possible to copy the `objdump` binaries of the target into the android emulator, but I did not find a way to pass the emulator any environment variables. One might be able to get those tests to run by hardcoding the paths to `objdump` within the emulator via a `build.rs` script. Since we are already testing these archs in other build bots, I just disabled here this part of the tests.

I've allowed the`x86_64-linux-android` target to fail because of issue: https://github.com/rust-lang-nursery/stdsimd/issues/368

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,7 +59,6 @@ matrix:
   allow_failures:
     - env: CLIPPY=On TARGET=x86_64-unknown-linux-gnu NO_ADD=1
     - env: RUSTFMT=On TARGET=x86_64-unknown-linux-gnu NO_ADD=1
-    - env: TARGET=x86_64-linux-android
 
 before_install:
   # FIXME (travis-ci/travis-ci#8920) shouldn't be necessary...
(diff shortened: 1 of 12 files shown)


alexcrichton:
Nice! Before we pile too much on our CI though I'd just want to confirm, do we expect this to find any bugs in stdsimd itself? Or to put it otherwise, will our effort needed to maintain all this CI be worth it in the end? 

I'd actually be sort of surprised if anything here breaks specifically on Android *other* than perhaps the runtime detection, right?

gnzlbg:
Yes these are only to test run time detection on android. FWIW I am
surprised it worked.

alexcrichton:
Ok! In that case I'd be fine landing, and would also be ok if we only tested stdsimd on the emulators as well

PR #371, opened 2018-03-13
Updates stdsimd to use the portable vector reduction intrinsics

gnzlbg:
See PR upstream: https://github.com/rust-lang/rust/pull/48983


gnzlbg:
Superseeded by #379

PR #379, opened 2018-03-17
Split protable vector types tests into multiple crates

gnzlbg:
The tests are still defined inline with the code but now there are a bunch of macros that control whether they are compiled or not.

The `crates/coresimd_v{...}_tests` then define these macros to include the tests.

To avoid recompiling `coresimd`, these crates add it as a dependency, and they include `ppsv` module using `#[path = "..."]`.


Cargo.toml
```diff
@@ -2,6 +2,12 @@
 members = [
   "crates/stdsimd-verify",
   "crates/stdsimd",
+  "crates/coresimd_v16_tests",
```
alexcrichton: Could these perhaps be organized under `crates/coresimd/tests/*.rs`?

That'd mean that `cargo test -p coresimd` should hopefully pick up everything

alexcrichton:
Looks great to me! My only real comment is about moving these to unit test crates rather than actual crates (which I think won't be too hard?), but otherwise looks good!

gnzlbg:
@alexcrichton done! I've never thought about the files in the tests/ directory as "crates" ! But that's exactly what they are, thanks!

gnzlbg:
I've merged the rustc reduction intrinsics here to remove the codegen modules.

Results51:
#379

Results51:
#492

Issue #383, opened 2018-03-19
vrsqrte_f32 fails to lower on ARMv7 NEON


gnzlbg:
The error is:

LLVM ERROR: Cannot select: intrinsic %llvm.aarch64.neon.frsqrte
error: Could not compile `coresimd`.

Issue #382, opened 2018-03-19
clz fails to produce correct assembly on armv7


gnzlbg:
Intrinsics affected are: `_clz_u{8,16,32}`.

```shell
failures:
---- coresimd::arm::v7::assert__clz_u16_clz stdout ----
	disassembly for coresimd::coresimd::arm::v7::assert__clz_u16_clz::_clz_u16_shim: 
	 0: push {fp, lr} 
	 1: mov fp, sp 
	 2: bl fdc8 <_ZN4core3num21_$LT$impl$u20$u16$GT$13leading_zeros17h3c918b237a87689fE> 
	 3: uxth r0, r0 
	 4: pop {fp, pc} 
thread 'coresimd::arm::v7::assert__clz_u16_clz' panicked at 'failed to find instruction `clz` in the disassembly', crates/stdsimd-test/src/lib.rs:370:9
note: Run with `RUST_BACKTRACE=1` for a backtrace.
---- coresimd::arm::v7::assert__clz_u32_clz stdout ----
	disassembly for coresimd::coresimd::arm::v7::assert__clz_u32_clz::_clz_u32_shim: 
	 0: b fddc <_ZN4core3num21_$LT$impl$u20$u32$GT$13leading_zeros17h1eadd5618945a0b2E> 
thread 'coresimd::arm::v7::assert__clz_u32_clz' panicked at 'failed to find instruction `clz` in the disassembly', crates/stdsimd-test/src/lib.rs:370:9
---- coresimd::arm::v7::assert__clz_u8_clz stdout ----
	disassembly for coresimd::coresimd::arm::v7::assert__clz_u8_clz::_clz_u8_shim: 
	 0: push {fp, lr} 
	 1: mov fp, sp 
	 2: bl fdb4 <_ZN4core3num20_$LT$impl$u20$u8$GT$13leading_zeros17hdd7115be42dec2f3E> 
	 3: uxtb r0, r0 
	 4: pop {fp, pc} 
thread 'coresimd::arm::v7::assert__clz_u8_clz' panicked at 'failed to find instruction `clz` in the disassembly', crates/stdsimd-test/src/lib.rs:370:9
```

sayantn:
I couldn't even find `_clz_u{8,16,32}`, so I guess this is just very old code, and also this generates the correct asm now

PR #384, opened 2018-03-19
Add arm neon vector types and fixes ARM tests

gnzlbg:
NEON is supported on:

* AArch64
* ARM when built with the `v7` and `neon` features. Building only with `v7` and trying to use run-time feature detection result in bad codegen. This might be a bug?

---

This PR adds ARM/AArch64 NEON architecture-specific vector types:

* ARM:
    * 64-bit wide: float32x2_t, int16x4_t, int32x2_t, int64x1_t, int8x8_t, poly16x4_t, poly8x8_t, uint16x4_t, uint32x2_t, uint64x1_t, uint8x8_t
    * 128-bit wide: float32x4_t, int16x8_t, int32x4_t, int64x2_t, int8x16_t, poly16x8_t, poly8x16_t, uint16x8_t, uint32x4_t, uint64x2_t, uint8x16_t

* AArch64:
    * 64-bit wide: float32x2_t, float64x1_t, int16x4_t, int32x2_t, int64x1_t, int8x8_t, poly16x4_t, poly8x8_t, uint16x4_t, uint32x2_t, uint64x1_t, uint8x8_t
  * 128-bit wide: float32x4_t, float64x2_t, int16x8_t, int32x4_t, int64x2_t, int8x16_t, poly16x8_t, poly8x16_t, uint16x8_t, uint32x4_t, uint64x2_t, uint8x16_t

It also adds appropriate conversions from/to portable vector types, and updates all the ARM/AArch64 intrinsics to the new vector types.

---

Other fixes:

* the armv7 tests weren't properly running before: this PR activates these tests and adds `assert_instr` for all armv7 intrinsics
* on armv7 clz intrinsics are broken, so this PR filled #382 and disabled these tests temporarily. The problem might be due to the core::intrinsic not being inlined here.
* the arm v7 neon `vrsqrte_f32` intrinsic fails to lower on ARM - this PR disables the intrinsic and filled #383
* rbit now uses the `core::intrinsic::bitreverse`
* properly error at compile-time when trying to perform run-time feature detection on ARM features that can't be detected at run-time

### DIFF
--- a/coresimd/arm/mod.rs
+++ b/coresimd/arm/mod.rs
@@ -7,14 +7,21 @@
 //! http://infocenter.arm.com/help/topic/com.arm.doc.
 //! ihi0073a/IHI0073A_arm_neon_intrinsics_ref.pdf
 //! [arm_dat]: https://developer.arm.com/technologies/neon/intrinsics
+#![allow(non_camel_case_types)]
 
 mod v6;
 pub use self::v6::*;
 
+#[cfg(any(target_arch = "aarch64", target_feature = "v7"))]
 mod v7;
+#[cfg(any(target_arch = "aarch64", target_feature = "v7"))]
 pub use self::v7::*;
 
-#[cfg(target_feature = "neon")]
+// NEON is supported on AArch64, and on ARM when built with the v7 and neon
+// features. Building ARM without neon produces incorrect codegen.
+#[cfg(any(target_arch = "aarch64",
+          all(target_feature = "v7", target_feature = "neon")))]
 mod neon;
-#[cfg(target_feature = "neon")]
+#[cfg(any(target_arch = "aarch64",
+          all(target_feature = "v7", target_feature = "neon")))]
 pub use self::neon::*;
--- a/coresimd/macros.rs
+++ b/coresimd/macros.rs
@@ -1,0 +1,23 @@
+//! Utility macros.
+
+#[allow(unused)]
+macro_rules! types {
+    ($(
+        $(#[$doc:meta])*
+        pub struct $name:ident($($fields:tt)*);
+    )*) => ($(
+        $(#[$doc])*
+            #[derive(Copy, Debug)]
+        #[allow(non_camel_case_types)]
+        #[repr(simd)]
+        pub struct $name($($fields)*);
+
+        #[cfg_attr(feature = "cargo-clippy", allow(expl_impl_clone_on_copy))]
+        impl ::clone::Clone for $name {
+            #[inline] // currently needed for correctness
+            fn clone(&self) -> $name {
+                *self
+            }
+        }
+    )*)
+}
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -12,6 +12,16 @@ export RUST_TEST_THREADS=1
 
 FEATURES="strict,$FEATURES"
 
+# FIXME: on armv7 neon intrinsics require the neon target-feature to be
+# unconditionally enabled.
+case ${TARGET} in
+    armv7*)
+        export RUSTFLAGS="${RUSTFLAGS} -C target-feature=+neon"
+        ;;
+    *)
+        ;;
+esac
+
 echo "RUSTFLAGS=${RUSTFLAGS}"
 echo "FEATURES=${FEATURES}"
 echo "OBJDUMP=${OBJDUMP}"
--- a/coresimd/aarch64/mod.rs
+++ b/coresimd/aarch64/mod.rs
@@ -11,7 +11,5 @@
 mod v8;
 pub use self::v8::*;
 
-#[cfg(target_feature = "neon")]
 mod neon;
-#[cfg(target_feature = "neon")]
 pub use self::neon::*;
--- a/coresimd/aarch64/neon.rs
+++ b/coresimd/aarch64/neon.rs
@@ -6,20 +6,51 @@
 use stdsimd_test::assert_instr;
 use coresimd::simd_llvm::simd_add;
 use coresimd::simd::*;
+use coresimd::arm::*;
+
+types! {
+    /// ARM-specific 64-bit wide vector of one packed `f64`.
+    pub struct float64x1_t(f64); // FIXME: check this!
+    /// ARM-specific 128-bit wide vector of two packed `f64`.
+    pub struct float64x2_t(f64, f64);
+}
+impl_from_bits_!(
+    float64x1_t: u32x2,
+    i32x2,
+    f32x2,
+    u16x4,
+    i16x4,
+    u8x8,
+    i8x8,
+    b8x8
+);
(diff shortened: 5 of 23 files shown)


coresimd/arm/mod.rs
```diff
@@ -7,14 +7,21 @@
...
 //! ihi0073a/IHI0073A_arm_neon_intrinsics_ref.pdf
 //! [arm_dat]: https://developer.arm.com/technologies/neon/intrinsics
+#![allow(non_camel_case_types)]
 
 mod v6;
 pub use self::v6::*;
 
+#[cfg(any(target_arch = "aarch64", target_feature = "v7"))]
 mod v7;
+#[cfg(any(target_arch = "aarch64", target_feature = "v7"))]
```
alexcrichton: I'm quite a newbie when it comes to ARM intrinsics, but how come the `#[cfg]` is needed here? 

It seems like `aarch64` implies the `v7` feature (although rustc doesn't print that out?) but is it a bug in rustc/llvm that we can only define the module when the `v7` feature is compiled in?

coresimd/arm/mod.rs
```diff
@@ -7,14 +7,21 @@
...
 pub use self::v7::*;
 
-#[cfg(target_feature = "neon")]
+// NEON is supported on AArch64, and on ARM when built with the v7 and neon
+// features. Building ARM without neon produces incorrect codegen.
```
alexcrichton: I guess this is probably related to the above comment, but what's the incorrect codegen here? Is that an LLVM/rustc bug?

coresimd/macros.rs
```diff
@@ -0,0 +1,23 @@
...
+    )*) => ($(
+        $(#[$doc])*
+            #[derive(Copy, Debug)]
+        #[allow(non_camel_case_types)]
+        #[repr(simd)]
+        pub struct $name($($fields)*);
+
+        #[cfg_attr(feature = "cargo-clippy", allow(expl_impl_clone_on_copy))]
+        impl ::clone::Clone for $name {
+            #[inline] // currently needed for correctness
```
alexcrichton: I think we can actually drop this now (the manual impl and just use `#[derive]`

coresimd/arm/mod.rs
gnzlbg: Yes, AArch64 always have a v7 feature enabled, so the LLVM backend does not advertise it as a feature (the v7 feature for AArch64 does not exist).

The cfg checks that either we are targeting for AArch64, or we have v7 enabled (that is, we are targetting ARM, but with v7). 

I don't think this is a bug. There are `arm-unknown...` and `armv7-unknown...` targets. I think this feature can only be enabled at compile time, and maybe linking between binaries compiled w/o this feature is not allowed ? (they might be different targets for a reason?). @japaric ?

coresimd/arm/mod.rs
gnzlbg: This might be an LLVM/rustc bug AFAICT.

On AArch64 we can compile without enabling the `neon` feature globally, use `#[target_feature(enable="neon")]` to enable neon for some functions, and call these with run-time feature detection without issues. The code generated looks pretty good.

ARMv6 and older do not support neon at all, so for ARM neon is only allowed if the `v7` feature is enabled (probably should check `v8` here as well but that's not supported by rustc yet?). 

Using the armv7-... target (with the v7 feature enabled) and tagging functions with `#[target_feature(enable="v7,neon")]` and calling them using run-time feature detection did work, but it produced horrible code (as in, way too many instructions). Enabling `neon` at compile-time produced the code I expected. So... maybe there is an LLVM/rustc bug where using`#[target_feature(enable = "v7,neon")]` is not enough for armv7 targets to generate good code? Maybe we need to enable some more features for these, I don't know yet.

Note that before, these modules where guarded with `#[cfg(target_feature = "neon")]`, so if these weren't enabled at compile-time we were not including them at all. Now it got a bit better for ARM, but I'd like to know what's going on with `neon` here.

coresimd/macros.rs
gnzlbg: We have these in at least 3 places throughout the library. I'll submit a PR that removes all of them.

coresimd/arm/mod.rs
alexcrichton: Oh hm ok so does v7 behave differently from like AVX on x86_64? It's not enabled in the baseline target but we still provide it all the time, right?

coresimd/arm/mod.rs
alexcrichton: Hm ok, so it sounds like v7 is very different from AVX in x86_64, but it's more like the 64-bit intrinsics on x86_64 wrt the i686 platform? (fundamentally not possible).

If that's the case, do you have some examples of the v7+neon code generation that are bad? I'd be curious to dig in there and see what's going on!

coresimd/arm/mod.rs
gnzlbg: So for `Aarch64` it is always enabled.

For `arm` AFAICT it must be enabled in the baseline target. That is, `arm-unknown...` does not support it at all, but `armv7-unknown` does. It is also not detectable at run-time AFAICT from looking at the linux hwcap files. So it looks to me that it is an option that `v7` must be enabled statically for the hole binary for `arm` targets. It might well be that `arm-unknown-... + target-feature=v7` is not equal to `armv7-unknown-...` because we are missing some features that are necessary (hence why the `armv7-..` target exists).

coresimd/arm/mod.rs
gnzlbg: > Hm ok, so it sounds like v7 is very different from AVX in x86_64, but it's more like the 64-bit intrinsics on x86_64 wrt the i686 platform? (fundamentally not possible).

I think `arm`, `armv7`, and `aarch64` are just different targets, as different at least as `x86` and `x86_64`.

> If that's the case, do you have some examples of the v7+neon code generation that are bad? I'd be curious to dig in there and see what's going on!

So  branched this, removed those checks, and disabled neon in ci/run.sh on this branch (https://github.com/gnzlbg/stdsimd/commit/9c9b2cf01c038e0f5c99e97e16fa3ebbee76fbab), which results in this travis failure(https://travis-ci.org/gnzlbg/stdsimd/jobs/355738848#L3159) .

Now I remember what was wrong: the run-time tests were not passing! The results generated were incorrect. This might well be an LLVM bug.

alexcrichton:
Ok thanks for all the info!

I think it's good to land this in the meantime, so let's do so!

coresimd/arm/mod.rs
gnzlbg: cc @petrochenkov @parched @Amanieu 

---

This also might be related to #139 where it was discussed that we might need to treat arm, armv7, armv8, as different architectures, some of which have a 64-bit mode.

PR #391, opened 2018-03-20
Documents arithmetic reduction semantics

gnzlbg:
This PR documents the semantics of the arithmetic reductions `sum`, `product`, `min`, and `max` for integers in the presence of integer overflow, and for floating point numbers in the presence of round-off errors and `NaN`s. 

It does so by adding test for these semantics for all portable vectors involved.


alexcrichton:
Looks good! I think CI is failing though?

gnzlbg:
@alexcrichton they are failing on release mode only because my implementation of these in rustc had a bug. I've fixed it here: https://github.com/rust-lang/rust/pull/49231

I'll close this until that lands, and reopen it again once the tests pass.

PR #394, opened 2018-03-22
add tests for endian-dependent behavior

gnzlbg:
The behavior of `mem::transmute` and memory load/stores is endian dependent. These tests document this behavior.

### DIFF
--- a/coresimd/mips/msa.rs
+++ b/coresimd/mips/msa.rs
@@ -17,7 +17,8 @@ extern "C" {
 
 /// Vector Add Absolute Values.
 ///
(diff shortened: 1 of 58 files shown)


alexcrichton:
r=me w/ green CI

gnzlbg:
@alexcrichton It looks like the PR that split the vector tests into multiple build bots broke the DOCUMENTATION build bot (it was broken due to rustc ICEs while that PR was merged, so we didn't realize this).

alexcrichton:
True! In that case let's merge and figure that out later

Issue #392, opened 2018-03-21
SHA extensions


jasondavies:
What needs to be done to support Intel's SHA extensions?

Happy to help if someone is willing to mentor/guide me!

I would also note that there are equivalent/similar extensions for ARM and POWER8.  But getting support for the Intel extensions would be a great start; they are already supported on AMD Ryzen.

gnzlbg:
I can mentor you. It is easier if you report after each step is done, but I am going to give you all the steps necessary here so that you can go at your own pace.

* Step 1: white-list the feature in rustc:
    * find the name of the cpuid bit feature / features, and how to query them
    * find the name of the LLVM backend feature / feature names - `llc --mcpu=help` gives you a least of all supported cpus by your LLVM installation, choose one that has the feature and write `llc --mcpu=<cpu_here> --mattr=help` to list all the features that the CPU supports. Find the ones you care about
    * come up with good names to name these features in rustc - the cpuid bit name, the llvm name, and or how clang and gcc name them are good candidates
    * white-list the feature in rustc by adding it to this list: https://github.com/rust-lang/rust/blob/36b66873187e37a9d79adad89563088a9cb86028/src/librustc_trans/llvm_util.rs#L88
   * submit a PR to rustc - you can work on the next steps but probably this PR will need to land first before any of your code will compile - this can take from a couple of hours to a couple of days, depending on how many PRs are on flight
* Step 2: implementing run-time feature detection for the intrinsic in `stdsimd`
    * you need to add the intrinsic to the supported x86 intrinsics in the `detect` module, for `x86` that would be here: https://github.com/rust-lang-nursery/stdsimd/blob/master/stdsimd/arch/detect/arch/x86.rs#L28
    * you need to check whether the intrinsic is enabled at run-time, and if that's the case, enable the feature. For x86's you need to check whether the cpuid bit of the intrinsic is set and if so, enable the intrinsic like the other code here does: https://github.com/rust-lang-nursery/stdsimd/blob/master/stdsimd/arch/detect/os/x86.rs#L125
* Step 3: implementing the intrinsic
  * come up with a place to implement the feature inside the `coresimd/x86` module. If the feature is its own self-standing feature, you probably want to add a new module there called `<featurename>.rs` 
[...]

PR #395, opened 2018-03-22
Add support for Intel SHA extensions.

jasondavies:
See #392.

### DIFF
--- a/coresimd/x86/mod.rs
+++ b/coresimd/x86/mod.rs
@@ -597,3 +597,6 @@ pub use self::aes::*;
 
 mod rdrand;
 pub use self::rdrand::*;
+
+mod sha;
+pub use self::sha::*;
--- a/coresimd/x86/sha.rs
+++ b/coresimd/x86/sha.rs
@@ -1,0 +1,201 @@
+use coresimd::simd::*;
+use coresimd::x86::*;
+use mem;
+
+#[allow(improper_ctypes)]
+extern "C" {
+    #[link_name = "llvm.x86.sha1msg1"]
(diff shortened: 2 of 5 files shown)


jasondavies:
Just for keeping track of progress.  I think I've implemented everything in "step 2" now.  I've also pulled in @alexcrichton's initial implementation (thank you!) so much of step 3 is already done aside from tests, which I think is all that remains.

For the tests, is there a preferred approach?  I was thinking of using some fixed test vectors.  I had a quick look at what GCC does, and they actually bring in an implementation of the intrinsics and test that they come up with the same results (though with fixed test vectors).

alexcrichton:
Looks great to me, thanks! For tests I think having fixes inputs/outputs is fine, we're mostly looking for "obviously wrong" bugs rather than smaller corner cases

gnzlbg:
The run-time support parts LGTM. For the tests, it is not about checking the result for every possible input value, but more about exercising that the intrinsics compile correctly and are able to generate correct results. So checking each intrinsic for two different inputs should be more than enough.

jasondavies:
One quick question: why are these assert instruction tests being ignored?

```sh
# CARGO_INCREMENTAL=0 cargo test sha
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs

     Running target/debug/deps/coresimd-3d30e14a0ab5fad2

running 14 tests
test coresimd::x86::sha::assert__mm_sha1msg1_epu32_sha1msg1 ... ignored
```

alexcrichton:
@jasondavies oh to run the instruction assertions you'll need to pass `--release`

alexcrichton:
Awesome, thanks!

PR #397, opened 2018-03-22
Fix default `cargo test` experience

alexcrichton:
Turns out Cargo doesn't automatically set `TARGET` for rustc invocations so
carry it forward manually from the build script over to the rustc invocation.

### DIFF
--- a/crates/coresimd/build.rs
+++ b/crates/coresimd/build.rs
@@ -1,0 +1,8 @@
+use std::env;
+
+fn main() {
+    println!(
+        "cargo:rustc-env=TARGET={}",
+        env::var("TARGET").unwrap()
+    );
+}
--- a/crates/simd-test-macro/src/lib.rs
+++ b/crates/simd-test-macro/src/lib.rs
@@ -60,19 +60,8 @@ pub fn simd_test(
         name.clone().as_str()
     ));
 
-    let default_target = if cfg!(target_os = "windows") {
-        Some("x86_64-pc-windows-msvc")
-    } else if cfg!(target_os = "linux") {
-        Some("x86_64-unknown-linux-gnu")
-    } else if cfg!(target_os = "macos") {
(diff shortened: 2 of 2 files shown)


crates/simd-test-macro/src/lib.rs
```diff
@@ -60,19 +60,8 @@ pub fn simd_test(
...
-    } else {
-        None
-    };
-
-    let target = env::var("TARGET").unwrap_or_else(|_| {
-        default_target.expect("TARGET environment variable not set and no default target known for the current target.").to_string()
-    });
+    let target = env::var("TARGET")
+        .expect("TARGET environment variable should be set for rustc");
     let mut force_test = false;
```
gnzlbg: So @alexcrichton merge as you wish but before these bits would set "TARGET" to sensitive defaults on all major platforms and otherwise complain.

How is just complaining any better?

gnzlbg: I mean, having to specify target for ci/run.sh is painful enough, if anything we should attempt to automatically detect target better.

That is, when someone writes `cargo test` things should just work.

alexcrichton: Oh that's what this PR is doing, automatically detecting `TARGET`! With this PR a bare `cargo test` should work.

The `TARGET` env var here in the procedural macro is carried over from the build script

gnzlbg: Ah nice, then let's merge this!

crates/coresimd/build.rs
```diff
@@ -0,0 +1,5 @@
+use std::env;
+
+fn main() {
+    println!("cargo:rustc-env=TARGET={}", env::var("TARGET").unwrap());
```
gnzlbg: I did not know about `cargo:rustc-env` this is cool!

PR #406, opened 2018-04-02
Fix the build on 0.0.4 branch

AdamNiederer:
This makes the before-vendor-types branch compile on newer versions of rustc. Would it be possible to release this a 0.0.4-1 or something similar? Everything on crates.io at the moment won't compile on the latest nightly.


AdamNiederer:
Oops, wrong branch.

gnzlbg:
I think it would be better to do a 0.0.5 release

PR #407, opened 2018-04-02
Fix the build on 0.0.4 branch

AdamNiederer:
This makes the before-vendor-types branch compile on newer versions of rustc. Would it be possible to release this a 0.0.4-1 or something similar? Everything on crates.io at the moment won't compile on the latest nightly.

### DIFF
--- a/ci/docker/i586-unknown-linux-gnu/Dockerfile
+++ b/ci/docker/i586-unknown-linux-gnu/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:17.04
+FROM ubuntu:17.10
 RUN apt-get update && apt-get install -y --no-install-recommends \
   gcc-multilib \
   libc6-dev \
--- a/ci/docker/i686-unknown-linux-gnu/Dockerfile
+++ b/ci/docker/i686-unknown-linux-gnu/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:17.04
+FROM ubuntu:17.10
 RUN apt-get update && apt-get install -y --no-install-recommends \
   gcc-multilib \
   libc6-dev \
--- a/ci/docker/x86_64-unknown-linux-gnu-emulated/Dockerfile
+++ b/ci/docker/x86_64-unknown-linux-gnu-emulated/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:17.04
+FROM ubuntu:17.10
 RUN apt-get update && apt-get install -y --no-install-recommends \
   gcc \
   libc6-dev \
--- a/ci/docker/x86_64-unknown-linux-gnu/Dockerfile
+++ b/ci/docker/x86_64-unknown-linux-gnu/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:17.04
+FROM ubuntu:17.10
 RUN apt-get update && apt-get install -y --no-install-recommends \
   gcc \
   libc6-dev \
--- a/ci/dox.sh
+++ b/ci/dox.sh
@@ -22,12 +22,18 @@ dox() {
   rm -rf target/doc/$arch
   mkdir target/doc/$arch
 
-  cargo build --target $target
+  cargo build --target $target --manifest-path crates/stdsimd/Cargo.toml
 
   rustdoc --target $target \
-          -o target/doc/$arch src/lib.rs \
-          --crate-name stdsimd \
+          -o target/doc/$arch crates/coresimd/src/lib.rs \
+          --crate-name coresimd \
           --library-path target/$target/debug/deps
+  rustdoc --target $target \
+          -o target/doc/$arch crates/stdsimd/src/lib.rs \
+          --crate-name stdsimd \
+          --library-path target/$target/debug/deps \
+          --extern cfg_if=`ls target/$target/debug/deps/libcfg_if-*.rlib` \
+          --extern libc=`ls target/$target/debug/deps/liblibc-*.rlib`
 }
 
 dox i686 i686-unknown-linux-gnu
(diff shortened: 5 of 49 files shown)


gnzlbg:
Something is wrong with the travis builds, I just restarted one build bot to see if it is spurious or not, but doesn't look good :/

gnzlbg:
@alexcrichton looks like something is wrong with the docker builds: 

```shell
E: Failed to fetch http://security.ubuntu.com/ubuntu/dists/zesty-security/universe/source/Sources  404  Not Found [IP: 91.189.88.162 80]
E: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/zesty/universe/source/Sources  404  Not Found [IP: 91.189.88.152 80]
E: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/zesty-updates/universe/source/Sources  404  Not Found [IP: 91.189.88.152 80]
E: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/zesty-backports/main/binary-amd64/Packages  404  Not Found [IP: 91.189.88.152 80]
E: Some index files failed to download. They have been ignored, or old ones used instead.
The command '/bin/sh -c apt-get update && apt-get install -y --no-install-recommends   gcc   libc6-dev   file   make   ca-certificates' returned a non-zero code: 100
The command "ci/run-docker.sh $TARGET $FEATURES" exited with 100.
```

AdamNiederer:
This branch is using a Travis config from a few months ago, so I wouldn't be surprised. I'll see if I can replay any travis-related commits between now and then.

gnzlbg:
@AdamNiederer as mentioned in the closed PR, I think time would be better spent in just doing a new release of the master branch with this fixed.

gnzlbg:
If you manage to get travis to run, I can merge this into that branch and do a 0.0.5 release, but that branch is really out-dated at this point.

gnzlbg:
@AdamNiederer I've pushed a couple of commits that make master green again, maybe those would help you here.

gnzlbg:
@AdamNiederer so we talked about this on IRC and decided that coresimd/stdsimd on crates.io are deprecated and our recommendation is to migrate to `std::arch` and `std::simd`. The `std::arch` module is on its path towards stabilization, while the `std::simd` module is still unstable, and might (and will) break every now and then. Now that we have an RFC, once the RFC is implemented it will hopefully stop breaking, but there are a couple of breaking changes in the RFC that aren't in `std::simd` yet.

We don't know whether maintaining a fork of `coresimd`/`stdsimd` might work good enough to be worth it. A couple of changes like this one will land in rustc, and while `std::arch`/`std::simd` will be kept in sync with those, it is hard to tell how hard it would be to do the same for the fork, or even if that will be at all possible. So the recommendation is basically:

* if you want as little breakage as possible, migrate to `std::arch` - this might be more or less work depending on how much you are currently relying on `std::simd`. 
* if you are willing to live with some breakage, migrate to `std::arch` + `std::simd`. This will obviously be the best thing for `std::simd`, since it will get more people to test it, but it might not be the best for `faster` users, depending on how often breaking changes land. We will obviously try to not break `std::simd` every day, but until now things are sync'ed with upstream every two weeks, so that's more or less how often things could break.

AdamNiederer:
Hm, I think I'll have to port to `std::simd` and `std::arch`, as faster relies quite heavily on both the types and the vendor intrinsics.

As an aside, it seems like `std::simd`'s RFC makes no mention of `std::arch`. Assuming the ABI guarantees of both make `std::simd`'s types generally safe to use in `std::arch`'s functions, it still looks like I'm going to have to do a lot of shimming to make both modules talk. Some thing like the following for each intrinsic on each arch:

```rust
fn _mm_add_ps(a: f32x4, b: f32x4) -> f32x4 {
    unsafe { transmute(std::arch::x86::_mm_add_ps(transmute(a), transmute(b))) }
}
```

That can easily be done with a code generator, but I'm wondering if there's a longer-term plan for interoperability between the two modules.

gnzlbg:
> As an aside, it seems like std::simd's RFC makes no mention of std::arch

IIRC `std::simd` does state that vector types of the same width are layout-compatible, so transmuting between them is safe.

What it doesn't specify is what the layout of vector types actually is, so while transmuting from an array/tuple to a vector type might work, it is not specified anywhere.

>  Something like the following for each intrinsic on each arch:

Yeah that's correct. In the mean time you can write: 

```rust
fn _mm_add_ps(a: f32x4, b: f32x4) -> f32x4 {
    std::arch::x86::_mm_add_ps(a.into_bits(), b.into_bits()).into_bits()
}
```

but that is not proposed for stabilization, so if you can use `mem::transmute`, you should prefer that for now.

gnzlbg:
@AdamNiederer so I've opened a repo to expose the `std::arch` intrinsics using the portable vector types: https://github.com/gnzlbg/typed_arch I've only have added a couple of examples, and will start adding all the CI necessary. Feel free to use that, and submit PR with the intrinsics you need. Once I finish adding CI, I will start the discussion in #rust-libs to add it to the nursery, because there are a couple of things that until resolved (e.g. `const function arguments` / type-level consts), are necessarily going to tie this repo to the compiler a bit. Ideally, by the time `std::simd` is stabilized this repo should work on stable Rust as well.

AdamNiederer:
Shoot, I wrote a [code generator](https://github.com/AdamNiederer/vektor-gen/blob/master/scrape.py) to do [exactly this](https://github.com/AdamNiederer/vektor) before checking my email today.

I'm not super excited about going through and manually re-typing stdsimd's intrinsics (and I'm sure you're not either), so wouldn't a more automatic solution be the way to go?

gnzlbg:
@AdamNiederer I generated the ones of typed_arch automatically as well, but the generator got it wrong in so many cases that I decided to manually verify all of them and only check in those who have been manually verified.

I checked your library to see how your generator did, and it did as bad as mine. Tip: your x86 code is not getting compiled for `x86_64` ;)

PR #414, opened 2018-04-04
Stabilize x86/x86_64 intrinsics

alexcrichton:
This commit stabilizes all intrinsics in the `x86` and `x86_64` modules, namely
allowing stabilization of the `arch::x86` and `arch::x86_64` module in libstd.
Stabilizations here were applied in an automated fashion using [this
script][scr], and notably everything related to `__m64` was omitted from this
round of stabilization

[scr]: https://gist.github.com/alexcrichton/5b456d495d6fe1df46a158754565c7a5

### DIFF
--- a/coresimd/x86/cpuid.rs
+++ b/coresimd/x86/cpuid.rs
@@ -10,14 +10,19 @@ use stdsimd_test::assert_instr;
 /// Result of the `cpuid` instruction.
 #[derive(Copy, Clone, Eq, Ord, PartialEq, PartialOrd)]
 #[cfg_attr(feature = "cargo-clippy", allow(stutter))]
+#[stable(feature = "simd_x86", since = "1.27.0")]
 pub struct CpuidResult {
     /// EAX register.
+    #[stable(feature = "simd_x86", since = "1.27.0")]
     pub eax: u32,
     /// EBX register.
+    #[stable(feature = "simd_x86", since = "1.27.0")]
     pub ebx: u32,
     /// ECX register.
+    #[stable(feature = "simd_x86", since = "1.27.0")]
     pub ecx: u32,
     /// EDX register.
+    #[stable(feature = "simd_x86", since = "1.27.0")]
     pub edx: u32,
 }
 
@@ -46,6 +51,7 @@ pub struct CpuidResult {
 /// [amd64_ref]: http://support.amd.com/TechDocs/24594.pdf
 #[inline]
 #[cfg_attr(test, assert_instr(cpuid))]
+#[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn __cpuid_count(leaf: u32, sub_leaf: u32) -> CpuidResult {
     let mut r = mem::uninitialized::<CpuidResult>();
     if cfg!(target_arch = "x86") {
@@ -66,6 +72,7 @@ pub unsafe fn __cpuid_count(leaf: u32, sub_leaf: u32) -> CpuidResult {
 /// See [`__cpuid_count`](fn.__cpuid_count.html).
 #[inline]
 #[cfg_attr(test, assert_instr(cpuid))]
+#[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn __cpuid(leaf: u32) -> CpuidResult {
     __cpuid_count(leaf, 0)
 }
@@ -114,6 +121,7 @@ pub fn has_cpuid() -> bool {
 /// See also [`__cpuid`](fn.__cpuid.html) and
 /// [`__cpuid_count`](fn.__cpuid_count.html).
 #[inline]
+#[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn __get_cpuid_max(leaf: u32) -> (u32, u32) {
     let CpuidResult { eax, ebx, .. } = __cpuid(leaf);
     (eax, ebx)
--- a/coresimd/x86/mod.rs
+++ b/coresimd/x86/mod.rs
@@ -33,7 +33,8 @@ types! {
     /// # Examples
     ///
     /// ```
-    /// # #![feature(cfg_target_feature, target_feature, stdsimd)]
+    /// # #![feature(stdsimd)]
+    /// # #![cfg_attr(not(dox), feature(cfg_target_feature, target_feature))]
     /// # #![cfg_attr(not(dox), no_std)]
     /// # #[cfg(not(dox))]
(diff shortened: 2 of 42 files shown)


coresimd/x86/cpuid.rs
```diff
@@ -44,8 +49,11 @@ pub struct CpuidResult {
 /// [wiki_cpuid]: https://en.wikipedia.org/wiki/CPUID
 /// [intel64_ref]: http://www.intel.de/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-instruction-set-reference-manual-325383.pdf
 /// [amd64_ref]: http://support.amd.com/TechDocs/24594.pdf
+///
+/// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=__cpuid_count)
```
gnzlbg: This is not an "official" Intel intrinsic, so this link returns empty. IIRC at least the cpuid intrinsics and the read/writeeflags intrinsics will suffer from this. There should be a list of these or their modules in the `stdsimd-verify` module.

coresimd/x86/mod.rs
```diff
@@ -598,5 +606,3 @@ pub use self::aes::*;
 mod rdrand;
 pub use self::rdrand::*;
 
-mod sha;
```
gnzlbg: What happened here?

coresimd/x86/sha.rs
```diff
@@ -26,29 +26,38 @@ use stdsimd_test::assert_instr;
 /// Perform an intermediate calculation for the next four SHA1 message values
```
gnzlbg: Looks like this module was removed from `x86/mod.rs` :/

coresimd/x86/sse4a.rs
```diff
@@ -33,9 +33,12 @@ extern "C" {
 ///
 /// If `length == 0 && index > 0` or `lenght + index > 64` the result is
 /// undefined.
+///
+/// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_extract_si64)
```
gnzlbg: So these intrinsics are part of the AMD intrinsics, and are not in the intel IntrinsicsGuide. AMD only provides a pdf with the spec AFAICT so the correct place to point them at would be:

```
/// [AMD64 Architecture Programmer's Manual, Volume 3: General-Purpose and System Instructions](http://support.amd.com/TechDocs/24594.pdf) 
```

coresimd/x86/tbm.rs
```diff
@@ -67,62 +71,80 @@ pub fn _bextr2_u64(a: u64, control: u64) -> u64 {
 /// Clears all bits below the least significant zero bit of `x`.
 ///
 /// If there is no zero bit in `x`, it returns zero.
+///
+/// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_blcfill_u32)
 #[inline]
```
gnzlbg: here the same as for sse4a applies, these are AMD only.

coresimd/x86/tbm.rs
```diff
@@ -37,6 +38,7 @@ pub fn _bextr_u32(a: u32, start: u32, len: u32) -> u32 {
 /// the least significant bits of the result.
 #[inline]
 #[target_feature(enable = "tbm")]
+#[stable(feature = "simd_x86", since = "1.27.0")]
```
gnzlbg: I think it would be better to remove this in this comment. Not that we fix it some day and it suddenly becomes stable.

coresimd/x86/test.rs
```diff
@@ -12,6 +12,7 @@ pub unsafe fn assert_eq_m64(a: __m64, b: __m64) {
 }
 
 #[target_feature(enable = "sse2")]
+#[stable(feature = "simd_x86", since = "1.27.0")]
```
gnzlbg: I don't think anything in this file should have this attribute.

coresimd/x86_64/bmi.rs
```diff
@@ -14,10 +14,13 @@ use stdsimd_test::assert_instr;
```
gnzlbg: OCD: I just realized that this file is still called bmi.rs, but this and the one in x86 should be renamed to bmi1.rs to match the target feature name.

coresimd/x86_64/mod.rs
```diff
@@ -38,5 +38,3 @@ pub use self::avx2::*;
 mod bswap;
 pub use self::bswap::*;
 
-mod rdrand;
```
gnzlbg: Here again, was this intended?

stdsimd/arch/detect/arch/x86.rs
```diff
@@ -16,8 +16,22 @@
...
+///
+/// This macro is provided in the standard library and will detect at runtime
+/// whether the specified CPU feature is detected. This does *not* resolve at
+/// compile time unless the specified feature is already enabled for the entire
+/// crate. Runtime detection currently relies mostly on the `cpuid` instruction.
+///
+/// This macro only takes one argument which is a string literal of the feature
+/// being tested for. The feature names supported are the lowercase versions of
+/// the ones defined by Intel in [their documentation][docs].
+///
```
gnzlbg: Do we want to list the whitelisted features here?

gnzlbg: Maybe also mentions that these names are the same as those used by `#[target_feature]`, and maybe it might be also worth mentioning twice that `is_x86_feature_detected("avx,avx2")` is not supported (even though it works for target feature.

gnzlbg:
So the first pass looks good! There are a couple of places where the script needs manual intervention, once those are finished and test pass I'll go over this one more time in more detail.

stdsimd/arch/detect/arch/x86.rs
alexcrichton: Excellent ideas!

alexcrichton:
Thanks for the close eye on review! I think I've addressed everything

alexcrichton:
Also FWIW we should hold off on merging this until the FCP has finished

alexcrichton:
Ok! I've merged this into a `next-breaking` branch and we'll update to that in the Rust repo. Once it's landed in nightly I'll circle back here to fix CI.

alexcrichton:
Er actually looks like it all pased! (except rustfmt)

gnzlbg:
Yay!! Congrats!

PR #415, opened 2018-04-05
Enable fxsr

gnzlbg:
The `fxsr` feature was already whitelisted upstream. If this fails, then it has other issues.


alexcrichton:
r=me, but looks like rustfmt needs an update?

PR #421, opened 2018-04-11
add missing from impls for u8x16 and i8x16

gnzlbg:


gnzlbg:
@alexcrichton this is all green, rustfmt is failing because of https://github.com/rust-lang-nursery/rustfmt/issues/2616

I'd say let's just live with that till it is fixed upstream and the fix lands on rustup.

PR #425, opened 2018-04-12
Workarounds for all/any mask reductions on x86, armv7, and aarch64

### LINKED ISSUE #362: Boolean vectors should use hand-tuned any() and all()
Thank you for adding portable SIMD and boolean vectors!

In `stdsimd`, `any()` and `all()` are based LLVM cross-ISA on bitwise reductions across the vector, so these operations fail to use the type system information that is the reason for their existence: The information that all the bits of each lane are the same. I.e. in `b8x16`, each lane is either 0 or 0xFF.

This means that looking at all the bits is unnecessary, since for each lane, examining any one bit is sufficient. In the SSE2 case, `_mm_movemask_epi8` takes one bit from each lane of a `u8x16`. It's what the `simd` crate uses.

I've uploaded a [demo crate](https://github.com/hsivonen/anyall) that exports uninlined versions of `any()` and `all()` on `u8x16` using both `simd` and `stdsimd`. The generated code for the `simd` crate versions is much simpler. Code generated using `stdsimd` should be as good.

```asm
    .text
    .file   "anyall0-66776763bbd7db1b14e447f414e2cf9a.rs"
    .section    .text.stdsimd_any_8x16,"ax",@progbits
    .globl  stdsimd_any_8x16
    .p2align    4, 0x90
    .type   stdsimd_any_8x16,@function
stdsimd_any_8x16:
    .cfi_startproc
    movdqa  (%rdi), %xmm0
    pshufd  $78, %xmm0, %xmm1
    por %xmm0, %xmm1
    pshufd  $229, %xmm1, %xmm0
    por %xmm1, %xmm0
    movdqa  %xmm0, %xmm1
    psrld   $16, %xmm1
    por %xmm0, %xmm1
    movdqa  %xmm1, %xmm0
    psrlw   $8, %xmm0
    por %xmm1, %xmm0
    movd    %xmm0, %eax
    testb   %al, %al
    setne   %al
    retq
.Lfunc_end0:
    .size   stdsimd_any_8x16, .Lfunc_end0-stdsimd_any_8x16
    .cfi_endproc

    .section    .text.stdsimd_all_8x16,"ax",@progbits
    .globl  stdsimd_all_8x16
    .p2align    4, 0x90
    .type   stdsimd_all_8x16,@function
stdsimd_all_8x16:
    .cfi_startproc
    movdqa  (%rdi), %xmm0
    pshufd  $78, %xmm0, %xmm1
    pand    %xmm0, %xmm1
    pshufd  $229, %xmm1, %xmm0
    pand    %xmm1, %xmm0
    movdqa  %xmm0, %xmm1
    psrld   $16, %xmm1
    pand    %xmm0, %xmm1
    movdqa  %xmm1, %xmm0
    psrlw   $8, %xmm0
    pand    %xmm1, %xmm0
    movd    %xmm0, %eax
    testb   %al, %al
    setne   %al
    retq
.Lfunc_end1:
    .size   stdsimd_all_8x16, .Lfunc_end1-stdsimd_all_8x16
    .cfi_endproc

    .section    .text.simd_any_8x16,"ax",@progbits
    .globl  simd_any_8x16
    .p2align    4, 0x90
    .type   simd_any_8x16,@function
simd_any_8x16:
    .cfi_startproc
    movdqa  (%rdi), %xmm0
    pmovmskb    %xmm0, %eax
    testl   %eax, %eax
    setne   %al
    retq
.Lfunc_end2:
    .size   simd_any_8x16, .Lfunc_end2-simd_any_8x16
    .cfi_endproc

    .section    .text.simd_all_8x16,"ax",@progbits
    .globl  simd_all_8x16
    .p2align    4, 0x90
    .type   simd_all_8x16,@function
simd_all_8x16:
    .cfi_startproc
    movdqa  (%rdi), %xmm0
    pmovmskb    %xmm0, %eax
    cmpl    $65535, %eax
    sete    %al
    retq
.Lfunc_end3:
    .size   simd_all_8x16, .Lfunc_end3-simd_all_8x16
    .cfi_endproc


    .section    ".note.GNU-stack","",@progbits
```

gnzlbg:
This commit adds workarounds for the mask reductions: `all` and `any`.

Note:

* not all LLVM bugs have been properly reported yet
* workarounds for 256-bit wide vector masks on armv7 and aarch64 are
  not provided in this PR

### `x86_64` with `MMX` enabled

```asm
all_8x8:
 push    rbp
 mov     rbp, rsp
 movzx   eax, byte, ptr, [rdi, +, 7]
 movd    xmm0, eax
 movzx   eax, byte, ptr, [rdi, +, 6]
 movd    xmm1, eax
 punpcklwd xmm1, xmm0
 movzx   eax, byte, ptr, [rdi, +, 5]
 movd    xmm0, eax
 movzx   eax, byte, ptr, [rdi, +, 4]
 movd    xmm2, eax
 punpcklwd xmm2, xmm0
 punpckldq xmm2, xmm1
 movzx   eax, byte, ptr, [rdi, +, 3]
 movd    xmm0, eax
 movzx   eax, byte, ptr, [rdi, +, 2]
 movd    xmm1, eax
 punpcklwd xmm1, xmm0
 movzx   eax, byte, ptr, [rdi, +, 1]
 movd    xmm0, eax
 movzx   eax, byte, ptr, [rdi]
 movd    xmm3, eax
 punpcklwd xmm3, xmm0
 punpckldq xmm3, xmm1
 punpcklqdq xmm3, xmm2
 movdqa  xmm0, xmmword, ptr, [rip, +, LCPI9_0]
 pand    xmm3, xmm0
 pcmpeqw xmm3, xmm0
 pshufd  xmm0, xmm3, 78
 pand    xmm0, xmm3
 pshufd  xmm1, xmm0, 229
 pand    xmm1, xmm0
 movdqa  xmm0, xmm1
 psrld   xmm0, 16
 pand    xmm0, xmm1
 movd    eax, xmm0
 and     al, 1
 pop     rbp
 ret
any_8x8:
 push    rbp
 mov     rbp, rsp
 movzx   eax, byte, ptr, [rdi, +, 7]
 movd    xmm0, eax
 movzx   eax, byte, ptr, [rdi, +, 6]
 movd    xmm1, eax
 punpcklwd xmm1, xmm0
 movzx   eax, byte, ptr, [rdi, +, 5]
 movd    xmm0, eax
 movzx   eax, byte, ptr, [rdi, +, 4]
 movd    xmm2, eax
 punpcklwd xmm2, xmm0
 punpckldq xmm2, xmm1
 movzx   eax, byte, ptr, [rdi, +, 3]
 movd    xmm0, eax
 movzx   eax, byte, ptr, [rdi, +, 2]
 movd    xmm1, eax
 punpcklwd xmm1, xmm0
 movzx   eax, byte, ptr, [rdi, +, 1]
 movd    xmm0, eax
 movzx   eax, byte, ptr, [rdi]
 movd    xmm3, eax
 punpcklwd xmm3, xmm0
 punpckldq xmm3, xmm1
 punpcklqdq xmm3, xmm2
 movdqa  xmm0, xmmword, ptr, [rip, +, LCPI8_0]
 pand    xmm3, xmm0
 pcmpeqw xmm3, xmm0
 pshufd  xmm0, xmm3, 78
 por     xmm0, xmm3
 pshufd  xmm1, xmm0, 229
 por     xmm1, xmm0
 movdqa  xmm0, xmm1
 psrld   xmm0, 16
 por     xmm0, xmm1
 movd    eax, xmm0
 and     al, 1
 pop     rbp
 ret
```

After this PR for `m8x8`, `m16x4`, `m32x2`:

```asm
all_8x8:
 push    rbp
 mov     rbp, rsp
 movq    mm0, qword, ptr, [rdi]
 pmovmskb eax, mm0
 cmp     eax, 255
 sete    al
 pop     rbp
 ret
any_8x8:
 push    rbp
 mov     rbp, rsp
 movq    mm0, qword, ptr, [rdi]
 pmovmskb eax, mm0
 test    eax, eax
 setne   al
 pop     rbp
 ret
```

### `x86` with `MMX` enabled

Before this PR:

```asm
all_8x8:
 call    L9$pb
L9$pb:
 pop     eax
 mov     ecx, dword, ptr, [esp, +, 4]
 movzx   edx, byte, ptr, [ecx, +, 7]
 movd    xmm0, edx
 movzx   edx, byte, ptr, [ecx, +, 6]
 movd    xmm1, edx
 punpcklwd xmm1, xmm0
 movzx   edx, byte, ptr, [ecx, +, 5]
 movd    xmm0, edx
 movzx   edx, byte, ptr, [ecx, +, 4]
 movd    xmm2, edx
 punpcklwd xmm2, xmm0
 punpckldq xmm2, xmm1
 movzx   edx, byte, ptr, [ecx, +, 3]
 movd    xmm0, edx
 movzx   edx, byte, ptr, [ecx, +, 2]
 movd    xmm1, edx
 punpcklwd xmm1, xmm0
 movzx   edx, byte, ptr, [ecx, +, 1]
 movd    xmm0, edx
 movzx   ecx, byte, ptr, [ecx]
 movd    xmm3, ecx
 punpcklwd xmm3, xmm0
 punpckldq xmm3, xmm1
 punpcklqdq xmm3, xmm2
 movdqa  xmm0, xmmword, ptr, [eax, +, LCPI9_0-L9$pb]
 pand    xmm3, xmm0
 pcmpeqw xmm3, xmm0
 pshufd  xmm0, xmm3, 78
 pand    xmm0, xmm3
 pshufd  xmm1, xmm0, 229
 pand    xmm1, xmm0
 movdqa  xmm0, xmm1
 psrld   xmm0, 16
 pand    xmm0, xmm1
 movd    eax, xmm0
 and     al, 1
 ret
any_8x8:
 call    L8$pb
L8$pb:
 pop     eax
 mov     ecx, dword, ptr, [esp, +, 4]
 movzx   edx, byte, ptr, [ecx, +, 7]
 movd    xmm0, edx
 movzx   edx, byte, ptr, [ecx, +, 6]
 movd    xmm1, edx
 punpcklwd xmm1, xmm0
 movzx   edx, byte, ptr, [ecx, +, 5]
 movd    xmm0, edx
 movzx   edx, byte, ptr, [ecx, +, 4]
 movd    xmm2, edx
 punpcklwd xmm2, xmm0
 punpckldq xmm2, xmm1
 movzx   edx, byte, ptr, [ecx, +, 3]
 movd    xmm0, edx
 movzx   edx, byte, ptr, [ecx, +, 2]
 movd    xmm1, edx
 punpcklwd xmm1, xmm0
 movzx   edx, byte, ptr, [ecx, +, 1]
 movd    xmm0, edx
 movzx   ecx, byte, ptr, [ecx]
 movd    xmm3, ecx
 punpcklwd xmm3, xmm0
 punpckldq xmm3, xmm1
 punpcklqdq xmm3, xmm2
 movdqa  xmm0, xmmword, ptr, [eax, +, LCPI8_0-L8$pb]
 pand    xmm3, xmm0
 pcmpeqw xmm3, xmm0
 pshufd  xmm0, xmm3, 78
 por     xmm0, xmm3
 pshufd  xmm1, xmm0, 229
 por     xmm1, xmm0
 movdqa  xmm0, xmm1
 psrld   xmm0, 16
 por     xmm0, xmm1
 movd    eax, xmm0
 and     al, 1
 ret
```

After this PR:

```asm
all_8x8:
 mov     eax, dword, ptr, [esp, +, 4]
 movq    mm0, qword, ptr, [eax]
 pmovmskb eax, mm0
 cmp     eax, 255
 sete    al
 ret
any_8x8:
 mov     eax, dword, ptr, [esp, +, 4]
 movq    mm0, qword, ptr, [eax]
 pmovmskb eax, mm0
 test    eax, eax
 setne   al
 ret 
```

### `aarch64`

Before this PR:

```asm
all_8x8:
 ldr     d0, [x0]
 umov    w8, v0.b[0]
 umov    w9, v0.b[1]
 tst     w8, #0xff
 umov    w10, v0.b[2]
 cset    w8, ne
 tst     w9, #0xff
 cset    w9, ne
 tst     w10, #0xff
 umov    w10, v0.b[3]
 and     w8, w8, w9
 cset    w9, ne
 tst     w10, #0xff
 umov    w10, v0.b[4]
 and     w8, w9, w8
 cset    w9, ne
 tst     w10, #0xff
 umov    w10, v0.b[5]
 and     w8, w9, w8
 cset    w9, ne
 tst     w10, #0xff
 umov    w10, v0.b[6]
 and     w8, w9, w8
 cset    w9, ne
 tst     w10, #0xff
 umov    w10, v0.b[7]
 and     w8, w9, w8
 cset    w9, ne
 tst     w10, #0xff
 and     w8, w9, w8
 cset    w9, ne
 and     w0, w9, w8
 ret
any_8x8:
 ldr     d0, [x0]
 umov    w8, v0.b[0]
 umov    w9, v0.b[1]
 orr     w8, w8, w9
 umov    w9, v0.b[2]
 orr     w8, w8, w9
 umov    w9, v0.b[3]
 orr     w8, w8, w9
 umov    w9, v0.b[4]
 orr     w8, w8, w9
 umov    w9, v0.b[5]
 orr     w8, w8, w9
 umov    w9, v0.b[6]
 orr     w8, w8, w9
 umov    w9, v0.b[7]
 orr     w8, w8, w9
 tst     w8, #0xff
 cset    w0, ne
 ret
```

After this PR:

```asm
all_8x8:
 ldr     d0, [x0]
 mov     v0.d[1], v0.d[0]
 uminv   b0, v0.16b
 fmov    w8, s0
 tst     w8, #0xff
 cset    w0, ne
 ret
any_8x8:
 ldr     d0, [x0]
 mov     v0.d[1], v0.d[0]
 umaxv   b0, v0.16b
 fmov    w8, s0
 tst     w8, #0xff
 cset    w0, ne
 ret
```

### `ARMv7` + `neon`

Before this PR:

```asm
all_8x8:
 vmov.i8 d0, #0x1
 vldr    d1, [r0]
 vtst.8  d0, d1, d0
 vext.8  d1, d0, d0, #4
 vand    d0, d0, d1
 vext.8  d1, d0, d0, #2
 vand    d0, d0, d1
 vdup.8  d1, d0[1]
 vand    d0, d0, d1
 vmov.u8 r0, d0[0]
 and     r0, r0, #1
 bx      lr
any_8x8:
 vmov.i8 d0, #0x1
 vldr    d1, [r0]
 vtst.8  d0, d1, d0
 vext.8  d1, d0, d0, #4
 vorr    d0, d0, d1
 vext.8  d1, d0, d0, #2
 vorr    d0, d0, d1
 vdup.8  d1, d0[1]
 vorr    d0, d0, d1
 vmov.u8 r0, d0[0]
 and     r0, r0, #1
 bx      lr
```

After this PR:

```asm
all_8x8:
 vldr    d0, [r0]
 b       <m8x8 as All>::all

<m8x8 as All>::all:
 vpmin.u8 d16, d0, d16
 vpmin.u8 d16, d16, d16
 vpmin.u8 d0, d16, d16
 b       m8x8::extract

any_8x8:
 vldr    d0, [r0]
 b       <m8x8 as Any>::any
 
<m8x8 as Any>::any:
 vpmax.u8 d16, d0, d16
 vpmax.u8 d16, d16, d16
 vpmax.u8 d0, d16, d16
 b       m8x8::extract
```

(note: inlining does not work properly on ARMv7)

### `x86_64` with SSE2 enabled

Before this PR:

```asm
all_8x16:
 push    rbp
 mov     rbp, rsp
 movdqa  xmm0, xmmword, ptr, [rip, +, LCPI9_0]
 movdqa  xmm1, xmmword, ptr, [rdi]
 pand    xmm1, xmm0
 pcmpeqb xmm1, xmm0
 pmovmskb eax, xmm1
 xor     ecx, ecx
 cmp     eax, 65535
 mov     eax, -1
 cmovne  eax, ecx
 and     al, 1
 pop     rbp
 ret
any_8x16:
 push    rbp
 mov     rbp, rsp
 movdqa  xmm0, xmmword, ptr, [rip, +, LCPI8_0]
 movdqa  xmm1, xmmword, ptr, [rdi]
 pand    xmm1, xmm0
 pcmpeqb xmm1, xmm0
 pmovmskb eax, xmm1
 neg     eax
 sbb     eax, eax
 and     al, 1
 pop     rbp
 ret
```

After this PR:

```asm
all_8x16:
 push    rbp
 mov     rbp, rsp
 movdqa  xmm0, xmmword, ptr, [rdi]
 pmovmskb eax, xmm0
 cmp     eax, 65535
 sete    al
 pop     rbp
 ret
any_8x16:
 push    rbp
 mov     rbp, rsp
 movdqa  xmm0, xmmword, ptr, [rdi]
 pmovmskb eax, xmm0
 test    eax, eax
 setne   al
 pop     rbp
 ret
```

### `aarch64`

Before this PR:

```asm
all_8x16:
 ldr     q0, [x0]
 umov    w8, v0.b[0]
 umov    w9, v0.b[1]
 tst     w8, #0xff
 umov    w10, v0.b[2]
 cset    w8, ne
 tst     w9, #0xff
 cset    w9, ne
 tst     w10, #0xff
 umov    w10, v0.b[3]
 and     w8, w8, w9
 cset    w9, ne
 tst     w10, #0xff
 umov    w10, v0.b[4]
 and     w8, w9, w8
 cset    w9, ne
 tst     w10, #0xff
 umov    w10, v0.b[5]
 and     w8, w9, w8
 cset    w9, ne
 tst     w10, #0xff
 umov    w10, v0.b[6]
 and     w8, w9, w8
 cset    w9, ne
 tst     w10, #0xff
 umov    w10, v0.b[7]
 and     w8, w9, w8
 cset    w9, ne
 tst     w10, #0xff
 umov    w10, v0.b[8]
 and     w8, w9, w8
 cset    w9, ne
 tst     w10, #0xff
 umov    w10, v0.b[9]
 and     w8, w9, w8
 cset    w9, ne
 tst     w10, #0xff
 umov    w10, v0.b[10]
 and     w8, w9, w8
 cset    w9, ne
 tst     w10, #0xff
 umov    w10, v0.b[11]
 and     w8, w9, w8
 cset    w9, ne
 tst     w10, #0xff
 umov    w10, v0.b[12]
 and     w8, w9, w8
 cset    w9, ne
 tst     w10, #0xff
 umov    w10, v0.b[13]
 and     w8, w9, w8
 cset    w9, ne
 tst     w10, #0xff
 umov    w10, v0.b[14]
 and     w8, w9, w8
 cset    w9, ne
 tst     w10, #0xff
 umov    w10, v0.b[15]
 and     w8, w9, w8
 cset    w9, ne
 tst     w10, #0xff
 and     w8, w9, w8
 cset    w9, ne
 and     w0, w9, w8
 ret
any_8x16:
 ldr     q0, [x0]
 umov    w8, v0.b[0]
 umov    w9, v0.b[1]
 orr     w8, w8, w9
 umov    w9, v0.b[2]
 orr     w8, w8, w9
 umov    w9, v0.b[3]
 orr     w8, w8, w9
 umov    w9, v0.b[4]
 orr     w8, w8, w9
 umov    w9, v0.b[5]
 orr     w8, w8, w9
 umov    w9, v0.b[6]
 orr     w8, w8, w9
 umov    w9, v0.b[7]
 orr     w8, w8, w9
 umov    w9, v0.b[8]
 orr     w8, w8, w9
 umov    w9, v0.b[9]
 orr     w8, w8, w9
 umov    w9, v0.b[10]
 orr     w8, w8, w9
 umov    w9, v0.b[11]
 orr     w8, w8, w9
 umov    w9, v0.b[12]
 orr     w8, w8, w9
 umov    w9, v0.b[13]
 orr     w8, w8, w9
 umov    w9, v0.b[14]
 orr     w8, w8, w9
 umov    w9, v0.b[15]
 orr     w8, w8, w9
 tst     w8, #0xff
 cset    w0, ne
 ret
```

After this PR:

```asm
all_8x16:
 ldr     q0, [x0]
 uminv   b0, v0.16b
 fmov    w8, s0
 tst     w8, #0xff
 cset    w0, ne
 ret
any_8x16:
 ldr     q0, [x0]
 umaxv   b0, v0.16b
 fmov    w8, s0
 tst     w8, #0xff
 cset    w0, ne
 ret
```

### `ARMv7` + `neon`

Before this PR:

```asm
all_8x16:
 vmov.i8 q0, #0x1
 vld1.64 {d2, d3}, [r0]
 vtst.8  q0, q1, q0
 vext.8  q1, q0, q0, #8
 vand    q0, q0, q1
 vext.8  q1, q0, q0, #4
 vand    q0, q0, q1
 vext.8  q1, q0, q0, #2
 vand    q0, q0, q1
 vdup.8  q1, d0[1]
 vand    q0, q0, q1
 vmov.u8 r0, d0[0]
 and     r0, r0, #1
 bx      lr
any_8x16:
 vmov.i8 q0, #0x1
 vld1.64 {d2, d3}, [r0]
 vtst.8  q0, q1, q0
 vext.8  q1, q0, q0, #8
 vorr    q0, q0, q1
 vext.8  q1, q0, q0, #4
 vorr    q0, q0, q1
 vext.8  q1, q0, q0, #2
 vorr    q0, q0, q1
 vdup.8  q1, d0[1]
 vorr    q0, q0, q1
 vmov.u8 r0, d0[0]
 and     r0, r0, #1
 bx      lr
```

After this PR:

```asm
all_8x16:
 vld1.64 {d0, d1}, [r0]
 b       <m8x16 as All>::all
 
<m8x16 as All>::all: 
 vpmin.u8 d0, d0, d
 b       <m8x8 as All>::all
any_8x16:
 vld1.64 {d0, d1}, [r0]
 b       <m8x16 as Any>::any
 
<m8x16 as Any>::any:
 vpmax.u8 d0, d0, d1
 b       <m8x8 as Any>::any
```

The inlining problems are pretty bad on ARMv7 + NEON.

#### With SSE2 enabled

Before this PR:

```asm
all_8x32:
 push    rbp
 mov     rbp, rsp
 movdqa  xmm0, xmmword, ptr, [rip, +, LCPI17_0]
 movdqa  xmm1, xmmword, ptr, [rdi]
 pand    xmm1, xmm0
 movdqa  xmm2, xmmword, ptr, [rdi, +, 16]
 pand    xmm2, xmm0
 pcmpeqb xmm2, xmm0
 pcmpeqb xmm1, xmm0
 pand    xmm1, xmm2
 pmovmskb eax, xmm1
 xor     ecx, ecx
 cmp     eax, 65535
 mov     eax, -1
 cmovne  eax, ecx
 and     al, 1
 pop     rbp
 ret
 any_8x32:
 push    rbp
 mov     rbp, rsp
 movdqa  xmm0, xmmword, ptr, [rdi]
 por     xmm0, xmmword, ptr, [rdi, +, 16]
 movdqa  xmm1, xmmword, ptr, [rip, +, LCPI16_0]
 pand    xmm0, xmm1
 pcmpeqb xmm0, xmm1
 pmovmskb eax, xmm0
 neg     eax
 sbb     eax, eax
 and     al, 1
 pop     rbp
 ret
```

After this PR:

```asm
all_8x32:
 push    rbp
 mov     rbp, rsp
 movdqa  xmm0, xmmword, ptr, [rdi]
 pmovmskb eax, xmm0
 cmp     eax, 65535
 jne     LBB17_1
 movdqa  xmm0, xmmword, ptr, [rdi, +, 16]
 pmovmskb ecx, xmm0
 mov     al, 1
 cmp     ecx, 65535
 je      LBB17_3
LBB17_1:
 xor     eax, eax
LBB17_3:
 pop     rbp
 ret
any_8x32:
 push    rbp
 mov     rbp, rsp
 movdqa  xmm0, xmmword, ptr, [rdi]
 pmovmskb ecx, xmm0
 mov     al, 1
 test    ecx, ecx
 je      LBB16_1
 pop     rbp
 ret
LBB16_1:
 movdqa  xmm0, xmmword, ptr, [rdi, +, 16]
 pmovmskb eax, xmm0
 test    eax, eax
 setne   al
 pop     rbp
 ret
```

#### With AVX enabled

Before this PR:

```asm
all_8x32:
 push    rbp
 mov     rbp, rsp
 vmovaps ymm0, ymmword, ptr, [rdi]
 vandps  ymm0, ymm0, ymmword, ptr, [rip, +, LCPI25_0]
 vextractf128 xmm1, ymm0, 1
 vpxor   xmm2, xmm2, xmm2
 vpcmpeqb xmm1, xmm1, xmm2
 vpcmpeqd xmm3, xmm3, xmm3
 vpxor   xmm1, xmm1, xmm3
 vpcmpeqb xmm0, xmm0, xmm2
 vpxor   xmm0, xmm0, xmm3
 vinsertf128 ymm0, ymm0, xmm1, 1
 vandps  ymm0, ymm0, ymm1
 vpermilps xmm1, xmm0, 78
 vandps  ymm0, ymm0, ymm1
 vpermilps xmm1, xmm0, 229
 vandps  ymm0, ymm0, ymm1
 vpsrld  xmm1, xmm0, 16
 vandps  ymm0, ymm0, ymm1
 vpsrlw  xmm1, xmm0, 8
 vandps  ymm0, ymm0, ymm1
 vpextrb eax, xmm0, 0
 and     al, 1
 pop     rbp
 vzeroupper
 ret
any_8x32:
 push    rbp
 mov     rbp, rsp
 vmovaps ymm0, ymmword, ptr, [rdi]
 vandps  ymm0, ymm0, ymmword, ptr, [rip, +, LCPI24_0]
 vextractf128 xmm1, ymm0, 1
 vpxor   xmm2, xmm2, xmm2
 vpcmpeqb xmm1, xmm1, xmm2
 vpcmpeqd xmm3, xmm3, xmm3
 vpxor   xmm1, xmm1, xmm3
 vpcmpeqb xmm0, xmm0, xmm2
 vpxor   xmm0, xmm0, xmm3
 vinsertf128 ymm0, ymm0, xmm1, 1
 vorps   ymm0, ymm0, ymm1
 vpermilps xmm1, xmm0, 78
 vorps   ymm0, ymm0, ymm1
 vpermilps xmm1, xmm0, 229
 vorps   ymm0, ymm0, ymm1
 vpsrld  xmm1, xmm0, 16
 vorps   ymm0, ymm0, ymm1
 vpsrlw  xmm1, xmm0, 8
 vorps   ymm0, ymm0, ymm1
 vpextrb eax, xmm0, 0
 and     al, 1
 pop     rbp
 vzeroupper
 ret
```

After this PR:

```asm
all_8x32:
 push    rbp
 mov     rbp, rsp
 vmovdqa ymm0, ymmword, ptr, [rdi]
 vxorps  xmm1, xmm1, xmm1
 vcmptrueps ymm1, ymm1, ymm1
 vptest  ymm0, ymm1
 setb    al
 pop     rbp
 vzeroupper
 ret
any_8x32:
 push    rbp
 mov     rbp, rsp
 vmovdqa ymm0, ymmword, ptr, [rdi]
 vptest  ymm0, ymm0
 setne   al
 pop     rbp
 vzeroupper
 ret
```

---

Closes #362 .

### DIFF
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -40,6 +40,7 @@ cargo_test "--release"
 case ${TARGET} in
     x86*)
         RUSTFLAGS="${RUSTFLAGS} -C target-feature=+avx"
+        export STDSIMD_DISABLE_ASSERT_INSTR=1
         cargo_test "--release"
         ;;
     *)
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -36,10 +36,10 @@ cargo_test() {
 cargo_test
 cargo_test "--release"
 
+# Test x86 targets compiled with AVX.
 case ${TARGET} in
     x86*)
-        RUSTFLAGS="${RUSTFLAGS} -C target-feature=+sse3,+ssse3,+sse4.1,+sse4.2,+avx,+popcnt,+lzcnt"
-        export STDSIMD_IGNORE_ASSERT_INSTR=1
+        RUSTFLAGS="${RUSTFLAGS} -C target-feature=+avx"
         cargo_test "--release"
         ;;
     *)
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -38,7 +38,7 @@ cargo_test "--release"
 
 case ${TARGET} in
     x86*)
-        RUSTFLAGS="${RUSTFLAGS} -C target-feature=+avx"
+        RUSTFLAGS="${RUSTFLAGS} -C target-feature=+sse3,+ssse3,+sse4.1,+sse4.2,+avx,+popcnt,+lzcnt"
         export STDSIMD_IGNORE_ASSERT_INSTR=1
         cargo_test "--release"
         ;;
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -35,3 +35,12 @@ cargo_test() {
 
 cargo_test
 cargo_test "--release"
+
+case ${TARGET} in
+    x86*)
+        RUSTFLAGS="${RUSTFLAGS} -C target-feature=+avx"
+        cargo_test "--release"
+        ;;
+    *)
+        ;;
+esac
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -39,6 +39,7 @@ cargo_test "--release"
 case ${TARGET} in
     x86*)
         RUSTFLAGS="${RUSTFLAGS} -C target-feature=+avx"
+        export STDSIMD_IGNORE_ASSERT_INSTR=1
         cargo_test "--release"
         ;;
     *)
--- a/coresimd/ppsv/codegen/masks_reductions.rs
+++ b/coresimd/ppsv/codegen/masks_reductions.rs
@@ -454,7 +454,7 @@ macro_rules! impl_mask_all_any {
     // 64-bit wide masks
     (m8x8) => {
         cfg_if! {
-            if #[cfg(any(target_arch = "x86", target_arch = "x86_64"))] {
+            if #[cfg(target_arch = "x86_64")] {
                 x86_64_mmx_movemask_impl!(m8x8, m8x16);
             } else if #[cfg(all(target_arch = "arm", target_feature = "v7", target_feature = "neon"))] {
                 arm_64_x8_v7_neon_impl!(m8x8, vpmin_u8, vpmax_u8);
@@ -467,7 +467,7 @@ macro_rules! impl_mask_all_any {
     };
     (m16x4) => {
         cfg_if! {
-            if #[cfg(any(target_arch = "x86", target_arch = "x86_64"))] {
+            if #[cfg(target_arch = "x86_64")] {
                 x86_64_mmx_movemask_impl!(m16x4, m16x8);
             } else if #[cfg(all(target_arch = "arm", target_feature = "v7", target_feature = "neon"))] {
                 arm_64_x4_v7_neon_impl!(m16x4, vpmin_u16, vpmax_u16);
@@ -480,7 +480,7 @@ macro_rules! impl_mask_all_any {
     };
     (m32x2) => {
         cfg_if! {
-            if #[cfg(any(target_arch = "x86", target_arch = "x86_64"))] {
+            if #[cfg(target_arch = "x86_64")] {
                 x86_64_mmx_movemask_impl!(m32x2, m32x4);
             } else if #[cfg(all(target_arch = "arm", target_feature = "v7", target_feature = "neon"))] {
                 arm_64_x2_v7_neon_impl!(m32x2, vpmin_u32, vpmax_u32);
--- a/coresimd/ppsv/codegen/masks_reductions.rs
+++ b/coresimd/ppsv/codegen/masks_reductions.rs
@@ -1,0 +1,603 @@
+//! LLVM6 currently generates sub-optimal code for the `all` mask reductions.
+//!
+//! See https://github.com/rust-lang-nursery/stdsimd/issues/362#issuecomment-372774371
+//! and the associated LLVM bug:
+//! https://bugs.llvm.org/show_bug.cgi?id=36702
+
+#![allow(unused)]
+
+use coresimd::simd::*;
+
+pub trait All: ::marker::Sized {
+    unsafe fn all(self) -> bool;
+}
+
+pub trait Any: ::marker::Sized {
+    unsafe fn any(self) -> bool;
+}
+
+// By default we use the simd_reduce_{all,any} intrinsics, which produces
+// sub-optimal code, except on aarch64 where that intrinsic is broken
+// due to https://bugs.llvm.org/show_bug.cgi?id=36796 so we just use
+// full-blown bitwise and/or reduction there.
+macro_rules! default_impl {
+    ($id:ident) => {
+        impl All for $id {
+            #[inline]
+            unsafe fn all(self) -> bool {
+                #[cfg(not(target_arch = "aarch64"))] {
+                    use coresimd::simd_llvm::simd_reduce_all;
+                    simd_reduce_all(self)
+                }
+                #[cfg(target_arch = "aarch64")] {
+                    // FIXME: Broken on AArch64
+                    // https://bugs.llvm.org/show_bug.cgi?id=36796
+                    self.and()
+                }
+            }
+        }
+
+        impl Any for $id {
+            #[inline]
+            unsafe fn any(self) -> bool {
+                #[cfg(not(target_arch = "aarch64"))] {
+                    use coresimd::simd_llvm::simd_reduce_any;
+                    simd_reduce_any(self)
+                }
+                #[cfg(target_arch = "aarch64")] {
+                    // FIXME: Broken on AArch64
+                    // https://bugs.llvm.org/show_bug.cgi?id=36796
+                    self.or()
+                }
+            }
+        }
+    };
+}
+
+// On x86 both SSE2 and AVX2 provide movemask instructions that can be used
+// here. The AVX2 instructions aren't necessarily better than the AVX
+// instructions below, so they aren't implemented here.
+//
+// FIXME: for mask generated from f32x4 LLVM6 emits pmovmskb but should emit
+// movmskps. Since the masks don't track whether they were produced by integer
+// or floating point vectors, we can't currently work around this yet. The
+// performance impact for this shouldn't be large, but this is filled as:
+// https://bugs.llvm.org/show_bug.cgi?id=37087
+#[cfg(all(any(target_arch = "x86", target_arch = "x86_64"), target_feature = "sse2"))]
+macro_rules! x86_128_sse2_movemask_impl {
+    ($id:ident) => {
+        impl All for $id {
+            #[inline]
+            #[target_feature(enable = "sse2")]
+            unsafe fn all(self) -> bool {
+                #[cfg(target_arch = "x86")]
+                use ::coresimd::arch::x86::_mm_movemask_epi8;
+                #[cfg(target_arch = "x86_64")]
+                use ::coresimd::arch::x86_64::_mm_movemask_epi8;
+                // _mm_movemask_epi8(a) creates a 16bit mask containing the most
+                // significant bit of each byte of `a`. If all bits are set,
+                // then all 16 lanes of the mask are true.
+                _mm_movemask_epi8(::mem::transmute(self)) == u16::max_value() as i32
+            }
+        }
+        impl Any for $id {
+            #[inline]
+            #[target_feature(enable = "sse2")]
+            unsafe fn any(self) -> bool {
+                #[cfg(target_arch = "x86")]
+                use ::coresimd::arch::x86::_mm_movemask_epi8;
+                #[cfg(target_arch = "x86_64")]
+                use ::coresimd::arch::x86_64::_mm_movemask_epi8;
+
+                _mm_movemask_epi8(::mem::transmute(self)) != 0
+            }
+        }
+    }
+}
+
+// On x86 with AVX we use _mm256_testc_si256 and _mm256_testz_si256.
+//
+// FIXME: for masks generated from floating point vectors one should use
+// x86_mm256_testc_ps, x86_mm256_testz_ps, x86_mm256_testc_pd,
+// x86_mm256_testz_pd.Since the masks don't track whether they were produced by
+// integer or floating point vectors, we can't currently work around this yet.
+//
+// TODO: investigate perf impact and fill LLVM bugs as necessary.
+#[cfg(all(any(target_arch = "x86", target_arch = "x86_64"), target_feature = "avx"))]
+macro_rules! x86_256_avx_test_impl {
+    ($id:ident) => {
+        impl All for $id {
+            #[inline]
+            #[target_feature(enable = "avx")]
+            unsafe fn all(self) -> bool {
+                #[cfg(target_arch = "x86")]
+                use ::coresimd::arch::x86::_mm256_testc_si256;
+                #[cfg(target_arch = "x86_64")]
+                use ::coresimd::arch::x86_64::_mm256_testc_si256;
+                _mm256_testc_si256(::mem::transmute(self),
+                                   ::mem::transmute($id::splat(true))) != 0
+            }
+        }
+        impl Any for $id {
+            #[inline]
+            #[target_feature(enable = "avx")]
+            unsafe fn any(self) -> bool {
+                #[cfg(target_arch = "x86")]
+                use ::coresimd::arch::x86::_mm256_testz_si256;
+                #[cfg(target_arch = "x86_64")]
+                use ::coresimd::arch::x86_64::_mm256_testz_si256;
+                _mm256_testz_si256(::mem::transmute(self),
+                                   ::mem::transmute(self)) == 0
+            }
+        }
+    }
+}
+
+// On x86 with SSE2 all/any for 256-bit wide vectors is implemented by executing
+// the algorithm for 128-bit on the higher and lower elements of the vector
+// independently.
+#[cfg(all(any(target_arch = "x86", target_arch = "x86_64"), target_feature = "sse2"))]
+macro_rules! x86_256_sse2_impl {
+    ($id:ident, $v128:ident) => {
+        impl All for $id {
+            #[inline]
+            #[target_feature(enable = "sse2")]
+            unsafe fn all(self) -> bool {
+                unsafe {
+                    union U {
+                        halves: ($v128, $v128),
+                        vec: $id
+                    }
+                    let halves = U {vec: self}.halves;
+                    halves.0.all() && halves.1.all()
+                }
+            }
+        }
+        impl Any for $id {
+            #[inline]
+            #[target_feature(enable = "sse2")]
+            unsafe fn any(self) -> bool {
+                unsafe {
+                    union U {
+                        halves: ($v128, $v128),
+                        vec: $id
+                    }
+                    let halves = U {vec: self}.halves;
+                    halves.0.any() || halves.1.any()
+                }
+            }
+        }
+    }
+}
+
+// Implementation for 64-bit wide masks on x86.
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+macro_rules! x86_64_mmx_movemask_impl {
+    ($id:ident, $vec128:ident) => {
+        impl All for $id {
+            #[inline]
+            #[target_feature(enable = "mmx")]
+            unsafe fn all(self) -> bool {
+                #[cfg(target_arch = "x86")]
+                use ::coresimd::arch::x86::_mm_movemask_pi8;
+                #[cfg(target_arch = "x86_64")]
+                use ::coresimd::arch::x86_64::_mm_movemask_pi8;
+                // _mm_movemask_pi8(a) creates an 8bit mask containing the most
+                // significant bit of each byte of `a`. If all bits are set,
+                // then all 8 lanes of the mask are true.
+                 _mm_movemask_pi8(::mem::transmute(self)) == u8::max_value() as i32
+            }
+        }
+        impl Any for $id {
+            #[inline]
+            #[target_feature(enable = "mmx")]
+            unsafe fn any(self) -> bool {
+                #[cfg(target_arch = "x86")]
+                use ::coresimd::arch::x86::_mm_movemask_pi8;
+                #[cfg(target_arch = "x86_64")]
+                use ::coresimd::arch::x86_64::_mm_movemask_pi8;
+
+                _mm_movemask_pi8(::mem::transmute(self)) != 0
+            }
+        }
+    }
+}
+
+// Implementation for 128-bit wide masks on x86
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+macro_rules! x86_128_impl {
+    ($id:ident) => {
+        cfg_if! {
+            if #[cfg(target_feature = "sse2")] {
+                x86_128_sse2_movemask_impl!($id);
+            }  else {
+                default_impl!($id);
+            }
+        }
+    }
+}
+
+// Implementation for 256-bit wide masks on x86
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+macro_rules! x86_256_impl {
+    ($id:ident, $half_id:ident) => {
+        cfg_if! {
+            if #[cfg(target_feature = "avx")] {
+                x86_256_avx_test_impl!($id);
+            } else if #[cfg(target_feature = "sse2")] {
+                x86_256_sse2_impl!($id, $half_id);
+            } else {
+                default_impl!($id);
+            }
+        }
+    }
+}
+
+// Implementation for ARM + v7 + NEON using vpmin and vpmax (folding
+// minimum/maximum of adjacent pairs) for 64-bit wide two-element vectors.
+#[cfg(all(target_arch = "arm", target_feature = "v7", target_feature = "neon"))]
+macro_rules! arm_64_x2_v7_neon_impl {
+    ($id:ident, $vpmin:ident, $vpmax:ident) => {
+        impl All for $id {
+            #[inline]
+            #[target_feature(enable = "v7,neon")]
+            unsafe fn all(self) -> bool {
+                use ::coresimd::arch::arm::$vpmin;
+                use ::mem::transmute;
+                // pmin((a, b), (-,-)) => (b, -).0 => b
+                let tmp: $id = transmute($vpmin(transmute(self), ::mem::uninitialized()));
+                tmp.extract(0)
+            }
+        }
+        impl Any for $id {
+            #[inline]
+            #[target_feature(enable = "v7,neon")]
+            unsafe fn any(self) -> bool {
+                use ::coresimd::arch::arm::$vpmax;
+                use ::mem::transmute;
+                // pmax((a, b), (-,-)) => (b, -).0 => b
+                let tmp: $id = transmute($vpmax(transmute(self), ::mem::uninitialized()));
+                tmp.extract(0)
+            }
+        }
+    }
+}
+
+// Implementation for ARM + v7 + NEON using vpmin and vpmax (folding
+// minimum/maximum of adjacent pairs) for 64-bit wide four-element vectors.
+#[cfg(all(target_arch = "arm", target_feature = "v7", target_feature = "neon"))]
+macro_rules! arm_64_x4_v7_neon_impl {
+    ($id:ident, $vpmin:ident, $vpmax:ident) => {
+        impl All for $id {
+            #[inline]
+            #[target_feature(enable = "v7,neon")]
+            unsafe fn all(self) -> bool {
+                use ::coresimd::arch::arm::$vpmin;
+                use ::mem::transmute;
+                // tmp = pmin((a, b, c, d), (-,-,-,-)) => (a, c, -, -)
+                let tmp = $vpmin(transmute(self), ::mem::uninitialized());
+                // tmp = pmin((a, b, -, -), (-,-,-,-)) => (c, -, -, -).0 => c
+                let tmp: $id = transmute($vpmin(tmp, ::mem::uninitialized()));
+                tmp.extract(0)
+            }
+        }
+        impl Any for $id {
+            #[inline]
+            #[target_feature(enable = "v7,neon")]
+            unsafe fn any(self) -> bool {
+                use ::coresimd::arch::arm::$vpmax;
+                use ::mem::transmute;
+                // tmp = pmax((a, b, c, d), (-,-,-,-)) => (a, c, -, -)
+                let tmp =  $vpmax(transmute(self), ::mem::uninitialized());
+                // tmp = pmax((a, b, -, -), (-,-,-,-)) => (c, -, -, -).0 => c
+                let tmp: $id = transmute($vpmax(tmp, ::mem::uninitialized()));
+                tmp.extract(0)
+            }
+        }
+    }
+}
+
+// Implementation for ARM + v7 + NEON using vpmin and vpmax (folding
+// minimum/maximum of adjacent pairs) for 64-bit wide eight-element vectors.
+#[cfg(all(target_arch = "arm", target_feature = "v7", target_feature = "neon"))]
+macro_rules! arm_64_x8_v7_neon_impl {
+    ($id:ident, $vpmin:ident, $vpmax:ident) => {
+        impl All for $id {
+            #[inline]
+            #[target_feature(enable = "v7,neon")]
+            unsafe fn all(self) -> bool {
+                use ::coresimd::arch::arm::$vpmin;
+                use ::mem::transmute;
+                // tmp = pmin(
+                //     (a, b, c, d, e, f, g, h),
+                //     (-, -, -, -, -, -, -, -)
+                // ) => (a, c, e, g, -, -, -, -)
+                let tmp = $vpmin(transmute(self), ::mem::uninitialized());
+                // tmp = pmin(
+                //     (a, c, e, g, -, -, -, -),
+                //     (-, -, -, -, -, -, -, -)
+                // ) => (c, g, -, -, -, -, -, -)
+                let tmp = $vpmin(tmp, ::mem::uninitialized());
+                // tmp = pmin(
+                //     (c, g, -, -, -, -, -, -),
+                //     (-, -, -, -, -, -, -, -)
+                // ) => (g, -, -, -, -, -, -, -).0 => g
+                let tmp: $id = transmute($vpmin(tmp, ::mem::uninitialized()));
+                tmp.extract(0)
+            }
+        }
+        impl Any for $id {
+            #[inline]
+            #[target_feature(enable = "v7,neon")]
+            unsafe fn any(self) -> bool {
+                use ::coresimd::arch::arm::$vpmax;
+                use ::mem::transmute;
+                // tmp = pmax(
+                //     (a, b, c, d, e, f, g, h),
+                //     (-, -, -, -, -, -, -, -)
+                // ) => (a, c, e, g, -, -, -, -)
+                let tmp = $vpmax(transmute(self), ::mem::uninitialized());
+                // tmp = pmax(
+                //     (a, c, e, g, -, -, -, -),
+                //     (-, -, -, -, -, -, -, -)
+                // ) => (c, g, -, -, -, -, -, -)
+                let tmp = $vpmax(tmp, ::mem::uninitialized());
+                // tmp = pmax(
+                //     (c, g, -, -, -, -, -, -),
+                //     (-, -, -, -, -, -, -, -)
+                // ) => (g, -, -, -, -, -, -, -).0 => g
+                let tmp: $id = transmute($vpmax(tmp, ::mem::uninitialized()));
+                tmp.extract(0)
+            }
+        }
+    }
+}
+
+
+// Implementation for ARM + v7 + NEON using vpmin and vpmax (folding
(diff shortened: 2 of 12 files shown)


gnzlbg:
cc @hsivonen 

--

@alexcrichton there are still some failures on x86, but I have no idea why they happen yet, the asm code generated for the x86 and the x86_64 functions looks identical to me.

alexcrichton:
Thanks for this! I'll admit though I'm **very** quickly getting lost in all this...


So purely on the surface I'm curious what's going on here? For example if we conditionalize these impls on `#[cfg(target_feature = ...)]` that's not too useful, right? Don't we want everything to be compiled depending on downstream users compilation environment?

I realize that we don't really have any better solution than this today, but this to me feels like we should be pursuing a different language-level solution. For example I feel like we should be able to monomorphize these implementations into downstream crates, reconfiguring the settings dynamically as they're monomorphized. That way if you compile with `-C target-feature=+neon` you get all the optimized versions (and still no runtime overhead of checking) from stdsimd. 

Does that makes sense? Is that perhaps too far out though?

gnzlbg:
> For example if we conditionalize these impls on #[cfg(target_feature = ...)] that's not too useful, right? Don't we want everything to be compiled depending on downstream users compilation environment?

Well `sse2` is available on `x86_64` by default, and `neon` is available on `aarch64` by default, so those would work with `std` out of the box. 

For `AVX` and `neon` on `armv7` users would need to recompile std with those features enabled.

> I realize that we don't really have any better solution than this today, but this to me feels like we should be pursuing a different language-level solution. For example I feel like we should be able to monomorphize these implementations into downstream crates, reconfiguring the settings dynamically as they're monomorphized. That way if you compile with -C target-feature=+neon you get all the optimized versions (and still no runtime overhead of checking) from stdsimd.

Sure, this makes sense. I was talking with @rkruppe about this the other day. The main problems are that:

* we can't do this with an attribute-based approach
* because any crate in the ecosystem can write `#[target_feature(enable = "avx")] fn foo() { ...}` and call anything from there, all the code of the whole program has to be available to the current translation unit, so that all of it can be monomorphized for the current set of target features enabled depending on where it is called. 

@centril is preparing an RFC for an effect-system for Rust. If we think of target-feature as an effect, for this to work all Rust code would need to always be polymorphic on the target-feature effect. This would fix things like `cfg(target_feature)` returning `false` for a feature that is currently in the function where it is used via `#[target_feature]`, and this would work even across functions and across crates. 

> Does that makes sense?

This is actually what users want (or at least what I want) so to me, yes, it makes a lot of sense.

>  Is that perhaps too far out though?

That depends. Three main goals for 2018 are effects: `async`, `const fn`, and `target-feature`. 

Rust code wants to be able to ask "am I being executed at compile-time" about the `const` effect, "am I being called synchronously or asynchronously" for async code (e.g. if the code is called with or without `await`), and none of these is that much different from asking "am I being compiled with AVX", and branch depending on the answer.

Instead of patching the hell out of the Rust boat with attributes for these three use cases **maybe** the PL PhDs working on Rust could try to combine their skills to propose an effect system that handles all of these in a principled way.

Honestly, I wish we could ship `std::arch` without shipping `#[target_feature]`, `cfg(target_feature)`, `is_..._feature_detected`, and company.

alexcrichton:
Hm ok this makes sense. I agree with @rkruppe that gaining an effect system is unlikely to happen any time soon so we probably don't want to plan on that happening.

I suppose another way to put my reaction is that I'm confused why we have to do this in the first place. Sort of the whole point of "portable SIMD" is to "pray the backend takes care of everything". If we end up defining everything ourselves and having massive swaths of `#[cfg]` then why not just do this all on crates.io?

It sounds like these are also considered LLVM bugs but presumably we're going to continuously run into these issues, right?

rkruppe:
Well there is a fundamental issue with putting such code into libraries that are called from multiple contexts: you either duplicate the function for each context to exploit knowledge, or you share the code by serving the lowest common denominator.

There are of course intermediates between these extremes: for some kinds of operations, regular inlining can help when it kicks in, and only generating LLVM IR when compiling leaf crates (https://github.com/rust-lang/rust/issues/38913) would effectively propagate `-C` options (but not `#[target_feature]`s) to all crates in the dependency graph. But if you want to *always* use the caller's target features, you have to duplicate the code for every set of target features, which is not obviously desirable even if restricted to functions that use the portable vector types (e.g., most algorithms on 128-bit vectors won't benefit from having AVX available).

gnzlbg:
> I suppose another way to put my reaction is that I'm confused why we have to do this in the first place. Sort of the whole point of "portable SIMD" is to "pray the backend takes care of everything". 

I am open to better ways to proceed. The current approach has been to: 

* fill these bugs (LLVM bugs: 37087, 36982, 36933, 36796, 36734, 36702, amongst others).
* get them fixed: two are fixed already, two have patches under review. 
* add workarounds for the critical ones, like these two, which prevent Firefox from moving off the simd crate

What do you propose that we do instead? We could not add workarounds and tell people to live with the bugs for a while or work around them in their own crates. That would be fine for me. Since we are not following LLVM trunk, "for a while" might be 6 months or 1.5 years, depending on how unlucky we get. 

> If we end up defining everything ourselves and having massive swaths of #[cfg] then why not just do this all on crates.io?

The workarounds here account for less than 1% of the vector methods. Over 99% of the methods currently use the LLVM intrinsics, so we would need to stabilize those to be able to do this on crates.io. 

The alternative would be to not add any methods with bugs to the vector types. Sadly, when the issue to add these methods (that were already part of the simd crate) was opened, nobody mentioned that their LLVM implementations had bugs. Nobody even mentioned that their implementation in the simd crate also had bugs.

Centril:
Small note on effects: I'm not preparing an RFC on a general effect system, but rather for `const` specifically, but I want to gain understanding of effects in general as well.

alexcrichton:
@gnzlbg hm ok, good to have numbers!

In my mind the "absolute ideal" solution here is to have a special node in MIR corresponding to "expand to `true` or `false` depending on target features when monomorphized". That would be an unstable language feature, obviously, but means you could write everything here as:

```rust
if __unstable_target_feature_activated("foo") {
    // ...
} else {
    // ...
}
```

and that wouldn't have the same runtime penalty of `is_x86_feature_detected!`. 

This is definitely very similar to MIR-only rlibs and what happens with overflow checking and debug assertions for primitives today. I think such a feature would at least help reduce the boilerplate here and also lift the need for recompiling std as these are all inlined and monomorphized on the fly anyway.

In any case I do think that we could probably land this in the meantime. I'd be hesitant to land anything that's not already exposed thorugh the standard targets though. For example the AVX-specific code here I don't think is ever tested or would be compiled?

In terms of our own CI as well, perhaps we should run `cargo test` for a few various levels of `-C target-feature=..` for the portable types?

alexcrichton:
Oh I'm not too worried about the codegen for the function itself but moreso am worried about the correctness. In that sense I think it's fine to hand-inspect assembly and otherwise just test for correctness

gnzlbg:
> am worried about the correctness.

I've updated the ci/run script to compile with avx as well on all x86 targets and run the release tests. 

These two intrinsics are tested by the portable vector types tests, and `all` is tested as well in the `std::arch` tests (because `PartialEq` is implemented as `a.eq(b).all()`). So if these pass that should cover the correctness.

> Oh I'm not too worried about the codegen for the function itself

I am not too worried either, but the whole point of these workarounds is to fix codegen bugs, so I'd rather check that they work as intended and that we don't inadvertently break this in the future.

coresimd/ppsv/codegen/masks_reductions.rs
```diff
@@ -0,0 +1,603 @@
...
+            #[inline]
+            #[target_feature(enable = "v7,neon")]
+            unsafe fn all(self) -> bool {
+                use ::coresimd::arch::arm::$vpmin;
+                use ::mem::transmute;
+                union U {
+                    halves: ($half, $half),
+                    vec: $id
+                }
+                let halves = U { vec: self }.halves;
```
hsivonen: Is this a valid way to split a vector into its halves? It seems that clang uses shuffles for this purpose.

With this implementation, I see a bunch of functions whose core follows this pattern fail unit tests:

```rust
        #[inline(always)]
        pub fn simd_is_ascii(s: u8x16) -> bool {
            let above_ascii = u8x16::splat(0x80);
            s.lt(above_ascii).all()
        }
```

hsivonen:
> I suppose another way to put my reaction is that I'm confused why we have to do this in the first place. Sort of the whole point of "portable SIMD" is to "pray the backend takes care of everything".

That position may make sense from the compiler front-end developer point of view, but from the point of view of a user of the language, the user of the language sees the standard library interface and doesn't care which layer (standard library, compiler front end or LLVM) does the Right Thing.

This PR makes the black box that the user of the language sees do the Right Thing (ignoring the concern I have about whether the ARMv7+NEON code path splits the vector into its halves using the right LLVM idiom; I didn't debug it fully yet).

It would be really sad not to get the user-of-the-language-facing Right Thing at this time even if ideally it could be moved to another layer later.

> If we end up defining everything ourselves and having massive swaths of #[cfg] then why not just do this all on crates.io?

These operations logically belong on the types provided by `std::simd`, so moving these operations to crates.io would be really bad API design.

The PR addresses the performance concern I've had regarding moving encoding_rs/Firefox from the `simd` crate to `stdsimd` on x86_64 and aarch64. (x86 benchmarks are still running. I'm running them on actual non-x86_64 hardware, so it's like watching paint dry.)

hsivonen:
> With this implementation, I see a bunch of functions whose core follows this pattern fail unit tests:
>
> ```rust
>        #[inline(always)]
>        pub fn simd_is_ascii(s: u8x16) -> bool {
>            let above_ascii = u8x16::splat(0x80);
>            s.lt(above_ascii).all()
>        }
> ```

I used

```sh
RUSTFLAGS='-C target_feature=+neon' cargo test --features simd-accel
```

With revision [d7013a3393611b678b9f4abc99510bc875de7b68](https://github.com/hsivonen/encoding_rs/commit/d7013a3393611b678b9f4abc99510bc875de7b68) of the [stdsimd branch](https://github.com/hsivonen/encoding_rs/tree/stdsimd) of encoding_rs.

> (x86 benchmarks are still running. I'm running them on actual non-x86_64 hardware, so it's like watching paint dry.)

And the x86 results are in.

This PR addresses my performance concerns on all the architectures that encoding_rs uses the `simd` crate on at present (x86, x86_64 and aarch64). 🎉 Thank you, @gnzlbg!

coresimd/ppsv/codegen/masks_reductions.rs
gnzlbg: > Is this a valid way to split a vector into its halves?

It is a valid way to do this. The RFC specifies that the layout of vectors is the same as arrays of the same element type and the same number of lanes, e.g., `f32x2` has the same layout as `[f32; 2]`. Since the layout of arrays is the same as the layout of tuples of the same element type and same number of element, e.g., `(f32, f32)` in the previous example, then yes, this is valid. I chose a tuple over an array here because I did not want to worry about doing `.get_unchecked`.

> With this implementation, I see a bunch of functions whose core follows this pattern fail unit tests:

On which arch? Can you post a minimal example that reproduces it here?

hsivonen: > On which arch? Can you post a minimal example that reproduces it here?

ARMv7 with `-C target_feature=+neon`. I'll try to minimize it.

gnzlbg: Does this work on the other targets like `x86` and `x86_64` ?

hsivonen: Yes. I tested that the pattern works on x86_64 by temporarily replacing the x86_64-specific code with the portable code.

ci/run.sh
```diff
@@ -36,3 +36,12 @@ cargo_test() {
 
 cargo_test
 cargo_test "--release"
+
+case ${TARGET} in
+    x86*)
+        RUSTFLAGS="${RUSTFLAGS} -C target-feature=+avx"
+        cargo_test "--release"
```
alexcrichton: This'll somehow need to disable all the `assert_instr` tests as well as anything beneath avx is likely to generate different instructions with avx enabled

coresimd/ppsv/codegen/masks_reductions.rs
hsivonen: Sorry about misattributing the bug. The bug was a misplaced parenthesis in my conditional compilation directives, which caused ARMv7+NEON to use a stride size constant intended for ARMv7 ALU code.

So: Not a blocker for getting this PR merged.

gnzlbg:
@hsivonen this PR has pretty much stalled since I was low on time, feel free to pick it up. Otherwise I might have some time to finish it next week.

gnzlbg:
@alexcrichton yes that's not working yet, but the issue i mentioned is unrelated. Raising the `-C target-feature=+avx` to AVX from SSE2 results in a lot of intrinsics that were previously inlined not being inlined anymore :/

I would expect the intrinsics to get inlined, and those asser instr to fail because different code is generated (e.g. because LLVM can do something better), but that's not why they are failing. They are failing because they contain a `call` instruction to the `#[target_feature]` intrinsics :/

alexcrichton:
Right yeah but I think your diagnosis of https://github.com/rust-lang-nursery/stdsimd/issues/427 is correct, and this should be fixed by https://github.com/rust-lang/rust/pull/50188

alexcrichton:
Looks like AppVeyor CI is failing for running too many tests?

alexcrichton:
64-bit osx has these failures:


```
failures:
---- ppsv::v64::f32x2_tests::arithmetic stdout ----
	thread 'ppsv::v64::f32x2_tests::arithmetic' panicked at 'assertion failed: `(left == right)`
  left: `f32x2(NaN, 0.0)`,
 right: `f32x2(0.0, 0.0)`', crates/coresimd/tests/../../../coresimd/ppsv/v64.rs:66:1
---- ppsv::v64::f32x2_tests::arithmetic_scalar stdout ----
	thread 'ppsv::v64::f32x2_tests::arithmetic_scalar' panicked at 'assertion failed: `(left == right)`
  left: `f32x2(NaN, 0.0)`,
 right: `f32x2(0.0, 0.0)`', crates/coresimd/tests/../../../coresimd/ppsv/v64.rs:66:1
failures:
    ppsv::v64::f32x2_tests::arithmetic
    ppsv::v64::f32x2_tests::arithmetic_scalar
```

The DOCUMENTATION build is gonna need a few updates  as well as [it's failing](https://travis-ci.org/rust-lang-nursery/stdsimd/jobs/374873347)

The x86_64-linux-emulated build I think like AppVeyor is running too many tests (and is failing)

The [wasm build](https://travis-ci.org/rust-lang-nursery/stdsimd/jobs/374873346) is failing but not due to issues here. It's ok to move that to allow failures and I'll fix it later

crates/coresimd/Cargo.toml
```diff
@@ -18,6 +18,9 @@ is-it-maintained-issue-resolution = { repository = "rust-lang-nursery/stdsimd" }
 is-it-maintained-open-issues = { repository = "rust-lang-nursery/stdsimd" }
 maintenance = { status = "experimental" }
 
+[dependencies]
+cfg-if = "0.1"
```
alexcrichton: For ease of integration into rust-lang/rust, could this macro be inlined? (It's actually quite a small macro)

gnzlbg: Sure, done.

gnzlbg:
All green :/ I'll open an issue to figure out what's happening in MacOSX on travis.

crates/coresimd/src/lib.rs
```diff
@@ -46,6 +46,50 @@ extern crate stdsimd_test;
 #[cfg(test)]
 extern crate test;
 
+#[doc(hidden)]
+macro_rules! cfg_if {
```
alexcrichton: Since this won't be available in libcore can this be moved diretly to the coresimd/mod.rs file to be used there?

gnzlbg: Sure :)

gnzlbg: Done :)

Results51:
#13579

Results51:
#msshb

Results51:
#MSS

PR #428, opened 2018-04-14
Shuffle around `stdsimd::arch::detect` bits

alexcrichton:
Compile more code on more platforms, tweak imports, try to catch mistakes
sooner.


alexcrichton:
This is fixing [this error](https://github.com/rust-lang/rust/pull/49664#issuecomment-381339968) which wasn't caught during x86/x86_64 builds. I'm hoping to build a maximal amount of code on all platforms to check as much as we can statically and then most of it just ends up being dead code that's eliminated.

gnzlbg:
It would be nice to triple-check that the cache is indeed eliminated on platforms that don't need it.

gnzlbg:
Also in which platforms is this code failing, and why isn't it failing in `stdsimd`'s CI ?

alexcrichton:
Oh yeah this is for sure all jettisoned from libstd, and it's an issue with integrating into libstd rather than stdsimd itself. In libstd there's no top-level `_std` item, but rather more tweaks need to happen for reexported items in the stdsimd crate's ci to integrate well into libstd

gnzlbg:
> but rather more tweaks need to happen for reexported items in the stdsimd crate's ci to integrate well into libstd

Ah crap, maybe I screwed it up last time I touched it. Could you add a comment to the top-level crates explaining a bit how things must be done to not break anything in `libstd` ?

alexcrichton:
Oh no worries, I've renamed the import of `std` to `__do_not_use_this_import` to discourage further use :)

PR #441, opened 2018-05-09
formatting

gnzlbg:


alexcrichton:
Looks like CI is failing?

gnzlbg:
Yeah, rustfmt broke stdsimd, again. 

See: https://github.com/rust-lang-nursery/rustfmt/pull/2693

Basically rustfmt tests that formatting something is indempotent, that is, that if you format something twice you get the same results that if you format something once. But it doesn't really test that the code after the formatting actually compiles. 

I'm clossing this until the issue is resolved upstream.

Issue #118, opened 2017-10-14
Allowing forks to build other branches


gnzlbg:
Currently the `.travis.yml` looks like this:

```yml

branches:
  only:
    # Pushes and PR to the master branch.
    - master
```

This prevents other branches from building in CI. The way I get my other branches to build is by:

- 1. changing the travis.yml file for a branch
- 2. working on the branch in my repo
- 3. change the travis.yml file back
- 4. pull-request

Is there a way to improve this ? I basically rely on travis and the docker images for running tests on all platforms.

alexcrichton:
I think it's fine to just delete that whitelist from teh config?

gnzlbg:
I though maybe it was there for "reasons" (the comment says something about pushing to master branch). I am fine with killing it.

gnzlbg:
Closed by https://github.com/rust-lang-nursery/stdsimd/commit/077f1f7d9b8ad88e56ffe9c8d6f887a5b51b7612 .

PR #444, opened 2018-05-14
Remove notes about stability

steveklabnik:
Hi there! So, I was glancing at https://doc.rust-lang.org/beta/std/arch/ today and noticed the "This module is currently unstable" bit. Now that some of it is stable, this is misleading, so we should remove it. Apparently that is here? If that's incorrect, please let me know! Also, backporting this change to beta would be great, but given the details here, I don't know if that's easy or hard.

### DIFF
--- a/stdsimd/mod.rs
+++ b/stdsimd/mod.rs
@@ -14,9 +14,6 @@
 /// `i686-pc-windows-msvc` target will have an `x86` module here, whereas
 /// `x86_64-pc-windows-msvc` has `x86_64`.
 ///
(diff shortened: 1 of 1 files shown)


alexcrichton:
👍

alexcrichton:
I think backporting would be fine yeah, I don't think much has changed! Should just need a few PRs to the master/beta branches

steveklabnik:
Great, I'll send those in.

Results51:
#118

PR #445, opened 2018-05-15
Make the Feature enum unstable

est31:
This is cleaner than just having a #[doc(hidden)] attribute


gnzlbg:
The `stdsimd` feature flag is intended to be used by users on nightly. Since we haven't planned to ever expose these, maybe they should use a different feature flag than `stdsimd`.

est31:
@gnzlbg changed it to use the `simd_internal` feature.

PR #446, opened 2018-05-15
Unstabilize some more #[doc(hidden)] items

est31:
follow up of #445


est31:
Also removed the `#[doc(hidden)]` attribute here because it doesn't make much sense anyway on `pub(crate)` items.

Now ripgrepping through the source gives me no stable `#[doc(hidden)]` items any more.

PR #451, opened 2018-05-16
Implement vec_perm

lu-zero:
It is pretty much the opposite of `vec_add`: there is a single instruction `vperm` that works with all the possible vector types as long the operands and return type match.

The instruction has an endianness bias and it is implemented only for little endian for now.


ci/run.sh
```diff
@@ -26,6 +26,7 @@ case ${TARGET} in
         export STDSIMD_DISABLE_ASSERT_INSTR=1
         ;;
     powerpc64-*)
+        export RUSTFLAGS="${RUSTFLAGS} -C target-feature=+altivec"
```
gnzlbg: This should not be required

lu-zero: That is some stray from testing, let me remove it now.

lu-zero:
Now using the arch-specific newtypes.

coresimd/powerpc/altivec.rs
```diff
@@ -355,10 +358,53 @@ impl_from_bits_!(
...
+}
+
 mod sealed {
 
     use super::*;
 
+    #[inline]
+    #[target_feature(enable = "altivec")]
+    #[cfg_attr(test, assert_instr(vperm))]
+    unsafe fn vec_vperm(a: vector_signed_int, b: vector_signed_int, c: vector_unsigned_char) -> vector_signed_int {
```
gnzlbg: I think we should add `assert_instr` tests for all vector types.

coresimd/powerpc/altivec.rs
```diff
@@ -655,6 +701,28 @@ where
...
 
+/// Vector permute.
+#[inline]
+#[target_feature(enable = "altivec")]
+pub unsafe fn vec_perm<T>(a: T, b: T, c: vector_unsigned_char) -> T
+where
+    T: sealed::VectorPerm,
+{
+
+    if cfg!(target_endian = "little") {
```
gnzlbg: prefer conditional compilation here: `#[cfg(target_endian = "little")]` and `#[cfg(not(target_endian = "little"))]` (or `#[cfg(target_endian = "big")]`).

coresimd/powerpc/altivec.rs
```diff
@@ -666,6 +734,18 @@ mod tests {
     use simd::*;
     use stdsimd_test::simd_test;
 
+    #[simd_test(enable = "altivec")]
+    unsafe fn vec_perm_u16x8() {
```
gnzlbg: there should be one run-time test for every vector type as well

I think that for consistency with how it is done in the other architectures we should name the run-time tests: `test_{intrinsic_name}_{short_vector_types}`, so this becomes `test_vec_perm_u16x8`.

coresimd/powerpc64/vsx.rs
```diff
@@ -216,3 +222,75 @@ impl_from_bits_!(
...
+    use super::*;
+
+    pub trait VectorPermDI {
+        unsafe fn vec_xxpermdi(self, b: Self, dm: u8) -> Self;
+    }
+
+    #[inline]
+    #[target_feature(enable = "vsx")]
+    #[cfg_attr(test, assert_instr(xxpermdi, dm = 0x0))]
+    unsafe fn xxpermdi(a: i64x2, b: i64x2, dm: u8) -> i64x2 {
```
gnzlbg: I think this is missing a `#[rustc_args_required_const(2)]` to require that the user passes a compile-time constant to `dm`.

gnzlbg: That won't work with the trait approach :/ 
crap

gnzlbg: Maybe make this function and the trait method impls `#[inline(always)]` ?

lu-zero:
Now we also have xxpermdi so we can check how vsx instructions fare.

coresimd/powerpc64/vsx.rs
```diff
@@ -216,3 +222,75 @@ impl_from_bits_!(
...
+    use coresimd::arch::powerpc::*;
+
+    #[cfg(target_arch = "powerpc64")]
+    use coresimd::arch::powerpc64::*;
+
+    use simd::*;
+    use stdsimd_test::simd_test;
+
+    #[simd_test(enable = "vsx")]
+    unsafe fn vec_xxpermdi_u62x2() {
```
gnzlbg: Also missing tests for the other vector types.

coresimd/powerpc/altivec.rs
```diff
@@ -330,13 +369,42 @@ where
     a.vec_add(b)
 }
 
+/// Vector permute.
+#[inline]
+#[target_feature(enable = "altivec")]
+pub unsafe fn vec_perm<T>(a: T, b: T, c: vector_unsigned_char) -> T
```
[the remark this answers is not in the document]
lu-zero: vec_vperm is the function used for the test.

coresimd/powerpc/altivec.rs
lu-zero: That function is the only one that is called for every vector type.

All the vector types are transmuted to i32x4.

coresimd/powerpc/altivec.rs
lu-zero: I can macro those tests, but it would basically test that transmute works for those types.

coresimd/powerpc/altivec.rs
lu-zero: Something along the lines of
``` rust
#[cfg(target_endian = "little")]
mod endian {

}
#[cfg(target_endian = "big")]
mod endian {

}
use self::endian::*;
```

lu-zero:
I refactored the endian-biased intrinsic support so it is a bit more structured.

coresimd/powerpc/altivec.rs
gnzlbg: That would be great, testing these is tedious, but sometimes optimizations happen that break this, and we also scan for assert_instr in `stdsimd-verify`. It would be great if the names of these functions follows a convention so that the code that scans them automatically becomes easier to write. (I followed one convention for vec_add, but maybe there is a better one).

gnzlbg:
So this looks good. I'll retry the tests once CI is working again.

lu-zero:
Now the set is rebased over https://github.com/rust-lang-nursery/stdsimd/pull/455

alexcrichton:
That's a... curious test failure!

```
---- coresimd::powerpc64::vsx::tests::vec_xxpermdi_u62x2 stdout ----
thread 'coresimd::powerpc64::vsx::tests::vec_xxpermdi_u62x2' panicked at 'assertion failed: `(left == right)`
  left: `u64x2(0, 2)`,
 right: `u64x2(0, 2)`', crates/coresimd/src/../../../coresimd/powerpc64/vsx.rs:293:9
```

lu-zero:
Yes... how do we implement equality for that type on BE?

gnzlbg:
It's here: https://github.com/rust-lang-nursery/stdsimd/blob/master/coresimd/ppsv/api/partial_eq.rs#L9

We do a vertical equality comparison and then test if all bits are set.

lu-zero:
I managed to find other unrelated problems while trying to reproduce the issue locally: https://github.com/rust-lang/rust/issues/50960

Shall we demote powerpc and powerpc64 for the time being?

gnzlbg:
> Shall we demote powerpc and powerpc64 for the time being?

That was the original plan any ways. We are currently only running the `assert_instr` tests on `ppc64le` so it makes sense to only run the run-time tests there as well. 

Since `ppc64le` is the only practical PowerPC target, I think we should just focus on that one first.

lu-zero:
Now it is all green (by cheating) :)

gnzlbg:
So I would prefer to just disable the run-time tests in CI e.g. by passing an environment variable that just ignores the `#[simd_test]`s  on ppc and ppc64 instead of completely allowing them to fail. Note that all of the portable packed vector tests are currently passing on both ppc and ppc64 and I'd like to know when they break.

lu-zero:
Sure, could you please point me how to do that and suggest a name for the variable?

gnzlbg:
We already have `STDSIMD_TEST_EVERYTHING=1` so maybe `STDSIMD_TEST_NORUN=1` ?

The place is the `stdsimd-test-macro` (https://github.com/rust-lang-nursery/stdsimd/blob/master/crates/simd-test-macro/src/lib.rs#L108) where you want to add an `#[ignore]` if a particular environment variable is defined, similar to how it is done in the `assert-instr-macro` here: https://github.com/rust-lang-nursery/stdsimd/blob/master/crates/assert-instr-macro/src/lib.rs#L42

lu-zero:
The run script doesn't forward the environment set on travis. I put the env variable directly in the script.

gnzlbg:
Thanks a lot for this! This is really good work!

PR #455, opened 2018-05-21
Update proc macro2

lu-zero:
Update to the latest version of proc_macro2, quote and syn.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,6 +55,7 @@ matrix:
         cargo clippy --all -- -D clippy-pedantic
   allow_failures:
     - env: CLIPPY=On TARGET=x86_64-unknown-linux-gnu NO_ADD=1
+    - env: RUSTFMT=On TARGET=x86_64-unknown-linux-gnu NO_ADD=1
 
 before_install:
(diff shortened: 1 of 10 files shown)


lu-zero:
looks like the new rustfmt has different ideas regarding the style and there is something quite fishy going on windows and x86 but it is unrelated to this set.

alexcrichton:
Unfortunately we're hit by a slew of bugs here:

* First the x86 tests seem to be failing I believe because mergefunc was recently landed. These are presumably duplicates with other functions being generated so they're merged and the disassembly is no longer showing us precisely which instruction is being used.
* Next proc-macro2 is [buggy on nightly](https://github.com/rust-lang/rust/issues/50942), causing the verify-x86 builder to build
* I think rustfmt with nightly needs to be run again?

alexcrichton:
I also don't know what's going on yet with Windows, that's just Linux :(

alexcrichton:
Ok Windows looks back to "normal", I think this PR is now purely blocked on figuring out mergefunc

alexcrichton:
Ok I've pushed a commit that will hopefully thwart mergefunc, let's see how it pans out.

alexcrichton:
Alright, all green now!

PR #458, opened 2018-05-22
Fix outdated docs

Alcaro:
The is_feature_detected! macro was renamed (#346), so "The name of the macro is the same across platforms" is false. Let's just remove that sentence.


gnzlbg:
Thanks!

Issue #170, opened 2017-11-04
Implement all MSA Intrinsics (MIPS SIMD Architecture)


gnzlbg:
The [reference](http://cdn2.imgtec.com/documentation/MD00866-2B-MSA32-AFP-01.12.pdf) and [gcc](https://gcc.gnu.org/onlinedocs/gcc/MIPS-SIMD-Architecture-Built-in-Functions.html#MIPS-SIMD-Architecture-Built-in-Functions) (generated with https://gist.github.com/gnzlbg/3a27cca666f4a9f8a51d2eb008ce914c from the gcc docs):

- [ ] `fn add_a_b(i8x16, i8x16) -> i8x16`
- [ ] `fn add_a_h(i16x8, i16x8) -> i16x8`
- [ ] `fn add_a_w(i32x4, i32x4) -> i32x4`
- [ ] `fn add_a_d(i64x2, i64x2) -> i64x2`
- [ ] `fn adds_a_b(i8x16, i8x16) -> i8x16`
- [ ] `fn adds_a_h(i16x8, i16x8) -> i16x8`
- [ ] `fn adds_a_w(i32x4, i32x4) -> i32x4`
- [ ] `fn adds_a_d(i64x2, i64x2) -> i64x2`
- [ ] `fn adds_s_b(i8x16, i8x16) -> i8x16`
- [ ] `fn adds_s_h(i16x8, i16x8) -> i16x8`
- [ ] `fn adds_s_w(i32x4, i32x4) -> i32x4`
- [ ] `fn adds_s_d(i64x2, i64x2) -> i64x2`
- [ ] `fn adds_u_b(u8x16, u8x16) -> u8x16`
- [ ] `fn adds_u_h(u16x8, u16x8) -> u16x8`
- [ ] `fn adds_u_w(u32x4, u32x4) -> u32x4`
- [ ] `fn adds_u_d(u64x2, u64x2) -> u64x2`
- [ ] `fn addv_b(i8x16, i8x16) -> i8x16`
- [ ] `fn addv_h(i16x8, i16x8) -> i16x8`
- [ ] `fn addv_w(i32x4, i32x4) -> i32x4`
- [ ] `fn addv_d(i64x2, i64x2) -> i64x2`
- [ ] `fn addvi_b(i8x16, imm0_31) -> i8x16`
- [ ] `fn addvi_h(i16x8, imm0_31) -> i16x8`
[...]

PR #460, opened 2018-05-24
Make SIMD tracking issue marked for `stdsimd` too

SimonSapin:
https://github.com/rust-lang/rust/issues/27731

### DIFF
--- a/coresimd/mod.rs
+++ b/coresimd/mod.rs
@@ -11,7 +11,7 @@ mod ppsv;
 /// This is an **unstable** module for portable SIMD operations. This module
 /// has not yet gone through an RFC and is likely to change, but feedback is
 /// always welcome!
-#[unstable(feature = "stdsimd", issue = "0")]
+#[unstable(feature = "stdsimd", issue = "27731")]
 pub mod simd {
     pub use coresimd::ppsv::*;
 }
@@ -76,7 +76,7 @@ pub mod arch {
     /// See the [module documentation](../index.html) for more details.
     #[cfg(any(target_arch = "arm", dox))]
     #[doc(cfg(target_arch = "arm"))]
-    #[unstable(feature = "stdsimd", issue = "0")]
+    #[unstable(feature = "stdsimd", issue = "27731")]
     pub mod arm {
         pub use coresimd::arm::*;
     }
@@ -86,7 +86,7 @@ pub mod arch {
     /// See the [module documentation](../index.html) for more details.
     #[cfg(any(target_arch = "aarch64", dox))]
     #[doc(cfg(target_arch = "aarch64"))]
-    #[unstable(feature = "stdsimd", issue = "0")]
+    #[unstable(feature = "stdsimd", issue = "27731")]
     pub mod aarch64 {
         pub use coresimd::aarch64::*;
         pub use coresimd::arm::*;
@@ -96,7 +96,7 @@ pub mod arch {
     ///
     /// See the [module documentation](../index.html) for more details.
     #[cfg(target_arch = "wasm32")]
-    #[unstable(feature = "stdsimd", issue = "0")]
+    #[unstable(feature = "stdsimd", issue = "27731")]
     pub mod wasm32 {
         pub use coresimd::wasm32::*;
     }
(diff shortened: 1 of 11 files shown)


coresimd/mod.rs
```diff
@@ -76,7 +76,7 @@ pub mod arch {
     /// See the [module documentation](../index.html) for more details.
     #[cfg(any(target_arch = "arm", dox))]
     #[doc(cfg(target_arch = "arm"))]
-    #[unstable(feature = "stdsimd", issue = "0")]
+    #[unstable(feature = "stdsimd", issue = "27731")]
```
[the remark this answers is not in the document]
gnzlbg: @alexcrichton the alternative here is to add a new feature to rustc, `arm_arch` ?

coresimd/mod.rs
```diff
@@ -86,7 +86,7 @@ pub mod arch {
     /// See the [module documentation](../index.html) for more details.
     #[cfg(any(target_arch = "aarch64", dox))]
     #[doc(cfg(target_arch = "aarch64"))]
-    #[unstable(feature = "stdsimd", issue = "0")]
+    #[unstable(feature = "stdsimd", issue = "27731")]
```
gnzlbg: The feature name that we could re-use is `aarch64_target_feature`, the alternative would be to add a new feature to rustc here (`aarch64_arch`?)

coresimd/mod.rs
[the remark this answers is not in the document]
gnzlbg: The library component is `arch::arm` (part of the std::arch` rfc), so maybe we just need new features for this :/

coresimd/mod.rs
```diff
@@ -106,7 +106,7 @@ pub mod arch {
     /// See the [module documentation](../index.html) for more details.
     #[cfg(any(target_arch = "mips", dox))]
     #[doc(cfg(target_arch = "mips"))]
-    #[unstable(feature = "stdsimd", issue = "0")]
+    #[unstable(feature = "stdsimd", issue = "27731")]
```
gnzlbg: Tracking issue here is #170 .

coresimd/mod.rs
```diff
@@ -116,7 +116,7 @@ pub mod arch {
     /// See the [module documentation](../index.html) for more details.
     #[cfg(any(target_arch = "mips64", dox))]
     #[doc(cfg(target_arch = "mips64"))]
-    #[unstable(feature = "stdsimd", issue = "0")]
+    #[unstable(feature = "stdsimd", issue = "27731")]
```
gnzlbg: Tracking issue here is #170 .

SimonSapin:
This is a drive-by PR that I’d rather keep minimal. Would you accept this as an incremental improvement without the feature name tweaking?

coresimd/mod.rs
```diff
@@ -126,7 +126,7 @@ pub mod arch {
     /// See the [module documentation](../index.html) for more details.
     #[cfg(any(target_arch = "powerpc", dox))]
     #[doc(cfg(target_arch = "powerpc"))]
-    #[unstable(feature = "stdsimd", issue = "0")]
+    #[unstable(feature = "stdsimd", issue = "27731")]
```
gnzlbg: For powerpc we don't have a tracking issue in the stdsimd repo but there is one issue open about altivec in the rust-lang/rust repo that we might want to recycle here: https://github.com/rust-lang/rust/issues/42743 cc @lu-zero I don't know if you wanted to open an issue in the stdsimd repo tracking powerpc status like we did for x86, arm, and msa, but if you do don't forget to cross-link that rust-lang/rust issue.

gnzlbg:
I've opened https://github.com/rust-lang-nursery/stdsimd/issues/461 to track the feature names.

coresimd/mod.rs
lu-zero: I can open an issue about this as well. Do you have a specific format for the tracker ?

gnzlbg: No specific format, what I've done in the past is hack a script that just translate the vendor C API to rust types, and then formats it as a list for the tracker. Each vendor is different, so there is not much to reuse here :/

PR #471, opened 2018-06-11
Mark appropriate const fns as promotable

oli-obk:
This is in preparation for doing the same change to rustc. The feature and attribute don't exist yet, so there'll be some warnings, but they'll go away once the actual change hits nightly

### DIFF
--- a/crates/stdsimd/src/lib.rs
+++ b/crates/stdsimd/src/lib.rs
@@ -9,6 +9,7 @@
 
 #![feature(const_fn, integer_atomics, staged_api, stdsimd)]
 #![feature(doc_cfg, allow_internal_unstable)]
+#![feature(promotable_const_fn, custom_attributes)]
 #![cfg_attr(feature = "cargo-clippy", allow(shadow_reuse))]
 #![cfg_attr(target_os = "linux", feature(linkage))]
 #![no_std]
--- a/stdsimd/arch/detect/cache.rs
+++ b/stdsimd/arch/detect/cache.rs
@@ -12,11 +12,13 @@ use core::sync::atomic::AtomicU64;
 use core::sync::atomic::AtomicU32;
 
 /// Sets the `bit` of `x`.
+#[promotable_const_fn]
 pub const fn set_bit(x: u64, bit: u32) -> u64 {
     x | 1 << bit
 }
 
 /// Tests the `bit` of `x`.
+#[promotable_const_fn]
 pub const fn test_bit(x: u64, bit: u32) -> bool {
     x & (1 << bit) != 0
 }


lu-zero:
Could you please add a link to the rustc commit? :)

oli-obk:
Can't, the PR isn't even open yet ;) I'll do that when I get there.

Could you place this PR into a branch so we're following the rustc rules for lockstep upgrades correctly?

oli-obk:
And now we have a PR: https://github.com/rust-lang/rust/pull/51500

lu-zero:
I'd ask @alexcrichton and @gnzlbg to do it since I never did before :)

oli-obk:
```
git fetch origin pull/471/head:oli-obk-lockstep
git checkout oli-obk-lockstep
git push origin oli-obk-lockstep
```

lu-zero:
That part I knew, the branch naming and where to put it is the part I'm completely oblivious of :)

gnzlbg:
It might be obvious, but what is a promotable const fn ? Also, it would be cool to add some tests that would break without the promotable attribute (you can add them inline, or in stdsimd/crates/coresimd/tests).

oli-obk:
A promotable const fn is one that is marked with the `#[promotable_const_fn]` attribute. Currently that means it can only call other promotable const fns and it cannot contain unions. This step is done as an intermediate solution for various fears we have around promoting calls to const fn and the attribute will likely never become stable, but be replaced by an automatic analysis.

We have such tests in the rustc repository, I'll check if you have compile-fail tests, if not, this is untestable anyway,

gnzlbg:
So does that mean that all `const fn`s that do not contain unions and only call other promotable const fns should use the attribute? Some of the functions you added the attribute to are private (`test_bit`, `set_bit`, etc.).

oli-obk:
> So does that mean that all const fns that do not contain unions and only call other promotable const fns should use the attribute?

No further functions should need this attribute. These two PRs just marked all the const fns callable from stable code (directly or indirectly) as promotable (because they already were stable). Now we can continue stabilizing const fns without stabilizing their promotability at the same time

coresimd/ppsv/api/masks.rs
```diff
@@ -11,25 +11,29 @@ macro_rules! impl_mask_minimal {
             /// Creates a new instance with each vector elements initialized
             /// with the provided values.
             #[inline]
+            #[promotable_const_fn]
             pub const fn new($($elem_name: bool),*) -> Self {
```
gnzlbg: which stable code can call the const functions in this file?

gnzlbg: is it possible to find that out?

oli-obk: Uh... other than removing the attribute and following the fallout?

gnzlbg: Could you post the fallout? I am curious because nothing in this module should be reachable from stable code.

gnzlbg: Actually, only the stdsimd/arch/detect/cache.rs functions should be reachable from stable code =/

oli-obk:
> I'll check if you have compile-fail tests

I have not found any such tests. So we'd have to add such tests, which is a lot of effort for little gain, as it's tested by the rustc test suite and will be checked here once the attribute is part of rustc

coresimd/ppsv/api/masks.rs
oli-obk: Oh... will do, I might have been a little overeager in marking things here, because it was hard to tell what was called from where.

gnzlbg: Might be that I screwed up. Worst case we can just merge this as is, and once the next nightly I can try removing the attribute from all places where it shouldn't be needed. Let me know when you want this merged.

oli-obk: ok, removed all unnecessary attributes

oli-obk:
We're going an alternative route, so no attributes are needed

Issue #475, opened 2018-06-12
Add _MM_SHUFFLE macro for x86/x86_64


bitshifter:
`xmmintrin.h` defines a _MM_SHUFFLE macro which is use to generate the mask used for `_mm_shuffle` and other similar instructions. https://clang.llvm.org/doxygen/xmmintrin_8h.html#a65a052b655bd49ff3fe128b61847df9f. I could not find this macro in `std::arch::x86`, perhaps it is there but doesn't show up in the docs?

In any case, if it's missing perhaps it would be good to add it. Creating masks by hand is cryptic and having an equivalent to `_MM_SHUFFLE` would make using the Rust intrinsics more like other languages.

gnzlbg:
Thanks for reporting this. Wanna send a PR? Should be pretty easy to implement (check the other macros and `const`s in the `coresimd/x86/sse.rs` module.

bitshifter:
Will do.

bitshifter:
@gnzlbg it turns out this is not so straight forward. I've added the macro and a test in this commit https://github.com/bitshifter/stdsimd/commit/727bc111c80f70ef103e0e1d93ad9cf37824a08a however it seems that coresimd is re-exported via stdsimd, but I do not know how to re-export the macro.

I was reminded about binary literals when looking at the tests, which are reasonably readable. Perhaps if this macro issue is not easily solved adding and boolean literal example would help.

gnzlbg:
@bitshifter you should add a `const fn` instead of a macro, or why does it need to be a macro?

gnzlbg:
Also, you should call it `_MM_SHUFFLE` and not `_mm_shuffle`.

PR #479, opened 2018-06-16
Added _MM_SHUFFLE utility function.

bitshifter:
Mimics the macro defined in xmmintrin.h defines a _MM_SHUFFLE macro which is use to generate the mask used for _mm_shuffle and other similar instructions. Fixes issue #475.

### DIFF
--- a/coresimd/x86/sse.rs
+++ b/coresimd/x86/sse.rs
@@ -968,6 +968,14 @@ pub unsafe fn _mm_setzero_ps() -> __m128 {
     __m128(0.0, 0.0, 0.0, 0.0)
 }
 
+/// A utility function for creating masks to use with Intel shuffle and permute intrinsics.
+#[inline]
+#[allow(non_snake_case)]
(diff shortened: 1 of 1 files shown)


coresimd/x86/sse.rs
```diff
@@ -968,6 +968,14 @@ pub unsafe fn _mm_setzero_ps() -> __m128 {
     __m128(0.0, 0.0, 0.0, 0.0)
 }
 
+/// A utility function for creating masks to use with Intel shuffle and permute intrinsics.
+#[inline]
+#[allow(non_snake_case)]
+#[stable(feature = "simd_x86", since = "1.27.0")]
```
gnzlbg: This will probably need to be 1.28.0 since 1.27 is in beta already and I don't think this will be back ported to beta.

bitshifter: Done

gnzlbg:
Thanks!

PR #486, opened 2018-06-22
Remove stray inconsistent #[cfg]

est31:
The cfg for the target is already covered a few lines below.
This cfg meant that even with the dox feature enabled, no
documentation for powerpc64 was generated except you were
actually targetting the powerpc64 arch.


est31:
This should fix a deadlink error I've encountered in https://github.com/rust-lang/rust/pull/51695#issuecomment-399293640

gnzlbg:
@est31 thanks!

PR #493, opened 2018-06-23
Fix some links

est31:
Always point to {target_name}/.../{target_name}/index.html


est31:
Given that the `dox` feature enables documentation for all targets, we could possibly remove that entire section. Not sure though whether the `dox` feature captures everything so I kept the section.

gnzlbg:
Thanks!

PR #494, opened 2018-06-23
deprecate read/write eflags

gnzlbg:
This deprecates `__{read,write}eflags`.

### DIFF
--- a/coresimd/x86/eflags.rs
+++ b/coresimd/x86/eflags.rs
@@ -6,6 +6,8 @@
 #[cfg(target_arch = "x86")]
 #[inline(always)]
 #[stable(feature = "simd_x86", since = "1.27.0")]
+#[rustc_deprecated(since = "1.29.0", reason = "See issue #51810 - use inline assembly instead")]
+#[doc(hidden)]
 pub unsafe fn __readeflags() -> u32 {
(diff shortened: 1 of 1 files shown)


Results51:
#999

Results51:
#984

gnzlbg:
Which rust version belongs exactly to the deprecation attribute? It might hit nightly in the next weeks, so it should be two stable versions from now (1.29) right?

BurntSushi:
At a high level this LGTM, but I don't know enough about the particulars of how exactly things are deprecated to feel comfortable signing off on this.

cc @rust-lang-nursery/libs @alexcrichton

alexcrichton:
FWIW these are covered by [this block](https://github.com/rust-lang-nursery/stdsimd/blob/87f25c76e12fc9636fdea3ae0bf52b727ee78381/crates/stdsimd-verify/tests/x86-intel.rs#L133-L139) during verification which is why they passed through to stable without actually being defined by Intel.

I think it's a compelling enough argument to deprecate these intrinsics currently and we can always undeprecate if we *really* need to (or if some LLVM "bug" is fixed or something like that)

Thanks for the PR @gnzlbg!

PR #496, opened 2018-06-24
add rotate_left/right

TheIronBorn:
(also removed some duplicate tests)

It would be good to test that on supported hardware this compiles to the correct instruction, but I don't know how to do that.

Also some scalar convenience impls would be good.

(https://github.com/rust-lang-nursery/rand/issues/377)

### DIFF
--- a/coresimd/ppsv/api/rotates.rs
+++ b/coresimd/ppsv/api/rotates.rs
@@ -9,8 +9,8 @@ macro_rules! impl_vector_rotates {
             /// the end of the resulting integer.
             ///
             /// Please note this isn't the same operation as `<<`!. Also note it
-            /// isn't equivalent to `slice::rotate_left` (that can be implemented
-            /// with vector shuffles).
+            /// isn't equivalent to `slice::rotate_left`, it doesn't move the vector's
+            /// lanes around. (that can be implemented with vector shuffles).
             #[inline]
             pub fn rotate_left(self, n: $id) -> $id {
                 const LANE_WIDTH: $elem_ty = ::mem::size_of::<$elem_ty>() as $elem_ty * 8;
--- a/coresimd/ppsv/api/rotates.rs
+++ b/coresimd/ppsv/api/rotates.rs
@@ -1,0 +1,83 @@
+//! Implements integer rotates.
+#![allow(unused)]
+
+macro_rules! impl_vector_rotates {
+    ($id:ident, $elem_ty:ident) => {
+        impl $id {
+            /// Shifts the bits of each lane to the left by the specified amount in
+            /// the corresponding lane of `n`, wrapping the truncated bits to
+            /// the end of the resulting integer.
+            ///
+            /// Please note this isn't the same operation as `<<`!. Also note it
+            /// isn't equivalent to `slice::rotate_left` (that can be implemented
+            /// with vector shuffles).
+            #[inline]
+            pub fn rotate_left(self, n: $id) -> $id {
+                const LANE_WIDTH: $elem_ty = ::mem::size_of::<$elem_ty>() as $elem_ty * 8;
+                // Protect against undefined behavior for over-long bit shifts
+                let n = n % LANE_WIDTH;
+                (self << n) | (self >> ((LANE_WIDTH - n) % LANE_WIDTH))
+            }
+
+            /// Shifts the bits of each lane to the right by the specified amount in
+            /// the corresponding lane of `n`, wrapping the truncated bits to
+            /// the beginning of the resulting integer.
+            ///
+            /// Please note this isn't the same operation as `>>`!. Also note it
+            /// isn't similar to `slice::rotate_right`, it doesn't move the vector's
+            /// lanes around. (that can be implemented with vector shuffles).
+            #[inline]
+            pub fn rotate_right(self, n: $id) -> $id {
+                const LANE_WIDTH: $elem_ty = ::mem::size_of::<$elem_ty>() as $elem_ty * 8;
+                // Protect against undefined behavior for over-long bit shifts
+                let n = n % LANE_WIDTH;
+                (self >> n) | (self << ((LANE_WIDTH - n) % LANE_WIDTH))
+            }
+        }
+    };
+}
+
+#[cfg(test)]
+macro_rules! test_vector_rotate_ops {
+    ($id:ident, $elem_ty:ident) => {
+        #[test]
+        fn rotate_ops() {
+            use coresimd::simd::$id;
+            use std::mem;
+            let z = $id::splat(0 as $elem_ty);
+            let o = $id::splat(1 as $elem_ty);
+            let t = $id::splat(2 as $elem_ty);
+            let f = $id::splat(4 as $elem_ty);
+
+            let max =
+                $id::splat((mem::size_of::<$elem_ty>() * 8 - 1) as $elem_ty);
+
+            // rotate_right
+            assert_eq!(z.rotate_right(z), z);
+            assert_eq!(z.rotate_right(o), z);
+            assert_eq!(z.rotate_right(t), z);
+
+            assert_eq!(o.rotate_right(z), o);
+            assert_eq!(t.rotate_right(z), t);
+            assert_eq!(f.rotate_right(z), f);
+            assert_eq!(f.rotate_right(max), f << 1);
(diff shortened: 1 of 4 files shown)


TheIronBorn:
Any concerns about organization etc?

TheIronBorn:
Huh, some of these are being compiled from `rotate_right` to `vprolq`. Not sure how to make that pass the tests.

coresimd/ppsv/api/rotates.rs
```diff
@@ -0,0 +1,104 @@
+//! Implements integer rotates.
+#![allow(unused)]
+
+/// Trait used for overloading rotates
+pub trait Rotate<T> {
```
gnzlbg: Given that you provide a specific implementation for each vector type below, this doesn't really need to be a trait.

In general we try to keep the traits we use to a minimum, and most of them are sealed (not exposed in the public API of the library).

coresimd/ppsv/api/rotates.rs
```diff
@@ -0,0 +1,104 @@
...
+            #[inline]
+            fn rotate_right(self, n: $id) -> $id {
+                const LANE_WIDTH: $elem_ty = ::mem::size_of::<$elem_ty>() as $elem_ty * 8;
+                // Protect against undefined behavior for over-long bit shifts
+                let n = n % LANE_WIDTH;
+                (self >> n) | (self << ((LANE_WIDTH - n) % LANE_WIDTH))
+            }
+        }
+
+        impl super::api::Rotate<$elem_ty> for $id {
```
gnzlbg: Aha, I see why you need the trait now.

crates/coresimd/tests/rotate_tests.rs
```diff
@@ -0,0 +1,49 @@
...
+extern crate stdsimd;
+
+use stdsimd::simd::*;
+use stdsimd_test::assert_instr;
+
+// Verify that supported hardware compiles rotates into single instructions
+
+#[inline]
+#[target_feature(enable = "avx512f")]
+#[cfg_attr(test, assert_instr(vprorvq))]
```
gnzlbg: The `assert_instr` and `target_feature` need to be guarded on `x86` and `x86_64`. Same for all other functions below.

TheIronBorn:
Could there be some way to seal the `Rotate` trait?

Otherwise I agree. Rotates probably aren't used that much anyway.

TheIronBorn:
Making progress. The only failures now are a couple shift implementations and left/right swapping.

gnzlbg:
> Could there be some way to seal the Rotate trait?

I am not sure. IIUC sealed traits that have to be imported by users don't work because they can't be imported. That is, if you have an implementation detail that's not exposed to users, you can make it a sealed trait. If you want to express a `where` bound to restrict an implementation, you can also make it a sealed trait. But if the user needs to call a method of the trait, then the user needs to import it, and the trait has to be public.

One could try to make the `rotate_left`/`rotate_right` functions on vectors generic, and then use a sealed trait to constrain it to vectors of the same type, or an scalar value. The sealed trait could then have a e.g. `to_vector()` method that for vectors does nothing and for scalars does an appropriate splat. 

But given that we could extend a non-generic `rotate_left`/`rotate_right` to allow taking scalars that way in a backwards compatible way, I'd prefer to get them merged first as inherent methods for vectors only, and figure out how to make these more ergonomic for scalars in a different PR.

TheIronBorn:
Is there some way to do `any(assert_instr(vprolq), assert_instr(vprorq))`?

gnzlbg:
@TheIronBorn 

> Is there some way to do any(assert_instr(vprolq), assert_instr(vprorq))?

Easily only using `#[cfg_attr(...), assert_instr(...)]` which means that you can only have them for different archs, target features, etc. 

For what exactly do you need it?

TheIronBorn:
Some compilations turn a left rotate into a right. Not sure why

gnzlbg:
That's weird. As long as the tests pass that might just be llvm optimizations at work. If you tell me which one exactly I can take a look.

TheIronBorn:
Here's one https://travis-ci.org/rust-lang-nursery/stdsimd/jobs/398348597

TheIronBorn:
Nearly there. Just a few still doing `shift + shift + or`.

TheIronBorn:
I've been trying to get codegen correct and found some weird behavior where just by clearing the cache and recompiling I get different results. https://godbolt.org/g/oJM7Tw
I'll report this to godbolt.

At the very least this makes progress on this difficult, and possibly even makes it impossible to guarantee codegen correctness.

TheIronBorn:
I think I've reached the end of my understanding of why these are still producing bad code

gnzlbg:
I see you are only enabling the avx512f target feature, could you try enabling avx512vl as well?

TheIronBorn:
That didn't seem to have any change

crates/coresimd/tests/rotate_tests.rs
```diff
@@ -0,0 +1,51 @@
...
+use stdsimd::simd::*;
+use stdsimd_test::assert_instr;
+
+// Verify that supported hardware compiles rotates into single instructions
+
+#[inline]
+#[cfg_attr(any(target_arch = "x86", target_arch = "x86_64"), target_feature(enable = "avx512f"))]
+#[cfg_attr(any(target_arch = "x86", target_arch = "x86_64"), target_feature(enable = "avx512vl"))]
+#[cfg_attr(any(target_arch = "x86", target_arch = "x86_64"), assert_instr(vpro))]
+unsafe fn rotate_right_variable(x: u64x8) -> u64x8 {
```
gnzlbg: Could you add tests here for 128-bit wide and 256-bit wide vectors as well?

gnzlbg:
Looking at the implementation of [`_mm_rolv_epi32`](https://github.com/llvm-mirror/clang/blob/15349a004a06780980702064429c03f23bd296c7/lib/Headers/avx512vlintrin.h#L4267) in clang, the only target feature attribute it has is [`avx512vl`](https://github.com/llvm-mirror/clang/blob/15349a004a06780980702064429c03f23bd296c7/lib/Headers/avx512vlintrin.h#L31). It is implemented in term of [`__builtin_ia32_prolvd128`](https://github.com/llvm-mirror/clang/blob/11b0a91ad96a1a636399660e3e99410af66ba553/include/clang/Basic/BuiltinsX86.def#L1365); the similar builtins are all next to it there and all require `avx512vl`.

The LLVM-IR intrinsic that these should generate is `llvm.x86.avx512.prolv`, there are a couple of tests in the LLVM sources about it:

* https://github.com/llvm-mirror/llvm/blob/6451a6d6f650f1435ab1e47898deb96f8e9c267a/test/CodeGen/X86/avx512vl-intrinsics-fast-isel.ll#L6748

* https://github.com/llvm-mirror/llvm/blob/8c0bb2f036502adec9989b1bef1fe19f76fb644a/test/CodeGen/X86/avx512-intrinsics.ll#L3854

So what you can do is emit the LLVM-IR that is being generated for the variable length rotates (e.g. using `RUSTFLAGS=--emit=llvm-ir`, `cargo llvm-ir ...`, etc.) and see how it differs from the IR in those tests.

For example, with [`llc trunk`](https://godbolt.org/g/32qfWv) it seems that only the `avx512f` attribute is required for the variable length rotates, but if I use `llc` 5.0 instead of trunk then I get pretty bad machine code, so it might well be that this only works on LLVM7 (the way to test this would be to use the `llc` shipped with Rust).

Also, using `i64` instead of `i32` to match your code seems to produce good machine code as well when the appropriate (`q`) intrinsic is used: https://godbolt.org/g/bF3DBt

TheIronBorn:
I think the Intel docs say `avx512vl` is only required for vectors < 512-bits wide

gnzlbg:
@TheIronBorn sorry about the organisational issues, but would you mind sending this PR to https://github.com/gnzlbg/ppv ? The ppsv module will be removed from `core` in the near future, and that library will probably be the replacement.

I'll leave this open here until we remove the ppsv module, and before removal will try to port this PR myself if you don't beat me to it.

gnzlbg:
The ppsv module has been removed and this has already been merged upstream.

@TheIronBorn I'll try to propagate the `assert_instr` tests upstream tomorrow.

PR #507, opened 2018-06-27
Add vector multiply round and add saturated

lu-zero:


coresimd/powerpc/altivec.rs
```diff
@@ -868,6 +882,28 @@ mod tests {
...
+            7 * 256,
+        ).into_bits();
+        let b: vector_signed_short =
+            i16x8::new(256, 256, 256, 256, 256, 256, 256, 256).into_bits();
+        let c: vector_signed_short =
+            i16x8::new(0, 1, 2, 3, 4, 5, 6, i16::max_value() - 1).into_bits();
+
+        let d = i16x8::new(0, 3, 6, 9, 12, 15, 18, i16::max_value());
+
+        assert_eq!(d, vec_madds(a, b, c).into_bits());
```
CryZe: Is this supposed to be vec_madds?

lu-zero: Good catch!
No, it should be an r that got forgotten...

PR #517, opened 2018-07-06
add swap_bytes/to_le/to_be

TheIronBorn:
We could add some `llvm.bswap` implementations (https://godbolt.org/g/jH2FCR suggests it might be better).

Testing could be better.

### DIFF
--- a/coresimd/ppsv/api/swap_bytes.rs
+++ b/coresimd/ppsv/api/swap_bytes.rs
@@ -1,0 +1,94 @@
+//! Horizontal swap bytes.
+
+macro_rules! impl_swap_bytes {
+    ($id:ident) => {
+        impl $id {
+            /// Reverses the byte order of the vector.
+            #[inline]
+            pub fn swap_bytes(self) -> Self {
+                unsafe { super::codegen::swap_bytes::SwapBytes::swap_bytes(self) }
+            }
+
+            /// Converts self to little endian from the target's endianness.
+            ///
+            /// On little endian this is a no-op. On big endian the bytes are swapped.
+            #[inline]
+            pub fn to_le(self) -> Self {
+                #[cfg(target_endian = "little")]
+                {
+                    self
+                }
+                #[cfg(not(target_endian = "little"))]
+                {
+                    self.swap_bytes()
+                }
+            }
+        }
+    };
+}
+
+#[cfg(test)]
+macro_rules! test_swap_bytes {
+    (u8x2) => {
+        #[test]
+        fn swap_bytes() {
+            use coresimd::simd::u8x2;
+            use std::mem;
+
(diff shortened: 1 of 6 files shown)


coresimd/ppsv/api/swap_bytes.rs
```diff
@@ -0,0 +1,94 @@
...
+            #[inline]
+            pub fn swap_bytes(self) -> Self {
+                unsafe { super::codegen::swap_bytes::SwapBytes::swap_bytes(self) }
+            }
+
+            /// Converts self to little endian from the target's endianness.
+            ///
+            /// On little endian this is a no-op. On big endian the bytes are swapped.
+            #[inline]
+            pub fn to_le(self) -> Self {
```
gnzlbg: Should we have a `to_be` as well ?

TheIronBorn: Oh right my bad.

coresimd/ppsv/api/swap_bytes.rs
```diff
@@ -0,0 +1,109 @@
...
+            let expected = x.swap_bytes();
+
+            let vec: u8x16 = unsafe { mem::transmute(x) };
+            let actual = unsafe { mem::transmute(vec.swap_bytes()) };
+
+            assert_eq!(expected, actual);
+        }
+    };
+    // testing larger vectors is less simple
+    ($id:ident) => {};
```
gnzlbg: This branch should always `panic!` so that we make sure that everything is properly tested.

TheIronBorn:
The only difficulty there is choosing the 8 lane-width vector without explicitly matching every possibility

coresimd/ppsv/api/swap_bytes.rs
```diff
@@ -42,68 +46,112 @@ macro_rules! impl_swap_bytes {
...
-            use coresimd::simd::u8x2;
-            use std::mem;
-
-            let x: u16 = 0x2d99;
-            let expected = x.swap_bytes();
-
-            let vec: u8x2 = unsafe { mem::transmute(x) };
-            let actual = unsafe { mem::transmute(vec.swap_bytes()) };
+macro_rules! impl_test {
+    ($id:ident, $vec8:ident) => {
```
gnzlbg: I think it would be better to just use a `[u8; 64]`, initialize that, and then use the `$id::{load,store}_unaligned` methods to load that into a vector, call `swap_bytes`/`to_le`/`to_be`, and then write the result back to an `[u8; 64]` array. Then you just compare the first `size_of::<$id>()` bytes of the array against the result that they should have.

TheIronBorn:
Like this?

gnzlbg:
Yes, this looks great, thank you a lot for working so hard on this!

PR #519, opened 2018-07-10
Add Vector Multiply Sum Saturated

lu-zero:


lu-zero:
the `DOCUMENTATION` instance fails this way:
```
error: component 'rust-std' for target 'x86_64-unknown-linux-gnu' was automatically added because it is required for toolchain 'nightly-x86_64-unknown-linux-gnu'
```
What is exactly the problem here?

lu-zero:
Sorry, the problem seems:
```
LLVM ERROR: Do not know how to split the result of this operator!
error: Could not compile `coresimd`.
```

That is a bit puzzling.

gnzlbg:
> the DOCUMENTATION instance fails this way:
>
> error: component 'rust-std' for target 'x86_64-unknown-linux-gnu' was automatically added because it is required for toolchain 'nightly-x86_64-unknown-linux-gnu'
>
> What is exactly the problem here?

There is no problem. `rustup target add` emits this error, but it is guarded with a `| true`.

---

> Sorry, the problem seems:
> 
> LLVM ERROR: Do not know how to split the result of this operator!
error: Could not compile `coresimd`.
>
> That is a bit puzzling.

This just means that there is an error somewhere. Could you set this line to set -ex ? https://github.com/rust-lang-nursery/stdsimd/blob/master/ci/dox.sh#L7

lu-zero:
I sprinkled some `--verbose` and got to this:

```
rustc --verbose --crate-name coresimd crates/coresimd/src/lib.rs --crate-type lib --emit=dep-info,link -C debuginfo=2 -C metadata=ae2adffe51c3f8ae -C extra-filename=-ae2adffe51c3f8ae --out-dir /home/ubuntu/rust/stdsimd/target/powerpc-unknown-linux-gnu/debug/deps --target powerpc-unknown-linux-gnu -C incremental=/home/ubuntu/rust/stdsimd/target/powerpc-unknown-linux-gnu/debug/incremental -L dependency=/home/ubuntu/rust/stdsimd/target/powerpc-unknown-linux-gnu/debug/deps -L dependency=/home/ubuntu/rust/stdsimd/target/debug/deps
LLVM ERROR: Do not know how to split the result of this operator!
```

lu-zero:
Since https://github.com/rust-lang/rust/issues/50960 is still open I would not devote much time on that.

PR #521, opened 2018-07-11
One more instruction and further refinements

lu-zero:
I'm wondering if we couldn't just use a single `target_feature` at module level.

### DIFF
--- a/coresimd/powerpc/altivec.rs
+++ b/coresimd/powerpc/altivec.rs
@@ -96,12 +96,16 @@ mod sealed {
     }
 
     impl VectorMsums<vector_unsigned_int> for vector_unsigned_short {
+        #[inline]
+        #[target_feature(enable = "altivec")]
         unsafe fn vec_msums(self, b: Self, c: vector_unsigned_int) -> vector_unsigned_int {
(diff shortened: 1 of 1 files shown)


coresimd/powerpc/altivec.rs
```diff
@@ -394,12 +394,16 @@ mod sealed {
     }
 
     impl VectorMsums<vector_unsigned_int> for vector_unsigned_short {
+        #[inline]
+        #[target_feature(enable = "altivec")]
```
gnzlbg: the `assert_instr` are missing

lu-zero: good call!

Actually looks like the portable instructions do not get mapped correctly.

lu-zero:
the CI meanwhile seems having its share of issues:
```
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

gnzlbg:
This needs rebasing on top of master.

lu-zero:
Rebased and added another instruction.

alexcrichton:
Thanks!

Issue #522, opened 2018-07-12
_MM_SHUFFLE has incorrect return type


Paul-E:
`_MM_SHUFFLE` was added recently in #479 . Currently [`_MM_SHUFFLE`](https://rust-lang-nursery.github.io/stdsimd/x86_64/stdsimd/arch/x86_64/fn._MM_SHUFFLE.html) returns a `u32`, but the functions it's output is used with take an `i32`, eg [`_mm_shuffle_epi32`](https://rust-lang-nursery.github.io/stdsimd/x86_64/stdsimd/arch/x86_64/fn._mm_shuffle_epi32.html)

gnzlbg:
cc @bitshifter

bitshifter:
Oops. Will look into it.

alexcrichton:
Hm it also looks like this isn't `unsafe` and it's also `const`, could it follow the existing intrisnics and be both `unsafe` and non-`const`?

gnzlbg:
@alexcrichton this is a helper macro in C. I think that making it const (or a macro in Rust) is not only fine, but necessary, since its result needs to be passed to other intrinsics taking immediate mode arguments, which only accept constant expressions (so if we don't make this const, or a macro, this becomes unusable). 

Also, this is not an intrinsic per se, in the sense that it does not require target features (it is a helper macro in C), so I don't know why it should be unsafe. We map other helper macros to just `const` variables, the difference here is that this helper macro takes some arguments, but that's about it.

alexcrichton:
Hm ok if that's the case can it be marked as unstable for now? I'd prefer to not stabilize this behavior immediately personally

gnzlbg:
>  I'd prefer to not stabilize this behavior immediately personally

Sure. 

---

FWIW, I think we screwed up. `_MM_SHUFFLE` is sometimes the recommended way of using the SSE and AVX shuffle intrinsics, but some of the Rust intrinsics take an `u32` (e.g. `_mm_shuffle_ps`, which is a bug, this should be `i32`) , and all others take an `i32` (`_mm256_shuffle_ps`, `_mm_shuffle_pi{16,32}`, ...).

Ideally, we should do a breaking change and fix the argument type of `_mm_shuffle_ps`. The alternative is to make `_MM_SHUFFLE!` a macro, so that the argument type can be infered from the expansion...

gnzlbg:
[...]

PR #524, opened 2018-07-14
Make _MM_SHUFFLE unstable.

bitshifter:
Making _MM_SHUFFLE unstable as it turns out some intrinsics take i32 and some u32 which means this helper doesn't return the correct type in all cases.

See #522


alexcrichton:
Thanks!

PR #528, opened 2018-07-18
remove std::simd 

### LINKED ISSUE #525: Shall we split the `std::simd` module into a crate separate from std?
I have a pretty much equivalent implementation of `stdsimd` that follows the RFC in https://github.com/gnzlbg/ppv

Given that:

* the stabilization path for `std::simd` is not clear yet, and that stabilization might take a while
* there are issues with some targets failing with `core` upstream, e.g., due to linkage issues
* `std::simd` costs most of the CI time of the `stdsimd` repo, 

I'd like to discuss whether we should split `std::simd` into a crate separate from `std` (not necessarily mine, we can put the current implementation 1:1 as is in the nursery).

Splitting the crate might let us sync upstream more often and without breakage with `stdsimd`, significantly reducing the amount of churn we are seeing.

cc @alexcrichton 

---

This basically turns `rust-lang-nursery/stdsimd` into `stdarch`, and the current `std::simd` into a crates on crates.io that people can just use with nightly in an opt-in way.

gnzlbg:
Closes #525 .


gnzlbg:
I've run out of CI capacity so I am posting this here a bit early, its still WIP but most things build already and most tests on most targets pass. 

I won't merge this till the `ppv` crate can be used as a fallback, but right now the only feature missing is the `FromBits/IntoBits` traits, which are not part of the RFC. My plan is to add those behind a feature flag, and then release the crate.

lu-zero:
Keeping FromBits and IntoBits would be less painful for our theoretical users (and for us when writing the tests) :)

CryZe:
This has the downside that you can't experiment with portable SIMD on Godbolt anymore, which was very useful. I guess someone should just add the ability to use crates to Godbolt though instead.

gnzlbg:
@Cryze yeah that was very useful, godbolt supports using libraries for other languages (e.g. C++) so maybe we should open an issue there.  Now that godbolt has gained cross-compilation support, this would be very very useful.

alexcrichton:
@gnzlbg would you be ok merging this for now to get CI green again, but we'll hold off on updating the submodule in rust-lang/rust until you've published?

gnzlbg:
Done, I might get time to get the From/IntoBits impl ported tomorrow. The submodule update will still take a while but maybe we could start doing that at the end of this week, since that should fix all of the linker errors that people were seeing due to missing definitions of some of the vector functions.

PR #531, opened 2018-07-20
Fix compile of stdsimd on powerpc with no flags

alexcrichton:
We're running into issues updating with rust-lang/rust#52535, so we need to get
this working without `RUSTFLAGS` enabling the `altivec` feature


alexcrichton:
cc @lu-zero

lu-zero:
Fine for me

PR #532, opened 2018-07-21
ARM DSP: add quad/double add/sub + exchange and select bytes intrinsics

paoloteti:
This PR add the following DSP intrinsics:

- Quad 8-bit addition/subtraction
- Double 8-bit addition/subtraction
- Saturating Add and Subtract with Exchange and
  Saturating Subtract and Add with Exchange, signed
- Select bytes based on GE bits

This patch bump the `assert_instr` limit to 22 (from 20) instead of add a lots of exception for all DSP intrinsics.

### DIFF
--- a/coresimd/arm/dsp.rs
+++ b/coresimd/arm/dsp.rs
@@ -14,24 +14,48 @@ types! {
     pub struct uint16x2_t(u16, u16);
 }
 
+macro_rules! dsp_call {
+    ($name:expr, $a:expr, $b:expr) => {
+        ::mem::transmute($name(::mem::transmute($a), ::mem::transmute($b)))
+    };
+}
+
 extern "C" {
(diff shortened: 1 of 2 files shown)


coresimd/arm/dsp.rs
```diff
@@ -135,7 +232,7 @@ mod tests {
             let a = i8x4::new(1, 2, 3, ::std::i8::MAX);
             let b = i8x4::new(2, -1, 0, 1);
             let c = i8x4::new(3, 1, 3, ::std::i8::MAX);
-            let r: i8x4 = ::mem::transmute(dsp::qadd8(::mem::transmute(a), ::mem::transmute(b)));
+            let r: i8x4 = dsp_call!(dsp::qadd8, a, b);
```
gnzlbg: why is the macro being used in the tests?

alexcrichton:
👍

PR #534, opened 2018-07-23
ARM DSP: fix intra_doc_link_resolution_failure

paoloteti:
Escape '[' and ']' used for arrays and not as Markdown syntax.


alexcrichton:
Thanks!

PR #536, opened 2018-07-24
generate docs for Cortex-M / Cortex-R intrinsics

japaric:
This should fix doc generation for the intrinsics added in #518, #529 and #532. Or, at least the docs show up for me when I invoke `cargo rustdoc -p coresimd --open -- --cfg dox`.


alexcrichton:
👍

japaric:
@alexcrichton do we need to update the submodule in rust-lang/rust for https://doc.rust-lang.org/nightly/core/arch/arm/index.html to be updated?

alexcrichton:
We do indeed!

PR #537, opened 2018-07-24
ARM  ARM DSP/SIMD32: Sum of 8-bit absolute differences

paoloteti:
Just add:

- `usad8`: Sum of 8-bit absolute differences
- `usad8a`: Sum of 8-bit absolute differences and constant (usad8(a, b) + c)


alexcrichton:
👍

PR #538, opened 2018-07-25
Restructure the PowerPC support

lu-zero:
Make the code not scattered between `powerpc` and `powerpc64` and factorize away some redundant target_feature now that the module imports are guarded.


coresimd/powerpc/mod.rs
```diff
@@ -5,5 +5,7 @@ mod altivec;
 #[cfg(target_feature = "altivec")]
 pub use self::altivec::*;
 
+#[cfg(target_feature = "vsx")]
 mod vsx;
+#[cfg(target_feature = "vsx")]
```
gnzlbg: both vsx and altivec can be detected at run-time, or is this not true for powerpc ?

Still, they can be detected at run-time for powerpc64 right?

gnzlbg: If so, the `#[target_feature(enable)]` are necessary.

If not, we have decided to keep `#[target_feature(enable)]` attributes on intrinsics even when they are guarded by a static `cfg(target_feature)`. 

Changing this should be something that is applied to all backends, not only powerpc.

gnzlbg: > They can be detected at runtime,

So this change breaks that because it forces them to be enabled at compile-time, for the whole binary. 

> Isn't #[target_feature(foo)] a short-hand for #[cfg(target_feature="foo")] ?

* `#[target_feature(enable = "x")]` is a function attribute that tells the compiler to generate code for the function with the feature `x` enabled.

* `cfg(target_feature = "x")` returns true if the feature `x` has been enabled at compile-time, and otherwise it returns false

lu-zero: The `#[cfg(target_feature = "altivec")]` comes from 5a20376f4, I just made the vsx support consistent and I assumed the two attributes are interchangeable and just conditionally compile what they decorate.

Also I assumed that the target_feature would map the same way `-maltivec`/`-mvsx` and relative macro symbol do. So it would just signal what the compiler actually supports for the given cpu target.

So, `#[target_feature(enable = "foo")]` would tell the compiler to use the specific feature notwithstanding what the current target_feature set is? The name is _quite_ misleading if that's the case (and I would just apply it to the whole module anyway)

lu-zero: I'm not sure if powerpc is still in non-really-building condition, I can keep the additional decorators around assuming they do not work on module-level and continue to implement the other intrinsics.

alexcrichton:
The current state at least looks good to me, thanks!

PR #546, opened 2018-08-01
add arm/aarch64 table lookup and vector combine intrinsics

gnzlbg:


gnzlbg:
@Amanieu please do a review of this when you have time just in case, I can fix things in a subsequent PR.

Amanieu:
One thing that concerns me is whether these intrinsics work correctly on big-endian targets (see [this link](https://github.com/llvm-mirror/clang/blob/master/utils/TableGen/NeonEmitter.cpp#L1864)).

For example, Clang's arm-neon.h has the following code:
```c
#ifdef __LITTLE_ENDIAN__
__ai int8x8_t vtbl1_s8(int8x8_t __p0, int8x8_t __p1) {
  int8x8_t __ret;
  __ret = (int8x8_t) __builtin_neon_vtbl1_v((int8x8_t)__p0, (int8x8_t)__p1, 0);
  return __ret;
}
#else
__ai int8x8_t vtbl1_s8(int8x8_t __p0, int8x8_t __p1) {
  int8x8_t __rev0;  __rev0 = __builtin_shufflevector(__p0, __p0, 7, 6, 5, 4, 3, 2, 1, 0);
  int8x8_t __rev1;  __rev1 = __builtin_shufflevector(__p1, __p1, 7, 6, 5, 4, 3, 2, 1, 0);
  int8x8_t __ret;
  __ret = (int8x8_t) __builtin_neon_vtbl1_v((int8x8_t)__rev0, (int8x8_t)__rev1, 0);
  __ret = __builtin_shufflevector(__ret, __ret, 7, 6, 5, 4, 3, 2, 1, 0);
  return __ret;
}
#endif
```

gnzlbg:
@Amanieu they definitely won't work correctly in big endian systems :/ 

Do we have any big endian arm and aarch64 targets in rustc ? I can't find any. If one can target these via `xargo` we should probably add them to rustup so that we can set them up in CI here. 

Without a system to test on, this would be something that I cannot really fix.

coresimd/aarch64/neon.rs
```diff
@@ -517,6 +633,936 @@ pub unsafe fn vpmaxq_f64(a: float64x2_t, b: float64x2_t) -> float64x2_t {
...
+#[target_feature(enable = "neon")]
+#[cfg_attr(test, assert_instr(tbx))]
+pub unsafe fn vtbx3_s8(a: int8x8_t, b: int8x8x3_t, c: int8x8_t) -> int8x8_t {
+    use coresimd::simd::i8x8;
+    let r = vqtbx2_s8(
+        a,
+        int8x16x2_t(vcombine_s8(b.0, b.1), vcombine_s8(b.2, ::mem::zeroed())),
+        ::mem::transmute(c),
+    );
+    let m: int8x8_t = simd_lt(c, ::mem::transmute(i8x8::splat(24)));
```
Amanieu: Should this be an unsigned comparison instead of a signed one? Negative index values are still out of range.

gnzlbg: Yes. This bug affects the arm and aarch64 `vtbx{N}_s8` intrinsics - the tests should be updated to test these with a negative index value.

Amanieu:
You are right that there are no big-endian targets at the moment, and it is somewhat a hassle to set one up since big-endian ARM systems are very rare.

So we don't forget about this in the future, maybe use [`compile_error`](https://doc.rust-lang.org/std/macro.compile_error.html) with a `#[cfg]` to indicate that these intrinsics are broken on big-endian.

Amanieu:
> I think we should be able to set up one on qemu as long as we had a proper target right ?

The tricky part is getting a big-endian toolchain (GCC and libc) to build libstd and libtest. These aren't available in distro repositories (AFAIK) but you can probably use one from [Linaro](http://releases.linaro.org/components/toolchain/binaries/latest/).

gnzlbg:
That sounds good, I'll send a PR to put them behind a cfg macro for little endian.

PR #547, opened 2018-08-01
Table lookup fixes

gnzlbg:
This provides the table lookup intrinsics only on little endian targets, and also adds tests for "negative indices" which are out-of-bounds.


alexcrichton:
👍

PR #548, opened 2018-08-04
Add `#` prefix to doc comment

topecongiro:
This will fix broken rustfmt in CI.

cc https://github.com/rust-lang-nursery/rustfmt/issues/2787.


gnzlbg:
@topecongiro IIUC that `#` prefix will not display that line, but both that line and the first line should be displayed (the `#` on the previous line is part of `#[cfg(...)]`), so I am not sure that this is correct?

PR #558, opened 2018-09-03
add __breakpoint

japaric:
this is an ARM compiler specific intrinsic which can be used to implement
CMSIS's __BKPT intrinsic

r? @gnzlbg


coresimd/arm/armclang.rs
```diff
@@ -0,0 +1,31 @@
...
+///
+/// It enables you to include a breakpoint instruction in your Rust code
+///
+/// `val` is a compile-time constant integer whose range is:
+///
+/// - `0...65535` if you are compiling source as A32 code.
+/// - `0...255` if you are compiling source as T32 code.
+///
+/// **NOTE** Due compiler limitations this function only supports the range `0...255` in A32 mode.
+// TODO support the extended range `0...65535` when compiling as A32 code (`cfg(not(target_feature =
```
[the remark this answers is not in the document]
japaric: Removed the TODO and updated the doc comment.

coresimd/arm/armclang.rs
```diff
@@ -0,0 +1,30 @@
...
+//!
+//! - [ARM Compiler v 6.10 - armclang Reference Guide](https://developer.arm.com/docs/100067/0610)
+
+/// This intrinsic inserts a BKPT instruction into the instruction stream generated by the compiler
+///
+/// It enables you to include a breakpoint instruction in your Rust code
+///
+/// `val` is a compile-time constant integer whose range is:
+///
+/// - `0...65535` if you are compiling source as A32 code.
```
gnzlbg: This should also be removed since it currently won't work and can be added later once it works in a backwards compatible way.

gnzlbg: Ah sorry, I see you also updated the Note below. So I think this is fine. I'll merge once CI is green. Let me know if there are any spurious CI failures.

coresimd/arm/armclang.rs
```diff
@@ -0,0 +1,30 @@
...
+/// - `0...255` if you are compiling source as T32 code.
+///
+/// [ARM's documentation](https://developer.arm.com/docs/100067/latest/compiler-specific-intrinsics/__breakpoint-intrinsic)
+///
+/// **NOTE** Due compiler limitations this function only supports the range `0...255` in A32 mode.
```
GabrielMajeri: Typo, should be `due to compiler...`. And maybe add a `:` after the **NOTE**?

gnzlbg:
This is failing on `aarch64` with:

```
error: <inline asm>:1:2: error: unrecognized instruction mnemonic
        BKPT 0
        ^
```

paoloteti:
> We might need a cfg on target arch and use a different instruction there.

If I'm not wrong is `BKPT` in `aarch32` and `BRK ` in `aarch64`.

andre-richter:
Yes, its `BRK`

http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0802a/BRK.html

alexcrichton:
Should this macro perhaps panic if any of the upper bits above `0xff` are set? (I think right now it silently masks them off)

japaric:
```
error: cannot find macro `panic!` in this scope
  --> crates/coresimd/src/../../../coresimd/arm/armclang.rs:43:5
   |
43 |     assert!(val >= 0 && val <= 255);
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Interesting.

Thinking more about it: Could we omit the assert! and document the behavior of `% 256`? I would prefer to avoid panicking branches in intrinsics as they can pull in the formatting machinery (few KBs) in unoptimized programs.

gnzlbg:
> Should this macro perhaps panic if any of the upper bits above 0xff are set? (I think right now it silently masks them off)

@alexcrichton it appears that it is not possible right now as @japaric mentioned. I've taken over this PR and just document that the only accepted values of `val` are in range `[0, 255]`, and that values out-of-range are undefined behavior. In the implementation, I've documented the behavior of the constify macro on out-of-range values but that's an implementation detail that we can change however we want.

PR #560, opened 2018-09-06
Rename `wasm32` memory intrinsics

alexcrichton:
The official name of the memory intrinsics has changed to `memory.size` and
`memory.grow`, so let's reflect that with our naming as well! Additionally they
have an argument of which memory to operate on with LLVM and must always be zero
currently.

### DIFF
--- a/coresimd/wasm32/mod.rs
+++ b/coresimd/wasm32/mod.rs
@@ -1,5 +1,6 @@
 //! WASM32 intrinsics
 
+#![allow(deprecated)]
 
 #[macro_use]
 #[cfg(all(not(test), feature = "wasm_simd128"))]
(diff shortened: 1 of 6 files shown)


coresimd/wasm32/mod.rs
```diff
@@ -15,35 +16,24 @@ use stdsimd_test::assert_instr;
...
-/// This function, when called, will return the current memory size in units of
-/// pages.
-///
-/// [instr]: https://github.com/WebAssembly/design/blob/master/Semantics.md#resizing
 #[inline]
 #[cfg_attr(test, assert_instr("memory.size"))]
+#[rustc_deprecated(reason = "renamed to memory::size", since = "1.30.0")]
+#[unstable(feature = "stdsimd", issue = "27731")]
+#[allow(deprecated)]
+#[doc(hidden)]
```
gnzlbg: We should open an issue to delete these by some date (e.g. halloween).

PR #561, opened 2018-09-06
Add wasm32 atomic intrinsics

alexcrichton:
Currently these are gated by the `atomics` feature unconditionally, but that may
be tweaked in the future! Otherwise this should enable building out some
primitives in the standard library using these intrinsics.

### DIFF
--- a/coresimd/wasm32/atomic.rs
+++ b/coresimd/wasm32/atomic.rs
@@ -1,0 +1,100 @@
+//! Intrinsics associated with WebAssembly's upcoming threads proposal.
+//!
+//! These intrinsics are all unstable because they're not actually stable in
+//! WebAssembly itself yet. The signatures may change as [the
+//! specification][spec] is updated.
+//!
+//! [spec]: https://github.com/WebAssembly/threads
+
+#![cfg(target_feature = "atomics")]
+
+#[cfg(test)]
+use stdsimd_test::assert_instr;
+#[cfg(test)]
+use wasm_bindgen_test::wasm_bindgen_test;
+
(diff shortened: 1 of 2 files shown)


coresimd/wasm32/atomic.rs
```diff
@@ -0,0 +1,100 @@
+//! Intrinsics associated with WebAssembly's upcoming threads proposal.
+//!
+//! These intrinsics are all unstable because they're not actually stable in
+//! WebAssembly itself yet. The signatures may change as [the
+//! specification][spec] is updated.
+//!
+//! [spec]: https://github.com/WebAssembly/threads
+
+#![cfg(target_feature = "atomics")]
```
gnzlbg: I think this should be in the parent hiding the module, and the re-export of these.

gnzlbg: Ah wait, I see you did the same in the memory module, do we want to export the modules themselves (`atomic`, `memory`, `simd128`, etc.) ?

alexcrichton: Sure yeah, do you mean just the attribute or the functions too?

On a possibly random tangent about function names...

I'm starting to get a little torn on the naming of the functions here. Technically the function name should be `wasm32::i32::atomic::wait` but that's a lot of `::`. Currently `wasm32::atomic::i32_wait` is sort of weirdly inconsistent. What do you think of `wasm32::i32_atomic_wait`? Basically translating all `.` to `_` and we have toplevel functions for all the operations? (this'd affect the SIMD proposal slightly)

PR #571, opened 2018-09-15
add the trap instrinsic

japaric:
as discussed in rust-lang/rfcs#2512

instead of the generic "trap" name the architecture specific name of the trap instruction has been used

### DIFF
--- a/coresimd/x86/mod.rs
+++ b/coresimd/x86/mod.rs
@@ -512,3 +512,13 @@ pub use self::rdrand::*;
 
 mod sha;
 pub use self::sha::*;
+
+#[cfg(test)]
+use stdsimd_test::assert_instr;
+
+/// Generates the trap instruction `UD2`
+#[cfg_attr(test, assert_instr(ud2))]
+#[inline]
+pub unsafe fn ud2() -> ! {
+    ::_core::intrinsics::abort()
+}
--- a/coresimd/aarch64/mod.rs
+++ b/coresimd/aarch64/mod.rs
@@ -14,3 +14,13 @@ pub use self::neon::*;
 
 mod crypto;
 pub use self::crypto::*;
+
+#[cfg(test)]
+use stdsimd_test::assert_instr;
+
+/// Generates the trap instruction `BRK 1`
+#[cfg_attr(test, assert_instr(brk))]
+#[inline]
+pub unsafe fn brk() -> ! {
+    ::_core::intrinsics::abort()
+}
--- a/coresimd/arm/mod.rs
+++ b/coresimd/arm/mod.rs
@@ -45,3 +45,14 @@ mod neon;
     dox
 ))]
 pub use self::neon::*;
+
+#[cfg(test)]
+use stdsimd_test::assert_instr;
+
+/// Generates the trap instruction `UDF`
+#[cfg(target_arch = "arm")]
+#[cfg_attr(test, assert_instr(udf))]
+#[inline]
+pub unsafe fn udf() -> ! {
+    ::_core::intrinsics::abort()
+}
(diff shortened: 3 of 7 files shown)


coresimd/x86/mod.rs
```diff
@@ -512,3 +512,13 @@ pub use self::rdrand::*;
...
 mod sha;
 pub use self::sha::*;
+
+#[cfg(test)]
+use stdsimd_test::assert_instr;
+
+/// Generates the trap instruction `UD2`
+#[cfg_attr(test, assert_instr(ud2))]
+#[inline]
+pub unsafe fn ud2() -> ! {
```
gnzlbg: The `stdsimd-verify` crate (https://github.com/rust-lang-nursery/stdsimd/tree/master/crates/stdsimd-verify) needs to be updated to support bottom.

gnzlbg: Here: https://github.com/rust-lang-nursery/stdsimd/blob/master/crates/stdsimd-verify/src/lib.rs#L127

alexcrichton:
This looks good to me!

For x86 I think this'll be initially unstable, right? If so, yay (can have a formal FCP shortly afterwards)

gnzlbg:
I've decided to take over this. I've rebased and hopefully fixed everything.

parched:
@gnzlbg I guess my concern is, for example on AArch64, there are now 2 `BRK` intrinsics (`std::arch::aarch64::brk` and `std::arch::aarch64::__breakpoint`), neither of which are standard vendor intrinsics in C. The "vendor intrinsic" for this in C is the portable `__builtin_trap`.

gnzlbg:
@parched 

Note that these APIs are unstable, and that we haven't signed off on their stabilization yet. The only thing I've signed on here is that they are good enough to be experimented on on nightly. We will need an RFC to stabilize these. 

> The "vendor intrinsic" for this in C is the portable __builtin_trap.

I do think that a portable `__builtin_trap` intrinsic is worth pursuing and recommend you to open a small issue about it in the RFC repo about it. After its discussed, I volunteer to implement it if nobody beats me to it. 

There were some unresolved questions in the previous issue about how portable can such an intrinsic actually be, and what to do in the platforms that might not support it. 

So the previous issues was mixing in my view two things:

* providing access to platform specific trap intrinsics / instructions
* providing a portable way to trap

Providing access to platform specific trap intrinsics and instructions was pretty uncontroversial, and its what's been merged here - this allows implementing a portable way to trap in an external crate (even though this might not be as good as `__builtin_trap`). 

Providing a portable way to trap had the unresolved questions mentioned above, and the discussion derailed a bit, which is why I think it is worth to have a focused discussion that's specific about that only.

I think that once we have implemented both alternatives on nightly, we will be in a much better place to consider whether we ought to expose both, even though that introduces some duplication, or whether a portable way to trap is enough, in which case we can just remove the platform specific intrinsics. Once we are there, it would be RFC time.

jethrogb:
When updating the rust submodule to the latest master, I get this error:
```
error[E0433]: failed to resolve. Maybe a missing `extern crate _core;`?
   --> libcore/../stdsimd/coresimd/x86/mod.rs:523:7
    |
523 |     ::_core::intrinsics::abort()
    |       ^^^^^ Maybe a missing `extern crate _core;`?
```

gnzlbg:
The paths added here might need some tuning to build upstream. Try the following:

* replace `::_core::intrinsics` with just intrinsics in all uses inside `coresimd`  (not only `x86`, from the looks, this will fail for all archs)
* add an appropriate import of the intrinsics module to `crates/coresimd/` 

You might need to tweak things a bit, so my recommendation is to just create a fork, change rust-lang/rust to use your fork, tweak things there until it builds properly, and then send a PR here to fix things. Let me know if you need help with this.

jethrogb:
Fixed in #599

PR #576, opened 2018-10-05
[breaking change] fixes unsoundness in avx non-temporal store APIs. 

### LINKED ISSUE #575: AVX stream intrinsics signatures are incorrect, maybe unsound
In the intel intrinsics guide, the signature is:

```c
void _mm256_stream_si256 (__m256i * mem_addr, __m256i a)
```

but in Rust the signature is:

```rust
pub unsafe fn _mm256_stream_si256(mem_addr: *const __m256i, a: __m256i)
```

Note that the `mem_addr` pointer is `const` in Rust, but not in C. Since the only purpose of this function is to write through this pointer, it is probably unsound for it to take a `const` pointer.

EDIT: `_mm256_stream_pd` and `_mm256_stream_ps`  also have this same issue. The SSE ` _mm_stream_ps`, `_mm_stream_pd`, and ` _mm_stream_pi` intrinsics correctly use `*mut` here.

A tangential bug is that the intrinsic verification failed hard here, since the C intrinsics do not use `const` pointers, it would have been safer to require these intrinsics to use `mut` pointers.

gnzlbg:
These APIs write through the pointer, so the pointer should be `*mut` and not `*const`.

Closes #575 .

### DIFF
--- a/coresimd/x86/avx.rs
+++ b/coresimd/x86/avx.rs
@@ -1941,8 +1941,8 @@ pub unsafe fn _mm256_lddqu_si256(mem_addr: *const __m256i) -> __m256i {
 #[target_feature(enable = "avx")]
 #[cfg_attr(test, assert_instr(vmovntps))] // FIXME vmovntdq
 #[stable(feature = "simd_x86", since = "1.27.0")]
-pub unsafe fn _mm256_stream_si256(mem_addr: *const __m256i, a: __m256i) {
-    intrinsics::nontemporal_store(mem::transmute(mem_addr), a);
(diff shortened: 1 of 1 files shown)


alexcrichton:
I've pushed this to https://github.com/rust-lang-nursery/stdsimd/tree/test-fix and yeah we'll need to send a PR to rust-lang/rust to run crater. This also looks good to me, thanks! Want to send a PR pointed to the branch at https://github.com/rust-lang-nursery/stdsimd/tree/test-fix to rust-lang/rust and we'll run crater?

alexcrichton:
Merging due to the lack of breakage evidence discovered in https://github.com/rust-lang/rust/pull/55538

alexcrichton:
er, the other button...

PR #578, opened 2018-10-07
Fix nightly build breakage

GabrielMajeri:
Building `coresimd` is broken on the latest nightlies, due to an unstable feature being renamed. This PR fixes that.


gnzlbg:
Thanks!

PR #580, opened 2018-10-09
Add wasm intrinsics to documentation

alexcrichton:


coresimd/wasm32/atomic.rs
```diff
@@ -43,6 +43,13 @@ extern "C" {
 ///   didn't block
 /// * 2 - the thread blocked, but the timeout expired.
 ///
+/// # Availability
```
gnzlbg: @QuietMisdreavus maybe we need some way to tell rustdoc about this, so that we can get a box similar to the ones we get for `target_feature` (https://doc.rust-lang.org/beta/core/arch/x86_64/fn._mm_rsqrt_ss.html), but for "cargo features".

QuietMisdreavus: Those boxes are given with `#[doc(cfg(...))]` attributes - however, it seems like in this case it's not just a "cargo feature" (e.g. `#[doc(cfg(feature = "atomics"))]`) but instead requires special setup thanks to being in the standard library. I don't think cargo features get special formatting like builtin cfgs, but that can be easily added to act similarly to target features. Using the attribute i suggested above will just show `feature = "atomics"` instead of any special text.

PR #581, opened 2018-10-23
fix _mm_castsi128_pd and _mm_castpd_si128 impls

kazcw:
The _mm_castX_Y SSE intrinsics are "reinterpreting" casts; LLVM's
simd_cast is a "converting" cast. Replace simd_casts with mem::transmute.
Fixes rust-lang/rust#55249

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -10,7 +10,7 @@ environment:
 install:
   # Install rust, x86_64-pc-windows-msvc host
   - appveyor-retry appveyor DownloadFile https://win.rustup.rs/ -FileName rustup-init.exe
-  - rustup-init.exe -y --default-host x86_64-pc-windows-msvc --default-toolchain nightly
+  - rustup-init.exe -y --default-host x86_64-pc-windows-msvc --default-toolchain nightly-2018-10-20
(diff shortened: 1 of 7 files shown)


alexcrichton:
Thanks for this! This looks correct to me, and I'm trying to dig into why CI is failing now.

alexcrichton:
Ok the segfault here [is not the fault of this PR](https://github.com/rust-lang/rust/pull/55281). I'm looking into pinning CI temoprarily to an older nightly to fix CI

alexcrichton:
@kazcw if you're extra enterprising, the next step here is to update the `src/stdsimd` submodule in rust-lang/rust, and that should be all that's necessary to get the fix into nightly!

kazcw:
@alexcrichton Cool! #55291

MaskRay:
>  `#` It's not clear why `-Z plt=yes` is required here. Probably a bug in LLVM. # If you can remove it and CI passes, please feel free to do so!`

Likely due to https://github.com/rust-lang/rust/pull/54592 switching the default.

If you use lld, it didn't support `-fno-plt` general-dynamic/local-dynamic TLS models until 2022-12-31 (https://github.com/llvm/llvm-project/commit/8dc73662ab5f7c992f9d7778de7c3611c1d5367e).

Issue #582, opened 2018-10-23
_mm_loadl_epi64 doesn't allow reads aligned to 8-byte boundaries


brian-armstrong:
`std::arch::x86_64::_mm_loadl_epi64` is kind of a weird case. Intel guides say it takes a `*const __m128i`, but the documentation is unclear on whether this needs to be aligned https://software.intel.com/en-us/node/524242 Note that `load` (must be aligned) and `loadu` (need not be aligned) are defined while `loadl` isn't. I believe the correct definition is that it need be aligned, but only to an 8-byte boundary (instead of the full 16) but I haven't been able to find documentation backing this up.

Clang's intrinsics header actually does go out of its way to allow this to be aligned to an 8-byte boundary, not a 16-byte boundary https://github.com/llvm-mirror/clang/blob/master/lib/Headers/emmintrin.h#L3587

This instruction produces the expected result on a 8- but not 16-byte aligned pointer in clang, but yields a segmentation fault on the same class of pointer in Rust.

gnzlbg:
You can ask Intel for clarification here: https://software.intel.com/en-us/forums/intel-isa-extensions/topic/363747

It might be interesting to survey what GCC, MSVC, and the Intel compiler do here. If they all support unaligned loads, then doing the same here is the right call.

brian-armstrong:
On a related note, `_mm_loadu_si64` appears to be missing.

gnzlbg:
Might be worth it to fill a separate issue to track that (or just send a PR).

brian-armstrong:
I went ahead and asked in that thread, and someone responded to point out that the manual says that in general, alignment isn't required unless otherwise specified. I'm not sure how much stock to put in that for this case, though.

I've been poking at godbolt to try to come up with a convincing argument about what C does one way or the other, but I'm finding the output is hard to reason about, and one _mm_loadl_epi64 seems to yield on average 5 instructions.

gnzlbg:
> I went ahead and asked in that thread, and someone responded to point out that the manual says that in general, alignment isn't required unless otherwise specified. I'm not sure how much stock to put in that for this case, though.

So IIUC, because `__m128` is a quadword, the pointer does not need to point to an aligned address ?

> I've been poking at godbolt to try to come up with a convincing argument about what C does one way or the other, but I'm finding the output is hard to reason about, and one _mm_loadl_epi64 seems to yield on average 5 instructions.

[...]

PR #584, opened 2018-10-27
allow unaligned reads with _mm_loadl_epi64

brian-armstrong:
As discussed in #582 it is "safe" to read unaligned memory locations with `_mm_loadl_epi64`. This code was suggested by scottmcm in the Rust Discord - I'm not familiar enough with this part of Rust to offer much commentary on it.

### DIFF
--- a/coresimd/x86/sse2.rs
+++ b/coresimd/x86/sse2.rs
@@ -1145,7 +1145,7 @@ pub unsafe fn _mm_setzero_si128() -> __m128i {
 )]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn _mm_loadl_epi64(mem_addr: *const __m128i) -> __m128i {
-    _mm_set_epi64x(0, simd_extract((*mem_addr).as_i64x2(), 0))
+    _mm_set_epi64x(0, simd_extract(ptr::read_unaligned::<__m128i >(mem_addr).as_i64x2(), 0))
 }
 
 /// Load 128-bits of integer data from memory into a new vector.


coresimd/x86/sse2.rs
```diff
@@ -1145,7 +1145,7 @@ pub unsafe fn _mm_setzero_si128() -> __m128i {
 )]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn _mm_loadl_epi64(mem_addr: *const __m128i) -> __m128i {
-    _mm_set_epi64x(0, simd_extract((*mem_addr).as_i64x2(), 0))
+    _mm_set_epi64x(0, simd_extract(ptr::read_unaligned::<__m128i >(mem_addr).as_i64x2(), 0))
```
gnzlbg: The type annotation here (`::<__m128i>`) shouldn't be necessary. 

Maybe one can just use `_mm_loadu_si128`  here ?

brian-armstrong: Depends - in my case, I need to do a shuffle afterwards anyway, so it comes out to be about the same. If you just wanted to load the lower 64 bits though, then adding in an unnecessary shuffle afterwards would likely cost some performance. At any rate, the fact that the instruction, when seemingly used correctly, yields a segfault is fairly surprising.

gnzlbg: The `ptr::read_unaligned` reads 128bits, and so does `_mm_loadu_si128`, or what am I missing?

brian-armstrong: Oh, yeah, you're right. I actually don't think I know how to write this PR, now that I think about it.

brian-armstrong:
So I cleaned up my example and got the right flags so that godbolt shows what I think is the right instructions. I'm not sure how to translate this into Rust though.

https://gcc.godbolt.org/z/EZv0qv

Notably, 
```
_mm_loadl_epi64(...);
_mm_storel_epi64(...);
```
becomes
```
        mov     rax, qword ptr [rdi + 24]
        mov     qword ptr [rdi + 40], rax
```
It's weird that it takes a qword ptr because I think it really only is loading a dword here, though I guess I have nothing to base that on.

coresimd/x86/sse2.rs
gnzlbg: You are doing great! I think you can replace the `ptr::read_unaligned(mem_addr)` with just `_mm_loadu_si128(mem_addr)` and that should just work.

gnzlbg:
> It's weird that it takes a qword ptr because I think it really only is loading a dword here, though I guess I have nothing to base that on.

So the Intrinsics Guide says these should lower to a `movq`, but these intrinsics are so simple, that the code that they will lower to will heavily depend on what the code at the call site around them is doing because LLVM can easily reason about them.

When we merge this, you can try them in your app, and if you see unexpected code generation we can revise their implementation.

brian-armstrong:
So I'm looking at the trace in my actual program, and I believe the C version which uses `_mm_loadl_epi64` is emitting movq while Rust (with this patch) is emitting `vmovqdu`. It's unclear if this is because llvm is unable to optimize the Rust version for some reason or if the C version has just correctly determined that only a quadword is needed, not a double quadword.

brian-armstrong:
I wonder if the right code here isn't just
```
    _mm_set_epi64x(0, ptr::read_unaligned(mem_addr as *const i64))
```

gnzlbg:
That could work too. Could you modify your stdsimd copy locally, and use it in your project, and report whether that works better?

brian-armstrong:
Yes, it does emit `movq`. It's hard to say whether that's generally faster, but it does seem closer to the original spirit of the intrinsic.

PR #586, opened 2018-11-03
[breaking change]: mask argument type of _mm_shuffle_ps

gnzlbg:
**DO NOT MERGE**

If CI passes, I'll push this as a branch, and send a PR to rust-lang/rust to schedule a crater run. Depending on how much code this change breaks, we can re-evaluate what to do with issue #522 .

### DIFF
--- a/crates/core_arch/src/x86/sse.rs
+++ b/crates/core_arch/src/x86/sse.rs
@@ -995,7 +995,7 @@ pub const fn _MM_SHUFFLE(z: u32, y: u32, x: u32, w: u32) -> i32 {
 #[cfg_attr(test, assert_instr(shufps, mask = 3))]
 #[rustc_args_required_const(2)]
 #[stable(feature = "simd_x86", since = "1.27.0")]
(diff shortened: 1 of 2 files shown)


gnzlbg:
@alexcrichton so CI passed, shall I send a PR to rust-lang/rust to schedule a crater run ?

alexcrichton:
@gnzlbg sure!

gnzlbg:
I think we should just merge this and deal with the fallout. cc @rust-lang-nursery/libs @Amanieu

alexcrichton:
I believe the crater fun from 6ish months ago was clean right? I forget now the original reason for closing, but I'm personally still in favor for this

Amanieu:
I think this is fine, we are just correcting an "incorrect" function signature and there are no crater failures.

gnzlbg:
I'll send a PR to update stdsimd upstream, and will do a new crater run before the update. There is another breaking change in the pipeline, so I'll try to merge both. If the crater run fails we can always revert this.

PR #589, opened 2018-11-06
Rename `llvm.cuda.syncthreads` -> `llvm.nvvm.barrier0`

peterhj:
Rename the NVPTX related intrinsic `llvm.cuda.syncthreads` -> `llvm.nvvm.barrier0` to keep up with changes in LLVM.

This should provide a workaround for https://github.com/rust-lang/rust/issues/54115, although there is also an analogous NVPTX related platform-intrinsic that still references the renamed `llvm.cuda.syncthreads` intrinsic.

Thanks to @rkruppe for pointing me here from the discord.

### DIFF
--- a/coresimd/nvptx/mod.rs
+++ b/coresimd/nvptx/mod.rs
@@ -13,7 +13,7 @@
 
 #[allow(improper_ctypes)]
 extern "C" {
-    #[link_name = "llvm.cuda.syncthreads"]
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

gnzlbg:
I wanted to get the nvptx target running in CI...

PR #590, opened 2018-11-09
update authors and version

gnzlbg:
The next version to be released on crates.io will be 0.1.3 (0.1.2 is the last version released).


gnzlbg:
It seems that windows is failing due to different instructions than the expected being generated.

PR #598, opened 2018-11-19
Add SGX target

jethrogb:
This adds support for the `x86_64-fortanix-unknown-sgx` target. See https://github.com/rust-lang/rust/pull/56066 for details.

CPUID is not supported inside SGX. In the future, I might want to add a way to supply CPUID information from the outside, but that can come later.


gnzlbg:
Is there a way to test SGX on travis-ci ?

gnzlbg:
Ideally, before updating `stdsimd` upstream we should add a CI build bot here that at least tests that this builds properly. We could use `xargo` for that. Once the tier-3 target is merged upstream, we should be able to remove that from here.

PR #599, opened 2018-11-19
Fix build when used as a submodule

jethrogb:
I tested this with a local checkout of https://github.com/rust-lang/rust/

Ideally CI would update the submodule and run `cargo +nightly check` inside `src/libstd`.

### DIFF
--- a/coresimd/aarch64/mod.rs
+++ b/coresimd/aarch64/mod.rs
@@ -22,5 +22,5 @@ use stdsimd_test::assert_instr;
 #[cfg_attr(test, assert_instr(brk))]
 #[inline]
 pub unsafe fn brk() -> ! {
-    ::_core::intrinsics::abort()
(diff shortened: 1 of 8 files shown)


jethrogb:
Please don't merge yet, I just found another issue

gnzlbg:
Ping me when I should merge. I'm going to try to add a test for updating the submodule to stdsimd's CI.

jethrogb:
Take 2

jethrogb:
Build failures: ARM & Windows is not this PR's fault, I think. I fixed rustfmt.

gnzlbg:
That windows job is broken on master as well, so don't worry about it, I'll take a look later. `rustfmt` still appears to be broken though, you might need to use `cargo fmt --all` to cover all crates.

jethrogb:
Ok one more time

PR #604, opened 2018-11-27
Revert #558. Fixes #603.

### LINKED ISSUE #603: Building as rust submodule is broken
See https://travis-ci.org/rust-lang/rust/jobs/458312853 and https://github.com/rust-lang/rust/pull/56178

Probably caused by @japaric's changes in #558.

jethrogb:
Revert "do not validate the argument to the __breakpoint intrinsic", "aarch64 support and range validation", "fix typo", "fix assert_instr test", "remove TODO; add link to doc comment", "add __breakpoint"

This reverts commits 7fbd636c3d768135e05641afd47d50dde98b262a, 72b41437d436e438dc0be3b9f701f5541daf0056, 9e5c960774ab828d753d0a43216e0a79dfdc8349, 8b84dc4031419c335d3f7fd98366c52220b9685a, af7c166fb8e274b02e5b4f62186703638fa61998, 40f61c77bc949d0a31b3420e456983185952577f.


jethrogb:
i586/i686 tests also fail on master: https://travis-ci.org/rust-lang-nursery/stdsimd/builds/460124295

gnzlbg:
We should fix this, not revert it. The question is: why does this break on rust-lang/rust but works fine here ?

gnzlbg:
IIRC the problem was that the `constify!` macro was not found on `armclang`, but the `macros` module is always included at the top of the library, and that happens before `armclang` is included.

alexcrichton:
Agreed yeah, would it be possible to get a gist of the error logs to see if we can fix it without reverting?

jethrogb:
See #603

alexcrichton:
That's fixed in https://github.com/rust-lang-nursery/stdsimd/pull/609

PR #605, opened 2018-11-27
Run CI on master

jethrogb:


gnzlbg:
@alexcrichton the errors here look like a regression, I'm not really able to reproduce this codegen (a sequence of two `shufps`) with `llc` trunk: https://rust.godbolt.org/z/Hk3MVx

PR #611, opened 2018-12-08
Add support for feature detection on FreeBSD/aarch64 (WIP)

myfreeweb:
On AArch64, FreeBSD does not currently provide `HWCAP` in the ELF auxiliary vector.
Reads of the ID registers are trapped though, and they are accessible in userspace.

- The register reading code could work on other OSes and even bare metal, so it's separated out into its own module.
- Seems like it could also be used on Linux.
- ~~TODO: The OS-dependent code should trap `SIGILL` before calling the reading function, in order to work on older OS versions that didn't expose the registers.~~  I guess it's fine™ to expect that to work.
- ~~TODO: The OS-dependent code should run the reading function on all cores (via thread affinity) and take a union of features, because there are [weird SoCs where cores have unequal features](https://medium.com/@jadr2ddude/a-big-little-problem-a-tale-of-big-little-gone-wrong-e7778ce744bb).~~ [On FreeBSD, the register view is the same on each core](https://reviews.freebsd.org/D17137#393947)

---
References:

- [Linux kernel docs: ARM64 CPU Feature Registers](https://www.kernel.org/doc/Documentation/arm64/cpu-feature-registers.txt)
- [FreeBSD r337739: Support reading from the arm64 ID registers from userspace.](https://reviews.freebsd.org/rS337739)
- [FreeBSD D17137: arm64: Add HWCAP support](https://reviews.freebsd.org/D17137)

### DIFF
--- a/stdsimd/arch/detect/mod.rs
+++ b/stdsimd/arch/detect/mod.rs
@@ -71,6 +71,12 @@ cfg_if! {
     } else if #[cfg(target_os = "linux")] {
         #[path = "os/linux/mod.rs"]
         mod os;
+    } else if #[cfg(target_os = "freebsd")] {
+        #[cfg(target_arch = "aarch64")]
+        #[path = "os/aarch64.rs"]
+        mod aarch64;
+        #[path = "os/freebsd/mod.rs"]
+        mod os;
     } else {
         #[path = "os/other.rs"]
         mod os;
--- a/stdsimd/arch/detect/os/aarch64.rs
+++ b/stdsimd/arch/detect/os/aarch64.rs
@@ -1,0 +1,80 @@
+//! Run-time feature detection for Aarch64 on any OS that emulates the mrs instruction.
+//!
+//! On FreeBSD >= 12.0, Linux >= 4.11 and other operating systems, it is possible to use
+//! privileged system registers from userspace to check CPU feature support.
+//!
+//! AArch64 system registers ID_AA64ISAR0_EL1, ID_AA64PFR0_EL1, ID_AA64ISAR1_EL1
+//! have bits dedicated to features like AdvSIMD, CRC32, AES, atomics (LSE), etc.
+//! Each part of the register indicates the level of support for a certain feature, e.g.
+//! when ID_AA64ISAR0_EL1[7:4] is >= 1, AES is supported; when it's >= 2, PMULL is supported.
+//!
+//! For proper support of [SoCs where different cores have different capabilities](https://medium.com/@jadr2ddude/a-big-little-problem-a-tale-of-big-little-gone-wrong-e7778ce744bb),
+//! the OS has to always report only the features supported by all cores, like [FreeBSD does](https://reviews.freebsd.org/D17137#393947).
+//!
+//! References:
+//!
+//! - [Zircon implementation](https://fuchsia.googlesource.com/zircon/+/master/kernel/arch/arm64/feature.cpp)
+//! - [Linux documentation](https://www.kernel.org/doc/Documentation/arm64/cpu-feature-registers.txt)
+
+use arch::detect::Feature;
+use arch::detect::cache;
+
+/// Try to read the features from the system registers.
+///
+/// This will cause SIGILL if the current OS is not trapping the mrs instruction.
+pub(crate) fn detect_features() -> cache::Initializer {
+    let mut value = cache::Initializer::default();
+
+    {
+        let mut enable_feature = |f, enable| {
+            if enable {
+                value.set(f as u32);
+            }
+        };
+
+        // ID_AA64ISAR0_EL1 - Instruction Set Attribute Register 0
+        let aa64isar0: u64;
+        unsafe { asm!("mrs $0, ID_AA64ISAR0_EL1" : "=r"(aa64isar0)); }
+
+        let aes = bits_shift(aa64isar0, 7, 4) >= 1;
+        let pmull = bits_shift(aa64isar0, 7, 4) >= 2;
+        let sha1 = bits_shift(aa64isar0, 11, 8) >= 1;
+        let sha2 = bits_shift(aa64isar0, 15, 12) >= 1;
+        enable_feature(Feature::pmull, pmull);
+        // Crypto is specified as AES + PMULL + SHA1 + SHA2 per LLVM/hosts.cpp
+        enable_feature(Feature::crypto, aes && pmull && sha1 && sha2);
+        enable_feature(Feature::lse, bits_shift(aa64isar0, 23, 20) >= 1);
+        enable_feature(Feature::crc, bits_shift(aa64isar0, 19, 16) >= 1);
+
+        // ID_AA64PFR0_EL1 - Processor Feature Register 0
+        let aa64pfr0: u64;
+        unsafe { asm!("mrs $0, ID_AA64PFR0_EL1" : "=r"(aa64pfr0)); }
+
+        let fp = bits_shift(aa64pfr0, 19, 16) < 0xF;
+        let fphp = bits_shift(aa64pfr0, 19, 16) >= 1;
+        let asimd = bits_shift(aa64pfr0, 23, 20) < 0xF;
+        let asimdhp = bits_shift(aa64pfr0, 23, 20) >= 1;
+        enable_feature(Feature::fp, fp);
+        enable_feature(Feature::fp16, fphp);
+        // SIMD support requires float support - if half-floats are
+        // supported, it also requires half-float support:
+        enable_feature(Feature::asimd, fp && asimd && (!fphp | asimdhp));
+        // SIMD extensions require SIMD support:
+        enable_feature(Feature::rdm, asimd && bits_shift(aa64isar0, 31, 28) >= 1);
+        enable_feature(Feature::dotprod, asimd && bits_shift(aa64isar0, 47, 44) >= 1);
+        enable_feature(Feature::sve, asimd && bits_shift(aa64pfr0, 35, 32) >= 1);
+
+        // ID_AA64ISAR1_EL1 - Instruction Set Attribute Register 1
+        let aa64isar1: u64;
+        unsafe { asm!("mrs $0, ID_AA64ISAR1_EL1" : "=r"(aa64isar1)); }
+
+        enable_feature(Feature::rcpc, bits_shift(aa64isar1, 23, 20) >= 1);
+    }
+
+    value
+}
+
+#[inline]
+fn bits_shift(x: u64, high: usize, low: usize) -> u64 {
+    (x >> low) & ((1 << (high - low + 1)) - 1)
+}
--- a/crates/stdsimd/src/lib.rs
+++ b/crates/stdsimd/src/lib.rs
@@ -11,6 +11,7 @@
     deny(clippy::missing_inline_in_public_items,)
 )]
 #![cfg_attr(target_os = "linux", feature(linkage))]
+#![cfg_attr(all(target_os = "freebsd", target_arch = "aarch64"), feature(asm))]
 #![no_std]
 #![unstable(feature = "stdsimd", issue = "27731")]
 
--- a/stdsimd/arch/detect/os/freebsd/aarch64.rs
+++ b/stdsimd/arch/detect/os/freebsd/aarch64.rs
@@ -1,0 +1,29 @@
+//! Run-time feature detection for Aarch64 on FreeBSD.
+
+use arch::detect::Feature;
+use arch::detect::cache;
+use super::super::aarch64::detect_features;
+
+/// Performs run-time feature detection.
+#[inline]
+pub fn check_for(x: Feature) -> bool {
+    cache::test(x as u32, detect_features)
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn dump() {
+        println!("asimd: {:?}", is_aarch64_feature_detected!("asimd"));
+        println!("pmull: {:?}", is_aarch64_feature_detected!("pmull"));
+        println!("fp: {:?}", is_aarch64_feature_detected!("fp"));
+        println!("fp16: {:?}", is_aarch64_feature_detected!("fp16"));
+        println!("sve: {:?}", is_aarch64_feature_detected!("sve"));
+        println!("crc: {:?}", is_aarch64_feature_detected!("crc"));
+        println!("crypto: {:?}", is_aarch64_feature_detected!("crypto"));
+        println!("lse: {:?}", is_aarch64_feature_detected!("lse"));
+        println!("rdm: {:?}", is_aarch64_feature_detected!("rdm"));
+        println!("rcpc: {:?}", is_aarch64_feature_detected!("rcpc"));
+        println!("dotprod: {:?}", is_aarch64_feature_detected!("dotprod"));
+    }
+}
--- a/stdsimd/arch/detect/os/freebsd/mod.rs
+++ b/stdsimd/arch/detect/os/freebsd/mod.rs
@@ -1,0 +1,14 @@
+//! Run-time feature detection on FreeBSD
+
+cfg_if! {
+    if #[cfg(target_arch = "aarch64")] {
+        mod aarch64;
+        pub use self::aarch64::check_for;
+    } else {
+        use arch::detect::Feature;
+        /// Performs run-time feature detection.
+        pub fn check_for(_x: Feature) -> bool {
+            false
+        }
+    }
+}


lu-zero:
How old are we talking regarding old oses? Trapping signals is _quite_ bad for libraries.

myfreeweb:
@lu-zero well, older than FreeBSD 12 (which is currently in release candidate phase). The number of people who use Rust + FreeBSD + AArch64 is currently quite small, especially considering that rustc currently doesn't build on FreeBSD/aarch64 without [a very patch that hasn't landed yet](https://reviews.freebsd.org/D18417) :) So it should be fine to ignore that.

stdsimd/arch/detect/os/aarch64.rs
```diff
@@ -0,0 +1,64 @@
...
+        unsafe { asm!("mrs $0, S3_0_C0_C6_0" : "=r"(aa64isar0)); }
+
+        let aes = bits_shift(aa64isar0, 7, 4) >= 1;
+        let pmull = bits_shift(aa64isar0, 7, 4) >= 2;
+        let sha1 = bits_shift(aa64isar0, 11, 8) >= 1;
+        let sha2 = bits_shift(aa64isar0, 15, 12) >= 1;
+        enable_feature(Feature::pmull, pmull);
+        // Crypto is specified as AES + PMULL + SHA1 + SHA2 per LLVM/hosts.cpp
+        enable_feature(Feature::crypto, aes && pmull && sha1 && sha2);
+        enable_feature(Feature::lse, bits_shift(aa64isar0, 23, 20) >= 1);
```
alexcrichton: According to [this page](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.100458_0200_00_en/lau1442496521966.html) the bits here correspond to the "atomic" feature, although I'm pretty unfamiliar with AArch64 feature names so `lse` may actually be that! Just wanted to confirm

stdsimd/arch/detect/os/aarch64.rs
```diff
@@ -0,0 +1,64 @@
...
+        enable_feature(Feature::crc, bits_shift(aa64isar0, 19, 16) >= 1);
```
alexcrichton: This says >= 1 but [these docs](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.100458_0200_00_en/lau1442496521966.html) indicate the value may need to be exactly one?

stdsimd/arch/detect/os/aarch64.rs
```diff
@@ -0,0 +1,64 @@
...
+                value.set(f as u32);
+            }
+        };
+
+        // ID_AA64ISAR0_EL1 - Instruction Set Attribute Register 0
+        let aa64isar0: u64;
+        unsafe { asm!("mrs $0, S3_0_C0_C6_0" : "=r"(aa64isar0)); }
+
+        let aes = bits_shift(aa64isar0, 7, 4) >= 1;
+        let pmull = bits_shift(aa64isar0, 7, 4) >= 2;
```
alexcrichton: According to [arm docs](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.100458_0200_00_en/lau1442496521966.html) should this only check for 0x2? 

Additionally, I wonder if this plus `crypto` below means that we should split up the `crypto` feature if AArch64 actually has multiple features here... (not necessary to do in this PR)

stdsimd/arch/detect/os/aarch64.rs
```diff
@@ -0,0 +1,64 @@
...
+        let fphp = bits_shift(aa64pfr0, 19, 16) >= 1;
+        let asimd = bits_shift(aa64pfr0, 23, 20) < 0xF;
+        let asimdhp = bits_shift(aa64pfr0, 23, 20) >= 1;
+        enable_feature(Feature::fp, fp);
+        enable_feature(Feature::fp16, fphp);
+        // SIMD support requires float support - if half-floats are
+        // supported, it also requires half-float support:
+        enable_feature(Feature::asimd, fp && asimd && (!fphp | asimdhp));
+        // SIMD extensions require SIMD support:
+        enable_feature(Feature::rdm, asimd && bits_shift(aa64isar0, 31, 28) >= 1);
```
alexcrichton: Hm on [these docs](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.100403_0200_00_en/lau1443437525908.html) this is called  the RAS feature...

gnzlbg:
I’d like to review this with calm tomorrow afternoon if it isn’t very
pressing.

w.r.t. signal-handling-based detection, there were some intrinsic issues
with this (e.g. how reliable is it that the proper exception will be raised
on a CPU that “doesn’t know” the instruction being executed?). However, in
some OSes the alternatives aren’t clear (e.g. windows), as long as “it
works for you” and doesn’t break anything I think it is something worth
getting experience with.

Would it be possible to get CI set up for aarch64-FreeBSD?
-- 
Dipl.-Ing. Gonzalo Brito Gadeschi
Institute of Aerodynamics and Chair of Fluid Mechanics
RWTH Aachen University
Wuellnerstraße 5a
D-52062 Aachen
Germany
Phone: ++49-(0)241-80-94821
Fax: ++49-(0)241-80-92257
E-mail:  g.brito@aia.rwth-aachen.de
Internet: www.aia.rwth-aachen.de

stdsimd/arch/detect/os/aarch64.rs
myfreeweb: `lse` is `atomic`, look at the comments in the `Feature` struct and the Linux implementation

stdsimd/arch/detect/os/aarch64.rs
myfreeweb: With these registers, the semantics is: higher number == higher level of support of this feature (which includes all previous ones). So e.g. if there's a next level of CRC32 implemented in the future, that CPU will have this value as 2.

e.g. Zircon implements checks as `>=`: https://fuchsia.googlesource.com/zircon/+/master/kernel/arch/arm64/feature.cpp#204

stdsimd/arch/detect/os/aarch64.rs
myfreeweb: (see above)

This is a better example actually: this is the same bits of the same register. `>= 1` means AES is supported, `>= 2` means PMULL is supported.

I like the idea of splitting `crypto`. Looks like the idea here was to align to LLVM's target features, but it's okay to e.g. check for AES only when writing an AES function that would use the `crypto` target feature

myfreeweb:
> I’d like to review this with calm tomorrow afternoon

Of course :)

> a CPU that “doesn’t know” the instruction being executed

Every aarch64 CPU knows `msr` with these arguments, they're just privileged instructions, and most (?) OSes trap them to allow them to succeed in userspace.

I can see how trying to detect the SIMD/etc instructions themselves via signal handling can be unreliable, yes.

>  Would it be possible to get CI set up for aarch64-FreeBSD?

I have a public buildbot..

stdsimd/arch/detect/os/aarch64.rs
myfreeweb: That link is to `PFR0`, not `ISAR0`

myfreeweb:
Update: no need to read the values on different cores, the kernel already presents the same view of the registers on all cores https://reviews.freebsd.org/D17137#393947 So this is good as is, pretty much. I can add some code comments about this I guess.

stdsimd/arch/detect/os/aarch64.rs
alexcrichton: Ah ok, good to know! Mind adding a comment somewhere in this file about that? (doesn't have to be on every comparison)

stdsimd/arch/detect/os/aarch64.rs
alexcrichton: Ok cool, thanks for the info!

stdsimd/arch/detect/os/aarch64.rs
alexcrichton: Ok cool, let's move [discussion for `crypto` to an issue](https://github.com/rust-lang-nursery/stdsimd/issues/614)

stdsimd/arch/detect/os/aarch64.rs
alexcrichton: Oh sorry I got mixed up with the blocks here, makes sense

alexcrichton:
FWIW I personally think that we shouldn't try to trap instructions here, if FreeBSD handles this and "makes the instruction work" that's fine by me and if we're not compatible with older versions that's probably ok for now too

stdsimd/arch/detect/mod.rs
```diff
@@ -63,6 +63,13 @@ pub use self::arch::Feature;
 mod bit;
 mod cache;
 
+cfg_if! {
+    if #[cfg(all(target_arch = "aarch64", any(target_os = "freebsd")))] {
```
gnzlbg: Why is the `any(target_os = "freebsd")` needed here?  (shouldn't it just be `target_os = "freebsd"` ?)

stdsimd/arch/detect/os/aarch64.rs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        let mut enable_feature = |f, enable| {
+            if enable {
+                value.set(f as u32);
+            }
+        };
+
+        // ID_AA64ISAR0_EL1 - Instruction Set Attribute Register 0
+        let aa64isar0: u64;
+        unsafe { asm!("mrs $0, S3_0_C0_C6_0" : "=r"(aa64isar0)); }
```
gnzlbg: @japaric do we have an intrinsic for this or is one planned ?

stdsimd/arch/detect/os/aarch64.rs
gnzlbg: This looks resolved to me (splitting `crypto` is an orthogonal issue).

stdsimd/arch/detect/os/aarch64.rs
gnzlbg: It would be good if the top level comments could contain links to zircon and every other place that was relevant when coming up with the implementation.

stdsimd/arch/detect/os/aarch64.rs
```diff
@@ -0,0 +1,64 @@
...
+        // supported, it also requires half-float support:
+        enable_feature(Feature::asimd, fp && asimd && (!fphp | asimdhp));
+        // SIMD extensions require SIMD support:
+        enable_feature(Feature::rdm, asimd && bits_shift(aa64isar0, 31, 28) >= 1);
+        enable_feature(Feature::dotprod, asimd && bits_shift(aa64isar0, 47, 44) >= 1);
+        enable_feature(Feature::sve, asimd && bits_shift(aa64pfr0, 35, 32) >= 1);
+
+        // ID_AA64ISAR1_EL1 - Instruction Set Attribute Register 1
+        let aa64isar1: u64;
+        unsafe { asm!("mrs $0, S3_0_C0_C6_1" : "=r"(aa64isar1)); }
```
gnzlbg: What's the "reference" for the contents of `msr` ? That is, how should it be queried, what do the bits in the output registers mean, etc. ? It might be worth it to link to the reference in the top-level comment of the module and minimally explain how this works (not that it is hard to follow, but it'll make lives easier of those having to update this in the future).

stdsimd/arch/detect/os/aarch64.rs
```diff
@@ -0,0 +1,64 @@
+//! Run-time feature detection for Aarch64 on any OS that traps the mrs instruction.
```
parched: Perhaps this file could go coresimd as it would be useful for bare metal too?

stdsimd/arch/detect/os/aarch64.rs
parched: @gnzlbg That's the `rsr64` intrinsic added in the other open PR but it doesn't this form yet.

@myfreeweb Probably better to use `ID_AA64ISAR0_EL1` instead of  `S3_0_C0_C6_0` in the `asm!` to make it more readable/easier to verify.

stdsimd/arch/detect/os/aarch64.rs
```diff
@@ -0,0 +1,64 @@
...
+
+        let fp = bits_shift(aa64pfr0, 19, 16) < 0xF;
+        let fphp = bits_shift(aa64pfr0, 19, 16) >= 1;
+        let asimd = bits_shift(aa64pfr0, 23, 20) < 0xF;
+        let asimdhp = bits_shift(aa64pfr0, 23, 20) >= 1;
+        enable_feature(Feature::fp, fp);
+        enable_feature(Feature::fp16, fphp);
+        // SIMD support requires float support - if half-floats are
+        // supported, it also requires half-float support:
+        enable_feature(Feature::asimd, fp && asimd && (!fphp | asimdhp));
```
parched: I think it would be better to just have `asimd` as the second arg here. Then add an assert for `fp == asimd` and `fphp == asimdhp` as these are required by the architecture.

gnzlbg:
> The register reading code could work on other OSes and even bare metal, so it's separated out into its own module.

So I think this is a good idea. We do the same for `x86` where we have it as its own os, I think it is worth it to do this for `aarch64` as well.  Since this is part of `libstd` , it is unlikely that people will use it directly on bare metal in the near future, but there is some interest to do this with `x86` as well and there are some ways to do that (e.g. split those into different `no_std` crates that people can include without pulling in std, portability check, etc.).

> Seems like it could also be used on Linux.

IIRC `msr` was a privileged instruction on Linux, but maybe I remember wrong?

> TODO: The OS-dependent code should run the reading function on all cores (via thread affinity) and take a union of features, because there are weird SoCs where cores have unequal features. On FreeBSD, the register view is the same on each core

`stdsimd` is used in `libstd`, so it requires OS support, we do have some workarounds over some specific known CPU bugs, but we don't have any workarounds that assume an "adversary" OS that tries to break your process. 

On Linux, we ask the OS "What is this process allowed to do?" via the OS interfaces (e.g. auxiliary vectors). If Linux decides to migrate the process to a different core, it has to make sure that the core is at least as capable as the one we are running on because we can cache the answer to the question. If the FreeBSD way of querying this information is via the `msr` register, then they have to make sure about that too. 

When the OS fails, all bets are off anyways. For example, see [A tale of big.LITTLE gone wrong](https://medium.com/@jadr2ddude/a-big-little-problem-a-tale-of-big-little-gone-wrong-e7778ce744bb), where Samsung Exynos aarch64 LITTLE cores had atomic support but the big cores did not (usually big cores are "better" than little cores, and the OS assumed that all SoCs were like this).  So the Apps would start in LITTLE cores, do run-time feature detection, detect that they can use atomics, cache that, start using them, and when some time the OS moved them to a big core, then they would SIGILL because the big cores did not support atomics. Working around "bugs" like these here feels like trying to do too much: we'd need OS APIs that allow us to query the set of potential cores the process might be moved to, and APIs to query the features of these cores.

> TODO: The OS-dependent code should trap SIGILL before calling the reading function, in order to work on older OS versions that didn't expose the registers. I guess it's fine™ to expect that to work.

What's the recommended FreeBSD way of handling this ? We don't support Windows+aarch64 but the recommended approach to do run-time feature detection there is to "try to use an intrinsic, and use SEH to handle the error if its not available" (see https://github.com/rust-lang-nursery/stdsimd/issues/120#issuecomment-339818841) . I agree with @lu-zero and @alexcrichton that we should avoid this if possible.

stdsimd/arch/detect/os/aarch64.rs
gnzlbg: @parched we don't really have a mechanism to expose this from `core::arch` yet (and it is unclear how to do that). I'd leave doing that to "future work".

stdsimd/arch/detect/os/aarch64.rs
gnzlbg: This is how we do it on Linux: https://github.com/rust-lang-nursery/stdsimd/blob/master/stdsimd/arch/detect/os/linux/aarch64.rs#L148

```rust
            let asimd = self.fp && self.asimd && (!self.fphp | self.asimdhp);
            enable_feature(Feature::asimd, asimd);
```

Maybe it would be worth it to somehow keep both implementations in sync.

parched: Ah ok hadn't seen that, yeah, best to keep them in sync.

parched:
>> Seems like it could also be used on Linux.
>>
>IIRC msr was a privileged instruction on Linux, but maybe I remember wrong?

Yes this works on Linux now too, since [version 4.11](https://github.com/torvalds/linux/commit/4aa8a472c33f4024381ade56bef008a9c566366a#diff-9764502b1d5418a9c00b3b235fd51bdc). 

Whether MRS is privileged depends on what register is being read, i.e., only ones ending with '_EL0' can possibly be read directly from userspace. EL1 ones like these cause an exception and must be emulated by the kernel if it wants (normally it will abort the process with SIGILL or similar).

gnzlbg:
> Yes this works on Linux now too, since version 4.11.

Cool! We should explore using this in Linux then, it is way easier than using auxiliary vectors.

stdsimd/arch/detect/mod.rs
myfreeweb: A subtle hint for potential future reuse of this module by other OSes :) I guess a comment would be better.

myfreeweb:
> If the FreeBSD way of querying this information is via the msr register, then they have to make sure about that too

Yep, the emulated msr only returns features supported by all cores. (Looks like you replied before I updated the original post.)

> see A tale of big.LITTLE gone wrong

That's where I linked with the "weird SoCs where cores have unequal features" text :)

> What's the recommended FreeBSD way of handling this ?

Similar to Windows, I guess. For an API that's literally "run a privileged instruction and expect it to be emulated", not much you can do other than trapping SIGILL.

But we can just ignore this. The FreeBSD/aarch64 userbase is small enough that "just upgrade your OS to >=12" is fine :)

stdsimd/arch/detect/mod.rs
gnzlbg: Note that because we use `#[path]` here any os wanting to use the `aarch64` module can just use it by including it directly below - so there is no need to split the per-OS logic into two `cfg_if!`s.

stdsimd/arch/detect/os/aarch64.rs
gnzlbg: @myfreeweb see @parched suggestion to use `ID_AA64ISAR0_EL1` instead of `S3_0_C0_C6_0`.

stdsimd/arch/detect/os/aarch64.rs
gnzlbg: You have some links in the PR message, please consider adding all that information to the top of the module to avoid people from having to find the commit / PR to get it.

stdsimd/arch/detect/os/aarch64.rs
gnzlbg: The links are in the PR, and there is already another comment suggesting adding these links to the module, so I am marking this as resolved.

gnzlbg:
> But we can just ignore this. The FreeBSD/aarch64 userbase is small enough that "just upgrade your OS to >=12" is fine :)

We can leave adding support for FreeBSD == 11 to "future work". If someone complains, I'd guess we can just disable this with a config flag or by querying the FreeBSD version somehow later. 

This is looking good, just a couple of nits open (don't forget to run `cargo fmt --all` since that is the only CI failure)!

myfreeweb:
Updated with comments and suggested `cfg`.

> cargo fmt --all

The CI rustfmt failure is:

```
error: component 'rustfmt' for target 'x86_64-unknown-linux-gnu' is unavailable for download
```

gnzlbg:
Thank you a lot @myfreeweb !

PR #613, opened 2018-12-09
Stabilize wasm32 memory-related intrinsics

alexcrichton:
This commit stabilizes the wasm32 memory-related intrinsics, as
specified in rust-lang/rust#56292. The old intrinsics were removed and
the current intrinsics were updated in place, but it's the last breaking
change!

### DIFF
--- a/coresimd/wasm32/memory.rs
+++ b/coresimd/wasm32/memory.rs
@@ -13,43 +13,52 @@ extern "C" {
 /// Corresponding intrinsic to wasm's [`memory.size` instruction][instr]
 ///
 /// This function, when called, will return the current memory size in units of
-/// pages.
+/// pages. The current WebAssembly page size is 65536 bytes (64 KB).
 ///
 /// The argument `mem` is the numerical index of which memory to return the
-/// size of. Note that currently wasm only supports one memory, so specifying
-/// a nonzero value will likely result in a runtime validation error of the
-/// wasm module.
+/// size of. Note that currently the WebAssembly specification only supports one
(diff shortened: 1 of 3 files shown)


coresimd/wasm32/memory.rs
```diff
@@ -47,9 +48,10 @@ pub unsafe fn size(mem: i32) -> i32 {
 #[inline]
 #[cfg_attr(test, assert_instr("memory.grow", mem = 0))]
 #[rustc_args_required_const(0)]
-pub unsafe fn grow(mem: i32, delta: i32) -> i32 {
+#[stable(feature = "simd_wasm32", since = "1.33.0")]
+pub unsafe fn memory_grow(mem: u32, delta: usize) -> isize {
```
sunfishcode: `memory.grow` conceptually returns the value that `memory.size` would return after the resize, which suggests that it should be usize rather than isize (with the error value being `usize::MAX` rather than `-1`). However, I don't think it'll make a big difference in practice, so I don't feel strongly about it.

gnzlbg: Good point. User programs can use `as` to go either way anyways, and to me it feels that both approaches have downsides - using `isize` might mean that negative values might indicate success, while using `usize` will confuse those trying to use `-1` here to catch errors like the spec suggests =/

alexcrichton: That sounds reasonable to me! I've switched to `usize` and documented the return value on failure. Thanks for the suggestions @sunfishcode!

PR #615, opened 2018-12-12
Derive Debug for CpuidResult

nvzqz:
Being able to `println!("{:#?}", cpuid_result)` and seeing all the registers in one go is nice to have.

### DIFF
--- a/coresimd/x86/cpuid.rs
+++ b/coresimd/x86/cpuid.rs
@@ -6,7 +6,7 @@
 use stdsimd_test::assert_instr;
 
 /// Result of the `cpuid` instruction.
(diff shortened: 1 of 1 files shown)


alexcrichton:
👍

gnzlbg:
Does one need to feature gate this or is it ok to just add the impl ?

nvzqz:
I am under the impression that trait impls are auto-stabilized. Idk whether that's still the case.

alexcrichton:
Ah yeah we have no way of gating this implementation and later stabilizing, but that's fine and we accept `Debug` without FCP/stabilization typically anyway

gnzlbg:
👍 just triple-checking, thanks!

PR #618, opened 2018-12-13
Start adding some avx512 intrinsics

alexcrichton:
First one is the quite simple `_mm512_abs_epi32` intrinsic!


alexcrichton:
I was actually quite surprised to learn but my local computer has avx512f at least, so I was even able to run and execute these tests!

coresimd/x86/mod.rs
```diff
@@ -391,6 +391,10 @@ types! {
     pub struct __m512d(f64, f64, f64, f64, f64, f64, f64, f64);
 }
 
+/// The `__mmask16` type used in AVX-512 intrinsics, a 16-bit integer
+#[allow(non_camel_case_types)]
+pub type __mmask16 = i16;
```
alexcrichton: FWIW this is definitely something we're going to want to debate/decide before stabilization. I've never used this type before (or these intrinsics) so this is the most naive way to translate it, but other options include:

* A newtype struct wrapper
* An unsigned number

I'm not sure what the best option is, but others I'm sure to!

Cargo.toml
```diff
@@ -10,7 +10,9 @@ exclude = [
 [profile.release]
 debug = true
 opt-level = 3
+incremental = true
```
gnzlbg: Weird, why is this needed?

alexcrichton: Oh it wasn't needed per se, but it made development a lot easier as it doesn't regress any codegen and it makes release builds (assert_instr) much faster

PR #623, opened 2018-12-14
Match clang for `_mm512_abs_epi32` intrinsics

alexcrichton:
This commit updates stdsimd's codegen to match Clang's for the
`_mm512_abs_epi32` intrinsic (and masked versions) which doesn't use any
LLVM intrinsic calls, but rather raw SIMD operations.

These are built on top of the new `simd_select_bitmask` intrinsic
introduced recently to the compiler!


alexcrichton:
Note that CI will fail here because the current nightly doesn't have support, but the next one will!

alexcrichton:
All green!

gnzlbg:
So this is how clang implement these. I still don't know why, but it probably is indeed better to do things like clang does here.

PR #624, opened 2018-12-14
Add a `x86_64::cmpxchg16b` intrinsic

alexcrichton:
This intrinsic isn't actually specified by Intel, but it's something
gated with CPUID and can otherwise be a useful thing to have when
building primitives!

There exists an `AtomicU128` type in the standard library but it's only
exposed currently (and it's unstable) when a platform fully supports
128-bit atomics. The x86_64 architecture does not support it *unless*
the `cmpxchg16b` instruction is available, and it isn't always available!

This commit is also a proposal for how we can include support for
128-bit atomics in the standard library on relevant platforms. I'm
thinking that we'll expose this one low-level intrinsic in
`std::arch::x86_64`, and then if desired a crate on crates.io can build
`AtomicU128` from this API.

In any case this is all unstable regardless!

### DIFF
--- a/coresimd/x86_64/mod.rs
+++ b/coresimd/x86_64/mod.rs
@@ -38,3 +38,6 @@ pub use self::bswap::*;
 
 mod rdrand;
 pub use self::rdrand::*;
+
+mod cmpxchg16b;
+pub use self::cmpxchg16b::*;
--- a/stdsimd/arch/detect/os/x86.rs
+++ b/stdsimd/arch/detect/os/x86.rs
@@ -116,6 +116,7 @@ fn detect_features() -> cache::Initializer {
 
         enable(proc_info_ecx, 0, Feature::sse3);
         enable(proc_info_ecx, 9, Feature::ssse3);
+        enable(proc_info_ecx, 13, Feature::cmpxchg16b);
         enable(proc_info_ecx, 19, Feature::sse4_1);
         enable(proc_info_ecx, 20, Feature::sse4_2);
         enable(proc_info_ecx, 23, Feature::popcnt);
@@ -288,6 +289,7 @@ mod tests {
         println!("xsaveopt: {:?}", is_x86_feature_detected!("xsaveopt"));
         println!("xsaves: {:?}", is_x86_feature_detected!("xsaves"));
         println!("xsavec: {:?}", is_x86_feature_detected!("xsavec"));
+        println!("cmpxchg16b: {:?}", is_x86_feature_detected!("cmpxchg16b"));
     }
 
     #[test]
@@ -343,6 +345,10 @@ mod tests {
         assert_eq!(
             is_x86_feature_detected!("xsaves"),
             information.xsaves_xrstors_and_ia32_xss()
+        );
+        assert_eq!(
+            is_x86_feature_detected!("cmpxchg16b"),
+            information.cmpxchg16b(),
         );
     }
 }
--- a/coresimd/x86_64/cmpxchg16b.rs
+++ b/coresimd/x86_64/cmpxchg16b.rs
@@ -1,0 +1,74 @@
+use sync::atomic::Ordering;
+
+#[cfg(test)]
+use stdsimd_test::assert_instr;
+
+/// Compare and exchange 16 bytes (128 bits) of data atomically.
+///
+/// This intrinsic corresponds to the `cmpxchg16b` instruction on x86_64
+/// processors. It performs an atomic compare-and-swap, updating the `ptr`
+/// memory location to `val` if the current value in memory equals `old`.
+///
+/// # Return value
+///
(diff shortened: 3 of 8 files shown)


alexcrichton:
Note that this is intended to be coupled with https://github.com/rust-lang/rust/pull/56826

coresimd/x86_64/mod.rs
```diff
@@ -38,3 +43,64 @@ pub use self::bswap::*;
 
 mod rdrand;
 pub use self::rdrand::*;
+
```
gnzlbg: I would put the intrinsic in its own module for consistency with the others, but.. the module would just contain one intrinsic..

coresimd/x86_64/mod.rs
```diff
@@ -38,3 +43,64 @@ pub use self::bswap::*;
...
+/// instead of just a pointer.
+///
+/// For more information on memory orderings here see the `compare_exchange`
+/// documentation for other `Atomic*` types in the standard library.
+///
+/// # Unsafety
+///
+/// This method is unsafe because it takes a raw pointer and will attempt to
+/// read and possibly write the memory at the pointer. The pointer must also be
+/// aligned on a 16-byte boundary.
```
gnzlbg: We probably want a `debug_assert!` here that checks that the pointer is 16-byte aligned. This won't be easily usable through `libcore`/`libstd`, but the `libstd` tests are run in CI with debug assertions enabled and this method might be used to implement some of the std atomic types, so it would at least be useful for that.

coresimd/x86_64/mod.rs
```diff
@@ -38,3 +43,64 @@ pub use self::bswap::*;
...
+///
+/// This method also requires the `cmpxchg16b` CPU feature to be available at
+/// runtime to work correctly, and executing this function if the CPU doesn't
```
gnzlbg: I would word this as "...correctly. If the CPU running the binary does not actually support `cmpxchg16b` and the program enters an execution path that eventually would reach this function the behavior is undefined." 

As in, the problem here is that an optimizing compiler is able to assume that the CPU supports `cmpxchg16b` in all execution paths that unconditionally reach this function, and is able to emit machine code under that assumption in those. That is, once the program enters such an execution path, all bets are off: there are no guarantees that this function will ever be reached.

cc @RalfJ

coresimd/x86_64/mod.rs
```diff
@@ -38,3 +43,64 @@ pub use self::bswap::*;
...
+        (SeqCst, SeqCst) => intrinsics::atomic_cxchg(dst, old, new),
+        (Acquire, Relaxed) => intrinsics::atomic_cxchg_acq_failrelaxed(dst, old, new),
+        (AcqRel, Relaxed) => intrinsics::atomic_cxchg_acqrel_failrelaxed(dst, old, new),
+        (SeqCst, Relaxed) => intrinsics::atomic_cxchg_failrelaxed(dst, old, new),
+        (SeqCst, Acquire) => intrinsics::atomic_cxchg_failacq(dst, old, new),
+
+        // The above block is all copied from libcore, and this statement is
+        // also copied from libcore except that it's a panic in libcore and we
+        // have a little bit more of a lightweight panic here.
+        _ => ::coresimd::x86::ud2(),
```
gnzlbg: The docs should probably contain a:

```rust
/// # Panics
///
/// If the `failure` ordering is `AcqRel` or `Release`, or if 
/// it is stronger than the `success` ordering.
```

stdsimd/arch/detect/os/x86.rs
```diff
@@ -116,6 +116,7 @@ fn detect_features() -> cache::Initializer {
 
         enable(proc_info_ecx, 0, Feature::sse3);
         enable(proc_info_ecx, 9, Feature::ssse3);
+        enable(proc_info_ecx, 13, Feature::cmpxchg16b);
```
gnzlbg: The short mnemonic for this is `cx16`. There is also a CPU feature `cx8` in `proc_info_edx = 8` that offers another CAS intrinsic. [clang](https://clang.llvm.org/docs/genindex.html) also calls this `cx16`.

alexcrichton:
Thanks for taking a look! I think I've taken care of everything now.

FWIW I also found the `cx16` feature name, but I couldn't actually find any information from Intel saying that it was the official name of this feature. All I could find with Intel was "cmpxchg16", so that's what I ended up naming it in the compiler itself.

alexcrichton:
Good idea! Updated now

gnzlbg:
Thanks!

gnzlbg:
I don’t think that allows using run time feature detection to , E.g., use
atomic if available or a lock otherwise.

the8472:
Yeah, but it's a simple change that's better than not providing it at all.

PR #628, opened 2018-12-26
avoid deprecated mem::zeroed

RalfJung:
Needed to make https://github.com/rust-lang/rust/pull/57045 go through


gnzlbg:
Thanks, sorry for the delay, christmas and all that.

PR #629, opened 2018-12-28
Add a build libcore-only nvptx64 test (using xargo)

peterhj:
This commit adds an nvptx64 build libcore-only test, using xargo. It involved some changes to the travis yml, as well as the CI run scripts. I use a $CROSS_TARGET env variable in the travis yml to force cross compilation on the host $TARGET.

This also disables the "integer_atomics" feature on nvptx/nvptx64. During debugging, I got a strange `unknown feature 'integer_atomics'` error (see bottom of this log: https://travis-ci.org/peterhj/stdsimd-nvptx/jobs/473037366). As the atomic int types aren't currently used w/ nvptx (although there do exist CUDA atomics on both integers and floating points, but they behave somewhat differently than the ones in sync::atomic), so I just disabled "integer_atomics" on the nvptx targets.

### DIFF
--- a/crates/coresimd/src/lib.rs
+++ b/crates/coresimd/src/lib.rs
@@ -18,7 +18,6 @@
     simd_ffi,
     asm,
     proc_macro_hygiene,
-    integer_atomics,
     stmt_expr_attributes,
     core_intrinsics,
     no_core,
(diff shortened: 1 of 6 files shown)


crates/coresimd/src/lib.rs
```diff
@@ -38,6 +37,10 @@
     powerpc_target_feature,
     wasm_target_feature
 )]
+#![cfg_attr(
```
gnzlbg: Could you add a comment explaining why unconditionally enabling this fails ?

peterhj: added

gnzlbg:
Thank you peter, this is a great start!

PR #630, opened 2019-01-04
Remove usage of `platform-intrinsics` where `unadjusted` works

alexcrichton:
As suggested [here]!

[here]: https://github.com/rust-lang/rust/pull/57048#issuecomment-449773525


alexcrichton:
Unfortunately not because of the aggregate return, which needs to be a specific LLVM type that otherwise doesn't manifest as such (since tuples aren't ABI-safe)

PR #640, opened 2019-01-21
Refactor stdsimd

gnzlbg:
This commit:

* renames `coresimd` to `core_arch` and `stdsimd` to `std_detect`

* `std_detect` does no longer depend on `core_arch` - it is a freestanding
  `no_std` (on some archs) library that only depends on `core`

* moves the top-level coresimd and stdsimd directories into the appropriate
  crates/... directories - this simplifies creating crate.io releases of these crates

* moves the top-level `coresimd` and `stdsimd` sub-directories into their
  corresponding crates in `crates/{core_arch, std_detect}`.

After this PR lands https://github.com/rust-lang/rust/pull/57808 could land. The main benefit is that we no longer build core::arch twice in rust-lang/rust, once in libstd, and once in libcore. Instead, these two PR change that to only build it once in libcore, and then libstd just re-export the `core::arch` module. The run-time feature detection is provided in `libstd` by just adding `std_detet` as a dependency, hopefully, without much hackery (~~since it is a `#[no_std]` library~~ it's only a `#[no_std]` library on some archs, so we'll see how this goes). This also allows those `#[no_std]` users that want to have the exact same run-time feature detection system as `libstd` to be able to do so by just adding `std_detect` as a dependency.


crates/std_detect/src/detect/cache.rs
```diff
@@ -78,7 +78,8 @@ struct Cache(AtomicU64);
 impl Cache {
     /// Creates an uninitialized cache.
     const fn uninitialized() -> Self {
-        Cache(AtomicU64::new(u64::max_value()))
+        const X: AtomicU64 = AtomicU64::new(u64::max_value());
+        Self(X)
```
gnzlbg: @Centril @oli-obk without this change `std_detect` fails to compile, which is weird. The error is "can only call other min_const_fn within a min_const_fn" but the crate has `feature(const_fn)` enabled =/

Centril: (repeating from Discord: I wasn't able to reproduce either... I'll see if I have time to test this locally; for the time being, this hack is fine, and `AtomicU64` is stable in 1.34 so hopefully that makes the hack go away... However, when using the hack in the future, please cc myself and @oli-obk...)

gnzlbg:
So @alexcrichton this should be all green now. In some targets `std_detect` needs to depend on `std`, to use `auxv`/`cpuinfo`/ etc. Is that possible at all when adding `std_detect` as a normal dependency to `libstd` ? In the future we might want to add it a `use_std` feature that falls back to something else if `std` is not available.

gnzlbg:
I am merging this. We can iterate on this in tree.

PR #647, opened 2019-01-22
Automatically insert emms after running each MMX test

gnzlbg:
After using MMX intrinsics the FPU must be
cleared by using `_mm_empty()` before interfacing
with any x87 code.

This commit makes the simd_test macro automatically do that
when one of the features enabled is "mmx".


alexcrichton:
This is a great idea!

PR #659, opened 2019-01-30
Fix docs.rs rustdoc arguments

gnzlbg:


gnzlbg:
Yay this worked ! https://docs.rs/core_arch/0.1.3/core_arch/

PR #661, opened 2019-01-30
Fix addcarry / subborrow intrinsic names

nikic:
Intended to fix https://github.com/rust-lang/rust/issues/57993. This does make the tests build for me, though a lot of them are failing. Not sure if that's normal or not.


alexcrichton:
Oh dear, I think I was a bit too trigger happy on the merging, sorry about that!

PR #662, opened 2019-01-30
Attempt to fix tests on master

alexcrichton:
* Make all doctests use items from the real `std` rather than this
  crate, it's just easier
* Handle debuginfo weirdness by flagging functions as `no_mangle` that
  we're looking for instructions within.

### DIFF
--- a/crates/assert-instr-macro/src/lib.rs
+++ b/crates/assert-instr-macro/src/lib.rs
@@ -49,7 +49,7 @@ pub fn assert_instr(
         .replace(':', "_")
         .replace(|c: char| c.is_whitespace(), "");
     let assert_name = syn::Ident::new(&format!("assert_{}_{}", name, instr_str), name.span());
-    let shim_name = syn::Ident::new(&format!("{}_shim", name), name.span());
+    let shim_name = syn::Ident::new(&format!("{}_shim_{}", name, instr_str), name.span());
     let mut inputs = Vec::new();
     let mut input_vals = Vec::new();
     let ret = &func.decl.output;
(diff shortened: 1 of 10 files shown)


crates/core_arch/src/core_arch_docs.md
```diff
@@ -251,16 +251,6 @@ Next up let's take a look at an example of manually using intrinsics. Here
...
 ```
-# #![cfg_attr(not(dox),feature(stdsimd))]
-# #![cfg_attr(not(dox), no_std)]
-# #[cfg(not(dox))]
-# extern crate std as real_std;
-# #[cfg(not(dox))]
-# extern crate core_arch as std;
-# #[cfg(not(dox))]
-# #[macro_use(is_x86_feature_detected)]
-# extern crate std_detect;
```
gnzlbg: Should we be doing this everywhere ? There are many doc examples where we do this dance.

crates/assert-instr-macro/src/lib.rs
```diff
@@ -99,6 +99,7 @@ pub fn assert_instr(
     let shim_name_str = format!("{}{}", shim_name, assert_name);
     let to_test = quote! {
         #attrs
+        #[no_mangle]
```
gnzlbg: We probably should add a comment around here about which exact problem `#[no_mangle]` solves (debug-info weirdness).

crates/core_arch/src/core_arch_docs.md
alexcrichton: Yeah I don't really know what was happening to have the doctests fail, but this is all in std now so none of this was necessary any more

PR #675, opened 2019-02-03
Add detect macros should support trailing commas (Fix #443)

### LINKED ISSUE #443: macros should support trailing commas
Since the macros from this crate are now showing up as reexported from std, I'm going to be taking it upon myself to make sure that they all support trailing commas.

(you know, the one trailing comma that nobody will ever care about or want to take advantage of because the only argument is a literal string of finite length)

botika:
#443


gnzlbg:
Thank you!

PR #677, opened 2019-02-05
Remove const workaround in std_detect cache

gnzlbg:


gnzlbg:
cc @eddyb looks like this passed now ¯\_(ツ)_/¯

Issue #655, opened 2019-01-28
Linux ARM (aarch64 and 32-bit) feature detection should avoid file I/O and `dlsym`


briansmith:
If `getauxval()` is provided by a statically-linked libc, like musl C, I think `dlsym` might not find it. (I'm not 100% sure on this.) If weak linkage is necessary then I think relying on the linker to do the weak linkage (using `[linkage = "extern_weak"]` if that's stable) would be better.

For `-musl` targets in particular, `getauxval()` should just be assumed to be available unless/until somebody mentions a real need to support old, old versions of musl.

For Android targets, Google now requires developers to target API level 26 (see https://developer.android.com/distribute/best-practices/develop/target-sdk) so IMO we should be able to always assume `getauxval()` is available for all Android targets.

For iOS targets, AFAICT no runtime detection is needed because iOS devices are guaranteed to be AAarch64 with crypto and SIMD extensions enabled.

For AAarch64 Linux targets, I think it is reasonable to require `getauxval()` be available, unless/until somebody brings a real-world example of where they are targeting AAarch64 with an ancient glibc, simply because AAarch64 itself is newer than the version of glibc that has `getauxval()`.

I also wonder if anybody is really targeting 32-bit ARM Linux with system glibc that is so old that it doesn't provide `getauxval()`. (This situation is different than x86_64 and x86 where we already know that there are LTS Linux distros that have ancient libcs.) IMO it would be better to be aggressive in simplifying the implementation to require `getauxval()` statically at link time unless/until somebody brings a real use case.

At the very least, it would be useful for me to have some way to disable the file I/O and parsing of `/proc/self/auxv` and/or `/proc/cpuinfo` in a crate depending on `stdsimd` directly, e.g. with a feature flag.

This would enable me to remove my own `getauxval()` code in *ring* in favor of using `stdsimd` on AAarch64/ARM Linux.

gnzlbg:
Thank you @briansmith , I think we can do most of this. 

> For Android targets, Google now requires developers to target API level 26 (see https://developer.android.com/distribute/best-practices/develop/target-sdk) so IMO we should be able to always assume getauxval() is available for all Android targets.

I think servo is still supporting a very old version of Android. cc @SimonSapin - do you have any thoughts about the minimum version that the std library should support here ?

> For iOS targets, AFAICT no runtime detection is needed because iOS devices are guaranteed to be AAarch64 with crypto and SIMD extensions enabled.

[...]

PR #678, opened 2019-02-07
Add cargo features to disable usage of file I/O and dlsym in std_detect

gnzlbg:


gnzlbg:
cc @briansmith - I think this could close #655 . Basically, adding `std_detect = { default-features = false }` as a dependency would not require `libc` nor `libstd`. In the platforms where `getauxval` is the preferred / only way of detecting features, it would just assume it's there.

PR #679, opened 2019-02-11
MaybeUninit: update to into_initialized

RalfJung:
This is the stdsimd side of https://github.com/rust-lang/rust/pull/58129.

(Having stdsimd in a separate repo means that what would usually be 1 PR turns into 3...)

### DIFF
--- a/crates/core_arch/src/x86/avx.rs
+++ b/crates/core_arch/src/x86/avx.rs
@@ -2959,7 +2959,7 @@ pub unsafe fn _mm256_zextpd128_pd256(a: __m128d) -> __m256d {
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn _mm256_undefined_ps() -> __m256 {
     // FIXME: this function should return MaybeUninit<__m256>
-    mem::MaybeUninit::<__m256>::uninitialized().into_inner()
+    mem::MaybeUninit::<__m256>::uninitialized().into_initialized()
 }
 
 /// Return vector of type `__m256d` with undefined elements.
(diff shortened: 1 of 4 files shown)


gnzlbg:
@RalfJung wait a sec, how did that PR get merged without updating stdsimd ? `stdsimd` should have failed to build as part of `libcore` because of this -.-

gnzlbg:
FWIW, there is a possible 1 PR approach. First you update `stdsimd` upstream to point to your own fork, then you fix the fork as part of your PR, and then you merge that. 

Then you send a PR here, which we merge, and we take care of updating `stdsimd` back to this repo master for you :)

RalfJung:
I left a libcore-private copy of `fn into_inner` which stdsimd uses (and which caused a bors failure because on non-x86 platforms that function is unused :/ ).

> First you update `stdsimd` upstream to point to your own fork

You mean as in, change the rustc branch it tests again?
That'll be "fun" when two people want to do it at the same time. ;)

gnzlbg:
> You mean as in, change the rustc branch it tests again?
> That'll be "fun" when two people want to do it at the same time. ;)

If you ping me or alex we can prevent two people from doing it at the same time. 

There is an alternative 2 PR method, which has the same problem if two people do it at the same time (depending on the changes).  You could update `stdsimd` in your PR to  `stdsimd/master` with your fix, if you send a PR here before that breaks stdsimd master temporarily.  If your PR is reviewed and ready to merge, we could also do just that. 

The safest method is the 3 PR approach, but if you ping us we can tell you if there are others trying to do the same right now or not :P

gnzlbg:
FYI @RalfJung windows is broken, so as long as travis passes everything should be ok.

RalfJung:
The Android and macOS build failure seem to be just timeouts?

gnzlbg:
I've restarted them.

PR #685, opened 2019-02-13
Fix nvptx64 libcore-only build on travis.

peterhj:


peterhj:
It looks like the wasm build is failing, but the nvptx64 build uses the right target: https://travis-ci.com/rust-lang-nursery/stdsimd/jobs/177594532#L727

gnzlbg:
Thanks! #684 should fix wasm.

PR #689, opened 2019-02-18
A few more cosmetic improvements

alexreg:
Sorry, just a few more I forgot to include! Not big at all.

r? @gnzlbg

### DIFF
--- a/crates/core_arch/src/x86_64/bmi.rs
+++ b/crates/core_arch/src/x86_64/bmi.rs
@@ -92,7 +92,7 @@ pub unsafe fn _blsr_u64(x: u64) -> u64 {
 
 /// Counts the number of trailing least significant zero bits.
 ///
-/// When the source operand is 0, it returns its size in bits.
+/// When the source operand is `0`, it returns its size in bits.
 ///
 /// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_tzcnt_u64)
 #[inline]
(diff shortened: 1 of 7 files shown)


gnzlbg:
Is there a way to automatically lint on these on CI ?

crates/core_arch/src/x86_64/bmi.rs
```diff
@@ -170,7 +170,7 @@ mod tests {
 
     #[simd_test(enable = "bmi1")]
     unsafe fn test_blsr_u64() {
-        // TODO: test the behavior when the input is 0
+        // TODO: test the behavior when the input is 0.
```
gnzlbg: might want to write "`0`" here

alexreg:
Done. Ta.

alexreg:
> Is there a way to automatically lint on these on CI ?

Not that I know of, but would be nice.

gnzlbg:
> Not that I know of, but would be nice.

Until we have one, what you are doing is an impossible battle to win: you can't fix all comments of all Rust contributors. 

I think it would be better to add a clippy lint that uses the system's spell and grammar checker (e.g. `aspell`) to correct spelling, grammar, etc. on comments, maybe with a way to control the spell-checker settings (locale, etc.). 

That would give different results on different systems, but we could choose to run clippy only on, e.g., Linux. Although if the spelling and grammar are correct on Linux, the chances that a Windows or MacOSX spell checker flag these as incorrect (and vice-versa) are probably slim as long as the locale's are configured to be the exact same.

alexreg:
@gnzlbg You're not wrong... hopefully these changes have a least set a precedent though, so others will try to copy and adhere to the style.

alexreg:
@gnzlbg You're not wrong... hopefully these changes have a least set a precedent though, so others will try to copy and adhere to the general style.

PR #693, opened 2019-02-19
Change imports in std_detect to edition-agnostic style

taiki-e:
This changes imports in `std_detect` to edition-agnostic style (This is **not** transition of `std_detect`).

This allows `libstd` to transition to Rust 2018 (https://github.com/rust-lang/rust/pull/58208).

cc https://github.com/rust-lang/rust/issues/58099

### DIFF
--- a/crates/std_detect/src/detect/cache.rs
+++ b/crates/std_detect/src/detect/cache.rs
@@ -3,13 +3,13 @@
 
 #![allow(dead_code)] // not used on all platforms
 
-use sync::atomic::Ordering;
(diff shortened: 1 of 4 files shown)


gnzlbg:
@taiki-e thank you for the PR, is the `crate::` necessary ? Wouldn't e.g. `::sync` suffice ?

gnzlbg:
> It cannot be imported with `::sync` in Rust 2018,

Ah! Indeed. Thank you !

PR #694, opened 2019-02-21
Include the ACLE module via path in aarch64 and arm

gnzlbg:
@japaric this might do the trick fixing the docs upstream. The ACLE module ends up being compiled twice when using rustdoc and that's it.


gnzlbg:
Ah no, this will have the same problem, the aarch64 module will show arm stuff, and viceversa. damn.

PR #699, opened 2019-02-25
NVPTX CUDA syscalls

denzp:
This change introduces basic CUDA (`nvptx64-nvidia-cuda` target) syscalls.

Following up discussing and implementation from [libc#1277](https://github.com/rust-lang/libc/pull/1277).

cc @gnzlbg

### DIFF
--- a/crates/core_arch/src/nvptx/mod.rs
+++ b/crates/core_arch/src/nvptx/mod.rs
@@ -11,6 +11,8 @@
 //! [llvm_docs]:
 //! https://llvm.org/docs/NVPTXUsage.html
 
+use core::ffi::c_void;
+
 #[allow(improper_ctypes)]
 extern "C" {
     #[link_name = "llvm.nvvm.barrier0"]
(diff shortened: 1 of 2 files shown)


gnzlbg:
cc @alexcrichton these are more intrinsics than syscalls. I'm not sure where would be a good place to expose these, but core::arch makes more sense than libc to me.

alexcrichton:
Agreed yeah, `stdsimd` and `core::arch` are all about exposing architecture-specific functionality, typically CPU instructions and such. For definitions of externally-available functionality which is commonly available that's best suited for `libc`

crates/core_arch/src/nvptx/mod.rs
```diff
@@ -11,6 +11,8 @@
 //! [llvm_docs]:
 //! https://llvm.org/docs/NVPTXUsage.html
 
+use core::ffi::c_void;
```
gnzlbg: `core_arch` is compiled inside libcore by using a `#[path = "..."]` to the `core_arch/src/mod.rs` file (the `lib.rs` is only used when building `core_arch` as a standalone library). 

So this needs to use `crate::ffi::c_void` to be able to compile in both Rust 2015 and Rust 2018 modes inside `libcore`.  You might need to bring the `ffi` module to the top module (check out `lib.rs` and `mod.rs`).

denzp: I've added a commit to address the issue. That's an incredibly clever solution, I must say :slightly_smiling_face:

gnzlbg: > That's an incredibly clever solution, I must say

@alexcrichton came up with it 🎉

gnzlbg:
Thanks!

PR #703, opened 2019-03-07
Add bittest instructions for x86

### LINKED ISSUE #553: Add bittest intrinsics
This is a feature request to add [`bittest`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#expand=12,79&text=bittest) intrinsics. I'm particularly interested in `_bittest64`.

mehcode:
Closes #553 

---

 * ~I don't think the `assert_instr` lines are doing anything so not sure if I'm using them wrong or they just aren't applicable here.~ Ah. It does run on CI.

 * I know the API differs slightly than what's prescribed in https://github.com/rust-lang-nursery/stdsimd/issues/553#issuecomment-415708627 but after looking around this feels more consistent.

 * Do we want to provide non-reference versions of `bittest`? Not sure if it'd be worth it. 
 
 * What's the policy on stuff like `debug_assert`s for checking invariants? If you test for a bit higher than the total bit size of the operand it runs right off the memory onto whatever is there at the moment.

### DIFF
--- a/crates/core_arch/src/x86/bt.rs
+++ b/crates/core_arch/src/x86/bt.rs
@@ -1,0 +1,102 @@
+#[cfg(test)]
+use stdsimd_test::assert_instr;
+
+/// Returns the bit in position `b` of the memory addressed by `p`.
+#[inline]
+#[cfg_attr(test, assert_instr(bt))]
+#[unstable(feature = "simd_x86_bittest", issue = "59414")]
+pub unsafe fn _bittest(p: *const i32, b: i32) -> u8 {
+    let r: u8;
+    asm!("btl $2, $1\n\tsetc ${0:b}"
+         : "=r"(r)
+         : "*m"(p), "r"(b)
+         : "cc", "memory");
+    r
+}
+
+/// Returns the bit in position `b` of the memory addressed by `p`, then sets the bit to `1`.
+#[inline]
+#[cfg_attr(test, assert_instr(bts))]
+#[unstable(feature = "simd_x86_bittest", issue = "59414")]
+pub unsafe fn _bittestandset(p: *mut i32, b: i32) -> u8 {
+    let r: u8;
+    asm!("btsl $2, $1\n\tsetc ${0:b}"
+         : "=r"(r), "+*m"(p)
+         : "r"(b)
+         : "cc", "memory");
+    r
+}
+
+/// Returns the bit in position `b` of the memory addressed by `p`, then resets that bit to `0`.
+#[inline]
+#[cfg_attr(test, assert_instr(btr))]
+#[unstable(feature = "simd_x86_bittest", issue = "59414")]
+pub unsafe fn _bittestandreset(p: *mut i32, b: i32) -> u8 {
+    let r: u8;
+    asm!("btrl $2, $1\n\tsetc ${0:b}"
+         : "=r"(r), "+*m"(p)
+         : "r"(b)
+         : "cc", "memory");
+    r
+}
+
+/// Returns the bit in position `b` of the memory addressed by `p`, then inverts that bit.
+#[inline]
+#[cfg_attr(test, assert_instr(btc))]
+#[unstable(feature = "simd_x86_bittest", issue = "59414")]
+pub unsafe fn _bittestandcomplement(p: *mut i32, b: i32) -> u8 {
+    let r: u8;
+    asm!("btcl $2, $1\n\tsetc ${0:b}"
+         : "=r"(r), "+*m"(p)
+         : "r"(b)
+         : "cc", "memory");
+    r
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::core_arch::x86::*;
+
+    #[test]
+    fn test_bittest() {
+        unsafe {
+            let a = 0b0101_0000i32;
+            assert_eq!(_bittest(&a as _, 4), 1);
+            assert_eq!(_bittest(&a as _, 5), 0);
+        }
+    }
+
+    #[test]
+    fn test_bittestandset() {
+        unsafe {
+            let mut a = 0b0101_0000i32;
+            assert_eq!(_bittestandset(&mut a as _, 4), 1);
+            assert_eq!(_bittestandset(&mut a as _, 4), 1);
+            assert_eq!(_bittestandset(&mut a as _, 5), 0);
+            assert_eq!(_bittestandset(&mut a as _, 5), 1);
+        }
+    }
+
+    #[test]
+    fn test_bittestandreset() {
+        unsafe {
+            let mut a = 0b0101_0000i32;
+            assert_eq!(_bittestandreset(&mut a as _, 4), 1);
+            assert_eq!(_bittestandreset(&mut a as _, 4), 0);
+            assert_eq!(_bittestandreset(&mut a as _, 5), 0);
+            assert_eq!(_bittestandreset(&mut a as _, 5), 0);
+        }
+    }
+
+    #[test]
+    fn test_bittestandcomplement() {
+        unsafe {
+            let mut a = 0b0101_0000i32;
+            assert_eq!(_bittestandcomplement(&mut a as _, 4), 1);
+            assert_eq!(_bittestandcomplement(&mut a as _, 4), 0);
+            assert_eq!(_bittestandcomplement(&mut a as _, 4), 1);
+            assert_eq!(_bittestandcomplement(&mut a as _, 5), 0);
+            assert_eq!(_bittestandcomplement(&mut a as _, 5), 1);
+        }
+    }
+}
--- a/crates/core_arch/src/x86_64/bt.rs
+++ b/crates/core_arch/src/x86_64/bt.rs
@@ -1,0 +1,102 @@
+#[cfg(test)]
+use stdsimd_test::assert_instr;
+
+/// Returns the bit in position `b` of the memory addressed by `p`.
+#[inline]
+#[cfg_attr(test, assert_instr(bt))]
+#[unstable(feature = "simd_x86_bittest", issue = "59414")]
+pub unsafe fn _bittest64(p: *const i64, b: i64) -> u8 {
+    let r: u8;
+    asm!("btq $2, $1\n\tsetc ${0:b}"
+         : "=r"(r)
+         : "*m"(p), "r"(b)
+         : "cc", "memory");
+    r
+}
+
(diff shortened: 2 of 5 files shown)


crates/core_arch/src/x86/bt.rs
```diff
@@ -0,0 +1,105 @@
+#[cfg(test)]
+use stdsimd_test::assert_instr;
+
+/// Return the bit at index `b` of 32-bit integer `a`.
+#[inline]
+#[cfg_attr(test, assert_instr(btl))]
+#[unstable(feature = "bittest", issue = "0")]
```
gnzlbg: probably want to name the feature `simd_x86_bittest` - this will need support upstream.

crates/core_arch/src/x86/bt.rs
```diff
@@ -0,0 +1,105 @@
+#[cfg(test)]
+use stdsimd_test::assert_instr;
+
+/// Return the bit at index `b` of 32-bit integer `a`.
+#[inline]
+#[cfg_attr(test, assert_instr(bt))]
+#[unstable(feature = "bittest", issue = "0")]
+pub unsafe fn _bittest_u32(a: &u32, b: u32) -> u8 {
```
gnzlbg: API validation suite will probably chuckle at the `&u32`.

gnzlbg: Also, AFACIT this intrinsic is called `_bittest` and not `_bittest_u32`.

crates/core_arch/src/x86/bt.rs
```diff
@@ -0,0 +1,105 @@
...
+    let r: u8;
+    asm!("btl $2, $1\nsetc ${0:b}"
+         : "=r"(r)
+         : "m"(a), "lr"(b)
+         : "cc", "flags", "fpsr");
```
gnzlbg: clang tags these with `sideeffect` - could you post a godbolt link to this Rust implementation so that we can check that the LLVM-IR is correct ?

crates/core_arch/src/x86_64/bt.rs
```diff
@@ -0,0 +1,105 @@
+#[cfg(test)]
+use stdsimd_test::assert_instr;
+
+/// Return the bit at index `b` of 64-bit integer `a`.
+#[inline]
+#[cfg_attr(test, assert_instr(bt))]
+#[unstable(feature = "bittest", issue = "0")]
+pub unsafe fn _bittest_u64(a: &u64, b: u64) -> u8 {
```
gnzlbg: This one is called `_bittest64`.

crates/core_arch/src/x86/bt.rs
mehcode: So I referenced Clang to update the inline assembly I was using. After doing so I'm reasonably confident that the reason Clang is using `sideffect` is because there is an observed write-back to `a` and clang only uses `r` as a constraint. I'm able to produce identical assembly using `=*r` which should indicate that a write-back. 

If I switch back to the inline asm I had before I think this should be valid without the `sideeffect` label (which is good as it doesn't look like rust inline asm supports that).

This is my first time trying to use inline asm like this so I may be totally wrong.

gnzlbg: > (which is good as it doesn't look like rust inline asm supports that

I think Rust uses `"volatile"` for this, but I wasn't suggesting that the older version was incorrect (clang might also as well just have a bug). I just wanted to compare the implementation with clang at the LLVM-IR level, because precisely what you mention about the inline assembly syntaxes being identical. If the LLVM-IR generated differs, then we can try to reason about why that's the case, and whether the bug is in our side or on clang's side. 

cc @Amanieu

Amanieu: Don't use "m" constraints, they are broken. Use "*m" and pass `&a` instead of `a`. "flags" and "fpsr" are not needed here, only "cc" is clobbered. Finally, use plain "r" instead of "lr".

crates/core_arch/src/x86/bt.rs
```diff
@@ -0,0 +1,105 @@
...
+/// Return the bit at index `b` of 32-bit integer `a`,
+/// and set that bit to one.
+#[inline]
+#[cfg_attr(test, assert_instr(bts))]
+#[unstable(feature = "simd_x86_bittest", issue = "0")]
+pub unsafe fn _bittestandset(a: &mut u32, b: u32) -> u8 {
+    let r: u8;
+    asm!("btsl $2, ($1)\n\tsetc ${0:b}"
+         : "=r"(r)
+         : "r"(a), "r"(b)
```
Amanieu: Use "+*m" for `a` to indicate read-write.

Amanieu:
What are the semantics of these intrinsics when the bit index is larger than the word size? The BT* instructions will happily continue on the next word.

If this is intended behavior then we need to use the "memory" clobber.

gnzlbg:
> The BT* instructions will happily continue on the next word.

I just read the docs of the `BT` instructions and I understood those as the correct semantics. 

The spec for the intrinsics says:

> Return the bit at index b of 64-bit integer a.

which hints that one can't "index out of bounds". This might be a bug in the spec of the intrinsics, or might be the intended behavior, I don't know.

@mehcode I would be more comfortable if these would take a raw pointer - in case that turns out to be a bug on the spec of the intrinsics, the raw pointer would just work, while the reference API would be weird. 

It would be great if somebody could ask for clarification here: https://software.intel.com/en-us/forums/intel-isa-extensions/topic/363747 (that's where spec bugs are reported).

Amanieu:
Alternatively if you first load the value into a register and then run the BT* instruction on that register then the instruction will automatically mask the index to fit in the register size.

gnzlbg:
> Alternatively if you first load the value into a register and then run the BT* instruction on that register then the instruction will automatically mask the index to fit in the register size.

Sure. I was just wondering whether the intent of the C intrinsics is to limit usage of the `bt` instructions to this case, or whether the intent is to enable "full" usage of the `bt` instruction (I don't know the answer).

Amanieu:
After reading some docs, I think the intended usage of this intrinsic is to expose the full BT instruction with an unlimited range. In this case the parameter must be changed to a raw pointer and the inline asm needs to use the "memory" clobber.

gnzlbg:
> After reading some docs,

Do you have a link?

Amanieu:
https://software.intel.com/en-us/node/523362

> Returns the bit in position b of the memory addressed by p.

It makes no mention of being limited to a single word, and I think the original intent of that intrinsic was to expose the BT instruction directly.

alexcrichton:
@gnzlbg FWIW there's already a number of intrinsics that use `&mut __m128` ([example](https://github.com/rust-lang-nursery/stdsimd/blob/f649570cb8d6b2ee56607418b07a00a8c7200f6e/crates/core_arch/src/x86/sse.rs#L1897-L1902)), and I don't think we've been too too consistent about the distinction between `*mut` and `&mut` historically. It should be fine to basically choose whatever's best as a local optimum

mehcode:
Updated.
 * Adjusted the constraints to indicate the read/write nature of the memory.
 * Changed to raw pointers instead of references. 

As this is meant to have access to a memory region and not be limited to a single primitive, a raw pointer makes sense to me.

crates/core_arch/src/x86_64/bt.rs
```diff
@@ -0,0 +1,102 @@
...
+pub unsafe fn _bittestandreset64(p: *mut u64, b: u64) -> u8 {
+    let r: u8;
+    asm!("btrq $2, $1\n\tsetc ${0:b}"
+         : "=r"(r), "+*m"(p)
+         : "r"(b)
+         : "cc", "memory");
+    r
+}
+
+/// Returns the bit in position `b` of the memory addressed by `p`, then compliments that bit.
```
Amanieu: *complements.

Then again, I personally think "inverts" would be clearer in this context.

mehcode: Fixed. I copied (without reading apparently) from the Intel link you added.

mehcode:
How do I fix the `missing cpuid` errors ?

```
failed to verify `_bittest`
  * missing cpuid for _bittest
failed to verify `_bittestandset`
  * missing cpuid for _bittestandset
failed to verify `_bittestandreset`
  * missing cpuid for _bittestandreset
failed to verify `_bittestandcomplement`
  * missing cpuid for _bittestandcomplement
failed to verify `_bittest64`
  * missing cpuid for _bittest64
failed to verify `_bittestandset64`
  * missing cpuid for _bittestandset64
failed to verify `_bittestandreset64`
  * missing cpuid for _bittestandreset64
failed to verify `_bittestandcomplement64`
  * missing cpuid for _bittestandcomplement64
```

gnzlbg:
Is there a `cpuid` bit flag to detect these? (I don't think so?). Otherwise you need to whitelist them here: https://github.com/rust-lang-nursery/stdsimd/blob/master/crates/stdsimd-verify/tests/x86-intel.rs#L248

gnzlbg:
@mehcode ping :)

mehcode:
Build should be green now. I had to use `i32` instead of `u32` to pass the intrinsic verification test.

crates/core_arch/src/x86/bt.rs
```diff
@@ -0,0 +1,102 @@
+#[cfg(test)]
+use stdsimd_test::assert_instr;
+
+/// Returns the bit in position `b` of the memory addressed by `p`.
+#[inline]
+#[cfg_attr(test, assert_instr(bt))]
+#[unstable(feature = "simd_x86_bittest", issue = "0")]
```
gnzlbg: Sorry that it took so long to review this again. Is there a PR for the `simd_x86_bittest` feature to rust-lang/rust ? Otherwise, I'd prefer to use the `stdsimd` feature here for now I think. Or what do you think @alexcrichton ? If we wanted to stabilize this in the near future, then adding their own feature flag might make more sense.

alexcrichton: Eh I think it'd be best to move away from `stdsimd` where we can, so I think landing this as a different feature name is fine. Before landing though we'll want to open an issue at rust-lang/rust to track this and fill in the issue number though

gnzlbg: I've opened a tracking issue https://github.com/rust-lang/rust/issues/59414 

@mehcode can you update the issue number to 59414 ?

PR #709, opened 2019-04-04
Implement all MSA Intrinsics

rbirdic:
Fixes #170

PR #710, opened 2019-04-09
Some MIPS improve

wzssyqa:
Support MIPS r6 triples
Remove workaround for MIPS CI: qemu bug fixed


wzssyqa:
This patch has been included in #709

PR #711, opened 2019-04-09
Automatic verification of MIPS MSA intrinsics

gnzlbg:
This PR adds automatic-verification of the MIPS MSA intrinsic type signatures

cc @rbirdic @wzssyqa

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,7 +89,7 @@ matrix:
     - name: "Documentation"
       install: true
       script: ci/dox.sh
-    - name: "Automatic verification: x86 and x86_64"
+    - name: "Automatic verification: x86 / x86_64 / arm / aarch64 / mips*"
       script: cargo test --manifest-path crates/stdsimd-verify/Cargo.toml
       install: true
     - name: "rustfmt"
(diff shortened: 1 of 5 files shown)


gnzlbg:
The problem is that the C headers that do not include the `__builtin_` use C macros to wrap those, so we can't extract any type information from them, e.g., see: https://github.com/gcc-mirror/gcc/blob/master/gcc/config/mips/msa.h#L52

gnzlbg:
The clang header does the same: https://clang.llvm.org/doxygen/msa_8h_source.html

alexcrichton:
Ah ok that at least means directly using the `__builtin` macros is probably a good idea! Seems fine by me

gnzlbg:
FYI i tried to enable checking the instructions, but something appears to be broken in the code that parses the `assert_instr` macros. It only finds intrinsics in 3/100 intrinsics. I've tried a bit of code golf there, but so much has changed in `syn` in the last years that I'm not really sure if the approach is correct anymore.

alexcrichton:
Ah no worries, if you want to land this an open an issue I can follow-up later poking around to see if I can enable this

gnzlbg:
I'll wait for @rbirdic to review the proposed changes to the `msa` module and merge afterwards.

PR #712, opened 2019-04-09
Fixing the GBA in the build script again

Lokathor:
This isn't ready, just a draft, but this is the easiest way to see what the CI itself is gonna do, so here we go.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,7 +85,7 @@ matrix:
         - cd ..
         # Pull the target spec up into the current directory and then build
         - mv ci/gba.json gba.json
-        - xargo build --all --target=gba
+        - xargo build --all --target=gba.json
     - name: "Documentation"
       install: true
       script: ci/dox.sh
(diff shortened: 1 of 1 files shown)


Lokathor:
Marked it as "ready" just because Travis wasn't running on it when it was draft only, but we're still not necessarily ready. If this current commit doesn't work out by pure luck I unfortunately can't get back to this for at least 36 hours.

Lokathor:
@gnzlbg looking at the latest build log, any way to get lazy_static out of the build?

gnzlbg:
Just explicitly build `core_arch` and `std_detect`, all other libraries don't make sense for GBA.

Lokathor:
@gnzlbg Well it can successfully build `core_arch`, which was the blocker that started this whole ordeal. We should be all good now. You can accept the PR if you don't have any other concerns.

gnzlbg:
It should be able to build `std_detect` as well (it should be empty) for that target. I'm merging this as is, but it might be worth it to investigate what's going on there.

gnzlbg:
Thank you!

PR #714, opened 2019-04-11
Fix verifying instructions for MIPS

### LINKED ISSUE #713: Verification proc macro fails to find many assert_instr
When running the verification test for MIPS MSA, most of the intrinsics have their instructions (content of `assert_instr`) set to empty, such that the test had to be disabled: https://github.com/rust-lang-nursery/stdsimd/blob/master/crates/stdsimd-verify/tests/mips.rs#L317 

It appears that the verification macro here is failing to match many of those: https://github.com/rust-lang-nursery/stdsimd/blob/master/crates/stdsimd-verify/src/lib.rs#L273

alexcrichton:
Looks like MIPS is the first architecture to have verification which
exercises the more flavorful forms of `assert_instr`, so the parsing
code for `assert_instr` needed an update.

Closes #713

### DIFF
--- a/crates/stdsimd-verify/src/lib.rs
+++ b/crates/stdsimd-verify/src/lib.rs
@@ -5,11 +5,12 @@ extern crate quote;
 #[macro_use]
 extern crate syn;
 
+use proc_macro::TokenStream;
+use proc_macro2::Span;
 use std::fs::File;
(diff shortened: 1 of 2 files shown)


gnzlbg:
Thanks!

PR #716, opened 2019-04-13
fix rustdoc link failures

RalfJung:
rustdoc link failures accidentally got downgraded to a warning some time ago, so these could slip in. I hope I got all of them.

### DIFF
--- a/crates/core_arch/src/acle/dsp.rs
+++ b/crates/core_arch/src/acle/dsp.rs
@@ -4,21 +4,21 @@
 //!
 //! Intrinsics that could live here:
 //!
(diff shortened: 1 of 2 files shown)


RalfJung:
CI failures don't look like this PR introduced them.

gnzlbg:
cc @japaric - i think it would be better to just fix all these comments. The lists at the top of each module are not necessary and could be removed, and the docs of the intrinsics use `[0]` and `[1]` to denote the arguments, but then they use `a` and `b` as variable names. The doc comments should only refer to those.

PR #726, opened 2019-04-18
add rtm cpu feature intrinsics

mtak-:
Blocked by https://github.com/rust-lang/rust/pull/60060

This does not include any HLE functions (xacquire/xrelease prefixed operations).

Let me know what needs changing.


crates/core_arch/src/x86/rtm.rs
```diff
@@ -0,0 +1,151 @@
...
+#[inline]
+#[target_feature(enable = "rtm")]
+#[cfg_attr(test, assert_instr(xtest))]
+pub unsafe fn _xtest() -> bool {
+    x86_xtest() != 0
+}
+
+/// Retrieves the parameter passed to [`_xabort`] when [`_xbegin`]'s status has the `_XABORT_EXPLICIT` flag set.
+#[inline]
+pub fn _xabort_code(status: u32) -> u32 {
```
mtak-: Wasn't sure whether to include this, but it was in the clang header. https://github.com/llvm-mirror/clang/blob/38057e002c45d115e9a9b0abf15fdb7d852537b2/lib/Headers/rtmintrin.h#L24

crates/core_arch/src/x86/rtm.rs
```diff
@@ -0,0 +1,151 @@
...
+    (status >> 24) & 0xFF
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::core_arch::x86::*;
+
+    #[test]
```
gnzlbg: Is there a reason not to use `#[simd_test]` here or just an oversight ?

---

`#[simd_test]` does run-time feature detection and executes these only if the hardware supports it.

crates/core_arch/src/x86/rtm.rs
```diff
@@ -0,0 +1,151 @@
+#[cfg(test)]
```
gnzlbg: Can you add a top-level comment here about these ?  E.g. to the spec, and maybe to a wikipedia page or some summary page explaining these ? Is for internal documentation only.

crates/core_arch/src/x86/rtm.rs
mtak-: Oversight! I'll look at the other tests more closely

crates/core_arch/src/x86/rtm.rs
mtak-: Pushed

mtak-:
Thanks! It's merged, but not yet in nightly AFAICT. I'll ping you.

crates/core_arch/src/x86/rtm.rs
```diff
@@ -0,0 +1,163 @@
...
+}
+
+/// Queries whether the processor is executing in a transactional region identified by restricted
+/// transactional memory (RTM) or hardware lock elision (HLE).
+///
+/// [Intel's documentation](https://software.intel.com/en-us/cpp-compiler-developer-guide-and-reference-xtest).
+#[inline]
+#[target_feature(enable = "rtm")]
+#[cfg_attr(test, assert_instr(xtest))]
+pub unsafe fn _xtest() -> u8 {
```
mtak-: Just changed this to return u8 to match intel headers more directly. It is specified to always be 0 or 1.

Additionally, if/when stdsimd later adds HLE intrinsics, `_xtest` should be enabled if either `hle` or `rtm` is supported.

gnzlbg: > Additionally, if/when stdsimd later adds HLE intrinsics, _xtest should be enabled if either hle or rtm is supported.

I don't think this is something that can be easily supported. We need to know which features are available when generating code for this intrinsic, so it is either rtm, or hle, or both. But if its rtm or hle, then we need to generate different code depending on the answer to which one is it.

mtak-: All the intel docs say either HLE or RTM. The [instruction set reference](https://www.intel.de/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-instruction-set-reference-manual-325383.pdf) for `xtest` says under "CPUID Feature Flag": HLE or RTM.

> we need to generate different code depending on the answer to which one is it.

Does that mean we need a duplicate of `_xtest` (`_hletest`) in any future `hle.rs` module?

@gnzlbg This should be ready to run through travis now.

gnzlbg: https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_xtest&expand=6168 says RTM CPUID flag only.

> Does that mean we need a duplicate of _xtest (_hletest) in any future hle.rs module?

That's what it would mean, but that would be very weird. It seems more likely that the link you posted is incorrect, you might want to ask Intel for clarification in their forums.

gnzlbg:
Can you rebase on top of master?

mtak-:
I have already done so. verify signatures on master is broken on CI after #733.
```
failed to verify `_rdtsc`
  * failed to equate: `__int64` and PrimUnsigned(64) for _rdtsc
thread 'verify_all_signatures' panicked at 'assertion failed: all_valid', crates/stdsimd-verify/tests/x86-intel.rs:186:5
```

gnzlbg:
Oh I thought the fix for that already had landed!

gnzlbg:
Thank you @mtak- for working on this!

PR #728, opened 2019-04-19
Tidy Rust2018 PR

gnzlbg:


gnzlbg:
@lu-zero the ppc build job has been flaky recently, it sometimes finish with a segfault, but restarting it fixes that =/ we didn't change anything here, so maybe the last LLVM upgrade broke something =/

PR #729, opened 2019-04-19
Unstabilize AMD SSE4A intrinsics

gnzlbg:


gnzlbg:
I'm closing this, I think we should just stabilize the target features instead.

PR #733, opened 2019-04-19
[breaking change] fix return type of _rdtsc

### LINKED ISSUE #559: x86 _rdtsc has wrong return type
This issue is related to the recently stabilised x86 intrinsic function [`_rdtsc`](https://doc.rust-lang.org/core/arch/x86/fn._rdtsc.html)

The issue is that the function is defined to return a signed `i64`. This is in accordance with [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=rdtsc&expand=4255,4255&othertechs=TSC). However, that is not how most people define it, including C/C++ compilers.

This is how GCC [defines it](https://github.com/gcc-mirror/gcc/blob/f9547530289255c6a4fb6b56c8efb0bcc5e5e1b3/gcc/config/i386/ia32intrin.h#L112):
```c
extern __inline unsigned long long // < note the return type
__attribute__((__gnu_inline__, __always_inline__, __artificial__))
__rdtsc (void)
{
  return __builtin_ia32_rdtsc ();
}
```

This [Clang example](https://gcc.godbolt.org/z/A833Yz) proves it also defines it as `unsigned long long`.

Now, I'm confused. Since the timestamp counter is a monotonically increasing timer, it wouldn't make sense to return negative values. In fact, I think it is allowed to return values with the most significant bit set (which would make the returned value negative, unless transmuted). Is the mismatch intended?

gnzlbg:
Closes #559 . 

The bug was reported to Intel (see #559), they haven't answered yet AFAICT, GCC does fix the bug, and so should we. We should do a crater run, and see if the fix is possible. Hopefully most people are just calling the intrinsic followed by `as u64` anyways to workaround this. We'll see.

### DIFF
--- a/crates/core_arch/src/x86/rdtsc.rs
+++ b/crates/core_arch/src/x86/rdtsc.rs
@@ -22,7 +22,7 @@ use stdsimd_test::assert_instr;
 #[inline]
 #[cfg_attr(test, assert_instr(rdtsc))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
-pub unsafe fn _rdtsc() -> i64 {
+pub unsafe fn _rdtsc() -> u64 {
     rdtsc()
 }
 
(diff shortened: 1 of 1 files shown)


gnzlbg:
I'm going to merge this, and update stdsimd upstream, doing a crater run.

PR #734, opened 2019-04-19
Disable assert_instr tests on Windows

gnzlbg:
They have been failing for too long. No idea what's going on.


gnzlbg:
cc @retep998 - if you know or are in contact with Windows users that might be able to help getting `core::arch` tested on Windows again, please send them our way. Or if there is a windows working group or something that we can ping, let us know.

PR #737, opened 2019-04-23
Implement F16C intrinsics

gnzlbg:
Note: `avx512f` enables `f16c` - once https://github.com/rust-lang/rust/pull/60191 is merged and available on nightly we can switch these to only requiring the `f16c` target-feature instead.

### DIFF
--- a/crates/core_arch/src/x86/f16c.rs
+++ b/crates/core_arch/src/x86/f16c.rs
@@ -1,0 +1,109 @@
+//! F16C intrinsics:
+//! https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=fp16&expand=1769
+
+use crate::{
+    core_arch::{simd::*, x86::*},
+    hint::unreachable_unchecked,
+    mem::transmute,
+};
+
+#[cfg(test)]
+use stdsimd_test::assert_instr;
+
+#[allow(improper_ctypes)]
+extern "unadjusted" {
+    #[link_name = "llvm.x86.vcvtph2ps.128"]
+    fn llvm_vcvtph2ps_128(a: i16x8) -> f32x4;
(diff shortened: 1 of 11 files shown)


crates/std_detect/src/detect/arch/x86.rs
```diff
@@ -177,6 +178,10 @@ macro_rules! is_x86_feature_detected {
         cfg!(target_feature = "avx512vpopcntdq") || $crate::detect::check_for(
             $crate::detect::Feature::avx512_vpopcntdq)
     };
+    ("f16c") => {
+        cfg!(target_feature = "avx512f") || $crate::detect::check_for(
```
alexcrichton: Could a to-do be added here to change avx when stable?

crates/core_arch/src/x86/f16c.rs
```diff
@@ -0,0 +1,134 @@
...
+    fn llvm_vcvtps2ph_128(a: f32x4, rounding: i32) -> i16x8;
+    #[link_name = "llvm.x86.vcvtps2ph.256"]
+    fn llvm_vcvtps2ph_256(a: f32x8, rounding: i32) -> i16x8;
+}
+
+/// Converts the 4 x 16-bit half-precision float values in the lowest 64-bit of
+/// the 128-bit vector `a` into 4 x 32-bit float values stored in a 128-bit wide
+/// vector.
+#[inline]
+#[target_feature(enable = "avx512f")]
```
alexcrichton: Could this and other functions be tagged with Todo to change this?

gnzlbg:
@alexcrichton since the F16C PR upstream has already been r+, I think it is better to not merge this until that lands, and then just update this PR to properly use `f16c`.

gnzlbg:
CI is green again.

PR #739, opened 2019-04-23
Add std_detect::detect::features() API that returns iterator of target-features and reject unstable features in feature-detection macros

### LINKED ISSUE #667: std_detect is_x86_feature_detected macro should reject unstable features if they are not enabled
E.g. "sse4a", we should check all others as well.

gnzlbg:
This PR implements an unstable `features()` function in `std::detect` that returns an iterator over the target features, returning a tuple of the target-feature name, and whether it is enabled or disabled in the current host. 

This can be used to implement a solution to https://github.com/rust-lang/rust/issues/54688 using our own run-time feature detection system.

The refactor required to implement this also allows us to close #667

### DIFF
--- a/ci/azure-install-rust.yml
+++ b/ci/azure-install-rust.yml
@@ -17,7 +17,7 @@ steps:
   - script: |
       @echo on
       if not defined TOOLCHAIN set TOOLCHAIN=nightly
-      rustup update %TOOLCHAIN%-%TARGET%
+      rustup update --no-self-update %TOOLCHAIN%-%TARGET%
       rustup default %TOOLCHAIN%-%TARGET%
     displayName: Install rust (windows)
     condition: eq( variables['Agent.OS'], 'Windows_NT' )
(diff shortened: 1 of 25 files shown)


lu-zero:
Do you plan to rebase this? I could use something like this to map the env var content to the features.

gnzlbg:
Yep, I'll merge this today.

lu-zero:
It rebases cleanly but if I try to build it I get:
```
    Checking core_arch v0.1.5 (/Users/lu_zero/Sources/rust/stdarch/crates/core_arch)
error[E0658]: non-builtin inner attributes are unstable
 --> crates/core_arch/src/simd.rs:3:1
  |
3 | #![rustfmt::skip]
  | ^^^^^^^^^^^^^^^^^
  |
  = note: for more information, see https://github.com/rust-lang/rust/issues/54726
  = help: add `#![feature(custom_inner_attributes)]` to the crate attributes to enable
```

gnzlbg:
Damn I forgot about this.

lu-zero:
Actually I had a problem with my local repo and origin/master was pointing to the wrong place, so a rebase isn't immediate :)

(if you have time to rebase and push would be great)

lu-zero:
If it survives I guess it should be good to go pending some shuffling/amending.

gnzlbg:
Gonna update stdarch upstream and do a crater run. There might be code in the wild accidentally relying on some of the unstable features.

PR #742, opened 2019-04-25
Mips msa build jobs

rbirdic:


gnzlbg:
These are failing with:

> error: could not create home directory: '/.rustup'
> info: caused by: Permission denied (os error 13)

gnzlbg:
@rbirdic tests are green, but just triple-checking: this does not add any `mips64r6` build jobs, so that is not tested. Is that intended?

rbirdic:
@gnzlbg You're right, I did not include mips64r6 build jobs into travis.yml. I will do that.

.travis.yml
```diff
@@ -33,6 +33,10 @@ matrix:
       env: TARGET=mips64-unknown-linux-gnuabi64 NORUN=1
     - name: "mips64el-unknown-linux-gnuabi64 - build-only"
       env: TARGET=mips64el-unknown-linux-gnuabi64 NORUN=1
+    - name: "mipsisa64r6el-unknown-linux-gnuabi64 - no assert_instr"
+      env: TARGET=mipsisa64r6el-unknown-linux-gnuabi64 STDSIMD_DISABLE_ASSERT_INSTR=1
```
gnzlbg: Why is `assert_instr` disabled for these ?

gnzlbg:
It appears that the new targets are not shipped with rustup. You might need to use `xargo` or `cargo xbuild` to build these.

.travis.yml
rbirdic: Assert instruction is not usable because llvm optimizes builtins into other instructions. So, there is no one to one mapping between builtins and resulting instruction.
Yes, this is redundant (already added this into run.sh file) but I wanted to make obvious.
I can remove it.

rbirdic:
@gnzlbg I already added xargo for mipsisa* targets
https://github.com/rust-lang-nursery/stdsimd/blob/84e9a2064abccf0ba9bbdd2117097d5e60e339b1/ci/run.sh#L30-L33
How do I disable rustup target add $TARGET?

.travis.yml
gnzlbg: For which builtins does this happen and for which targets ?

gnzlbg:
The `rustup target add $TARGET` happens in a "global" `install:` key in the travis.yml file. You can just add a `install: true` key to each build job to override the global one (where `true`does nothing and succeeds).

.travis.yml
rbirdic: This happens for mipisa64r6 and mipisa64r6el targets, for following builtins:

```
    core_arch::mips::msa::assert___msa_aver_u_b_aver_u_b
    core_arch::mips::msa::assert___msa_aver_u_d_aver_u_d
    core_arch::mips::msa::assert___msa_aver_u_h_aver_u_h
    core_arch::mips::msa::assert___msa_aver_u_w_aver_u_w
    core_arch::mips::msa::assert___msa_bclr_d_bclr_d
    core_arch::mips::msa::assert___msa_bclri_b_bclri_b
    core_arch::mips::msa::assert___msa_binsli_b_binsli_b
    core_arch::mips::msa::assert___msa_binsli_d_binsli_d
    core_arch::mips::msa::assert___msa_binsli_h_binsli_h
    core_arch::mips::msa::assert___msa_binsli_w_binsli_w
    core_arch::mips::msa::assert___msa_binsri_b_binsri_b
    core_arch::mips::msa::assert___msa_binsri_d_binsri_d
    core_arch::mips::msa::assert___msa_binsri_h_binsri_h
    core_arch::mips::msa::assert___msa_binsri_w_binsri_w
    core_arch::mips::msa::assert___msa_bmnzi_b_bmnzi_b
    core_arch::mips::msa::assert___msa_bmz_v_bmz_v
    core_arch::mips::msa::assert___msa_bmzi_b_bmzi_b
    core_arch::mips::msa::assert___msa_bneg_d_bneg_d
    core_arch::mips::msa::assert___msa_bsel_v_bsel_v
    core_arch::mips::msa::assert___msa_bset_d_bset_d
    core_arch::mips::msa::assert___msa_clei_s_d_clei_s_d
    core_arch::mips::msa::assert___msa_clti_s_d_clti_s_d
    core_arch::mips::msa::assert___msa_copy_u_d_copy_u_d
    core_arch::mips::msa::assert___msa_dpadd_u_h_dpadd_u_h
    core_arch::mips::msa::assert___msa_ld_b_ld_b
    core_arch::mips::msa::assert___msa_ld_h_ld_h
    core_arch::mips::msa::assert___msa_ld_w_ld_w
    core_arch::mips::msa::assert___msa_ldi_d_ldi_d
    core_arch::mips::msa::assert___msa_ldi_h_ldi_h
    core_arch::mips::msa::assert___msa_ldi_w_ldi_w
    core_arch::mips::msa::assert___msa_maxi_s_d_maxi_s_d
    core_arch::mips::msa::assert___msa_mini_s_d_mini_s_d
    core_arch::mips::msa::assert___msa_ori_b_ori_b
    core_arch::mips::msa::assert___msa_slli_b_slli_b
    core_arch::mips::msa::assert___msa_srli_b_srli_b
    core_arch::mips::msa::assert___msa_st_b_st_b
    core_arch::mips::msa::assert___msa_st_h_st_h
    core_arch::mips::msa::assert___msa_st_w_st_w
```
Example:
```

disassembly for msa_st_w_shim_st_w: 
	 0: lui at,0x14 
	 1: ldi.b $w0,0 
	 2: daddiu v1,zero,42 
	 3: daddu at,at,t9 
	 4: insert.d $w0[0],a0 
	 5: daddiu at,at,-18472 
	 6: ld v0,-11128(at) 
	 7: ld at,-11136(at) 
	 8: insert.d $w0[1],a1 
	 9: daddiu v0,v0,-3997 
	10: sd v1,8(at) 
	11: sd v0,0(at) 
	12: daddiu at,a2,2044 
	13: jr ra 
	14: st.d $w0,0(at) 
thread 'core_arch::mips::msa::assert___msa_st_w_st_w' panicked at 'failed to find instruction `st.w` in the disassembly'
```

gnzlbg: So why isn't `st.w` being generated? Is the generated code more efficient for a CPU with MSA than the MSA instruction itself? Do we have a bug in the toolchain ? (In Rust, LLVM, etc.)

rbirdic: I don't know if the generated code is more efficient (it should be the same), but it makes no difference on little-endian target. It's LLVM thing. Because of calling convention every vector gets passed the same as it would be v2i64, so I am guessing that this makes LLVM prefer st.d over st.w.

gnzlbg: > Because of calling convention every vector gets passed the same as it would be v2i64,

Is a different calling convention possible? The vector types appear to be properly defined (using different element types and number of elements). 

Looking at the LLVM intrinsics, they appear to be incorrectly declared to use the `"C"` ABI, and should use the unnamed ABI instead (e.g. here: https://github.com/rust-lang-nursery/stdsimd/blob/master/crates/core_arch/src/mips/msa.rs#L71). But I don't know if that will make a difference for these particular targets.

If that isn't causing this (or if that makes no sense), maybe `stdsimd-verify` should just check that the `prefix` in the `prefix.suffix` instruction names is found, instead of trying to match `prefix.suffix`.

rbirdic: > If that isn't causing this (or if that makes no sense), maybe stdsimd-verify should just check that the prefix in the prefix.suffix instruction names is found, instead of trying to match prefix.suffix (maybe we should just change the assert_instr to only contain the prefix).

This would help with st.w but the other issues.
Some of the other issues https://github.com/Microsoft/llvm/blob/master/lib/Target/Mips/MSA.txt
I think that this would not be possible to reliably assert every builtin on all optimization levels.
x86 has done the same for AVX.

gnzlbg: > I think that this would not be possible to reliably assert every builtin on all optimization levels.

`assert_instr` is only checked in `--release`

rbirdic: If  `prefix` would be used instead of `prefix.suffix `in `assert_instr`, release mode build would be successful, but the problem would appear in stdsimd-verify:
https://github.com/rust-lang-nursery/stdsimd/blob/master/crates/stdsimd-verify/tests/mips.rs#L315-L325

gnzlbg: So maybe the best we can do is, for mips targets, in the `assert_instr` macro, only check the prefix. That way, the prefix.suffix is kept in the `assert_instr` annotation, and `stdsimd-verify` has all the information. 

That could happen somewhere around here: https://github.com/rust-lang-nursery/stdsimd/blob/master/crates/stdsimd-test/src/lib.rs#L88

rbirdic: There is another problem. For some builtins, even this is not enough.
For example, for a binsli.d builtin, the situation is as follows:

```
disassembly for msa_binsli_d_shim_binsli_d: 
	 0: ldi.b $w0,0 
	 1: lui at,0x14 
	 2: daddiu a1,zero,58 
	 3: insert.d $w0[0],a2 
	 4: daddu at,at,t9 
	 5: daddiu at,at,-4200 
	 6: insert.d $w0[1],a3 
	 7: ld a0,-10752(at) 
	 8: ld at,-10744(at) 
	 9: copy_s.d v0,$w0[0] 
	10: copy_s.d v1,$w0[1] 
	11: daddiu at,at,22660 
	12: sd a1,8(a0) 
	13: jr ra 
	14: sd at,0(a0) 
thread 'core_arch::mips::msa::assert___msa_binsli_d_binsli_d' panicked at 'failed to find instruction `binsli.d` in the disassembly'
```
The `prefix` does not appear, so, checking only `prefix` will not solve the problem for this situation.
Most of the problematic builtins are listed here:
https://github.com/Microsoft/llvm/blob/master/lib/Target/Mips/MSA.txt

gnzlbg: So what we do for other targets here, is comment the `assert_instr` out, and white-list those in `stdsimd-verfiy` as "its ok that this does not have an assert_instr` because of X". 

The comment that comments the `assert_instr` out, should explain why: is this an LLVM bug (link to the bug) ? Is LLVM generating more optimized code here? Then explain why this code is more optimized, add an assert_instr for the more optimized version, and white-list that in stdsimd-verify, etc. 

---

Off-topic: it's a bit off-topic, but I think that when this PR is merged, we can start talking about stabilizing this intrinsics. That might need a small RFC, or an FCP, but I can help with that and with pushing this through the process. You have done an awesome job here, implementing all intrinsics, adding tests, automatic-verification, etc. I've barely have had to help you, and I think when this is stabilized, maybe it would be worthy of being part of the rust blog release announcement to explain what you did, how you did it, how much time it costed, what were the issues that we could do better process wise, etc. Right now, we only have x86 intrinsics, and showing that we are committed to shipping intrinsics for other archs like WASM, MIPS, ARM, etc. would be a good thing IMO.

rbirdic:
@gnzlbg 
I currently have a problem with the xargo build.
The error I get is as follows:
```
Running `rustc --edition=2018 --crate-name rustc_std_workspace_alloc /cargo-h/registry/src/github.com-1ecc6299db9ec823/rustc-std-workspace-alloc-1.0.0/src/lib.rs --color always --crate-type lib --emit=dep-info,link -C opt-level=3 -C debuginfo=2 -C metadata=6f00195cab321fc8 -C extra-filename=-6f00195cab321fc8 --out-dir /tmp/xargo.LSLHMki8mX7L/target/mipsisa64r6el-unknown-linux-gnuabi64/release/deps --target mipsisa64r6el-unknown-linux-gnuabi64 -C linker=mipsisa64r6el-linux-gnuabi64-gcc -L dependency=/tmp/xargo.LSLHMki8mX7L/target/mipsisa64r6el-unknown-linux-gnuabi64/release/deps -L dependency=/tmp/xargo.LSLHMki8mX7L/target/release/deps --cap-lints allow --sysroot /root/.xargo -Z force-unstable-if-unmarked`
error[E0463]: can't find crate for `std`

For more information about this error, try `rustc --explain E0463`.
error: Could not compile `rustc-std-workspace-alloc`.

Caused by:
  process didn't exit successfully: `rustc --edition=2018 --crate-name rustc_std_workspace_alloc /cargo-h/registry/src/github.com-1ecc6299db9ec823/rustc-std-workspace-alloc-1.0.0/src/lib.rs --color always --crate-type lib --emit=dep-info,link -C opt-level=3 -C debuginfo=2 -C metadata=6f00195cab321fc8 -C extra-filename=-6f00195cab321fc8 --out-dir /tmp/xargo.LSLHMki8mX7L/target/mipsisa64r6el-unknown-linux-gnuabi64/release/deps --target mipsisa64r6el-unknown-linux-gnuabi64 -C linker=mipsisa64r6el-linux-gnuabi64-gcc -L dependency=/tmp/xargo.LSLHMki8mX7L/target/mipsisa64r6el-unknown-linux-gnuabi64/release/deps -L dependency=/tmp/xargo.LSLHMki8mX7L/target/release/deps --cap-lints allow --sysroot /root/.xargo -Z force-unstable-if-unmarked` (exit code: 1)
warning: build failed, waiting for other jobs to finish...
error: build failed
error: `"cargo" "build" "--release" "--manifest-path" "/tmp/xargo.LSLHMki8mX7L/Cargo.toml" "--target" "mipsisa64r6el-unknown-linux-gnuabi64" "-v" "-p" "std"` failed with exit code: Some(101)
note: run with `RUST_BACKTRACE=1` for a backtrace
root@7f82a80d3d7f:/checkout# exit

```
I found something about this, but I can not solve the problem:
https://github.com/japaric/xargo/issues/240
Do you maybe have any suggestion?

gnzlbg:
> Do you maybe have any suggestion?

Do you have the rust-src component installed ? (`rustup install rust-src` or something like that?) IIRC `xargo` needs to build libcore/liballoc/libstd, so it needs their source code for that. I don't remember if it pulls it automatically for you, or if you had to install that manually, but you can try running the script manually on your machine and checking whether that passes.

crates/stdsimd-test/src/lib.rs
```diff
@@ -154,7 +154,9 @@ pub fn assert(fnptr: usize, fnname: &str, expected: &str) {
                 // Original limit was 20 instructions, but ARM DSP Intrinsics
                 // are exactly 20 instructions long. So, bump the limit to 22
                 // instead of adding here a long list of exceptions.
-                _ => 22,
+		// Some MIPS MSA Intrinsics are 22 instructions long,
+		// so, previous limit of 22 is increased to 23.
+                _ => 23,
```
rbirdic: @gnzlbg This limit of 23 instructions suits for the mipsisa64r6el target.
For mipsisa64r6 target, there are 98 intrinsics that are longer than this limit.
The length of these intrinsic is as follows:

```
10 intrinsics are 24 instructions long
54 intrinsics are 25 instructions long
24 intrinsics are 26 instructions long
4 intrinsics are 28 instructions long
6 intrinsics are 30 instructions long
```
Is it ok to just increase this limit to 31 or it is better to add a list of exceptions for these intrinsics?

rbirdic:
> > Do you maybe have any suggestion?
> 
> Do you have the rust-src component installed ? (`rustup install rust-src` or something like that?) IIRC `xargo` needs to build libcore/liballoc/libstd, so it needs their source code for that. I don't remember if it pulls it automatically for you, or if you had to install that manually, but you can try running the script manually on your machine and checking whether that passes.

This is fixed with new

crates/stdsimd-test/src/lib.rs
gnzlbg: The exceptions are here: https://github.com/rust-lang-nursery/stdsimd/blob/master/crates/stdsimd-test/src/lib.rs#L153

You can just add a new match arm there, that does something like:

```rust
 _ if target.contains("mips") => 31,
```

and make this exception specific for mips only.

rbirdic:
@gnzlbg Sorry. I accidentally closed this PR.

crates/stdsimd-test/src/lib.rs
rbirdic: I used `fnname` and this solves my problem:

`_ if fnname.contains("msa") => 31,`

rbirdic:
@gnzlbg 
Build for mipsisa64r6* targets fails with:
```
+ cmd=/cargo-h/bin/xargo test --target=mipsisa64r6-unknown-linux-gnuabi64 --manifest-path=crates/core_arch/Cargo.toml
+ cmd=/cargo-h/bin/xargo test --target=mipsisa64r6-unknown-linux-gnuabi64 --manifest-path=crates/core_arch/Cargo.toml -- 
+ /cargo-h/bin/xargo test --target=mipsisa64r6-unknown-linux-gnuabi64 --manifest-path=crates/core_arch/Cargo.toml --
error: couldn't lock mipsisa64r6-unknown-linux-gnuabi64's sysroot as read-only
caused by: Permission denied (os error 13)
```
I can not reproduce this on my setup.
Problem seems to be in this line:
https://github.com/japaric/xargo/blob/5f673d9a74607e03702350e64a53758a7d4e18fa/src/xargo.rs#L69
Below is strace that should correspond to it:

```
openat(AT_FDCWD, "/root/.xargo/lib/rustlib/mipsisa64r6el-unknown-linux-gnuabi64/.sentinel", O_RDWR|O_CREAT|O_CLOEXEC, 0666) = 3
statfs("/root/.xargo/lib/rustlib/mipsisa64r6el-unknown-linux-gnuabi64/.sentinel", {f_type=OVERLAYFS_SUPER_MAGIC, f_bsize=4096, f_blocks=23997379, f_bfree=7249000, f_bavail=6028316, f_files=6111232, f_ffree=4783234, f_fsid={val=[4004146751, 385313266]}, f_namelen=255, f_frsize=4096, f_flags=ST_VALID|ST_RELATIME}) = 0
flock(3, LOCK_EX|LOCK_NB) 
```
I'm not sure what syscall could fail here.
Do you maybe have any suggestion?

rbirdic:
@gnzlbg
Did you maybe look at this?

gnzlbg:
>/ root/.xargo/lib/rustlib/mipsisa64r6el-unknown-linux-gnuabi64/.

So it appears that `xargo` is trying to write to that directory, but this directory is probably read only. That happens in `run_docker.sh` around here: https://github.com/rust-lang-nursery/stdsimd/blob/master/ci/run-docker.sh#L19

Check out the `--volume` parameters, if a volume (e.g. a directory on the HOST where the docker container runs) is not mounted, you can read or write from/to it. If it's mounted as `ro`, then it's read-only, etc. 

If you are testing locally using `./ci/run_docker.sh mipsisa64r6el-unknown-linux-gnuabi64` you should be seeing this same error. Let me know if you can reproduce it with that. A way to fix it might be to add the `.xargo` directory as a volume to `run_docker.sh`.

rbirdic:
@gnzlbg 
Yes, I am testing in that way, with that command and cannot reproduce it.
I added a new patch which provides mounting `.xargo`directory, but the same problem still exists.
Now docker command looks like:
`docker run --user 2000:2000 --rm --init --volume /home/travis/.cargo:/cargo-h --volume /home/travis/.xargo:/root/.xargo ...`
Is this the right path `--volume /home/travis/.xargo:/root/.xargo` ?

gnzlbg:
> Is this the right path --volume /home/travis/.xargo:/root/.xargo ?

I don't think so, I think the path should be just `.xargo` (local to the current directory).

> Yes, I am testing in that way, with that command and cannot reproduce it.

Weird. @dylan-dpc @japaric @jamesmunns could you be able to take a look ? I think this line in cross might hint at what to do: https://github.com/rust-embedded/cross/blob/540dc4f5a1ca8d6d95bbf4da66ce763a2352d36f/src/docker.rs#L84

gnzlbg:
This needs rebasing.

gnzlbg:
Could you please rebase this on top of master?

rbirdic:
@gnzlbg Yes, I will do this today.

gnzlbg:
@rbirdic you need to "migrate" the changes from the `.travis.yml` file to `azure-pipelines.yml`, and also need to rebase the changes to `run-docker.sh` and `run.sh`.

rbirdic:
@gnzlbg I have a question about `azure-pipelines.yml`
Which is the right way to add the following options for mipsisa* architectures("cargo install xargo" etc.)?:
https://github.com/rust-lang/stdarch/blob/f01110afb3acb452a65eb7385c21d38a131acd7e/.travis.yml#L37:L40
Is something like this appropriate?:
```
mipsisa64r6el-unknown-linux-gnuabi64:
   TARGET: mipsisa64el-unknown-linux-gnuabi64
steps:
  - script: cargo install xargo
  displayName: Install xargo
  - script: rustup component add rust-src
  displayName: Add rust-src
```

gnzlbg:
@rbirdic just add them always at the top level for all targets, I'll refactor them later when this is merged

Amanieu:
Hi, are you still working on this PR. I can merge it if you rebase.

rbirdic:
> 
> 
> Hi, are you still working on this PR. I can merge it if you rebase.

Hi, unfortunately, I don't work on this project anymore.

PR #746, opened 2019-04-26
[breaking change] Fix pointer validation in stdsimd-verify

gnzlbg:
@alexcrichton this PR makes stdsimd-verify treat mut and const pointer and references differently. We had a breaking changes due to incorrect pointer mutability already (e.g. #576), and I hope this will prevent at least the more obvious ones as the AVX-512 intrinsics get added. 

Two breaking changes are introduced:

* #743 is fixed, which means two API breaking changes to those intrinsics
* `_mm_clflush` had an incorrect API, the pointer should have been a `const` pointer, but it was a `mut` one

I'll propose to merge something like this to stdsimd, and then do a crater run in the PR to rust-lang/rust.

### DIFF
--- a/crates/core_arch/src/lib.rs
+++ b/crates/core_arch/src/lib.rs
@@ -17,7 +17,6 @@
     rustc_attrs,
     stdsimd,
     staged_api,
-    align_offset,
     maybe_uninit,
     doc_cfg,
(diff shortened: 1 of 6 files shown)


gnzlbg:
cc @dchammond

PR #751, opened 2019-05-04
Add Altivec vec_max

lu-zero:
Since `paste` seems not ideal I came up with some creative macro to reduce a little more the boilerplating.


gnzlbg:
LGTM!

lu-zero:
The  failed checks seem unrelated, is it ok to merge?

lu-zero:
(actually the other set)

PR #752, opened 2019-05-05
Altivec update

lu-zero:
Here a larger set with additional macros to reduce the number of lines of code.

- [x] vec_max
- [x] vec_min
- [x] vec_sub
- [x] vec_subs
- [x] vec_splats
- [x] vec_abs
- [x] vec_abss

### DIFF
--- a/ci/docker/powerpc64-unknown-linux-gnu/Dockerfile
+++ b/ci/docker/powerpc64-unknown-linux-gnu/Dockerfile
@@ -6,6 +6,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         qemu-system-ppc file make
 
 ENV CARGO_TARGET_POWERPC64_UNKNOWN_LINUX_GNU_LINKER=powerpc64-linux-gnu-gcc \
-    CARGO_TARGET_POWERPC64_UNKNOWN_LINUX_GNU_RUNNER="qemu-ppc64 -L /usr/powerpc64-linux-gnu" \
+    CARGO_TARGET_POWERPC64_UNKNOWN_LINUX_GNU_RUNNER="qemu-ppc64 -cpu power9 -L /usr/powerpc64-linux-gnu" \
     CC=powerpc64-linux-gnu-gcc \
     OBJDUMP=powerpc64-linux-gnu-objdump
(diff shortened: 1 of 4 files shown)


lu-zero:
I'm seeing 
```
failures:
    core_arch::x86::rtm::tests::test_xabort
    core_arch::x86::rtm::tests::test_xbegin_xend
    core_arch::x86::rtm::tests::test_xtest
```

```
error: called `filter_map(p).next()` on an `Iterator`. This is more succinctly expressed by calling `.find_map(p)` instead.
   --> crates/stdsimd-verify/src/lib.rs:319:5
    |
319 | /     attrs
320 | |         .iter()
321 | |         .flat_map(|a| {
322 | |             if let Some(a) = a.interpret_meta() {
...   |
338 | |         })
339 | |         .next()
    | |_______________^
```

And this is not something I touched.

If this extended set is fine for you I'd land it and then push more instructions :)

lu-zero:
the failed test
```
    core_arch::x86::rtm::tests::test_xabort
    core_arch::x86::rtm::tests::test_xbegin_xend
    core_arch::x86::rtm::tests::test_xtest
```

Seems unrelated to this patch.

lu-zero:
@gnzlbg I would push this set today.

gnzlbg:
The apple build failure looked spurious, so I've restarted it, but the ppc failures look legit :/

lu-zero:
```
qemu: uncaught target signal 4 (Illegal instruction) - core dumped
```

It is quite fishy (since abss is the same as abs but with the saturating sub)

Obviously it does not happen on real hardware.

gnzlbg:
That might be the case. It might also mean that the version of QEMU used just does not support that instruction, and needs to be updated, or maybe some other thing. 

In any case, CI is currently green, we shouldn't break it again. Maybe extract those changes into a different PR and investigate further there?

lu-zero:
I'll try to drop one patch after the other.

Is it possible to update the qemu in the docker image?

gnzlbg:
> Is it possible to update the qemu in the docker image?

It is possible, but it might be easier to update, e.g., the ubuntu version, to one that just ships a newer version of qemu. 

It might be helpful to find out first what the bug actually is.

lu-zero:
I tested locally installing docker and friends and seems that the problem is `vec_subs`.
bumping to the newer ubuntu gets us `qemu-3.1`, now I'm setting up a cross-environment to test  locally qemu-4.

lu-zero:
Apparently the qemu quirky behaviour stops if you target a non-generic cpu.

gnzlbg:
Thanks!

PR #762, opened 2019-05-13
add Hygon Dhyana support to enable sse4a/tbm feature

hygonsoc:
add Hygon Dhyana CPU Vendor ID("HygonGenuine") checking to enable Hygon sse4a/TBM feature

As Hygon Dhyana originates from AMD technology and shares most of the architecture with AMD's family 17h, but with different CPU Vendor ID("HygonGenuine")/Family series number(Family 18h).

for CPUID feature bits, Hygon Dhyana(family 18h) share the same definition with AMD family 17h.
AMD CPUID specification is https://www.amd.com/system/files/TechDocs/25481.pdf.

Related Hygon kernel patch can be found on
http://lkml.kernel.org/r/5ce86123a7b9dad925ac583d88d2f921040e859b.1538583282.git.puwen@hygon.cn

### DIFF
--- a/crates/std_detect/src/detect/os/x86.rs
+++ b/crates/std_detect/src/detect/os/x86.rs
@@ -234,8 +234,18 @@ fn detect_features() -> cache::Initializer {
         // The `is_x86_feature_detected!("lzcnt")` macro then
         // internally maps to Feature::abm.
         enable(extended_proc_info_ecx, 5, Feature::abm);
-        if vendor_id == *b"AuthenticAMD" {
-            // These features are only available on AMD CPUs:
+        // As Hygon Dhyana originates from AMD technology and shares most of the architecture with
+        // AMD's family 17h, but with different CPU Vendor ID("HygonGenuine")/Family series
+        // number(Family 18h).
+        //
+        // For CPUID feature bits, Hygon Dhyana(family 18h) share the same definition with AMD
+        // family 17h.
+        //
+        // Related AMD CPUID specification is https://www.amd.com/system/files/TechDocs/25481.pdf.
+        // Related Hygon kernel patch can be found on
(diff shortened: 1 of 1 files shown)


crates/std_detect/src/detect/os/x86.rs
```diff
@@ -234,7 +234,7 @@ fn detect_features() -> cache::Initializer {
         // The `is_x86_feature_detected!("lzcnt")` macro then
         // internally maps to Feature::abm.
         enable(extended_proc_info_ecx, 5, Feature::abm);
-        if vendor_id == *b"AuthenticAMD" {
+        if vendor_id == *b"AuthenticAMD" || vendor_id == *b"HygonGenuine" {
             // These features are only available on AMD CPUs:
```
gnzlbg: This comment should be updated. It would be nice to add a link to AMD and Hygon cpuid specifications, so that it is easy to verify that these are the correct cpuid bits to read for these features in these CPUs.

gnzlbg: @hygonsoc did you forgot to push a comment resolving this comment ?

hygonsoc: @gnzlbg already push the commit resolving this comment, thanks!

gnzlbg: Hmm, I don't think its pushed. Like, I meant that the comment in the code should be updated with the information, not only the commit message 😆

hygonsoc: @gnzlbg sorry for the misunderstanding, update the comment in codes in commit f7da713, thanks!

gnzlbg:
Thank you!

PR #763, opened 2019-05-14
Altivec update 2

lu-zero:
Another batch of functions

- [x] vec_addc
- [x] vec_adds
- [x] vec_and
- [x] vec_andc
- [x] vec_avg
- [x] vec_ceil
- [x] vec_cmpb
- [x] vec_cmpeq
- [x] vec_cmpge
- [x] vec_cmpgt

### DIFF
--- a/crates/core_arch/src/lib.rs
+++ b/crates/core_arch/src/lib.rs
@@ -17,7 +17,6 @@
     rustc_attrs,
     stdsimd,
     staged_api,
-    maybe_uninit,
     doc_cfg,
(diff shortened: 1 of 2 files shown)


lu-zero:
There is another spurious failure because of android.

lu-zero:
The android runners keep timing out. Can we please demote them until somebody unbreak them?

lu-zero:
The wasm target is broken. I'd land this set since it is growing a bit too much.

PR #764, opened 2019-05-22
Drop maybe_uninit

lu-zero:


lu-zero:
`wasm32-unknown-unknown` seems to fail to download or unpack something.

PR #765, opened 2019-05-23
Fix build of auxvec.rs on FreeBSD/powerpc64

t6:
```
error[E0432]: unresolved import `mem`
  --> src/libstd/../stdsimd/crates/std_detect/src/detect/os/freebsd/auxvec.rs:45:9
   |
45 |     use mem;
   |         ^^^ no `mem` external crate

For more information about this error, try `rustc --explain E0432`.
error: Could not compile `std`.
```
Tested by @pkubaj in https://reviews.freebsd.org/D20332

### DIFF
--- a/crates/std_detect/src/detect/os/freebsd/auxvec.rs
+++ b/crates/std_detect/src/detect/os/freebsd/auxvec.rs
@@ -42,7 +42,7 @@ pub(crate) fn auxv() -> Result<AuxVec, ()> {
 
 /// Tries to read the `key` from the auxiliary vector.
 fn archauxv(key: usize) -> Result<usize, ()> {
-    use mem;
+    use crate::mem;
 
     #[derive (Copy, Clone)]
     #[repr(C)]


gnzlbg:
Can you also add a couple of build jobs for these targets ? Check out the `ci/build.sh` script of the `rust-lang/libc` crate for how to do that.

t6:
Unlikely to happen anytime soon from my side. powerpc64 support is not a high priority for me (I don't even have hardware). Maybe @pkubaj or @mikaelurankar can do it.

gnzlbg:
Without having at least a build job that makes sure that these targets build correctly, chances are that the next commit to stdsimd will break them again.

pkubaj:
Do we need to donate some hardware (local or remote) for this? I can't do that myself, but I could talk with Raptor Engineering about that, maybe they could give some VPS for tests.

gnzlbg:
You should be able to just cross-compile to the targets from Linux on travis. That should check that they at least build without build errors. 

Running tests is a different story, but with Cirrus-CI you have native FreeBSD 12 amd64 VMs, so you can test i686 and x86_64-unknown-freebsd there. For ppc64 and aarch64 on FreeBSD, you could run qemu-user or similar natively in the FreeBSD12 VMs just like we do on our linux vms.

pkubaj:
`qemu-user` for powerpc* platforms is broken on FreeBSD.

gnzlbg:
Damn. Do you know if it works for aarch64 ?

pkubaj:
@MikaelUrankar probably knows :)

MikaelUrankar:
> Damn. Do you know if it works for aarch64 ?

No, qemu-user is broken and not actively maintained on our side.

gnzlbg:
Is there an alternative on FreeBSD ? I suppose jails only work for the same target as the host right ?

MikaelUrankar:
You are correct for the jails. You can still run a guest with "full emulation" with qemu-system-ppc64 or qemu-system-aarch64 but it's damn slow.

gnzlbg:
Does freebsd have a cross-compilation toolchain (with headers, etc.) from `x86_64-unknown-freebsd` to the `ppc64` and `aarch64` targets ? We only need to run the tests under qemu-system, and that's quite fast, as long as we can cross-compile properly for those targets before. 

If we had to compile under qemu-system, e.g., because there is no cross-compilation toolchain, then that would be impractically slow.

MikaelUrankar:
I used clang to cross compile rust for armv6/7/8 using the documentation at [1]
'make buildworld' is not necessary to build the base system, you can grab some tarballs [2] [3] and extract them somewhere.

For powerpc64, it's possible to cross compile C code but not C++  (I had to 'ssh' build jobs on my powermac g5 to cross-compile rust)

[1] https://gist.github.com/dumbbell/b587da50ef014078da9e732a4331ebad
[2] ftp://ftp.freebsd.org/pub/FreeBSD/releases/powerpc/powerpc64/11.2-RELEASE/base.txz
[3] ftp://ftp.freebsd.org/pub/FreeBSD/releases/arm64/aarch64/11.2-RELEASE/base.txz

gnzlbg:
So for `aarch64` that looks actually quite good. Using Cirrus we could ship a jail that sets the cross-compilation environment up, and then we should just be able to cross-compile `core_arch` and `std_detect` tests using `xargo` which should re-compile libcore/liballoc/libstd from scratch first. Then we need an `aarch64` FreeBSD image that we can set up in qemu system, run the tests in there, and copy the results back. We do this for `s390x` and `sparc64` in libc (check [libc/ci/test-runner-linux](https://github.com/rust-lang/libc/blob/master/ci/test-runner-linux#L24) and the `linux-s390x.sh` and `linux-sparc64.sh` files for how we set that up as well as the docker containers for those targets).

PR #766, opened 2019-05-26
Altivec update (again)

lu-zero:
- vec_cmple
- vec_cmplt
- vec_expte
- vec_floor

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -113,6 +113,7 @@ matrix:
 
   allow_failures:
     - name: "powerpc-unknown-linux-gnu - no assert_instr"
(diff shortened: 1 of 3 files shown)


lu-zero:
wasm and arm aren't green. One has a codegen fault while the other timedout.

gnzlbg:
LGTM, wasm32 is failing because nightly is broken.

lu-zero:
thank you!

gnzlbg:
ppc64 big endian has a failure, I don't remember why, but it was in allowed failures.

lu-zero:
ppc64 big endiang is not really supported by qemu and I'm not keen in fixing it =/

PR #767, opened 2019-06-03
Stabilize `unreachable`.

pepyakin:
Tracking issue: https://github.com/rust-lang/rust/issues/61119


gnzlbg:
The wasm job is failing, but that appears unrelated.

PR #768, opened 2019-06-07
Fix libcore build

gnzlbg:


gnzlbg:
Something is awfully wrong with the disassembly on linux :/

PR #769, opened 2019-06-22
Adjust #[doc(include)] paths for rustdoc change

jonas-schievink:
Needed for https://github.com/rust-lang/rust/pull/60938 to land


gnzlbg:
Thanks, the library tests currently do not pass on nightly, so I'm not sure updating stdsimd upstream can be done until #768 is fixed.

jonas-schievink:
Yeah I've seen. Unfortunate, but I can wait until that's resolved.

PR #772, opened 2019-07-07
Fix typo in crates/core_arch/src/x86/sse2.rs

JohnTitor:
Pointed out in https://github.com/rust-lang/rust/issues/62454.


gnzlbg:
Thanks!

PR #774, opened 2019-07-08
Fix assert_instr tests

gnzlbg:
This fixes the assert_instr tests, at least on macosx. It's WIP because it will need a bit of code golf till it works everywhere else.

I have no idea what was wrong before, nor what was the code actually doing. 

What we do now is the following: dump the binary with the appropriate tool (e.g. objdump), parse the instruction dumps of the shims into a hashset of (function_name -> instructions), and then query this hashset for the shim name. 

I have no idea what was `backtrace` used for. We know the name of the shim we want to disassembly, so there is no need for us to inspect the backtrace AFAICT.

### DIFF
--- a/crates/assert-instr-macro/src/lib.rs
+++ b/crates/assert-instr-macro/src/lib.rs
@@ -43,13 +43,23 @@ pub fn assert_instr(
     // testing for.
     let disable_assert_instr = std::env::var("STDSIMD_DISABLE_ASSERT_INSTR").is_ok();
 
+    // If instruction tests are disabled avoid emitting this shim at all, just
+    // return the original item without our attribute.
+    if !cfg!(optimized) || disable_assert_instr {
+        return (quote! { #item }).into();
+    }
+
     let instr_str = instr
         .replace('.', "_")
         .replace('/', "_")
         .replace(':', "_")
         .replace(char::is_whitespace, "");
     let assert_name = syn::Ident::new(&format!("assert_{}_{}", name, instr_str), name.span());
-    let shim_name = syn::Ident::new(&format!("{}_shim_{}", name, instr_str), name.span());
+    // These name has to be unique enough for us to find it in the disassembly later on:
+    let shim_name = syn::Ident::new(
+        &format!("stdsimd_test_shim_{}_{}", name, instr_str),
+        name.span(),
+    );
     let mut inputs = Vec::new();
     let mut input_vals = Vec::new();
     let ret = &func.decl.output;
@@ -100,7 +110,8 @@ pub fn assert_instr(
     let to_test = quote! {
         #attrs
         #[no_mangle]
-        unsafe extern #abi fn #shim_name(#(#inputs),*) #ret {
+        #[inline(never)]
+        pub unsafe extern #abi fn #shim_name(#(#inputs),*) #ret {
             // The compiler in optimized mode by default runs a pass called
             // "mergefunc" where it'll merge functions that look identical.
             // Turns out some intrinsics produce identical code and they're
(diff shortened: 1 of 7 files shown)


crates/assert-instr-macro/src/lib.rs
```diff
@@ -49,7 +49,7 @@ pub fn assert_instr(
         .replace(':', "_")
         .replace(char::is_whitespace, "");
     let assert_name = syn::Ident::new(&format!("assert_{}_{}", name, instr_str), name.span());
-    let shim_name = syn::Ident::new(&format!("{}_shim_{}", name, instr_str), name.span());
+    let shim_name = syn::Ident::new(&format!("stdsimd_test_shim_{}_{}", name, instr_str), name.span());
```
gnzlbg: turns out `shim` is a really common name, and I was getting collisions against a `vtable.shim` inside `std::once::Once`... this disambiguates that.

TODO: add a comment here, or refactor this.

crates/assert-instr-macro/src/lib.rs
```diff
@@ -100,7 +100,8 @@ pub fn assert_instr(
     let to_test = quote! {
         #attrs
         #[no_mangle]
-        unsafe extern #abi fn #shim_name(#(#inputs),*) #ret {
+        #[inline(never)]
+        pub unsafe extern #abi fn #shim_name(#(#inputs),*) #ret {
```
gnzlbg: I'm not really sure why this wasn't `#[inline(never)]` before, but we really do want to prevent this shim from being removed / inlined.

crates/assert-instr-macro/src/lib.rs
```diff
@@ -133,6 +134,8 @@ pub fn assert_instr(
         fn #assert_name() {
             #to_test
 
+            unsafe { asm!("" : : "r"(#shim_name as usize) : "memory" : "volatile") };
```
gnzlbg: Before we never called the shim for anything, we passed a pointer to it to `stdsimd_test::assert`, but AFAICT that function never called that pointer either (maybe that's what `backtrace` was doing?). 

Either way, we leak the address of this pointer to "all memory" here, and perform arbitrary side-effects, which could include calling this function. So with this LLVM should never be able to remove the function.

crates/stdsimd-test/src/lib.rs
```diff
@@ -46,38 +45,56 @@ lazy_static! {
...
     instrs: Vec<Instruction>,
 }
 
+#[derive(Debug)]
 struct Instruction {
     parts: Vec<String>,
 }
 
-fn normalize(symbol: &str) -> String {
+fn normalize(mut symbol: &str) -> String {
```
gnzlbg: TODO: document what this does - in a nutshell, it brings the shim function names from potentially mangled names with paths, etc. down to the `stdsimd_test_shim_...` form, so that we can query these easily by doing HashSet lookup.

crates/stdsimd-test/src/disassembly.rs
```diff
@@ -108,14 +116,18 @@ fn parse_objdump(output: &str) -> HashMap<String, Vec<Function>> {
...
             .or_insert_with(Vec::new)
             .push(Function {
                 addr: None,
                 instrs: instructions,
             });
     }
 
+    for (_k, v) in &mut ret {
+        assert_eq!(v.len(), 1);
+    }
```
gnzlbg: So right now these are still `HashMap<String, Vec<Function>>`, but this should become `HashSet<String, Function>` in the future. 

If we are finding multiple duplicated symbols of the same shim, things already went wrong somewhere.

PR #778, opened 2019-07-08
Setup Azure Pipelines

gnzlbg:


gnzlbg:
@alexcrichton this sets up azure pipelines - I didn't manage to get the window jobs to pass (wrong rust toolchain being picked), but we can enable them later.

gnzlbg:
@alexcrichton this is blocked on setting up the access tokens to publish to github pages.

alexcrichton:
I think the tokens are only necessary for publishing documentation, which isn't necesarily critical nowadays, so if you want to debug the other various issues I can cover that later once this lands.

gnzlbg:
@alexcrichton @pietroalbini already set up deploying to gh-pages

PR #786, opened 2019-07-30
Make the _mm_movemask_epi8 test a non palindrome

### LINKED ISSUE #785: _mm_movemask_epi8 test result is a palindrome
https://github.com/rust-lang/stdarch/blob/65c9fa511acb1841bac37f92ad9b586ca45f1df7/crates/core_arch/src/x86/sse2.rs#L4162

This means that reversing result order wont cause the test to fail. I noticed this, because the test succeeded despite the order being wrong when emulating it in https://github.com/bjorn3/rustc_codegen_cranelift.

bjorn3:
Fixes #785


gnzlbg:
@pietroalbini it seems that azure pipelines is not running anymore for this repo?

PR #787, opened 2019-07-30
Remove simd_rem intrinsic definition

bjorn3:
It is unused


gnzlbg:
LGTM. Its use has been moved to the `packed_simd` crate / std::simd.

bjorn3:
There doesn't appear to be a `std::simd` module on stable and nightly.

gnzlbg:
It was temporarily moved into a separate crate because it was increasing the binary size of libcore by quite a bit.

bjorn3:
The [Azure (color)](https://en.wikipedia.org/wiki/Azure_(color)) turned green :)

gnzlbg:
Land all the things!

Issue #788, opened 2019-07-30
Use simd_* intrinsics instead of llvm intrinsics where possible


bjorn3:
* [ ] `_mm_sqrt_ps`: `llvm.x86.sse.sqrt.ps` -> `simd_fsqrt`
* [ ] `_mm_min_ps`: `llvm.x86.sse.min.ps` -> `simd_fmin`
* [ ] ...

This makes it easier for other codegen backends to support stuff, as all variants of the same operation are nicely bundled together. (eg single and double precision sqrt)

gnzlbg:
We used to do that, but only for arm and x86 we needed to add ~10.000 intrinsics. This meant modifying rustc every time a new intrinsic should be added, and doing CI of the intrinsics in rustc, which would add a dozen build jobs to run the tests on less common targets, and increase CI time of rustc build jobs by quite a while (I think the largest tests here take ~20 min.

It would be simpler if other backends could just either hijack the llvm intrinsics, or add their own `#[link_foo_intrinsic]` attributes that can be used here to link against other intrinsics. Particularly for the more complex algorithms exposed that require calling different intrinsics.

bjorn3:
> Particularly for the more complex algorithms exposed that require calling different intrinsics.

I meant the simple ones which already have a `simd_*` `platform-intrinsic`.

gnzlbg:
This PR was the last one removing support for that: https://github.com/rust-lang/rust/pull/57416 in case you are considering adding it again. If that happens, we could incrementally remove the `#[link_llvm_intrinsics]` from this crate, and start calling those.

gnzlbg:
> I meant the simple ones which already have a simd_* platform-intrinsic.

Your list appeared unbounded, it would then maybe be helpful if you could write a complete list of operations that you want to propose. The `packed_simd` crate already implements all the basic stuff portably, and for that the currently available intrinsics mostly suffice.

gnzlbg:
>  The packed_simd crate already implements all the basic stuff portably, and for that the currently available intrinsics mostly suffice

Expanding on that, it implements the two operations you mention, and they currently produce optimal code generation with LLVM. For doing a `min` we just use `<` and `simd_select`. For doing a `sqrt` we just use the vector `llvm.fsqrt` intrinsics. We probably need to add an LLVM pass to lower those to libmvec calls at some point, packed_simd just optionally links a vector math library, and lowers those to FFI calls.

bjorn3:
[...]

PR #790, opened 2019-07-31
Use more simd_* intrinsics

bjorn3:
I currently only did this for x86. Also I skipped `_mm_sqrt_ps` and some more, as llvm emitted `rsqrtps` combined with a lot of extra instructions instead of `sqrtps`, causing slight rounding errors and non optimal codegen.

cc #788

### DIFF
--- a/crates/core_arch/src/x86/avx.rs
+++ b/crates/core_arch/src/x86/avx.rs
@@ -513,7 +513,7 @@ pub unsafe fn _mm256_sqrt_ps(a: __m256) -> __m256 {
 #[cfg_attr(test, assert_instr(vsqrtpd))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn _mm256_sqrt_pd(a: __m256d) -> __m256d {
-    sqrtpd256(a)
+    simd_fsqrt(a)
 }
 
 /// Blends packed double-precision (64-bit) floating-point elements from
--- a/crates/core_arch/src/x86/avx.rs
+++ b/crates/core_arch/src/x86/avx.rs
@@ -426,7 +426,7 @@ pub unsafe fn _mm256_round_pd(a: __m256d, b: i32) -> __m256d {
 #[cfg_attr(test, assert_instr(vroundpd))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn _mm256_ceil_pd(a: __m256d) -> __m256d {
-    roundpd256(a, 0x02)
+    simd_ceil(a)
 }
 
 /// Rounds packed double-precision (64-bit) floating point elements in `a`
@@ -438,7 +438,7 @@ pub unsafe fn _mm256_ceil_pd(a: __m256d) -> __m256d {
 #[cfg_attr(test, assert_instr(vroundpd))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn _mm256_floor_pd(a: __m256d) -> __m256d {
-    roundpd256(a, 0x01)
+    simd_floor(a)
 }
 
 /// Rounds packed single-precision (32-bit) floating point elements in `a`
@@ -477,7 +477,7 @@ pub unsafe fn _mm256_round_ps(a: __m256, b: i32) -> __m256 {
 #[cfg_attr(test, assert_instr(vroundps))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn _mm256_ceil_ps(a: __m256) -> __m256 {
-    roundps256(a, 0x02)
+    simd_ceil(a)
 }
 
 /// Rounds packed single-precision (32-bit) floating point elements in `a`
@@ -489,7 +489,7 @@ pub unsafe fn _mm256_ceil_ps(a: __m256) -> __m256 {
 #[cfg_attr(test, assert_instr(vroundps))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn _mm256_floor_ps(a: __m256) -> __m256 {
-    roundps256(a, 0x01)
+    simd_floor(a)
 }
 
 /// Returns the square root of packed single-precision (32-bit) floating point
--- a/crates/core_arch/src/x86/avx.rs
+++ b/crates/core_arch/src/x86/avx.rs
@@ -255,7 +255,7 @@ pub unsafe fn _mm256_andnot_ps(a: __m256, b: __m256) -> __m256 {
 #[cfg_attr(test, assert_instr(vmaxpd))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn _mm256_max_pd(a: __m256d, b: __m256d) -> __m256d {
-    maxpd256(a, b)
+    simd_fmax(a, b)
 }
 
 /// Compares packed single-precision (32-bit) floating-point elements in `a`
@@ -267,7 +267,7 @@ pub unsafe fn _mm256_max_pd(a: __m256d, b: __m256d) -> __m256d {
 #[cfg_attr(test, assert_instr(vmaxps))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn _mm256_max_ps(a: __m256, b: __m256) -> __m256 {
-    maxps256(a, b)
+    simd_fmax(a, b)
 }
 
 /// Compares packed double-precision (64-bit) floating-point elements
@@ -279,7 +279,7 @@ pub unsafe fn _mm256_max_ps(a: __m256, b: __m256) -> __m256 {
 #[cfg_attr(test, assert_instr(vminpd))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn _mm256_min_pd(a: __m256d, b: __m256d) -> __m256d {
-    minpd256(a, b)
+    simd_fmin(a, b)
 }
 
 /// Compares packed single-precision (32-bit) floating-point elements in `a`
@@ -291,7 +291,7 @@ pub unsafe fn _mm256_min_pd(a: __m256d, b: __m256d) -> __m256d {
 #[cfg_attr(test, assert_instr(vminps))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn _mm256_min_ps(a: __m256, b: __m256) -> __m256 {
-    minps256(a, b)
+    simd_fmin(a, b)
 }
 
 /// Multiplies packed double-precision (64-bit) floating-point elements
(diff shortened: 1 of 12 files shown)


gnzlbg:
> Also I skipped _mm_sqrt_ps and some more, as llvm emitted rsqrtps combined with a lot of extra instructions instead of sqrtps, causing slight rounding errors and non optimal codegen.

I'll look into that.

bjorn3:
Got the same for `_mm256_sqrt_ps`.

bjorn3:
```rust
#![feature(platform_intrinsics)]

extern crate core;

use core::arch::x86_64::__m128;

extern "platform-intrinsic" {
    fn simd_fsqrt<T>(a: T) -> T;
}

pub unsafe fn sqrt(a: __m128) -> __m128 {
    simd_fsqrt(a)
}
```

Optimized LLVM:

```llvm
; playground::sqrt
; Function Attrs: nofree nounwind nonlazybind uwtable
define void @_ZN10playground4sqrt17h5d635885a5180697E(<4 x float>* noalias nocapture sret dereferenceable(16), <4 x float>* noalias nocapture readonly dereferenceable(16) %a) unnamed_addr #0 {
start:
  %1 = load <4 x float>, <4 x float>* %a, align 16
  %2 = tail call fast <4 x float> @llvm.sqrt.v4f32(<4 x float> %1)
  store <4 x float> %2, <4 x float>* %0, align 16
  ret void
}
```

Optimized asm:

```asm
.LCPI0_0:
	.long	3204448256              # float -0.5
	.long	3204448256              # float -0.5
	.long	3204448256              # float -0.5
	.long	3204448256              # float -0.5

.LCPI0_1:
	.long	3225419776              # float -3
	.long	3225419776              # float -3
	.long	3225419776              # float -3
	.long	3225419776              # float -3

playground::sqrt: # @playground::sqrt
# %bb.0:
	movq	%rdi, %rax
	movaps	(%rsi), %xmm0
	rsqrtps	%xmm0, %xmm1
	movaps	%xmm0, %xmm2
	mulps	%xmm1, %xmm2
	movaps	.LCPI0_0(%rip), %xmm3   # xmm3 = [-5.0E-1,-5.0E-1,-5.0E-1,-5.0E-1]
	mulps	%xmm2, %xmm3
	mulps	%xmm1, %xmm2
	addps	.LCPI0_1(%rip), %xmm2
	xorps	%xmm1, %xmm1
	cmpneqps	%xmm0, %xmm1
	mulps	%xmm3, %xmm2
	andps	%xmm2, %xmm1
	movaps	%xmm1, (%rdi)
	retq
```

bjorn3:
I have gone through every llvm intrinsic for x86 and x86_64 to see if there is a `simd_*` replacement.

crates/core_arch/src/x86/avx.rs
```diff
@@ -426,7 +426,7 @@ pub unsafe fn _mm256_round_pd(a: __m256d, b: i32) -> __m256d {
 #[cfg_attr(test, assert_instr(vroundpd))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn _mm256_ceil_pd(a: __m256d) -> __m256d {
-    roundpd256(a, 0x02)
+    simd_ceil(a)
```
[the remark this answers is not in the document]
bjorn3: No, `_mm256_round_ps` and `_mm256_round_pd` still use them.

https://github.com/bjorn3/stdarch/blob/cb05d6fba4b50a04f433f808a4eac36b29ea761b/crates/core_arch/src/x86/avx.rs#L465

crates/core_arch/src/x86/fma.rs
```diff
@@ -32,7 +33,7 @@ use stdarch_test::assert_instr;
 #[cfg_attr(test, assert_instr(vfmadd))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn _mm_fmadd_pd(a: __m128d, b: __m128d, c: __m128d) -> __m128d {
-    vfmaddpd(a, b, c)
+    simd_fma(a, b, c)
```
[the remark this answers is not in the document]
bjorn3: It just lowers to `llvm.fma.v*` with the llvm codegen backend. According to https://llvm.org/docs/LangRef.html#llvm-fma-intrinsic, it returns the same result as the corresponding libm `fma` function, which I think means that for vectors it will return the same result as applying `fma` to every lane.

crates/core_arch/src/x86/sse.rs
```diff
@@ -194,7 +194,7 @@ pub unsafe fn _mm_min_ss(a: __m128, b: __m128) -> __m128 {
 #[cfg_attr(test, assert_instr(minps))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn _mm_min_ps(a: __m128, b: __m128) -> __m128 {
-    minps(a, b)
+    simd_fmin(a, b)
```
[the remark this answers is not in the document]
bjorn3: `simd_fmin` get lowered to [`minnum`](https://llvm.org/docs/LangRef.html#llvm-minnum-intrinsic).The only possible difference with `minps` it the behaviour with `-0.0` and `0.0`. LLVM doesnt guarantee that one is smaller than the other, while `minps` always picks the second operand it seems. The intrinsic docs at intel unhelpfully list the result as `MIN(a, b)`

crates/core_arch/src/x86/fma.rs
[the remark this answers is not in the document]
gnzlbg: Thanks, that sounds correct.

crates/core_arch/src/x86/sse.rs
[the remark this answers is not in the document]
gnzlbg: Maybe we can add a test for this and verify that it passes before this change and after it ?

cc @rkruppe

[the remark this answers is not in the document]
bjorn3: I reverted the change, as `simd_fmin` does behave differently from `minps` for `-0.0`.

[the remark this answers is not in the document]
gnzlbg: I'm resolving this.

 If you want to improve this, you might need to add a different `simd_` intrinsic to rust, that instead of generating a `llvm.minnum`, it instead generates an `fcmp ult` followed by a `select`. I recommend checking out first whether it is at all possible to match `minps` semantics with that, e.g., depending on how these handle denormals, and the case where one of the arguments is a nan.

crates/core_arch/src/x86/avx.rs
```diff
@@ -255,7 +255,7 @@ pub unsafe fn _mm256_andnot_ps(a: __m256, b: __m256) -> __m256 {
 #[cfg_attr(test, assert_instr(vmaxpd))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn _mm256_max_pd(a: __m256d, b: __m256d) -> __m256d {
-    maxpd256(a, b)
+    simd_fmax(a, b)
```
gnzlbg: Is the behavior of these the same, e.g., for subnormals, when one argument contain NaNs, etc. ?

crates/core_arch/src/x86/sse.rs
```diff
@@ -219,6 +220,7 @@ pub unsafe fn _mm_max_ss(a: __m128, b: __m128) -> __m128 {
 #[cfg_attr(test, assert_instr(maxps))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn _mm_max_ps(a: __m128, b: __m128) -> __m128 {
+    // See the `test_mm_min_ps` test why this can't be implemented using `simd_fmax`.
```
gnzlbg: I think it would be better to add similar tests to the other intrinsics using `simd_fmax` and `simd_fmin`, that check subnormals, and also that check the behavior when the first argument is nan, and the second non-nan, and viceversa.

crates/core_arch/src/x86/sse.rs
```diff
@@ -2638,6 +2640,21 @@ mod tests {
...
+        // `r1` to `a` and `r2` to `b`.
+        let a = _mm_setr_ps(-0.0, 0.0, 0.0, 0.0);
+        let b = _mm_setr_ps(0.0, 0.0, 0.0, 0.0);
+        let r1: [u8; 16] = transmute(_mm_min_ps(a, b));
+        let r2: [u8; 16] = transmute(_mm_min_ps(b, a));
+        let a: [u8; 16] = transmute(a);
+        let b: [u8; 16] = transmute(b);
+        assert_eq!(r1, b);
+        assert_eq!(r2, a);
+        assert_ne!(a, b); // sanity check that -0.0 is actually present
```
gnzlbg: I think we need to also test here the behavior when the first argument is nan and the second is not, and vice versa (e.g. if the result the Nan? the second argument ? always the non-nan ? etc.).

crates/stdarch-test/src/lib.rs
```diff
@@ -91,7 +91,7 @@ pub fn assert(_fnptr: usize, fnname: &str, expected: &str) {
 
     // Look for `expected` as the first part of any instruction in this
     // function, e.g., tzcntl in tzcntl %rax,%rax.
-    let found = instrs.iter().any(|s| s.contains(expected));
+    let found = instrs.iter().any(|s| s.starts_with(expected));
```
gnzlbg: No instruction in any `assert_instr` was changed, so I'm not sure why this is needed ? Is there an instruction that was matching exactly before but isn't matching exactly anymore ?

crates/core_arch/src/x86/sse.rs
bjorn3: How do I create a subnormal? As far as I understand they are close to zero, but I don't know how close.

crates/stdarch-test/src/lib.rs
bjorn3: Cause otherwise `rsqrt` matches when `sqrt` is expected.

gnzlbg: Makes sense, thanks for catching this!

crates/core_arch/src/x86/sse.rs
gnzlbg: > How do I create a subnormal? 

Check out the docs for `f{32,64}::is_normal()`. Each floating-point type has a `MIN_POSITIVE` number, and all numbers between that one and zero (I think in range: `(-MIN_POSITIVE, MIN_POSITIVE)`) are subnormal. I don't know if creating them from a literal returns 0.0 or not. But if they do, then checking permutations of -0.0, 0.0, and NaN should be enough, e.g., (-0.0, 0.0), (0.0, -0.0), (1.0, NaN), (NaN, 1.0).

bjorn3: `0.000000000000000000000000000000000000000000001f32.is_normal()` returns false and transmuting it to `[u8; 4]` gives `[1, 0, 0, 0]`. Do you want to check permutations with that number too? Or should I just use `0.0`?

gnzlbg: > Do you want to check permutations with that number too? 

Yes, we should check that too :)

bjorn3:
LLVM somehow doesn't use the simd instructions for certain intrinsics on i586.

gnzlbg:
@bjorn3 maybe we could use the generic intrinsics in some cases (e.g. `#[cfg(target_feature = "sse2")]` ?), and the specific ones in others ?

bjorn3:
Rebased to trigger CI, as the old logs are no longer available.

bjorn3:
Windows build failed while installing rust:

```
Run rustup update nightly --no-self-update && rustup default nightly
At D:\a\_temp\0855049a-8a0a-4cb8-bf51-de53a4f07b31.ps1:2 char:40
+ rustup update nightly --no-self-update && rustup default nightly
+                                        ~~
The token '&&' is not a valid statement separator in this version.
+ CategoryInfo          : ParserError: (:) [], ParseException
+ FullyQualifiedErrorId : InvalidEndOfLine
```

gnzlbg:
> rustup update nightly --no-self-update && rustup default nightly

Can you split this statement into two different lines and try again? 

```
rustup update nightly --no-self-update
rustup default nightly
```

makotokato:
> Windows build failed while installing rust:

This is fixed by https://github.com/rust-lang/stdarch/commit/ac59837bc11fdcc9557f739b901505b7d802686c

gnzlbg:
Closing / reopening to re-trigger CI.

bjorn3:
Reverted the mmx changes, as those are the ones not compiling to the required instruction.

bjorn3:
CI is finally happy!

gnzlbg:
> Reverted the mmx changes, as those are the ones not compiling to the required instruction.

Uh, sorry, my fault, I should have caught this. Yes, `mmx` intrinsics (or those using the `_m64` type in general) won't work with the generic `simd_` intrinsics. I wouldn't worry about that, `_m64` creates so many headaches that few people are using it, and also, chances are we will never stabilize it.

gnzlbg:
Thank you @bjorn3 for working on this!

PR #796, opened 2019-08-12
Apply snap of bootstrap compiler in rust-lang/rust

Mark-Simulacrum:
FWIW, in the future we should try to avoid this as I'm assuming we'll break CI trying to land this here (and we can't swap bootstrap compilers in Rust until this lands).

### DIFF
--- a/crates/core_arch/src/mod.rs
+++ b/crates/core_arch/src/mod.rs
@@ -8,11 +8,7 @@ mod acle;
 
 mod simd;
 
-#[cfg_attr(
-    bootstrap,
-    doc(include = "../stdarch/crates/core_arch/src/core_arch_docs.md")
-)]
-#[cfg_attr(not(bootstrap), doc(include = "core_arch_docs.md"))]
+#[doc(include = "core_arch_docs.md")]
 #[stable(feature = "simd_arch", since = "1.27.0")]
 pub mod arch {
     /// Platform-specific intrinsics for the `x86` platform.


Mark-Simulacrum:
Ah just as a note here this isn't actually needed quite yet (probably Wednesday, Tuesday) but I presume it's good to get the ball rolling nevertheless.

Mark-Simulacrum:
Ah, I've come up with a work around by disabling `--cfg bootstrap` entirely as an option to libcore for now so this isn't actually all that urgent, though I'd love to be able to avoid such hacks in the future.

gnzlbg:
@Mark-Simulacrum it appears that azure doesn't run on PRs, and bors doesn't work with Azure - I don't have a timeline for when this will be fixed

Mark-Simulacrum:
We'll want to land this (perhaps by just directly merging) within the next few days for sure, since the hack I invented to work around it in rust-lang/rust is not long-term tenable.

gnzlbg:
I've raised the issue with infra. The libc repo is also blocked due to Azure Pipelines failing.

pietroalbini:
Well, this is a legit timeout.

RalfJung:
50min later, 4 out of 18 jobs are done, and right now none of the remaining jobs is actually running... "Waiting for an available agent".

gnzlbg:
This is a known problem. For a smooth libc workflow we would need 20
concurrent jobs. This was eating the 70 job queue of rust-lang/rust. Now we
have a 10 concurrent job queue for libc and other similar projects. That
solved the rust-Lang/rust problem, but made everything else worse.

Worst part, pretty much constantly we have ~2 Rust-Lang/rust PRs blocked on
libc or stdarch fixes that take days to merge because we have too few jobs.

PR #799, opened 2019-08-17
Verify that all intrinsics have a run-time test

gnzlbg:
Add a check to stdsimd-verify to check that all intrinsics have a run-time test. This is not the case right now, but we should at least not add intrinsics without tests.


crates/stdarch-verify/src/lib.rs
```diff
@@ -37,14 +37,33 @@ fn functions(input: TokenStream, dirs: &[&str]) -> TokenStream {
 
     let mut functions = Vec::new();
     for &mut (ref mut file, ref path) in &mut files {
-        for item in file.items.drain(..) {
-            if let syn::Item::Fn(f) = item {
-                functions.push((f, path))
+        for mut item in file.items.drain(..) {
+            match item {
+                syn::Item::Fn(f) => functions.push((f, path)),
+                syn::Item::Mod(ref mut m) => {
```
gnzlbg: stdsimd-verify did not scan for function items within inline modules (`mod { ... }`) - since tests are inside a `mod tests { ... }` this is required to add them.

crates/core_arch/src/x86/sse.rs
```diff
@@ -3437,7 +3437,7 @@ mod tests {
     }
 
     #[simd_test(enable = "sse")]
-    pub unsafe fn test_mm_cvtsi32_ss() {
+    unsafe fn test_mm_cvtsi32_ss() {
```
gnzlbg: These tests were `pub` items for some reason and the check errored that they also needed tests..

crates/stdarch-verify/tests/x86-intel.rs
```diff
@@ -140,6 +141,156 @@ fn verify_all_signatures() {
 
     let mut all_valid = true;
     'outer: for rust in FUNCTIONS {
+        if !rust.has_test {
+            // FIXME: this list should be almost empty
+            let skip = [
+                "__readeflags",
```
gnzlbg: We should try to make this list smaller over time. It might well be that some of these are already tested in other tests, or that the test names have typos or similar.

PR #792, opened 2019-08-01
Add more ARM SIMD intrinsics

Licenser:
I'm trying to add some more SIMD intrinsics for arm. It's still very much WIP - I'm also not sure how to test them locally.

PR #802, opened 2019-08-20
Add stdarch-gen

Licenser:
Add a generator for neon-simd instructions. See #792 for reference.


crates/stdarch-gen/Makefile
```diff
@@ -0,0 +1,5 @@
+all:
+	cargo run
+	rustfmt arm.rs aarch64.rs
```
bjorn3: Why not run rustfmt directly from `main.rs`?

Licenser: `main.rs` isn't generated `arm.rs` and `aarch64.rs` both are so they're not human edited.

bjorn3: I mean that `main.rs` will run rustfmt on the files it generated itself.

Licenser: oh I dind't think of that!

crates/stdarch-gen/src/main.rs
```diff
@@ -0,0 +1,502 @@
...
+    let mut file_arm = File::create(ARM_OUT)?;
+    file_arm.write_all(out_arm.as_bytes())?;
+    file_arm.write_all(tests_arm.as_bytes())?;
+
+    let mut file_aarch = File::create(AARCH64_OUT)?;
+    file_aarch.write_all(out_aarch64.as_bytes())?;
+    file_aarch.write_all(tests_aarch64.as_bytes())?;
+    Command::new("sh")
+            .arg("-c")
+            .arg(format!("rustfmt {} {}", ARM_OUT, AARCH64_OUT))
```
bjorn3: Please don't use `sh` as command to execute. Windows doesnt support it for example.

crates/stdarch-gen/src/main.rs
```diff
@@ -0,0 +1,502 @@
...
+            .output()
```
bjorn3: Please use `.status()` instead.

crates/stdarch-gen/src/main.rs
```diff
@@ -0,0 +1,502 @@
...
+        "TRUE" => match &t[..3] {
+            "u8x" => "0xFF",
+            "u16" => "0xFF_FF",
+            "u32" => "0xFF_FF_FF_FF",
+            "u64" => "0xFF_FF_FF_FF_FF_FF_FF_FF",
+            _ => panic!("No TRUE for type {}", t),
+        },
+        o => o,
+    }
+}
```
bjorn3: Missing trailing newline.

crates/stdarch-gen/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "stdarch-gen"
+version = "0.1.0"
+authors = ["Heinz N. Gies <heinz@licenser.net>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
gnzlbg: You can remove this line.

crates/stdarch-gen/Makefile
```diff
@@ -0,0 +1,4 @@
+all:
```
gnzlbg: I'd rather not use makefiles. `cargo run` should generate the files, and `cargo test` should verify that the files in the other repos are uptodate.

crates/stdarch-gen/src/main.rs
```diff
@@ -0,0 +1,502 @@
...
+    let mut tests_aarch64 = String::from(
+        r#"
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::core_arch::simd::*;
+    use std::mem::transmute;
+    use stdarch_test::simd_test;
+"#,
+    );
```
gnzlbg: The aarch64 and arm versions are identical. Can we simplify this ?

crates/stdarch-gen/Makefile
Licenser: That's a really nice idea! I didn't think of that!

crates/stdarch-gen/src/main.rs
Licenser: good call!

Licenser:
Note: The tests here fail because the generated functions are not included to don't create a conflict with #792

gnzlbg:
While this is something we could do, it's unclear to me that this is the right way to do it. A couple of global notes:

* Why invent your own format? The solution we used to have used json for the specification format, and that removes the need to do any parsing since you can just use `serde_json`. As I mentioned before, the benefits of a custom format are unclear to me. They add a lot of overhead right now, and make future extensions, like at least partially generating the specification files from the vendor documents (e.g. like stdarch-verify does) harder.

* Why invent your own DSL? AFAICT we could just use regex to match types. That would remove a big chunk of the DSL. It's unclear to me how we can adapt this DSL to work for AVX-512. The DSL we used to have to auto-generate the intrinsics worked for both x86 and neon intrinsics.

Licenser:
> While this is something we could do, it's unclear to me that this is the right way to do it. A couple of thoughts:
> 
> * Why invent your own format? The solution we used to have used json for the specification format, and that removes the need to do any parsing since you can just use `serde_json`.

In the entire history of mankind JSON has never improved anything.

> As I mentioned before, the benefits of a custom format are unclear to me. They add a lot of overhead right now, and make future extensions, like at least partially generating the specification files from the vendor documents (e.g. like stdarch-verify does) harder.

Let me clear that up then.

First of, they factually do not, add a lot of overhead - about 30 odd lines, call it 50 and it makes 10% of the entirety of `main.rs`. That would not go to 0 with any other representation either.

The majority of the code is about generating the intrinsics and that's pretty much independent of how the specification is written - that's on purpose.

That said. I'm not married to this particular format. It exists in it's current form because it's easy to read, easy to write and most of all easy to change and evolve - those things I do care about!

To put it in perspective. The 400 and odd lines of spec generate over or so intrinsics, more then doubling the current coverage in `std::arch::{arm, aarch64}` even counting the 500 odd lines for the `main.rs` (that will amortise as the spec grows) that's less then 5 lins per intrinsic + test.

There is one big reason so, it allows fitting the format to the needs, not the other way around. Picking any pre-existing format means the final solution is, to a large degree, pre-determined and that limits what it can be, the format will change as it covers more intrinsics and types.

> * Why invent your own DSL? AFAICT we could just use regex to match types. That would remove a big chunk of the DSL. It's unclear to me how we can adapt this DSL to work for AVX-512. The DSL we used to have to auto-generate the intrinsics worked for both x86 and neon intrinsics.

That's a good point a regexp would do a better job here!

gnzlbg:
I started playing with this, and using `serde_json`, the whole parser becomes:

```rust
#[derive(Serialize, Deserialize)]
struct Intrinsic {
    name: String,
    doc: String,
    fn: String,
    arm: Option<String>,
    aarch64: Option<String>,
    a: Option<Vec<Value>>,
    b: Option<Vec<Value>>,
    e:  Option<Vec<Value>>,
    generate: Vec<String>,
}

#[derive(Serialize,Deserialize)]
#[serde(untagged)]
enum Value { Int(i128), Float(f64), Bool(bool) }

fn parse(file: &str) -> Vec<Intrinsic> {
    serde_json::parse(file).unwrap()
} 
```

by trading of the spec format to something that already has a valid parser, like JSON, but as I mentioned, there are many formats that are already supported (JSON is just what we used to use in the past to solve this exact same problem):

```json
[ 
  {
  "doc": "Vector bitwise and",
  "name": "vand",
  "fn": "simd_and",
  "arm": "vand",
  "aarch64": "and",
  "a": [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15 ],
  "b": [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15 ],
  "e": [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15 ],
  "generate": ["int*_t", "uint*_t", "int64x*_t", "uint64x*_t"]
  }
]
```

I don't really care about which format is used, but I think there is value in using an existing format with a working serializer and deserializer, not only because it removes all parsing code here, but also because it would make it trivial to automatically generate at least part of the `spec` files from the vendor intrinsic specifications that we already parse in the `stdarch-verify` crate (we just need to build a `Vec<Intrinsic>` from them, and write that down to a file, that this crate can consume). 

Sure, writing a custom parser here for deserialization, and writing a custom serializer somewhere else for this custom format isn't hard, but I don't really see much value in doing that when json would just work.

Licenser:
JSON would not just work, if you look at it will just get more verbose. A human should never have to write JSON it's not a human format.

Please take a look at some of the more complex entries and not just the most trivial one.

- https://github.com/Licenser/neongen/blob/master/neon.spec#L104 re-uses the whole header to generate different intrinsics for arm v7 and aarch64
- https://github.com/Licenser/neongen/blob/master/neon.spec#L68 uses a partial replacement for the tests to test or for both `or` with ones and zeros.

The advantage is not how it's loader, I honestly don't care much about that part as its automated one way or another. What i care about is the user, the human having to write those, so the advantage comes form making it easier and less error prone to write it. To co-locate tests and code and to reuse as much as possible to prevent copy&paste (errors).

It is more then just a purely mechanical specification, it does include code in the sense that test are specified - JSON was not meant to create code (hello react...)

If you are not convinced this is easier to write then a JSON give it a try, write a JSON specification that covers the same intrinsics and tests then compare it to the .spec file.

sunnygleason:
@Licenser @gnzlbg I think you're both onto something. JSON is a crazy amount of data
that would have to be authored somewhere else anyway. A custom format is potentially
exotic and esoteric, and isn't the actual definitive source either.

Would you be open to an approach that uses the arm_neon C file, possibly in coordination
with a flat file/CSV?

I think @Licenser 's spec format is nice, but the test data itself is actually very repetitive
and in fact dependent on the operation category and input types (so specifying it repeatedly
is redundant).

If you look at the tests I wrote from https://github.com/rust-lang/stdarch/pull/792, you'll
see that the test data is a function of "operation type", "input type", and an "element
reference closure" which compresses a lot of the text. Not sure if that's good or not though! 😄 

In any event, very interested in your thoughts!

gnzlbg:
As i've repeated many times, I don't really care about the format as long as it is a standard one. My first comment suggested TOML, which is intended to be human readable and editable.

> Please take a look at some of the more complex entries and not just the most trivial one.

I see nothing there that isn't **trivially** implemented in JSON. 

> What i care about is the user, the human having to write those, so the advantage comes form making it easier and less error prone to write it. 

I care about this too, and at 4000 intrinsics, with dozens of relevant tests per intrinsic, my main priority would be for humans not to have to write these at all.

Licenser:
I think we have fundamentally different goals here. I would like a format that’s easy to edit and satisfy the growing requirements for tests - that inherently makes it impossible to automatically generate the entire content since tests require human interaction. You are looking for extracting the intrinsics from a doc as far as I can see it which i don’t find very important, 4000 intrinsic sounds much but the spec file already implements ~250 plus sims tests.

You think a standard format is a must, I’m think it will get in the way since what it’s about is the tests and that’s code not data. In the end it’s your crate so ¯\_(ツ)_/¯ it’s your call I’m just trying ti contribute. That said it’s my time I contribute so I don’t want to waste it on something I believe is a bad choice.

So clueing this as I don’t believe it will be a fruitful discussion.

Licenser:
I think we have fundamentally different goals here. I would like a format that’s easy to edit and satisfy the growing requirements for tests - that inherently makes it impossible to automatically generate the entire content since tests require human interaction. You are looking for extracting the intrinsics from a doc as far as I can see it which i don’t find very important, the spec file i linked implements

Issue #804, opened 2019-09-02
Add an env override to std_detect


lu-zero:
Sometimes it can be useful to disable certain extensions even if the host cpu sports them.

- [x] Refactor the cached check logic in `src/detect/mod.rs`
- [ ] Add a `RUST_STD_DETECT_DISABLE` `env::var` check
  - Takes a space separate list of strings
  - For each of them mapping to an arch Feature, turn it off.
- [ ] Document its existence

@gnzlbg Do you see any problem in this approach?

gnzlbg:
Shall this env variable be read at run-time, or at build time ?

Either way, I think enabling this should be put behind a cargo feature that's disabled by default.

bjorn3:
I think it should be during runtime, so you dont have to recompile the world.

lu-zero:
> Shall this env variable be read at run-time, or at build time ?

run-time

> Either way, I think enabling this should be put behind a cargo feature that's disabled by default.

Not sure how to name the feature. (and how to advertise it)

bjorn3:
> Not sure how to name the feature. (and how to advertise it)

`allow-runtime-detect-override`?

lu-zero:
I'm sold on it, now, which crate should expose it?

lu-zero:
Also: should we make the cpu feature checks macro behave like log?

gnzlbg:
> I'm sold on it, now, which crate should expose it?

Maybe `std_detect` ? 

> Also: should we make the cpu feature checks macro behave like log?

There is an RFC that allows something like that: https://github.com/rust-lang/rfcs/pull/2725 but I'm not sure if that's what you mean.

lu-zero:
It is exactly what I meant.
Also this use-case might be mentioned [here](https://github.com/gnzlbg/rfcs/blob/target_feature_runtime/text/0000-target-feature-runtime.md#the-target_feature_detection_runtime-attribute)

I'll start sending the refactoring patches and the initial impl soon.

lu-zero:
I started with https://github.com/rust-lang/stdarch/pull/806

Where should I put the documentation about the env variable?

Can I use [this](https://crates.io/crates/arg_enum_proc_macro) to implement the mapping `name -> feature` mapping?

PR #808, opened 2019-09-17
Env override support

lu-zero:
See #804 The env var is `RUST_STD_DETECT_UNSTABLE` for now.


crates/std_detect/src/detect/cache.rs
```diff
@@ -63,6 +69,19 @@ impl Initializer {
         let v = self.0;
         self.0 = set_bit(v, bit);
     }
+
+    /// Unsets the `bit` of the cache.
+    #[inline]
+    pub(crate) fn unset(&mut self, bit: u32) {
+        // FIXME: this way of making sure that the cache is large enough is
+        // brittle.
+        debug_assert!(
```
gnzlbg: I think this will detect errors here in CI, which is where we need to catch these issues and where we try to test all features, so I think this is fine. 

Did you have in mind any other way to improve this ?

crates/std_detect/src/detect/mod.rs
```diff
@@ -97,10 +97,27 @@ cfg_if! {
...
+                let _ = Feature::from_str(v).map(|v| value.unset(v as u32));
+            }
+            value
+        } else {
+            value
+        }
+    } else {
+        value
+    }
+}
```
gnzlbg: I think it would be better to do it right after initializing the cache. Ideally you'll put the logic in a separate function that's `#[inline(never)]` to keep things tight.

crates/std_detect/src/detect/cache.rs
lu-zero: Not right now, I copied the FIXME from the other functions, we could test this in the macro even if I'm afraid wouldn't be that fun to implement it.

crates/std_detect/src/detect/mod.rs
lu-zero: I'm not sure where you'd like to put it, but sure I can spin the code out of the branch.

gnzlbg: I think it should happen inside this branch: https://github.com/rust-lang/stdarch/blob/master/crates/std_detect/src/detect/cache.rs#L161

crates/std_detect/src/detect/cache.rs
gnzlbg: I'd just remove the FIXMEs and just replace this with a comment saying what this does (catch if we have outgrown the cache during testing or so).

crates/std_detect/src/lib.rs
```diff
@@ -24,12 +24,23 @@
 extern crate cfg_if;
 
 cfg_if! {
-    if #[cfg(feature = "std_detect_file_io")] {
+    if #[cfg(all(feature = "std_detect_file_io", feature = "std_detect_env_override"))] {
```
gnzlbg: You can just use `any` here.

crates/std_detect/src/detect/cache.rs
lu-zero: Added a commit about it

crates/std_detect/src/lib.rs
lu-zero: Sure

crates/std_detect/src/detect/cache.rs
```diff
@@ -152,13 +187,17 @@ impl Cache {
...
+/// variable `RUST_STD_DETECT_UNSTABLE` and uses its its content to disable
+/// Features that would had been otherwise detected.
 #[inline]
 pub(crate) fn test<F>(bit: u32, f: F) -> bool
 where
     F: FnOnce() -> Initializer,
 {
     if CACHE.is_uninitialized() {
-        CACHE.initialize(f());
+        CACHE.initialize(env_override(f()));
```
gnzlbg: Could you move this line to a different `#[inline(never)]` function that gets called here ?

lu-zero: Sure

crates/std_detect/src/detect/mod.rs
```diff
@@ -135,3 +135,10 @@ pub fn features() -> impl Iterator<Item = (&'static str, bool)> {
     }
     impl_()
 }
+
+#[test]
```
gnzlbg: Please put tests in the tests/ directory -this gets included in libstd as is.

lu-zero: Moved everything now.

gnzlbg:
So this looks good. The only thing missing is a CI buildjob that uses the feature, e.g., on x86_64-unknown-linux-gnu, and then disables e.g. AVX and checks that AVX2 is enabled but AVX isn't. 

The simplest thing I think is what you suggested, copying this buildjob (https://github.com/rust-lang/stdarch/blob/master/ci/azure.yml#L182) and just running the pertinent test.

lu-zero:
I added the buildjob and shuffled around the style fixes.

crates/std_detect/src/detect/mod.rs
```diff
@@ -61,6 +61,13 @@ cfg_if! {
             }
             #[doc(hidden)]
             pub mod __is_feature_detected {}
+
+            impl Feature {
+                #[doc(hidden)]
+                pub fn from_str(_s: &str) -> Result<Feature, ()> { Err<()> }
```
gnzlbg: syntax error

gnzlbg:
I think you merged this and now master is broken :D

gnzlbg:
Don't stress, keep calm. Just create a new branch from master, and submit a PR that fixes it :)

lu-zero:
Ouch, when you ctrl^Rgit p+Enter on the wrong shell...

lu-zero:
Hopefully I fixed the syntax error before the push.

PR #811, opened 2019-09-18
Soundness fix on 32-bit to cache.rs

TethysSvensson:
This PR fixes a soundness issue on 32-bit.

The problem is that on 32-bit relaxed semantics are not good enough as the code is currently written. The problem is that it is possible for a thread to observe the initialized value for CACHE.1 while still observing the uninitialized value for CACHE.0.

The way I fixed it was by putting a bit in both of the `AtomicU32` values that shows whether it has been initialized or not.

While changing the code, I also changed it so only a single atomic load is needed on the fast path on both 32 and 64 bit.

### DIFF
--- a/crates/std_detect/src/detect/cache.rs
+++ b/crates/std_detect/src/detect/cache.rs
@@ -29,9 +29,6 @@ const fn unset_bit(x: u64, bit: u32) -> u64 {
     x & !(1 << bit)
 }
 
-/// Maximum number of features that can be cached.
-const CACHE_CAPACITY: u32 = 63;
-
 /// This type is used to initialize the cache
 #[derive(Copy, Clone)]
 pub(crate) struct Initializer(u64);
@@ -91,29 +88,41 @@ static CACHE: Cache = Cache::uninitialized();
 struct Cache(AtomicU64);
 
 #[cfg(target_pointer_width = "64")]
+const UNINITIALIZED_VALUE: u64 = u64::max_value();
+
+#[cfg(target_pointer_width = "64")]
+const UNINITIALIZED_BIT: u32 = 63;
+
+/// Maximum number of features that can be cached.
+/// We reserve a single bit to check if the cache is initialized
+#[cfg(target_pointer_width = "64")]
+const CACHE_CAPACITY: u32 = 63;
+
+#[cfg(target_pointer_width = "64")]
 #[allow(clippy::use_self)]
 impl Cache {
     /// Creates an uninitialized cache.
     #[allow(clippy::declare_interior_mutable_const)]
     const fn uninitialized() -> Self {
-        Cache(AtomicU64::new(u64::max_value()))
-    }
-    /// Is the cache uninitialized?
-    #[inline]
-    pub(crate) fn is_uninitialized(&self) -> bool {
-        self.0.load(Ordering::Relaxed) == u64::max_value()
+        Cache(AtomicU64::new(UNINITIALIZED_VALUE))
     }
 
     /// Is the `bit` in the cache set?
     #[inline]
-    pub(crate) fn test(&self, bit: u32) -> bool {
-        test_bit(CACHE.0.load(Ordering::Relaxed), bit)
+    fn test(&self, bit: u32) -> bool {
+        let cache_value = CACHE.0.load(Ordering::Relaxed);
+        if !test_bit(cache_value, UNINITIALIZED_BIT) {
+            test_bit(cache_value, bit)
+        } else {
+            initialize(bit)
+        }
     }
 
     /// Initializes the cache.
     #[inline]
-    pub(crate) fn initialize(&self, value: Initializer) {
-        self.0.store(value.0, Ordering::Relaxed);
+    fn initialize(&self, value: Initializer) {
+        let value = unset_bit(value.0, UNINITIALIZED_BIT);
+        self.0.store(value, Ordering::Relaxed);
     }
 }
 
@@ -125,35 +134,52 @@ impl Cache {
 struct Cache(AtomicU32, AtomicU32);
 
 #[cfg(target_pointer_width = "32")]
+const UNINITIALIZED_VALUE: u32 = u32::max_value();
+
+#[cfg(target_pointer_width = "32")]
+const UNINITIALIZED_BIT: u32 = 31;
+
+/// Maximum number of features that can be cached.
+/// We reserve a single bit in each of the two atomic values
+/// to check if the cache is initialized
+#[cfg(target_pointer_width = "32")]
+const CACHE_CAPACITY: u32 = 62;
+
+#[cfg(target_pointer_width = "32")]
 impl Cache {
     /// Creates an uninitialized cache.
     const fn uninitialized() -> Self {
         Cache(
-            AtomicU32::new(u32::max_value()),
-            AtomicU32::new(u32::max_value()),
+            AtomicU32::new(UNINITIALIZED_VALUE),
+            AtomicU32::new(UNINITIALIZED_VALUE),
         )
-    }
-    /// Is the cache uninitialized?
-    #[inline]
-    pub(crate) fn is_uninitialized(&self) -> bool {
-        self.1.load(Ordering::Relaxed) == u32::max_value()
     }
 
     /// Is the `bit` in the cache set?
     #[inline]
-    pub(crate) fn test(&self, bit: u32) -> bool {
-        if bit < 32 {
-            test_bit(CACHE.0.load(Ordering::Relaxed) as u64, bit)
+    fn test(&self, bit: u32) -> bool {
+        if bit < 31 {
+            let cache_value = CACHE.0.load(Ordering::Relaxed);
+            if !test_bit(cache_value, UNINITIALIZED_BIT) {
+                test_bit(cache_value, bit)
+            } else {
+                initializer(bit)
+            }
         } else {
-            test_bit(CACHE.1.load(Ordering::Relaxed) as u64, bit - 32)
+            let cache_value = CACHE.1.load(Ordering::Relaxed);
+            if !test_bit(cache_value, UNINITIALIZED_BIT) {
+                test_bit(cache_value, bit - 31)
+            } else {
+                initializer(bit)
+            }
         }
     }
 
     /// Initializes the cache.
     #[inline]
-    pub(crate) fn initialize(&self, value: Initializer) {
-        let lo: u32 = value.0 as u32;
-        let hi: u32 = (value.0 >> 32) as u32;
+    fn initialize(&self, value: Initializer) {
+        let lo: u32 = unset_bit(value.0, UNINITIALIZED_BIT) as u32;
+        let hi: u32 = unset_bit(value.0 >> 31, UNINITIALIZED_BIT) as u32;
         self.0.store(lo, Ordering::Relaxed);
         self.1.store(hi, Ordering::Relaxed);
     }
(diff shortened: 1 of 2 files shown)


crates/std_detect/src/detect/cache.rs
```diff
@@ -124,61 +133,82 @@ impl Cache {
...
-            AtomicU32::new(u32::max_value()),
-            AtomicU32::new(u32::max_value()),
+            AtomicU32::new(UNINITIALIZED_VALUE),
+            AtomicU32::new(UNINITIALIZED_VALUE),
         )
     }
-    /// Is the cache uninitialized?
-    #[inline]
-    pub(crate) fn is_uninitialized(&self) -> bool {
-        self.1.load(Ordering::Relaxed) == u32::max_value()
```
gnzlbg: So IIUC the problem was here right? This checks whether the second atomic has been initialized, and if that is the case, it assumes both atomics to be initialized. However, the function that initializes the two atomics below performs two relaxes writes, and while this test returns true if the second write has happened, there is no guarantee that the first write has happened, and therefore the UB. Is that correct?

gnzlbg: If so, wouldn't it suffice to have the first write happen before the second write ?

TethysSvensson: Yes. This could also be fixed by using `Acquire` here and `Release` when doing `self.1.store` further down. However doing it this way also allows you to get away with just a single atomic load on the fast path.

crates/std_detect/src/detect/cache.rs
```diff
@@ -124,61 +133,82 @@ impl Cache {
...
     /// Initializes the cache.
     #[inline]
-    pub(crate) fn initialize(&self, value: Initializer) {
-        let lo: u32 = value.0 as u32;
-        let hi: u32 = (value.0 >> 32) as u32;
+    fn initialize(&self, value: Initializer) {
+        let lo: u32 = unset_bit(value.0, UNINITIALIZED_BIT) as u32;
+        let hi: u32 = unset_bit(value.0 >> 31, UNINITIALIZED_BIT) as u32;
         self.0.store(lo, Ordering::Relaxed);
         self.1.store(hi, Ordering::Relaxed);
```
gnzlbg: Would it be possible here to have the `self.0.store` happen before the `self.1.store` instead of doing all other changes ?

crates/std_detect/src/detect/cache.rs
TethysSvensson: If you wish to do it with `Release` semantics, then the fast path would contain a single load with `Release` semantics and a single load with `Relaxed` semantics. This PR only uses a single `Relaxed` load on the fast path.

crates/std_detect/src/detect/cache.rs
TethysSvensson: No, I don' think that would be enough. As far as I can tell, you need to use `Release` and `Acquire` semantics to solve it this way.

crates/std_detect/src/detect/cache.rs
gnzlbg: We currently have this code:

```rust
// thread 1
self.0.store(rel);
self.1.store(rel); 

// thread 2
self.1.load(rel); // assumes self.0.store(rel) has happened
```

Is there a way to have `self.0.store` happen-before `self.1.store` such that a relaxed load in thread 2 of `self.1.load` can only observe the initialized state if both stores have happened ?

TethysSvensson: `self.0.store` is already happen-before `self.1.store` given that they are run in the same thread. I am not sure I understand your question.

There are two options as far as I can tell:
- Have a single bit that signifies in one of the atomics that initialization is done. You first write all the other data, then write this bit using release-semantics. If you ever observe this bit being set using acquire-semantics, then you can use relaxed-semantics to get all the other values.
- Have a single bit in each atomic value. That way you can load and check them independently of each other and you will only ever need relaxed loads and stores.

gnzlbg: > Have a single bit that signifies in one of the atomics that initialization is done. You first write all the other data, then write this bit using release-semantics. If you ever observe this bit being set using acquire-semantics, then you can use relaxed-semantics to get all the other values.

So then I don't think I'm understanding the issue. Isn't this what we currently do?

gnzlbg: Currently we have:

```rust
// compile-time:
AtomicU32 x, y = {0xFFFFFFFF, 0xFFFFFFFF);

// run-time:
if test(y.load(relaxed) == 0xFFFFFFFF) {
      x.store(relaxed) = ..; 
      y.store(relaxed) = ..; // this bit pattern is never == 0xFFFFFFFF }
}
```

Since x.store happens-before the y.store, the only way for a thread to observe an `y.load(relaxed) != 0xFFFFFFFF` is if there has been a write to both `x` and `y`, right?

gnzlbg: What this allows is something like this to happen: 

```
y.load(relaxed) == 0xFFFFFFFF // thread 1
x.store(relaxed) = ...; // thread 1
y.load(relaxed) == 0xFFFFFFFF // thread 2
x.store(relaxed) = ...; // thread 2
y.store(relaxed) = ...; // thread 2
y.store(relaxed) = ...; // thread 1
```

where two threads both end up writing to x and y. But that's ok AFAICT.

bjorn3: Cant the compiler reorder the store of y before the store of x, because both are relaxed?

gnzlbg: That's why I asked above if there is a way to make the store of x happen before the store of y. That's all we would need here for correctness AFAICT.

bjorn3: https://bartoszmilewski.com/2008/12/01/c-atomics-and-memory-ordering/

> memory_order_release: preceding stores are not moved past the current store or any subsequent stores.

gnzlbg: So would it suffice to change this to:

```rust
// run-time:
if y.load(relaxed) == 0xFFFFFFFF) {
      x.store(relaxed) = ..; 
      y.store(release) = ..never 0xFFFFFFFF..;
}
```

?

gnzlbg: @bjorn3 that blog post also says:

> the x86 guarantees acquire semantics for loads and release semantics for stores

so... maybe it only matters for compiler re-orderings, but we don't really want any here anyways. So we can probably just write:

```rust
// run-time:
if y.load(acquire) == 0xFFFFFFFF) {
      x.store(relaxed) = ..; 
      y.store(release) = ..never 0xFFFFFFFF..;
}
```

and call it a day.

TethysSvensson: What we have currently is roughly equivalent to this:

https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=81bab903b3535435ad9f33afbcfaf252

Your solution with using acquire and release is correct basically what I suggested in my first option. The downside of it is that you need to use acquire and release.

gnzlbg: @TethysSvensson I think it is more like [this](https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=d64b6a6d105924a8633912b9db6aba44) (we never change the value of the atomics after the first assignment). 

I also tried to add Acquire / Release to your example without much luck, I still get a bad count.

TethysSvensson: I cannot get my example to work either. I think it was broken.

You are right that the value is never reset in the current code. I only reset it to run the experiment multiple times.

The point I am trying to make is that there is currently nothing introduces a synchronization point across threads. So even though it works on x86 which has a very strong memory model, I think the code can give the wrong result when only relying on the LLVM memory model.

gnzlbg: cc @RalfJung

gnzlbg: > The point I am trying to make is that there is currently nothing introduces a synchronization point across threads

Yes, I am able to follow this much. I also understand that Release+Acquire would suffice here. 

What I'm still not sure about is why do we need the Acquire at all? IIUC having a relaxed store followed by a release store, and then doing a relaxed load on the second store, should be enough to guarantee that both stores happen before the relaxed load that observes the initialized cache.

talchas: As far as the C11 memory model (and thus presumably the primary target of the llvm memory model) goes, a release without an acquire is meaningless. As a practical example, the problematic code is actually something like 
```rust
if y.load(relaxed) == 0xFFFFFFFF) {
      x.store(relaxed) = foo; 
      y.store(release) = bar; //..never 0xFFFFFFFF..
}
let r = x.load(relaxed);
```

and the `x.load(relaxed)` could be moved before the if statement, so long as it's corrected for the write inside of the branch, ie it could be:

```rust
let mut r = x.load(relaxed);
if y.load(relaxed) == 0xFFFFFFFF) {
      r = foo;
      x.store(relaxed) = r; 
      y.store(release) = bar; //..never 0xFFFFFFFF..
}
```

making it be `y.load(acquire)` prevents this

gnzlbg:
Thanks, that clarifies it for me! I think I would prefer to pursue the
acquire/release for x86 32-bit because it is simple and without a benchmark
that shows and improvement it is hard to tell weather complicating the
implementation is worth it, particularly for 32-bit x86 which is a target
with many performance pitfalls. If someone wants to work on a benchmark,
i’d Still recommend to try to land a commented acquire / release approach
first, so that we can fix the soundness bug in the simplest way possible,
and also have a baseline to which compare the optimizations.

For x86-64 I’d prefer to keep the code as simple as possible.

TethysSvensson:
I really don't see any possibility of the current code having faster in the common case for either 32-bit or 64-bit. Here are a summary of the atomic counts in the different versions:

|     | Fast path | Initialization path |
| --- | --- | --- |
| Current 64-bit | 2 relaxed loads | 2 relaxed loads, 1 relaxed store |
| My 64-bit | 1 relaxed load | 1 relaxed load, 1 relaxed store |
| Current 32-bit | 2 relaxed loads | 2 relaxed loads, 2 relaxed stores |
| 32-bit with acquire/release | 1 release load, 1 relaxed load | 1 release load, 1 relaxed store, 1 acquire store, 1 relaxed load |
| My 32-bit | 1 relaxed load | 1 relaxed load, 2 relaxed stores |

As I see it I try to achieve multiple things with this PR, some of which you might not want and some of which should probably be in a different PR. I think that is the main reason why you think it is too complicated.

Please let me know which of the following you think are worth pursuing and whether you want them in a separate PR.

- General cleanup, such as removing the second argument to `cache::test` and increasing clarity by defining more constants.
- Introduction of some arguable unnecessarily defensive code (e.g. [this](https://github.com/rust-lang/stdarch/pull/811/files#diff-178500f226a696efe9b031eab45d8a2eR181) and [this](https://github.com/rust-lang/stdarch/pull/811/files#diff-178500f226a696efe9b031eab45d8a2eR124))
- Getting rid of the double loads on 64-bit.
- Solving the soundness problem on 32-bit by getting rid of double loads. This sacrifices a one of the 63 feature-bits.
- Solving the soundness problem on 32-bit by introducing acquire/release semantics. This sacrifices some performance.

gnzlbg:
So I think it would be better to split the 64-bit and 32-bit version of this PR as follows. 

For 64-bit, I think we should try to get rid of the double relaxed-load in the simplest way possible, on its own PR.

For 32-bit, I think I would prefer to split the work in two PRs. First, fix the soundness bug in the simplest way possible, using Acquire/Release - don't worry about performance here. 

Once the 32-bit implementation is sound, then figure out how to make it faster on its own different PR.

I hope that by proceeding this way the tradeoffs of each problem will be clearer, and we can make progress faster. How does that sound?

crates/std_detect/src/detect/cache.rs
RalfJung: I'd personally always use release/acquire unless it is either blatantly obvious that relaxed is correct (that's clearly not the case here) or we have benchmarks showing that release/acquire is too expensive.

TethysSvensson:
This was fixed in #837 by changing the `Relaxed` semantics to `SeqCst` semantics.

If this at some point ends begin enough of a performance problem that you would like to pursue more complicated solution based on `Relaxed` semantics, let me know and I can try to resurrect this PR.

PR #813, opened 2019-09-23
Migrate CI to GitHub Actions

alexcrichton:
This involves less secret and user management than azure pipelines, has
more concurrency by default for repos, and in general has a bit more
modern syntax!


.github/workflows/main.yml
```diff
@@ -0,0 +1,204 @@
...
+    - run: ci/dox.sh
+      env:
+        CI: 1
+    - name: Publish documentation
+      run: |
+        cd target/doc
+        git init
+        git add .
+        git -c user.name='ci' -c user.email='ci' commit -m init
+        git push -f -q https://git:${{ secrets.github_token }}@github.com/${{ github.repository }} HEAD:gh-pages
```
gnzlbg: Shouldn't this maybe use a deploy key ?

.github/workflows/main.yml
```diff
@@ -0,0 +1,204 @@
...
+    name: Check Style
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - name: Install Rust
+      run: rustup update nightly && rustup default nightly
+    - run: ci/style.sh
+
+  docs:
+    name: Build Documentation
```
alexcrichton: I split this out from the "style" step above because it was taking so long, and I figured the "sequence style before everything" was intended to filter out rustfmt/clippy issues.

.github/workflows/main.yml
alexcrichton: This is implicitly set by GitHub Actions, so no need to out of our way to set extra tokens and whatnot

.github/workflows/main.yml
```diff
@@ -0,0 +1,204 @@
...
+    - name: Install Rust
+      run: rustup update nightly && rustup default nightly
+    - run: RUST_STD_DETECT_UNSTABLE=avx cargo test --features=std_detect_env_override --manifest-path crates/std_detect/Cargo.toml env_override_no_avx
+
+  test:
+    needs: [style]
+    name: Test
+    runs-on: ${{ matrix.os }}
+    strategy:
+      max-parallel: 10
```
alexcrichton: I put everything into the same matrix here, and I did away with the stage0/stage1 tests to instead just limit the concurrency, which I believe achieves the same goal as before

.github/workflows/main.yml
```diff
@@ -0,0 +1,204 @@
...
+        - nvptx64-nvidia-cuda
+        - thumbv6m-none-eabi
+        - thumbv7m-none-eabi
+        - thumbv7em-none-eabi
+        - thumbv7em-none-eabihf
+
+        # macOS targets
+        - x86_64-apple-darwin
+        # FIXME: gh-actions build environment doesn't have linker support
+        # - i686-apple-darwin
```
gnzlbg: Maybe instead of using `macos-latest` there is a slightly older image that one can use ?

.github/workflows/main.yml
```diff
@@ -0,0 +1,204 @@
...
+          os: ubuntu-latest
+        - target: thumbv7em-none-eabi
+          os: ubuntu-latest
+        - target: thumbv7em-none-eabihf
+          os: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@master
+    - name: Install Rust (rustup)
+      run: rustup update nightly --no-self-update && rustup default nightly
```
gnzlbg: Is --no-self-update required on github actions?

.github/workflows/main.yml
gnzlbg: The concurrency limit was there because of cross-repo queues, but IIUC github actions has a per-repo queue with a per-repo limit right ?

.github/workflows/main.yml
gnzlbg: Ah, wow, this is great!

.github/workflows/main.yml
alexcrichton: I'm not 100% sure about the queueing algorithm of github actions, and I suspect that it's still in flux. I believe it's per repo though yeah now that you mention this, so I can certainly remove this. Worth mentioning if it's needed in the future though!

.github/workflows/main.yml
alexcrichton: Oh I [commented above](https://github.com/rust-lang/stdarch/pull/813#discussion_r327336365) but this is implicitly set by github actions so we should be good to go without a deploy key.

.github/workflows/main.yml
alexcrichton: Unfortunately there's only one image right now (aliased as `macos-latest`) and it doesn't have support for i686-apple-darwin, but I don't think i686-apple-darwin is covering much that i686-unknown-linux-gnu isn't already covering.

.github/workflows/main.yml
alexcrichton: Unfortunately it is for now on Windows yeah. The windows environment has permissions such that you can't self-update the `rustup.exe` executable. They're purportedly working on fixing it, but it's not fixed yet :(

.github/workflows/main.yml
gnzlbg: The main reason the latest image doesn't support it is because the target is deprecated, so...

.github/workflows/main.yml
gnzlbg: Azure had the same issue :/

alexcrichton: Oh it's worth pointing out that azure pipelines and github actions share most of their infrastructure, so limitations in one tend to be in the other as well...

.github/workflows/main.yml
alexcrichton: Yeah at some point we're gonna need to start talking in rust-lang/rust about dropping the target, but that's still a bit down the road!

gnzlbg: Wait I was wrong. MacOSX 10.13 deprecated the target, and now it is removed in 10.14, and 10.15 is in beta.. As long Azure or Travis offer VMs with MacOSX <= 10.13 rust-lang/rust can produce binaries for it, but it wouldn't surprise me if github actions doesn't add VMs for it.

alexcrichton: True yeah, at this time Github Actions [only supports 10.14](https://help.github.com/en/articles/virtual-environments-for-github-actions#supported-virtual-environments-and-hardware-resources) and doesn't have old enough Xcode sdks to target i686-apple-darwing (or at least not that I've been able to figure out). GitHub Actions is still somewhat in flux though, so you're right that they may eventually add support

PR #817, opened 2019-09-30
Deprecate MXCSR-related intrinsics

### LINKED ISSUE #781: Deprecate MXCSR intrinsics
Modifying MXCSR is insta-UB in LLVM, and inspecting it is kind of pointless since LLVM can change it behind your back however it wants (e.g. depending on the opt-level). 

We should deprecate these and tell people not to use them, since they won't do what they want them to do.

Users that want to mess with MXCSR, should do so in such a way that doesn't trigger UB in LLVM. For example, by opening a single inline `asm!` block, where they save the register, modify it, execute some code, and then restore it, before leaving the inline assembly block.

cc @rkruppe - thanks for suggesting this.

GabrielMajeri:
Fixes #781.

### DIFF
--- a/crates/core_arch/src/x86/sse.rs
+++ b/crates/core_arch/src/x86/sse.rs
@@ -1372,7 +1372,7 @@ pub unsafe fn _mm_sfence() {
 ///
 /// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_getcsr)
 #[rustc_deprecated(
-    since = "1.40.0",
+    since = "1.59.0",
     reason = "accessing floating point exception state causes undefined behavior in LLVM, see https://github.com/rust-lang/stdarch/issues/781"
 )]
 #[inline]
@@ -1391,7 +1391,7 @@ pub unsafe fn _mm_getcsr() -> u32 {
 ///
 /// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_setcsr)
 #[rustc_deprecated(
-    since = "1.40.0",
+    since = "1.59.0",
     reason = "accessing floating point exception state causes undefined behavior in LLVM, see https://github.com/rust-lang/stdarch/issues/781"
 )]
 #[inline]
--- a/crates/core_arch/src/x86/sse.rs
+++ b/crates/core_arch/src/x86/sse.rs
@@ -1371,14 +1371,14 @@ pub unsafe fn _mm_sfence() {
 /// For more info see [`_mm_setcsr`](fn._mm_setcsr.html)
 ///
 /// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_getcsr)
-#[rustc_deprecated(
-    since = "1.59.0",
-    reason = "accessing floating point exception state causes undefined behavior in LLVM, see https://github.com/rust-lang/stdarch/issues/781"
-)]
 #[inline]
 #[target_feature(enable = "sse")]
 #[cfg_attr(test, assert_instr(stmxcsr))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
+#[rustc_deprecated(
+    since = "1.59.0",
+    reason = "accessing floating point exception state causes undefined behavior in LLVM, see https://github.com/rust-lang/stdarch/issues/781"
+)]
 pub unsafe fn _mm_getcsr() -> u32 {
     let mut result = 0_i32;
     stmxcsr((&mut result) as *mut _ as *mut i8);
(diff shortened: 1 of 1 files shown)


crates/core_arch/src/x86/sse.rs
```diff
@@ -1470,6 +1470,7 @@ pub unsafe fn _mm_sfence() {
 /// For more info see [`_mm_setcsr`](fn._mm_setcsr.html)
 ///
 /// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_getcsr)
+#[rustc_deprecated(since = "1.38.0", reason = "modifying MXCSR causes undefined behavior in LLVM, see https://github.com/rust-lang/stdarch/issues/781")]
```
bjorn3: If this is merged into rustc, it should hit stable at version 1.40.0, so please use that as since version.

GabrielMajeri: Thanks, fixed now

gnzlbg:
cc @rkruppe 

Is reading the MXCSR ok ? If not, why is that UB ?

rkruppe:
[As far as LLVM is concerned](https://llvm.org/docs/LangRef.html#floating-point-environment):

> The default LLVM floating-point environment assumes that floating-point instructions do not have side effects. Results assume the round-to-nearest rounding mode. No floating-point exception state is maintained in this environment.

This doesn't contain the magic words "behavior is undefined" but if you believe that *writing* MXCSR is UB, then this wording indicates that reading it is also UB because the UB flows from the *assumption* that the LLVM IR instructions are side-effect free and both reading and writing FP exception state conflicts with that.

crates/core_arch/src/x86/sse.rs
```diff
@@ -1470,6 +1470,10 @@ pub unsafe fn _mm_sfence() {
 /// For more info see [`_mm_setcsr`](fn._mm_setcsr.html)
 ///
 /// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_getcsr)
+#[rustc_deprecated(
+    since = "1.40.0",
+    reason = "modifying MXCSR causes undefined behavior in LLVM, see https://github.com/rust-lang/stdarch/issues/781"
```
rkruppe: I do think this should be deprecated but the current reason isn't very convincing on this function since it doesn't modify MXCSR. Perhaps reword it like this:

> accessing floating point exception state causes undefined behavior in LLVM

crates/core_arch/src/x86/sse.rs
```diff
@@ -1607,6 +1611,10 @@ pub unsafe fn _mm_getcsr() -> u32 {
 ///
 ///
 /// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_setcsr)
```
rkruppe: I think if this function is going to be deprecated as "impossible to use correctly" its docs should be mostly removed or replaced with some text saying that much. Otherwise, the function's rustdoc page is a small bar saying it's deprecated dwarfed by multiple screens full of text and examples showing what wonderful things it can supposedly be used for.

gnzlbg:
> This doesn't contain the magic words "behavior is undefined" but if you believe that writing MXCSR is UB, then this wording indicates that reading it is also UB because the UB flows from the assumption that the LLVM IR instructions are side-effect free and both reading and writing FP exception state conflicts with that.

What's the side-effect of reading the MXCSR register ?

rkruppe:
> What's the side-effect of reading the MXCSR register ?

Reading doesn't have side effects AFAIK, but doing reads contradicts the assumption that floating point instructions don't have side effects (because they affect which value gets read from MXCSR).

GabrielMajeri:
I've rebased the PR, the build failures seem to be unrelated to the MXCSR changes.

Amanieu:
The build failures should be fixed now, can you rebase again?

crates/core_arch/src/x86/sse.rs
```diff
@@ -1371,6 +1371,10 @@ pub unsafe fn _mm_sfence() {
 /// For more info see [`_mm_setcsr`](fn._mm_setcsr.html)
 ///
 /// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_getcsr)
+#[rustc_deprecated(
+    since = "1.40.0",
```
bjorn3, line 1375:
```suggestion
    since = "1.59.0",
```
this is the current nightly.

crates/core_arch/src/x86/sse.rs
```diff
@@ -1383,131 +1387,13 @@ pub unsafe fn _mm_getcsr() -> u32 {
...
-/// ```rust,ignore
-/// _MM_SET_FLUSH_ZERO_MODE(_MM_FLUSH_ZERO_OFF); // turn off (default)
-/// _MM_SET_FLUSH_ZERO_MODE(_MM_FLUSH_ZERO_ON); // turn on
-/// ```
-///
+/// This function cannot be used safely as it causes undefined behavior in LLVM.
 ///
 /// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_setcsr)
+#[rustc_deprecated(
+    since = "1.40.0",
```
bjorn3, line 1394:
```suggestion
    since = "1.59.0",
```

crates/core_arch/src/x86/sse.rs
GabrielMajeri: Whoops, I didn't catch that! Thanks.

GabrielMajeri:
@Amanieu I've fixed some of the tests, but there are still others which use `_MM_GET_EXCEPTION_STATE`/`_MM_SET_EXCEPTION_STATE` (for example, `test_mm_comieq_ss_vs_ucomieq_ss`). Should I just get rid of those tests as well, or allow them to use the deprecated functions (even though it's sort-of undefined behavior)?

Amanieu:
I think it's fine to leave the tests as they are for now and just allow the warning.

Amanieu:
With that said, there have been [changes](https://reviews.llvm.org/D68121) in LLVM since this PR was opened in 2019 and I am not sure if the original reasons for deprecating these functions still apply.

GabrielMajeri:
@Amanieu Hmm, you're right, I'm not convinced using them is instantly undefined behavior anymore. I'm going to close this PR for now and leave a comment on the original issue, if someone's interested in further investigating the problem.

PR #818, opened 2019-10-06
Remove azure pipelines badges

taiki-e:


gnzlbg:
@alexcrichton are there any github action badges that we should add ?

PR #820, opened 2019-10-06
Update proc-macro2, syn, and quote to 1.0

taiki-e:


gnzlbg:
We need to fix CI before landing this.

taiki-e:
It seems that building documentation has also failed (related to #815). 
It probably need to specify an edition.

lu-zero:
I guess it should be good now?

gnzlbg:
@lu-zero do the emulated test still fail ?

mati865:
Emulated tests hangs and is killed because of 6 hour timeout.

PR #821, opened 2019-10-10
Add the github Action badge

lu-zero:
Here the action badge :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,9 @@
 stdarch - Rust's standard library SIMD components
 =======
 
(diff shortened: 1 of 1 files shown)


bjorn3:
It says "no status".

lu-zero:
When I tested it it said failure.

gnzlbg:
Let's try this out!

lu-zero:
Works as intended, when the master is being tested it shows "no status", otherwise the result of the run.

PR #824, opened 2019-10-22
Replace rustfmt::skip custom inner attribute with rustfmt.toml

mati865:
Fixes errors like https://travis-ci.com/rust-lang-nursery/packed_simd/jobs/248339096#L367


mati865:
Style check is green, documentation is being fixed in https://github.com/rust-lang/stdarch/pull/820 and emulated test remains broken.

PR #825, opened 2019-10-29
Add `#[rustc_args_required_const]` to `simd_shuffle`

ecstatic-morse:
Currently, these have to be special-cased in the promotion logic in `rustc`, as they predate the existence of the attribute. Add the attribute now so we can clean up the special-case later in the compiler.

I'm not sure if there are additional considerations here. Perhaps someone could fill me in if so?


gnzlbg:
Thanks, this is good.

PR #829, opened 2019-11-27
Run-time feature detection for Aarch64 on Windows.

makotokato:
Windows has API to detect CPU features for crc32 and ARM crypto extension.

### DIFF
--- a/crates/std_detect/src/detect/os/windows/aarch64.rs
+++ b/crates/std_detect/src/detect/os/windows/aarch64.rs
@@ -1,0 +1,55 @@
+//! Run-time feature detection for Aarch64 on Windows.
+
+use crate::detect::{cache, Feature};
+
+/// Try to read the features using IsProcessorFeaturePresent.
+pub(crate) fn detect_features() -> cache::Initializer {
+    type DWORD = u32;
+    type BOOL = i32;
+
+    const FALSE: BOOL = 0;
+    // The following Microsoft documents isn't updated for aarch64.
+    // https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-isprocessorfeaturepresent
+    // These are defined in winnt.h of Windows SDK
+    const PF_ARM_NEON_INSTRUCTIONS_AVAILABLE: u32 = 19;
+    const PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE: u32 = 30;
+    const PF_ARM_V8_CRC32_INSTRUCTIONS_AVAILABLE: u32 = 31;
+
+    extern "system" {
+        pub fn IsProcessorFeaturePresent(ProcessorFeature: DWORD) -> BOOL;
+    }
+
+    let mut value = cache::Initializer::default();
+    {
+        let mut enable_feature = |f, enable| {
+            if enable {
+                value.set(f as u32);
+            }
+        };
+
+        // Some features such Feature::fp may be supported on current CPU,
+        // but no way to detect it by OS API.
+        // Also, we require unsafe block for the extern "system" calls.
+        unsafe {
+            enable_feature(
+                Feature::asimd,
+                IsProcessorFeaturePresent(PF_ARM_NEON_INSTRUCTIONS_AVAILABLE) != FALSE,
+            );
+            enable_feature(
+                Feature::crc,
+                IsProcessorFeaturePresent(PF_ARM_V8_CRC32_INSTRUCTIONS_AVAILABLE) != FALSE,
+            );
+            // PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE means aes, sha1, sha2 and
+            // pmull support
+            enable_feature(
+                Feature::crypto,
+                IsProcessorFeaturePresent(PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE) != FALSE,
+            );
+            enable_feature(
+                Feature::pmull,
+                IsProcessorFeaturePresent(PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE) != FALSE,
+            );
+        }
+    }
+    value
+}
--- a/crates/std_detect/src/detect/mod.rs
+++ b/crates/std_detect/src/detect/mod.rs
@@ -106,6 +106,9 @@ cfg_if! {
         mod aarch64;
         #[path = "os/freebsd/mod.rs"]
         mod os;
+    } else if #[cfg(all(target_os = "windows", target_arch = "aarch64"))] {
+        #[path = "os/windows/aarch64.rs"]
+        mod os;
     } else {
         #[path = "os/other.rs"]
         mod os;


crates/std_detect/src/detect/os/windows/aarch64.rs
```diff
@@ -0,0 +1,30 @@
...
+            IsProcessorFeaturePresent(PF_ARM_NEON_INSTRUCTIONS_AVAILABLE) != 0
+        });
+        enable_feature(Feature::crc, unsafe {
+            IsProcessorFeaturePresent(PF_ARM_V8_CRC32_INSTRUCTIONS_AVAILABLE) != 0
+        });
+        enable_feature(Feature::crypto, unsafe {
+            IsProcessorFeaturePresent(PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE) != 0
+        });
+        enable_feature(Feature::pmull, unsafe {
+            IsProcessorFeaturePresent(PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE) != 0
```
bjorn3: copy paste error?

crates/std_detect/Cargo.toml
```diff
@@ -25,6 +25,9 @@ maintenance = { status = "experimental" }
 libc = { version = "0.2", optional = true, default-features = false }
 cfg-if = "0.1.10"
 
+[target.'cfg(target_os="windows")'.dependencies]
+winapi = {version = "0.3.8", features = ["processthreadsapi"]}
```
bjorn3: can this depend on aarch64 too?

crates/std_detect/src/detect/os/windows/aarch64.rs
makotokato: On Windows/aarch64, `PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE` means that aes, sha1, sha2 and pmull support.

crates/std_detect/Cargo.toml
makotokato: winapi uses on windows/aarch64 only. Should I add additional condition?

crates/std_detect/src/detect/os/windows/aarch64.rs
bjorn3: Maybe add a comment?

crates/std_detect/Cargo.toml
```diff
@@ -25,6 +25,9 @@ maintenance = { status = "experimental" }
 libc = { version = "0.2", optional = true, default-features = false }
 cfg-if = "0.1.10"
 
+[target.'cfg(target_os="windows")'.dependencies]
```
gnzlbg:
```suggestion
[target.'cfg(all(target_os = "windows", target_arch = "aarch64"))'.dependencies]
```

crates/std_detect/Cargo.toml
gnzlbg: cc @alexcrichton @retep998: this would make `winapi` a dependency of `libstd`, but IIUC this is currently not the case:

* license of winapi looks compatible with the standard runtime
* winapi has a build.rs that we'd need to make sure works within rust's build system
* winapi might not be worth pulling into libstd for this particular feature only due to other issues (build times, dependency, etc.)

IIRC, pulling a new dependency into libstd requires the libs team to sign on on it. If you both think this is worth doing, I can try to find answers to these questions, and we can have a mini FCP on it.

makotokato:
If I shouldn't use winapi, I remove this, then I add FFI to call Win32 API without winapi.

gnzlbg:
> If I shouldn't use winapi, I remove this, then I add FFI to call Win32 API without winapi.

I think this feature is important, and removing winapi would definitely help landing this faster. After this has landed on rust-lang/rust, you can submit another PR that changes this to use winapi, and we can work out through all organizational issues, but that would usually take on the order of weeks if an FCP is required.

crates/std_detect/src/detect/os/windows/aarch64.rs
```diff
@@ -0,0 +1,42 @@
+//! Run-time feature detection for Aarch64 on Windows.
+
+use crate::detect::{cache, Feature};
+
+/// Try to read the features from OS API.
+pub(crate) fn detect_features() -> cache::Initializer {
+    const PF_ARM_NEON_INSTRUCTIONS_AVAILABLE: u32 = 19;
```
gnzlbg: Could we add a comment with a link to where you got these from? Or in which file they are? That would help with maintenance.

crates/std_detect/src/detect/os/windows/aarch64.rs
```diff
@@ -0,0 +1,42 @@
...
+    const PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE: u32 = 30;
+    const PF_ARM_V8_CRC32_INSTRUCTIONS_AVAILABLE: u32 = 31;
+
+    extern "system" {
+        pub fn IsProcessorFeaturePresent(ProcessorFeature: u32) -> u32;
```
gnzlbg: the first `u32` is a `DWORD` and the second one is a `BOOL`, right ? Could we maybe use the same ctypes that win API uses for clarity? (e.g. just use a type alias `type DWORD = u32;` at the beginning of the file).

crates/std_detect/src/detect/os/windows/aarch64.rs
```diff
@@ -0,0 +1,42 @@
...
+        let mut enable_feature = |f, enable| {
+            if enable {
+                value.set(f as u32);
+            }
+        };
+
+        // Some features such Feature::fp may be supported on current CPU,
+        // but no way to detect it by OS API.
+
+        enable_feature(Feature::asimd, unsafe {
```
gnzlbg: It's ok to put all of these inside a single `unsafe { ... }` block to avoid repeating unsafe - just add a comment stating that `unsafe` is required for the extern "system" calls.

crates/std_detect/Cargo.toml
retep998: Currently I believe `winapi` being a dependency of `std` is blocked on upgrading `winapi`'s minimum supported Rust version which will happen in `winapi` 0.4.

gnzlbg: Thanks @retep998, I think we can land this as is, and refactor it into using `winapi` once it becomes a dependency of `std`. I'll open a tracking issue in this repo to not forget about this once this is merged.

crates/std_detect/src/detect/os/windows/aarch64.rs
makotokato: This API's document is https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-isprocessorfeaturepresent, but that document isn't updated for aarch64.

gnzlbg: Thanks, where did you got the values for `aarch64` from ? The windows actual header files? Can those be viewed online somewhere?

makotokato: no online document.  But chromium has copy of define as https://chromium.googlesource.com/crashpad/crashpad/+/doc/compat/non_win/winnt.h

gnzlbg: Thanks, can you add that link to the comment as well? After that, this is good to merge!

makotokato: The latest PR is already added for this.

gnzlbg:
Thank you!

PR #830, opened 2019-11-30
Add CRC32 detection to arm32

makotokato:
Cortex-A53 etc has 32-bit mode, and it can use crc32 instruction sets even if 32-bit (A32 mode) and CPU is supported. So we should add crc32 detection for arm32.


gnzlbg:
Thank you!

PR #832, opened 2019-12-07
Don't use `&&` on GitHub Actions.

makotokato:
According to [the document](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idsteps), default shell of Windows is PowerShell. But `&&` isn't supported on PowerShell yet.

So we should use multi line run syntax instead.


gnzlbg:
Thanks!

Issue #614, opened 2018-12-10
Split up aarch64's "crypto" feature?


alexcrichton:
First brought up [here](https://github.com/rust-lang-nursery/stdsimd/pull/611/files#r240009793), but we may not want to follow in LLVM's footsteps for this!

gnzlbg:
I think this was already discussed somewhere else, e.g., https://github.com/rust-lang/rust/issues/29717#issuecomment-267526911 .

That is, we offer `crypto` because ARM "specifies" this feature (and GCC and Clang also do so), but there is nothing standing in the way of exposing each of the features that `crypto` enables independently. What would need to be done is:

* white list the new features in `rust-lang/rust`
* update the `#[target_feature]` of intrinsics requiring `crypto` to only require exactly the "sub-features" they need 

And that's it, everything should still work "as is" for those enabling `crypto`.

myfreeweb:
hmm, are there any actual chips that only expose some of the crypto features but not others? I've only ever seen `AES+PMULL,SHA1,SHA2,CRC32` together..

makotokato:
> hmm, are there any actual chips that only expose some of the crypto features but not others? I've only ever seen `AES+PMULL,SHA1,SHA2,CRC32` together..

RasPi3 (and RasPi4?) doesn't support ARM crypto extension, so it only support CRC32.

newpavlov:
The `crypto` feature looks like a small mess to me. The problem is that it means different features for different ARM versions. See [here](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.100067_0610_00_en/chr1392632801932.html) (click on the "cryptographic extensions" link). For ARMv8.4-A it includes SHA1, SHA256, SHA512, SHA3, AES, SM3, SM4, while for older versions only SHA1, SHA256, AES. Thus I think we should use finer-grained features to avoid confusion.

briansmith:
> hmm, are there any actual chips that only expose some of the crypto features but not others? 

Linux removed `aes` but not `pmull` or other crypto features from being reported on `getauxval()` to work around a bug in some ARM64 processors (in 32-bit mode): https://github.com/torvalds/linux/commit/44b3834b2eed595af07021b1c64e6f9bc396398b.

[...]

PR #833, opened 2019-12-20
Add crypto target feature detection to arm32

makotokato:
ARM crypto extension is supported on A32 mode.


Amanieu:
I'm happy to accept this as it is, but eventually we will want to split the `crypto` features up: #614

PR #834, opened 2019-12-20
Support crc32 even if on arm32

makotokato:
CRC32 is supported on A32 and T32.

### DIFF
--- a/crates/core_arch/src/arm/mod.rs
+++ b/crates/core_arch/src/arm/mod.rs
@@ -34,6 +34,11 @@ mod neon;
 ))]
 pub use self::neon::*;
 
+#[cfg(any(target_arch = "aarch64", target_feature = "v7"))]
+mod crc;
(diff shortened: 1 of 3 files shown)


makotokato:
> LLVM ERROR: Cannot select: intrinsic %llvm.arm.crc32b

Hmm, I will recheck this error on my armv8 aarch32 linux machine.

gnzlbg:
Let me know if you need help. Maybe a target feature is missing on arm, or the intrinsic has a likely different name, or LLVM does not support the intrinsic for arm 32-bit yet.

crates/core_arch/src/arm/mod.rs
```diff
@@ -34,6 +34,11 @@ mod neon;
 ))]
 pub use self::neon::*;
 
+#[cfg(any(target_arch = "aarch64", target_feature = "v7"))]
```
Amanieu: Why is `v7` required here?

PR #835, opened 2019-12-20
Add const unstability attributes

oli-obk:
These are needed for rustc to be able to correctly handle stability of constness of intrinsics. Without either `rustc_const_unstable` or `rustc_const_stable` an intrinsic is not const evaluable at all.

related: https://github.com/rust-lang/rust/issues/61495


gnzlbg:
Thank you!

PR #837, opened 2020-01-05
Fix race condition in feature cache on 32 platforms

matklad:
If we observe that the second word is initialized, we can't really
assume that the first is initialized as well. So check each word
separately.

r? @gnzlbg 

I've found this by just looking at the code, and I didn't actually test this (`cargo test --target i686-unknown-linux-gnu` failed with some build.rs error which I haven't looked into).

### DIFF
--- a/crates/std_detect/src/detect/cache.rs
+++ b/crates/std_detect/src/detect/cache.rs
@@ -89,9 +89,10 @@ static CACHE: [Cache; 2] = [Cache::uninitialized(), Cache::uninitialized()];
 /// Note: the last feature bit is used to represent an
 /// uninitialized cache.
 ///
-/// Note: we use `Relaxed` atomic operations, because we are only interested
+/// Note: we can use `Relaxed` atomic operations, because we are only interested
 /// in the effects of operations on a single memory location. That is, we only
-/// need "modification order", and not the full-blown "happens before".
+/// need "modification order", and not the full-blown "happens before". However,
+/// we use `SeqCst` just to be on the safe side.
 struct Cache(AtomicUsize);
 
 impl Cache {
@@ -106,19 +107,19 @@ impl Cache {
     /// Is the cache uninitialized?
     #[inline]
     pub(crate) fn is_uninitialized(&self) -> bool {
-        self.0.load(Ordering::Relaxed) == usize::max_value()
+        self.0.load(Ordering::SeqCst) == usize::max_value()
     }
 
     /// Is the `bit` in the cache set?
     #[inline]
     pub(crate) fn test(&self, bit: u32) -> bool {
-        test_bit(self.0.load(Ordering::Relaxed) as u64, bit)
+        test_bit(self.0.load(Ordering::SeqCst) as u64, bit)
     }
 
     /// Initializes the cache.
     #[inline]
     fn initialize(&self, value: usize) {
-        self.0.store(value, Ordering::Relaxed);
+        self.0.store(value, Ordering::SeqCst);
     }
 }
 
--- a/crates/std_detect/src/detect/cache.rs
+++ b/crates/std_detect/src/detect/cache.rs
@@ -5,11 +5,7 @@
 
 use crate::sync::atomic::Ordering;
 
-#[cfg(target_pointer_width = "64")]
-use crate::sync::atomic::AtomicU64;
-
-#[cfg(target_pointer_width = "32")]
-use crate::sync::atomic::AtomicU32;
+use crate::sync::atomic::AtomicUsize;
 
 /// Sets the `bit` of `x`.
 #[inline]
@@ -30,7 +26,7 @@ const fn unset_bit(x: u64, bit: u32) -> u64 {
 }
 
 /// Maximum number of features that can be cached.
-const CACHE_CAPACITY: u32 = 63;
+const CACHE_CAPACITY: u32 = 62;
 
 /// This type is used to initialize the cache
 #[derive(Copy, Clone)]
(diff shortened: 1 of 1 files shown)


matklad:
Oh, this is another spin on https://github.com/rust-lang/stdarch/pull/811 actually! But this version of the seems simpler.

lu-zero:
Can you please try to rebase and see if now the tests pass?

matklad:
rebased!

crates/std_detect/src/detect/cache.rs
```diff
@@ -81,83 +77,44 @@ impl Initializer {
 }
 
 /// This global variable is a cache of the features supported by the CPU.
+#[cfg(target_pointer_width = "64")]
 static CACHE: Cache = Cache::uninitialized();
 
-/// Feature cache with capacity for `CACHE_CAPACITY` features.
+/// This global variable is a cache of the features supported by the CPU.
+#[cfg(target_pointer_width = "32")]
+static CACHE: [Cache; 2] = [Cache::uninitialized(), Cache::uninitialized()];
```
alexcrichton: Instead of defining this differently on each platform, could this perhaps be unified a bit like so?

```
#[cfg(target_pointer_width = "64")]
 static CACHE: [Cache; 1] = [Cache::uninitialized()];
#[cfg(target_pointer_width = "32")]
 static CACHE: [Cache; 2] = [Cache::uninitialized(), Cache::uninitialized()];
```

I think that'd help cut down on the `#[cfg]` traffic

crates/std_detect/src/detect/cache.rs
```diff
@@ -81,83 +77,44 @@ impl Initializer {
...
-        test_bit(CACHE.0.load(Ordering::Relaxed), bit)
+        test_bit(self.0.load(Ordering::Relaxed) as u64, bit)
     }
 
     /// Initializes the cache.
     #[inline]
-    pub(crate) fn initialize(&self, value: Initializer) {
-        self.0.store(value.0, Ordering::Relaxed);
+    fn initialize(&self, value: usize) {
+        self.0.store(value, Ordering::Relaxed);
```
alexcrichton: I would personally say this should all become `SeqCst`. Especially on 32-bit platforms where a write to one location is trying to imply a write to a different location. If it's necessarily discovered later I think we can start relaxing this.

crates/std_detect/src/detect/cache.rs
matklad: Awesome hint, thanks!

crates/std_detect/src/detect/cache.rs
matklad: The idea behind this PR is exactly to write a code in such a way that each cache instance is independent, so that we are in the single memory location world, where Relaxed works.

I think using `SeqCst` would obscure this intention. Instead, I've added a comment explaining why we use the most reckless memory ordering. 

I can still change this to SeqCst, but I think that would not guarantee anything on top without *also* changing `test` to use `SeqCst`, and `test` looks like a potentially hot method.

crates/std_detect/src/detect/cache.rs
```diff
@@ -167,16 +128,32 @@ cfg_if::cfg_if! {
...
-            CACHE.initialize(value);
+            do_initialize(value);
         }
     }
 }
 
+#[cfg(target_pointer_width = "32")]
+const CACHE_BITS: u32 = 31;
+#[cfg(target_pointer_width = "64")]
+const CACHE_BITS: u32 = 63;
```
alexcrichton: Could this constant either be removed or defined in terms of `mem::size_of::<usize>()` instead of using `#[cfg]`?

crates/std_detect/src/detect/cache.rs
```diff
@@ -194,8 +171,18 @@ pub(crate) fn test<F>(bit: u32, f: F) -> bool
 where
     F: FnOnce() -> Initializer,
 {
-    if CACHE.is_uninitialized() {
-        initialize(f());
+    #[cfg(target_pointer_width = "32")]
```
alexcrichton: I think this `#[cfg]` could be avoided and the math here could be done with `mem::size_of::<usize>()` as well

crates/std_detect/src/detect/cache.rs
```diff
@@ -167,16 +128,32 @@ cfg_if::cfg_if! {
...
+const CACHE_BITS: u32 = 31;
+#[cfg(target_pointer_width = "64")]
+const CACHE_BITS: u32 = 63;
+
+const CACHE_MASK: usize = (1 << CACHE_BITS) - 1;
+
+#[inline]
+fn do_initialize(value: Initializer) {
+    CACHE[0].initialize((value.0) as usize & CACHE_MASK);
+    #[cfg(target_pointer_width = "32")]
```
alexcrichton: Like elsewhere, can the `#[cfg]` be removed here to work with shifts instead?

crates/std_detect/src/detect/cache.rs
alexcrichton: Ah I think I was mistaken in looking at how the testing was done, I thought when the test for "am I initialized" was done by looking at the first word, not the word we're interested in.

I would still personally prefer to see `SeqCst` everywhere because it hardly matters here in terms of performance and anything that makes atomics "a bit more correct" is something I'm personally a fan of at least.

crates/std_detect/src/detect/cache.rs
matklad: Actually it could!

matklad:
I've tried to get rid of cfgs using clever indexing, but unfortunately rustc was just smart enough to figure out that I try to index out of bounds into array and just dumb enough to not realize that it's dead code :)

In the last commit, I've just made both paths use two slots.

alexcrichton:
👍

PR #838, opened 2020-01-25
Add Icelake avx512 features

lu-zero:
As documented in https://software.intel.com/sites/default/files/managed/c5/15//architecture-instruction-set-extensions-programming-reference.pdf

### DIFF
--- a/crates/std_detect/src/detect/os/x86.rs
+++ b/crates/std_detect/src/detect/os/x86.rs
@@ -210,6 +210,14 @@ pub(crate) fn detect_features() -> cache::Initializer {
                         enable(extended_features_ebx, 31, Feature::avx512vl);
                         enable(extended_features_ecx, 1, Feature::avx512vbmi);
                         enable(extended_features_ecx, 14, Feature::avx512vpopcntdq);
+                        enable(extended_features_ecx, 6, Feature::avx512vbmi2);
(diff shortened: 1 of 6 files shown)


crates/std_detect/src/detect/os/x86.rs
```diff
@@ -210,6 +210,14 @@ pub(crate) fn detect_features() -> cache::Initializer {
...
                         enable(extended_features_ecx, 1, Feature::avx512vbmi);
                         enable(extended_features_ecx, 14, Feature::avx512vpopcntdq);
+                        enable(extended_features_ecx, 6, Feature::avx512vbmi2);
+                        enable(extended_features_ecx, 8, Feature::avx512gfni);
+                        enable(extended_features_ecx, 9, Feature::avx512vaes);
+                        enable(extended_features_ecx, 10, Feature::avx512vpclmulqdq);
+                        enable(extended_features_ecx, 11, Feature::avx512vnni);
+                        enable(extended_features_ecx, 12, Feature::avx512bitalg);
+                        enable(extended_features_ecx, 5, Feature::avx512bf16);
+                        enable(extended_features_ecx, 8, Feature::avx512vp2intersect);
```
bjorn3: Should this be sorted by which bit is used?

lu-zero: I put it in the order the document provides, but sounds a better idea to sort by bit.

PR #842, opened 2020-02-27
re-stabilize the AVX-512 features that were stabilized in Rust 1.27.0

oconnor663:
https://github.com/rust-lang/stdarch/pull/739 added per-feature
stabilization of runtime CPU feature detection. In so doing, it
de-stabilized some detection features that had been stable since Rust
1.27.0, breaking some published crates (on nightly). This commit
re-stabilizes the subset of AVX-512 detection features that were
included in 1.27.0 (that is, the pre-Ice-Lake subset). Other instruction
sets (MMX in particular) remain de-stabilized, pending a decision about
whether should ever stabilize them.

See https://github.com/rust-lang/rust/issues/68905.

### DIFF
--- a/crates/std_detect/src/detect/arch/x86.rs
+++ b/crates/std_detect/src/detect/arch/x86.rs
@@ -114,25 +114,25 @@ features! {
     /// AVX (Advanced Vector Extensions)
     @FEATURE: #[stable(feature = "simd_x86", since = "1.27.0")] avx2: "avx2";
     /// AVX2 (Advanced Vector Extensions 2)
-    @FEATURE: #[unstable(feature = "stdsimd", issue = "27731")] avx512f: "avx512f" ;
+    @FEATURE: #[stable(feature = "simd_x86", since = "1.27.0")] avx512f: "avx512f" ;
     /// AVX-512 F (Foundation)
-    @FEATURE: #[unstable(feature = "stdsimd", issue = "27731")] avx512cd: "avx512cd" ;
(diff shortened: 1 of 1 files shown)


oconnor663:
(CI failure appears unrelated.)

Amanieu:
Could you add a comment saying that the AVX512 *intrinsics* are not stable yet, only the feature check. And that these were stabilized to avoid breaking code that was already relying on it. Add a reference to rust-lang/rust#68905.

This doesn't have to be a doc comment, just a small note in the code for the next person who stumbles across this code wondering why these are stable while the intrinsics aren't.

oconnor663:
In light of https://github.com/rust-lang/rust/issues/68905#issuecomment-592176692, should we expand this PR to stabilize the entire set of AVX-512 features, and also mmx/sse4a/tbm? I don't know whether detection for those things actually works.

Amanieu:
cc @joshtriplett

joshtriplett:
:+1:

oconnor663:
Also "cmpxchg16b" and "rtm"? I haven't seen those discussed yet.

If someone who knows what they're doing would like to land this PR, and then make follow-up changes on top of it, that might be a better option :) I don't have any expertise in this area, and I'm worried about breaking things.

oconnor663:
This PR is now up to speed with https://github.com/rust-lang/rust/issues/68905#issuecomment-595376319 and is potentially ready to merge. I've sanity checked it locally by building the `blake3` crate with it, using the nightly toolchain.

PR #849, opened 2020-04-23
Stabilize all remaining x86 features for feature detection

Amanieu:
Follow-up on #842 since rav1e (https://github.com/xiph/rav1e/pull/2250) requires some additional AVX512 features that weren't stabilized.

cc https://github.com/rust-lang/rust/issues/71473


joshtriplett:
LGTM.

PR #854, opened 2020-05-07
Fix non-fp64 libcore builds

overdrivenpotato:
Building the MIPS MSA module for non-`fp64` targets fails with an LLVM error. This PR blacklists PSP OS targets from MSA support in order to fix building libcore.

### DIFF
--- a/crates/core_arch/src/mips/mod.rs
+++ b/crates/core_arch/src/mips/mod.rs
@@ -1,7 +1,12 @@
 //! MIPS
 
-mod msa;
-pub use self::msa::*;
+// Building this module (even if unused) for non-fp64 targets such as the Sony
+// PSP fails with an LLVM error. There doesn't seem to be a good way to detect
+// fp64 support as it is sometimes implied by the target cpu, so
+// `#[cfg(target_feature = "fp64")]` will unfortunately not work. This is a
+// fairly conservative workaround that only disables MSA intrinsics for the PSP.
+#[cfg(not(target_os = "psp"))] mod msa;
+#[cfg(not(target_os = "psp"))] pub use self::msa::*;
 
 #[cfg(test)]
 use stdarch_test::assert_instr;
--- a/crates/core_arch/src/mips/mod.rs
+++ b/crates/core_arch/src/mips/mod.rs
@@ -1,13 +1,10 @@
 //! MIPS
 
-// Building this module (even if unused) for non-fp64 targets such as the Sony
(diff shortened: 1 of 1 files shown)


bjorn3:
Can you use `-Ctarget_feature=-fp64`?

overdrivenpotato:
@bjorn3 Unfortunately not. Disabling fp64 via `-fp64` flag still fails with the same error:

```
LLVM ERROR: MSA requires a 64-bit FPU register file (FR=1 mode). See -mattr=+fp64.
```

It looks like building anything involving MSA intrinsics fails unless you have `+fp64`, which is not possible on the PSP.

bjorn3:
And `-Ctarget-feature=-fp64` combined with `#[target_feature = "fp64"]` on the module?

Amanieu:
Did you check if `#[cfg(target_feature = "fp64")]` actually works? We get our target features from LLVM, which should enable `fp64` if the selected CPU supports it.

overdrivenpotato:
@bjorn3 That works, I can update this PR. It does removes the module when building on non-fp64 targets, though. Is this desired? I assumed it would break runtime detection for anybody targeting non-fp64 as a baseline.

CC @Amanieu

Amanieu:
If disabling `fp64` causes LLVM errors when compiling MSA code, then there is no point in exposing those intrinsics since they are unusable anyways.

overdrivenpotato:
Fair enough. Just pushed a commit to use `#[cfg(target_feature = "fp64")]`

Amanieu:
Thanks!

overdrivenpotato:
I did a bit more testing, and it looks like this breaks `mipsel-unknown-linux-gnu` builds (and possibly others without `softfloat`) that used MSA. They were previously able to build the `msa` module without an explicit `fp64` feature being set, but this module is now not compiled in by default. I'm not sure whether this is even an issue, so I'm just dropping this comment for future reference.

Amanieu:
How was that target able to compile? You just said that LLVM can't compile the msa module if fp64 is not available.

overdrivenpotato:
@Amanieu My best guess is that `fp64` is somehow implied by the rest of the target configuration, possibly due to different CPUs.

overdrivenpotato:
I can open a new PR to revert to the OS-based blacklist if needed.

Amanieu:
I think this is fine as it is for now, but we can revert if there are any complaints.

PR #855, opened 2020-05-21
Implementation for Aarch64 TME intrinsics

vertexclique:
This enables Aarch64 TME intrinsics and documents them. We need to enable `tme` feature at `rust-lang/rust` before merging this.

### DIFF
--- a/crates/core_arch/src/aarch64/tme.rs
+++ b/crates/core_arch/src/aarch64/tme.rs
@@ -19,50 +19,50 @@ use stdarch_test::assert_instr;
 
 extern "C" {
     #[link_name = "llvm.aarch64.tstart"]
-    fn aarch64_tstart() -> i32;
+    fn aarch64_tstart() -> u64;
     #[link_name = "llvm.aarch64.tcommit"]
     fn aarch64_tcommit() -> ();
     #[link_name = "llvm.aarch64.tcancel"]
-    fn aarch64_tcancel(imm0: i64) -> ();
+    fn aarch64_tcancel(imm0: u64) -> ();
     #[link_name = "llvm.aarch64.ttest"]
-    fn aarch64_ttest() -> i32;
+    fn aarch64_ttest() -> u64;
 }
 
 /// Transaction successfully started.
-pub const _TMSTART_SUCCESS: u32 = 0x00_u32;
+pub const _TMSTART_SUCCESS: u64 = 0x00_u64;
 
 /// Extraction mask for failure reason
-pub const _TMFAILURE_REASON: u32 = 0x00007FFF_u32;
+pub const _TMFAILURE_REASON: u64 = 0x00007FFF_u64;
 
 /// Transaction retry is possible.
-pub const _TMFAILURE_RTRY: u32 = 1 << 15;
+pub const _TMFAILURE_RTRY: u64 = 1 << 15;
 
 /// Transaction executed a TCANCEL instruction
-pub const _TMFAILURE_CNCL: u32 = 1 << 16;
+pub const _TMFAILURE_CNCL: u64 = 1 << 16;
 
 /// Transaction aborted because a conflict occurred
-pub const _TMFAILURE_MEM: u32 = 1 << 17;
+pub const _TMFAILURE_MEM: u64 = 1 << 17;
 
 /// Fallback error type for any other reason
-pub const _TMFAILURE_IMP: u32 = 1 << 18;
+pub const _TMFAILURE_IMP: u64 = 1 << 18;
 
 /// Transaction aborted because a non-permissible operation was attempted
-pub const _TMFAILURE_ERR: u32 = 1 << 19;
+pub const _TMFAILURE_ERR: u64 = 1 << 19;
 
 /// Transaction aborted due to read or write set limit was exceeded
-pub const _TMFAILURE_SIZE: u32 = 1 << 20;
+pub const _TMFAILURE_SIZE: u64 = 1 << 20;
 
 /// Transaction aborted due to transactional nesting level was exceeded
-pub const _TMFAILURE_NEST: u32 = 1 << 21;
+pub const _TMFAILURE_NEST: u64 = 1 << 21;
 
 /// Transaction aborted due to a debug trap.
-pub const _TMFAILURE_DBG: u32 = 1 << 22;
+pub const _TMFAILURE_DBG: u64 = 1 << 22;
 
 /// Transaction failed from interrupt
-pub const _TMFAILURE_INT: u32 = 1 << 23;
+pub const _TMFAILURE_INT: u64 = 1 << 23;
 
 /// Indicates a TRIVIAL version of TM is available
-pub const _TMFAILURE_TRIVIAL: u32 = 1 << 24;
+pub const _TMFAILURE_TRIVIAL: u64 = 1 << 24;
 
 /// Starts a new transaction. When the transaction starts successfully the return value is 0.
 /// If the transaction fails, all state modifications are discarded and a cause of the failure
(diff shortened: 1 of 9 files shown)


crates/core_arch/src/aarch64/tme.rs
```diff
@@ -0,0 +1,124 @@
...
+/// Transaction retry is possible.
+pub const _TMFAILURE_RTRY: u32 = 1 << 15;
+
+/// Transaction cancelled.
+pub const _TMFAILURE_CNCL: u32 = 1 << 16;
+
+/// Transaction cancelled due to high memory usage.
+pub const _TMFAILURE_MEM: u32 = 1 << 17;
+
+///
```
Amanieu, line 46: Missing doc comments?

crates/core_arch/src/aarch64/tme.rs
```diff
@@ -0,0 +1,124 @@
...
+#[target_feature(enable = "tme")]
+#[cfg_attr(test, assert_instr(ttest))]
+pub unsafe fn __ttest() -> u32 {
+    aarch64_ttest() as _
+}
+
+/// Encodes cancellation reason, which is the parameter passed to [`__tcancel`]
+/// Takes cancellation reason flags and retry-ability.
+#[inline]
+pub const fn _tcancel_code(reason: u32, retryable: bool) -> u32 {
```
Amanieu, line 122: I don't see this in any ARM documentation. As a general rule we try to avoid defining our own helper functions.

crates/core_arch/src/aarch64/tme.rs
```diff
@@ -0,0 +1,124 @@
...
+//!
+//! This CPU feature is available on Aarch64 – ARMv8-A arch. onwards.
+//! This feature is in the non-neon feature set. TME specific vendor documentation can
+//! be found [TME Intrinsics Introduction][tme_intrinsics_intro].
+//!
+//! The reference is [ACLE Q4 2019][acle_q4_2019_ref].
+//!
+//! ACLE has a section for TME extensions and state masks for aborts and failure codes.
+//! In addition to that [LLVM Aarch64 Intrinsics][llvm_aarch64_int] are
+//! self explanatory for what needs to be exported.
```
Amanieu, line 11: This feels more like documentation of the implementation than documentation for the user of these intrinsics. Doc comments should try to be the latter.

crates/core_arch/src/aarch64/tme.rs
```diff
@@ -0,0 +1,124 @@
+//! ARM's Transactional Memory Extensions (TME).
+//!
+//! This CPU feature is available on Aarch64 – ARMv8-A arch. onwards.
```
Amanieu, line 3: This sentence doesn't really make sense?

vertexclique:
@Amanieu There is a problem with current tests for neon intrinsics in the test suite. Any way to get over it?

Amanieu:
I'll have a look at the tests, it doesn't seem to be due to your code.

Amanieu:
I fixed the CI failures in #857, please rebase.

vertexclique:
Rebased, it seems QEMU doesn't come with this feature extension.

crates/std_detect/tests/cpu-detection.rs
```diff
@@ -45,6 +45,7 @@ fn aarch64_linux() {
     println!("rdm: {}", is_aarch64_feature_detected!("rdm"));
     println!("rcpc: {}", is_aarch64_feature_detected!("rcpc"));
     println!("dotprod: {}", is_aarch64_feature_detected!("dotprod"));
+    println!("tme: {}", is_aarch64_feature_detected!("tme"));
```
Amanieu, line 48: You need to implement detection for this feature. Do a search for `dotprod` to see how it is done.

vertexclique: Updated with feature detection, didn't find the feature tbh. Even instructions and signatures are the same. https://github.com/llvm/llvm-project/commit/a36d31478c182903523e04eb271bbf102bfab2cc#diff-3674dac5c4241890431fece8594c3143R25-R34

What am I doing wrong? (I still suspect from qemu, checked source code, it doesn't exist there yet. Might be wrong.)

crates/core_arch/src/aarch64/tme.rs
```diff
@@ -0,0 +1,183 @@
...
+//! [tme_intrinsics_intro]: https://developer.arm.com/docs/101028/0010/transactional-memory-extension-tme-intrinsics
+//! [llvm_aarch64_int]: https://github.com/llvm/llvm-project/commit/a36d31478c182903523e04eb271bbf102bfab2cc#diff-ff24e1c35f4d54f1110ce5d90c709319R626-R646
+//! [a_profile_future]: https://static.docs.arm.com/ddi0601/a/SysReg_xml_futureA-2019-04.pdf?_ga=2.116560387.441514988.1590524918-1110153136.1588469296
+
+#[cfg(test)]
+use stdarch_test::assert_instr;
+
+extern "C" {
+    #[link_name = "llvm.aarch64.tstart"]
+    fn aarch64_tstart() -> i64;
```
Amanieu, line 22: According to the ACLE documentation these should all be `u64` instead of `i64`.

crates/core_arch/src/aarch64/tme.rs
```diff
@@ -0,0 +1,183 @@
...
+/// Starts a new transaction. When the transaction starts successfully the return value is 0.
+/// If the transaction fails, all state modifications are discarded and a cause of the failure
+/// is encoded in the return value.
+///
+/// [ARM TME Intrinsics](https://developer.arm.com/docs/101028/0010/transactional-memory-extension-tme-intrinsics).
+#[inline]
+#[target_feature(enable = "tme")]
+#[cfg_attr(test, assert_instr(tstart))]
+pub unsafe fn __tstart() -> i64 {
+    aarch64_tstart() as _
```
Amanieu, line 76: Unnecessary cast.

crates/core_arch/src/aarch64/tme.rs
```diff
@@ -0,0 +1,183 @@
...
+
+/// Tests if executing inside a transaction. If no transaction is currently executing,
+/// the return value is 0. Otherwise, this intrinsic returns the depth of the transaction.
+///
+/// [ARM TME Intrinsics](https://developer.arm.com/docs/101028/0010/transactional-memory-extension-tme-intrinsics).
+#[inline]
+#[target_feature(enable = "tme")]
+#[cfg_attr(test, assert_instr(ttest))]
+pub unsafe fn __ttest() -> i64 {
+    aarch64_ttest() as _
```
Amanieu, line 115: Unnecessary cast.

crates/std_detect/tests/cpu-detection.rs
Amanieu: The current feature detection is fine. Linux doesn't have the TME flag in /proc/cpuinfo or AT_HWCAPS yet, so there's nothing we can do about that for now.

Amanieu:
It seems that the tests are failing because the `objdump` version in CI doesn't support TME instructions.

Try updating the ubuntu version to 20.04 in `ci/docker/aarch64-unknown-linux-gnu/Dockerfile`.

crates/std_detect/tests/cpu-detection.rs
vertexclique: Yes, when it enters to kernel I will add here.

vertexclique: Addressed your comments. Thanks for the help! @Amanieu

Amanieu:
Thanks! You'll need to send a PR to rust-lang/rust to update the stdarch submodule if you want to have this in the next nightly.

RalfJung:
I am not sure if we can have intrinsics like these in Rust. After all, the compiler may reorder unrelated memory accesses into the transaction block, which would then be un-done when the transaction is aborted, which can cause arbitrary program misbehavior. Or am I misunderstanding how these instructions work?

See https://github.com/rust-lang/stdarch/issues/1521 for discussion.

PR #857, opened 2020-05-28
Fix CI issues caused by updated nightly

Amanieu:
Rust bug: https://github.com/rust-lang/rust/issues/72545


Amanieu:
Seems like there's still quite a bit of fallout from rust-lang/rust#72545 in our dependencies. We'll just have to wait it out.

PR #858, opened 2020-05-28
Mark __mm_prefetch as safe

imoldfella:
Prefetch should be safe to use, even when the address given it is invalid.

### DIFF
--- a/crates/core_arch/src/x86/sse.rs
+++ b/crates/core_arch/src/x86/sse.rs
@@ -1840,7 +1840,7 @@ pub const _MM_HINT_NTA: i32 = 0;
 #[cfg_attr(test, assert_instr(prefetchnta, strategy = _MM_HINT_NTA))]
 #[rustc_args_required_const(1)]
 #[stable(feature = "simd_x86", since = "1.27.0")]
-pub unsafe fn _mm_prefetch(p: *const i8, strategy: i32) {
+pub fn _mm_prefetch(p: *const i8, strategy: i32) {
     // The `strategy` must be a compile-time constant, so we use a short form
     // of `constify_imm8!` for now.
     // We use the `llvm.prefetch` instrinsic with `rw` = 0 (read), and


BurntSushi:
> Prefetch should be safe to use, even when the address given it is invalid.

Could you please elaborate? Is there some kind of source you can link to? It is not at all obvious to me that this is true.

Amanieu:
The prefetch instruction has no effect on memory (it doesn't write) and does nothing if it would have caused a page fault.

However the feature is only available with SSE, it is not a valid instruction on older architectures.

In fact, the compiler seems to reject the code for this very reason:
```
error[E0658]: `#[target_feature(..)]` can only be applied to `unsafe` functions
    --> crates/core_arch/src/x86/sse.rs:1836:1
     |
1836 |   #[target_feature(enable = "sse")]
     |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
1843 | / pub fn _mm_prefetch(p: *const i8, strategy: i32) {
1844 | |     // The `strategy` must be a compile-time constant, so we use a short form
1845 | |     // of `constify_imm8!` for now.
1846 | |     // We use the `llvm.prefetch` instrinsic with `rw` = 0 (read), and
...    |
1858 | |     pref!(strategy)
1859 | | }
     | |_- not an `unsafe` function
     |
     = note: see issue #69098 <https://github.com/rust-lang/rust/issues/69098> for more information
     = help: add `#![feature(target_feature_11)]` to the crate attributes to enable
```

imoldfella:
According to Intel its true, but it fails because target_feature requires unsafe. According to Lokauthor this will change with target_feature 1.1, so I'm jumping the gun on this and closing the request. Thanks for considering!

BurntSushi:
The instruction itself might accept any value, but the function itself is defined in Rust code. I don't know if there is any UB to consider there. Maybe not since it's a raw pointer, but I wouldn't be surprised if there was.

PR #859, opened 2020-05-28
Convert __mmask16 to use an unsigned type

Daniel-B-Smith:
To be consistent with Clang:

https://github.com/llvm/llvm-project/blob/1b02db52b79e01f038775f59193a49850a34184d/clang/lib/Headers/avx512fintrin.h#L37

and to simplify literals where you want to set the top bit to 1.


Amanieu:
Thanks!

PR #860, opened 2020-05-29
Disable bootstrap for stage0

vertexclique:
Amanieu mentioned that stage0 build should exclude TME for Aarch64 in here: https://github.com/rust-lang/rust/pull/72749#issuecomment-636148778

r? @Amanieu

### DIFF
--- a/crates/core_arch/src/aarch64/mod.rs
+++ b/crates/core_arch/src/aarch64/mod.rs
@@ -15,7 +15,9 @@ pub use self::neon::*;
 mod crypto;
 pub use self::crypto::*;
 
+#[cfg(not(bootstrap))]
 mod tme;
+#[cfg(not(bootstrap))]
 pub use self::tme::*;
 
 mod crc;


Amanieu:
You also need to `#[cfg]` out the `use` line below.

vertexclique:
Yup, forgot that.

Mark-Simulacrum:
Before we land this -- I want to be sure that I'm following correctly -- this will *not* break when we update the bootstrap compiler, right? It'll just retain less functionality? Historically having cfg(bootstrap) outside of rust-lang/rust proper is pretty painful so it would be good to avoid this if possible (say, by waiting ~1 week after bootstrap is bumped in rust-lang/rust).

Amanieu:
Actually I think the issue is not due to stage0. Looking at the [CI logs](https://github.com/rust-lang/rust/pull/72749/checks?check_run_id=721377106) it seems that rustdoc is failing because it doesn't recognize the "tme" target feature. Why is stage0 rustdoc used here?

Amanieu:
Sorry it seems I was wrong about the issue. I'll close this and we can continue discussion in https://github.com/rust-lang/rust/pull/72749.

Amanieu:
Reopening as per https://github.com/rust-lang/rust/pull/72749#issuecomment-636173102

Amanieu:
Please update your submodule in the other PR to include this.

PR #864, opened 2020-05-31
Add remaining integer AVX 512f comparison intrinsics

Daniel-B-Smith:
This is just a draft. I created the PR in order to get CI to run.


crates/core_arch/src/x86/avx512f.rs
```diff
@@ -199,6 +199,68 @@ pub unsafe fn _mm512_mask_cmpeq_epu64_mask(m: __mmask8, a: __m512i, b: __m512i)
...
+
+/// Compare packed unsigned 64-bit integers in a and b based on the comparison operand specified by op,
+///  using zeromask m (elements are zeroed out when the corresponding mask bit is not set).
+///
+/// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#expand=727,1063,4909,1062,1062,1063&text=_mm512_mask_cmp_epu64_mask)
+#[inline]
+#[target_feature(enable = "avx512f")]
+#[rustc_args_required_const(3)]
+#[cfg_attr(test, assert_instr(vpcmp, op = 0))]
+pub unsafe fn _mm512_mask_cmp_epu64_mask(
```
Daniel-B-Smith, line 249: Arguably, all of the other intrinsics should be implemented on top of this intrinsic. My preference would be to declare victory and move on, but I can make the change if you would like.

crates/core_arch/src/x86/avx512f.rs
```diff
@@ -199,6 +199,68 @@ pub unsafe fn _mm512_mask_cmpeq_epu64_mask(m: __mmask8, a: __m512i, b: __m512i)
...
+}
+
+/// Compare packed unsigned 64-bit integers in a and b based on the comparison operand specified by op.
+///
+/// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#expand=727,1063,4909,1062,1062,1063&text=_mm512_mask_cmp_epu64_mask)
+#[inline]
+#[target_feature(enable = "avx512f")]
+#[rustc_args_required_const(2)]
+#[cfg_attr(test, assert_instr(vpcmp, op = 0))]
+pub unsafe fn _mm512_cmp_epu64_mask(a: __m512i, b: __m512i, op: _MM_CMPINT_ENUM) -> __mmask8 {
```
Daniel-B-Smith, line 230: It's kind of annoying to have the full implementation duplicated here. When I attempted to call `_mm512_mask_cmp_epu64_mask` here, the constness of `op` was not propagated along, so I got compiler errors that the argument needed to be a constant.

Amanieu:
Are you planning on adding the epi32/epu32 intrinsics as well. Those are also part of AVX512F.

Daniel-B-Smith:
I'm a little confused by the documentation for the 32 bit comparisons. Most of the 32 bit comparisons read `CPUID Flags: AVX512F for AVX-512, KNCNI for KNC`, e.g.:

https://software.intel.com/sites/landingpage/IntrinsicsGuide/#expand=727,1063,4909,1062,1062,1063,874,850,779&avx512techs=AVX512F&text=_mm512_cmpeq

The `cmplt_epi32` variants have separate AVX 512f and KNCNI documentation entries (which is why those ended up on Alex Crichton's list):

https://software.intel.com/sites/landingpage/IntrinsicsGuide/#expand=727,1063,4909,1062,1062,1063,874,850,779,1029,1029,1030&text=_mm512_cmplt_epi32_mask

I have been reading that documentation as being an implicit `AND` between AVX512f and Knight's Corner, but it may be an implicit `OR` ([clang](https://github.com/llvm/llvm-project/blob/master/clang/lib/Headers/avx512fintrin.h) implements all of the combined comparisons in the AVX512f header). If it's the latter, we will also need to update the verification tests to accept `avx512f` for `avx512f/kncni` (see CI results).

Amanieu:
I believe the condition is AVX512F OR KNC. I would ignore KNC for now and just gate it on AVX512F only.

Daniel-B-Smith:
I've added the 32 bit integer comparisons and updated `stdarch-verify` to accept "avx512f/kncni" intrinsics to be gated by solely "avx512f". Now that I know it is an OR, I will update my list so that I get all of the mixed intrinsics going forward.

Daniel-B-Smith:
Closing since https://github.com/rust-lang/stdarch/pull/866 also contains these changes.

PR #865, opened 2020-06-04
Clarify documentation about wasm32 target_feature gates

Ngalstyan4:
Most wasm32 SIMD intrinsics are available behind a feature gate only after recompiling libcore with relevant `target_feature`. This is documented for [atomic_notify](https://doc.rust-lang.org/core/arch/wasm32/fn.atomic_notify.html), for example but is not documented for any of the `simd128`-gated intrinsics. Unlike the atomics, it would probably not make sense to add this comment above each function so here is my attempt at putting it somewhere visible globally to make sure it is clear that recompilation of libcore is needed.

### DIFF
--- a/crates/core_arch/src/mod.rs
+++ b/crates/core_arch/src/mod.rs
@@ -57,6 +57,13 @@ pub mod arch {
 
     /// Platform-specific intrinsics for the `wasm32` platform.
     ///
+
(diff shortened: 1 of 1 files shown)


lu-zero:
Please fix the style check :)

PR #868, opened 2020-06-09
Fix x86 extract_epi{8,16} functions

### LINKED ISSUE #867: x86 Various issues with extract_epiNN functions
## 1. Inconsistent function signatures

The SSE versions have the following signatures:
```rust
pub unsafe fn _mm_extract_epi8(a: __m128i, imm8: i32) -> i32;
pub unsafe fn _mm_extract_epi16(a: __m128i, imm8: i32) -> i32;
```
But the AVX versions have the following:
```rust
pub unsafe fn _mm256_extract_epi8(a: __m256i, imm8: i32) -> i8;
pub unsafe fn _mm256_extract_epi16(a: __m256i, imm8: i32) -> i16;
```
Note the difference in return type. Since [Intel specifies](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=extract_epi) these all return `int`, it's not clear how this happened. I thought the function signatures are supposed to be machine-checked?

## 2. Incorrect sign extension

```rust
use core::arch::x86_64::*;

fn main() {
    println!("{:x}", unsafe{_mm_extract_epi16(_mm_set1_epi8(!0), 0)});
}
```

This prints: `ffffffff`. [Intel specifies](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_extract_epi16&expand=2426,2426) the upper 16 bits should be 0.

cc @AdrianCX

jethrogb:
* Update Intel intrinsics definitions with the latest version
  * Surprisingly, the only thing that changed in terms of function signatures is exactly the inconsistency I pointed out in #867.
* Update _mm256_extract_epi{8,16} to match latest definition
* Fix _mm_extract_epi16 sign extension

Fixes #867

### DIFF
--- a/crates/core_arch/src/x86/avx2.rs
+++ b/crates/core_arch/src/x86/avx2.rs
@@ -3743,9 +3743,9 @@ pub unsafe fn _mm256_xor_si256(a: __m256i, b: __m256i) -> __m256i {
 // This intrinsic has no corresponding instruction.
 #[rustc_args_required_const(1)]
 #[stable(feature = "simd_x86", since = "1.27.0")]
-pub unsafe fn _mm256_extract_epi8(a: __m256i, imm8: i32) -> i8 {
+pub unsafe fn _mm256_extract_epi8(a: __m256i, imm8: i32) -> i32 {
     let imm8 = (imm8 & 31) as u32;
-    simd_extract(a.as_i8x32(), imm8)
+    simd_extract::<_, u8>(a.as_u8x32(), imm8) as i32
 }
 
 /// Extracts a 16-bit integer from `a`, selected with `imm8`. Returns a 32-bit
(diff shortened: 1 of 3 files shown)


crates/core_arch/src/x86/avx2.rs
```diff
@@ -3743,9 +3743,9 @@ pub unsafe fn _mm256_xor_si256(a: __m256i, b: __m256i) -> __m256i {
 // This intrinsic has no corresponding instruction.
 #[rustc_args_required_const(1)]
 #[stable(feature = "simd_x86", since = "1.27.0")]
-pub unsafe fn _mm256_extract_epi8(a: __m256i, imm8: i32) -> i8 {
+pub unsafe fn _mm256_extract_epi8(a: __m256i, imm8: i32) -> i32 {
```
jethrogb, line 3746: This and the change below is technically a breaking change in the public API. However we are fixing the function here to match the definition. Thoughts?

Amanieu: I think that we need a crater run at the very least for this. cc @rust-lang/libs

Amanieu:
I'm going to merge this for now, let's do a crater run on the PR to update the submodule in rust-lang/rust.

crates/core_arch/src/x86/avx2.rs
BurntSushi: I think this is probably going to break anyone that uses `_mm256_extract_epi8`, no?

It seems like this is probably an okay "wontfix" bug, since I think `i8` is big enough to hold all possible bit patterns returned here.

Amanieu: I expect most users of these intrinsics to cast the returned value immediately, which would hide the issue.

alexcrichton:
FWIW this blocks further updates to the submodule in https://github.com/rust-lang/rust/pull/74482, so for future crater runs could the rust-lang/rust PR point to a fork with the change? That way the master branch of this repository could be updated by rust-lang/rust PRs as necessary for other features.

PR #869, opened 2020-06-14
Implement AVX512f floating point comparisons

Daniel-B-Smith:
This adds the AVX 512f floating point comparison intrinsics. This should finish all of the AVX 512f comparison intrinsics.


Daniel-B-Smith:
This should be ready for review. There's one remaining rough edge here. If you pass an invalid value to `sae`, the intrinsic won't link and llvm gives a very unhelpful error message. In particular, passing `0` in the `assert_instr` failed to link.

Amanieu:
It seems that not all values are valid for the `sae` argument. You should replicate the exact rules (from [here](https://github.com/llvm/llvm-project/blob/bd50cf905fa7c0c7caa134301c6ca0658c81eeb1/clang/lib/Sema/SemaChecking.cpp#L3497)) in a custom `constify` macro.

Daniel-B-Smith:
Done. Thanks! I tried to find where/how the sae parameter was validated, but I struggled to find it.

crates/core_arch/src/x86/avx512f.rs
```diff
@@ -965,6 +965,646 @@ pub unsafe fn _mm512_set1_epi64(a: i64) -> __m512i {
...
+#[rustc_args_required_const(2)]
+#[cfg_attr(test, assert_instr(vcmp, op = 0))]
+pub unsafe fn _mm512_cmp_ps_mask(a: __m512, b: __m512, op: i32) -> __mmask16 {
+    let neg_one = -1;
+    macro_rules! call {
+        ($imm5:expr, $imm4:expr) => {
+            vcmpps(a.as_f32x16(), b.as_f32x16(), $imm5, neg_one, $imm4)
+        };
+    }
+    let r = constify_imm5_sae!(op, _MM_FROUND_CUR_DIRECTION, call);
```
Amanieu, line 1108: Why not just use `constify_imm5` here and hard-code the rounding mode in the `vcmpps` call above? This code creates many unnecessary match branches.

Daniel-B-Smith: Just a bad refactoring. I had it like that originally. Should be fixed now.

Daniel-B-Smith:
CI is fixed now.

Amanieu:
Thanks!

PR #870, opened 2020-06-21
Add _mm_loadu_si64

### LINKED ISSUE #40: Implement all x86 vendor intrinsics

pickfire:
Fix #40

### DIFF
--- a/crates/core_arch/src/x86/sse.rs
+++ b/crates/core_arch/src/x86/sse.rs
@@ -1258,7 +1258,7 @@ pub unsafe fn _mm_loadr_ps(p: *const f32) -> __m128 {
 /// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_loadu_si64)
 #[inline]
 #[target_feature(enable = "sse")]
-#[cfg_attr(test, assert_instr(movq))]
+#[cfg_attr(all(test, not(target = "x86")), assert_instr(movq))]
 #[stable(feature = "simd_x86_mm_loadu_si64", since = "1.46.0")]
 pub unsafe fn _mm_loadu_si64(mem_addr: *const u8) -> __m128i {
     transmute(i64x2(0, ptr::read_unaligned(mem_addr as *const i64)))
--- a/crates/core_arch/src/x86/sse.rs
+++ b/crates/core_arch/src/x86/sse.rs
@@ -3673,9 +3673,9 @@ mod tests {
 
     #[simd_test(enable = "sse2")]
     unsafe fn test_mm_loadu_si64() {
-        let a = _mm_set_epi64x(5, 0);
+        let a = _mm_setr_epi64x(5, 6);
         let r = _mm_loadu_si64(&a as *const _ as *const _);
-        assert_eq_m128i(a, r);
+        assert_eq_m128i(r, _mm_set_epi64x(5, 0));
     }
 
     #[simd_test(enable = "sse")]
--- a/crates/core_arch/src/x86/sse.rs
+++ b/crates/core_arch/src/x86/sse.rs
@@ -1258,7 +1258,7 @@ pub unsafe fn _mm_loadr_ps(p: *const f32) -> __m128 {
 /// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_loadu_si64)
 #[inline]
 #[target_feature(enable = "sse")]
-#[cfg_attr(test, assert_instr(movups))]
+#[cfg_attr(test, assert_instr(movq))]
 #[stable(feature = "simd_x86_mm_loadu_si64", since = "1.46.0")]
 pub unsafe fn _mm_loadu_si64(mem_addr: *const u8) -> __m128i {
     _mm_set_epi64x(ptr::read_unaligned(mem_addr as *const i64), 0)
--- a/crates/core_arch/src/x86/sse.rs
+++ b/crates/core_arch/src/x86/sse.rs
@@ -1258,7 +1258,7 @@ pub unsafe fn _mm_loadr_ps(p: *const f32) -> __m128 {
 /// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_loadu_si64)
 #[inline]
 #[target_feature(enable = "sse")]
-#[cfg_attr(all(test, not(target = "x86")), assert_instr(movq))]
+#[cfg_attr(all(test, not(target_arch = "x86")), assert_instr(movq))]
 #[stable(feature = "simd_x86_mm_loadu_si64", since = "1.46.0")]
 pub unsafe fn _mm_loadu_si64(mem_addr: *const u8) -> __m128i {
     transmute(i64x2(0, ptr::read_unaligned(mem_addr as *const i64)))
--- a/crates/core_arch/src/x86/sse.rs
+++ b/crates/core_arch/src/x86/sse.rs
@@ -1251,6 +1251,25 @@ pub unsafe fn _mm_loadr_ps(p: *const f32) -> __m128 {
     simd_shuffle4(a, a, [3, 2, 1, 0])
 }
 
+/// Loads unaligned 64-bits of integer data from memory into new vector.
+///
+/// `mem_addr` does not need to be aligned on any particular boundary.
+///
+/// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_loadu_si64)
+#[inline]
+#[target_feature(enable = "sse")]
+#[cfg_attr(test, assert_instr(movups))]
+#[stable(feature = "simd_x86", since = "1.46.0")]
+pub unsafe fn _mm_loadu_si64(mem_addr: *const u8) -> __m128i {
+    let mut dst = _mm_setzero_si128();
+    ptr::copy_nonoverlapping(
+        mem_addr,
+        &mut dst as *mut __m128i as *mut u8,
+        8, // == 64 bits == mem::size_of::<__m128i>() / 2
+    );
+    dst
+}
+
 /// Stores the upper half of `a` (64 bits) into memory.
 ///
 /// This intrinsic corresponds to the `MOVHPS` instruction. The compiler may
@@ -3656,6 +3675,13 @@ mod tests {
         let r = _mm_loadr_ps(p);
         let e = _mm_add_ps(_mm_setr_ps(4.0, 3.0, 2.0, 1.0), _mm_set1_ps(fixup));
         assert_eq_m128(r, e);
+    }
+
+    #[simd_test(enable = "sse2")]
+    unsafe fn test_mm_loadu_si64() {
+        let a = _mm_set_epi64x(5, 0);
+        let r = _mm_loadu_si64(&a as *const _ as *const _);
+        assert_eq_m128i(a, r);
     }
 
     #[simd_test(enable = "sse")]
--- a/crates/core_arch/src/x86/sse.rs
+++ b/crates/core_arch/src/x86/sse.rs
@@ -1259,15 +1259,9 @@ pub unsafe fn _mm_loadr_ps(p: *const f32) -> __m128 {
 #[inline]
 #[target_feature(enable = "sse")]
 #[cfg_attr(test, assert_instr(movups))]
-#[stable(feature = "simd_x86", since = "1.46.0")]
+#[stable(feature = "simd_x86_mm_loadu_si64", since = "1.46.0")]
 pub unsafe fn _mm_loadu_si64(mem_addr: *const u8) -> __m128i {
-    let mut dst = _mm_setzero_si128();
-    ptr::copy_nonoverlapping(
-        mem_addr,
-        &mut dst as *mut __m128i as *mut u8,
-        8, // == 64 bits == mem::size_of::<__m128i>() / 2
-    );
-    dst
+    _mm_set_epi64x(ptr::read_unaligned(mem_addr as *const i64), 0)
 }
 
 /// Stores the upper half of `a` (64 bits) into memory.
--- a/crates/core_arch/src/x86/sse.rs
+++ b/crates/core_arch/src/x86/sse.rs
@@ -1261,7 +1261,7 @@ pub unsafe fn _mm_loadr_ps(p: *const f32) -> __m128 {
 #[cfg_attr(test, assert_instr(movq))]
 #[stable(feature = "simd_x86_mm_loadu_si64", since = "1.46.0")]
 pub unsafe fn _mm_loadu_si64(mem_addr: *const u8) -> __m128i {
-    _mm_set_epi64x(ptr::read_unaligned(mem_addr as *const i64), 0)
+    transmute(i64x2(0, ptr::read_unaligned(mem_addr as *const i64)))
 }
 
 /// Stores the upper half of `a` (64 bits) into memory.
--- a/crates/stdarch-test/src/lib.rs
+++ b/crates/stdarch-test/src/lib.rs
@@ -3,7 +3,6 @@
 //! This basically just disassembles the current executable and then parses the
 //! output once globally and then provides the `assert` function which makes
 //! assertions about the disassembly of a function.
-#![feature(const_transmute)]
 #![feature(vec_leak)]
 #![allow(clippy::missing_docs_in_private_items, clippy::print_stdout)]
 


crates/core_arch/src/x86/sse.rs
```diff
@@ -1251,6 +1251,25 @@ pub unsafe fn _mm_loadr_ps(p: *const f32) -> __m128 {
...
+#[cfg_attr(test, assert_instr(movups))]
+#[stable(feature = "simd_x86", since = "1.46.0")]
+pub unsafe fn _mm_loadu_si64(mem_addr: *const u8) -> __m128i {
+    let mut dst = _mm_setzero_si128();
+    ptr::copy_nonoverlapping(
+        mem_addr,
+        &mut dst as *mut __m128i as *mut u8,
+        8, // == 64 bits == mem::size_of::<__m128i>() / 2
+    );
+    dst
```
Amanieu, line 1270: The implementation could be simplified:
```rust
let val = ptr::read_unaligned(mem_addr as *const i64);
_mm_set_epi64x(val, 0)
```

pickfire: Thanks for the suggestion, but do we need `as *const i64`?

bjorn3: Yes, otherwise only an `u8` would be read and not a full `i64`.

Amanieu: In this particular case I think we can just change the argument type to `*const i64`. The original intrinsic uses `const void*` but we don't have that in Rust.

pickfire: But why is it `*const i64` instead of `*const u64`?

Amanieu: All the other intrinsics seem to prefer using signed types over unsigned ones (e.g. `_mm_set_epi64x`). It makes no difference in practice, the bits loaded from memory are the same.

Amanieu: Hmm apparently the intrinsic verifier doesn't like it so just change it back to `*mut u8` and use a cast like I suggested.

Amanieu:
Can you rebase this? The CI failures should be fixed now.

pickfire:
> thread 'core_arch::x86::sse::assert__mm_loadu_si64_movups' panicked at 'failed to find instruction `movups` in the disassembly', crates/stdarch-test/src/lib.rs:157:9

What should I do? Should I change ` #[cfg_attr(test, assert_instr(movups))]`?

crates/core_arch/src/x86/sse.rs
```diff
@@ -3658,6 +3672,13 @@ mod tests {
         assert_eq_m128(r, e);
     }
 
+    #[simd_test(enable = "sse2")]
+    unsafe fn test_mm_loadu_si64() {
+        let a = _mm_set_epi64x(5, 0);
```
bjorn3, line 3677: In memory this is represented as `[0, 0, 0, 0, 0, 0, 0, 0, 5, 0, 0, 0, 0, 0, 0, 0]`, not `[5, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]` as you expected.

Amanieu:
`movups` is a 128-bit move. You want `movq` which is a 64-bit move.

crates/core_arch/src/x86/sse.rs
pickfire: Oh, now it still failed with

```rust
        let a = _mm_set_epi8(0, 0, 0, 0, 0, 0, 0, 0, 5, 0, 0, 0, 0, 0, 0, 0);
        let r = _mm_loadu_si64(&a as *const _ as *const _);
        assert_eq_m128i(a, r);
```

Or did I declare `[0, 0, 0, 0, 0, 0, 0, 0, 5, 0, 0, 0, 0, 0, 0, 0]` wrongly, or is it endianness issue?

```
thread 'core_arch::x86::sse::tests::test_mm_loadu_si64' panicked at 'assertion failed: `(left == right)`
  left: `[360287970189639680, 0]`,
 right: `[0, 360287970189639680]`', crates/core_arch/src/x86/test.rs:20:5
```

Amanieu: I would recommend testing the equivalent C code.

pickfire:
Oh, it didn't use `movq` in the resulting code. It should be though.

crates/core_arch/src/x86/sse.rs
```diff
@@ -1251,6 +1251,19 @@ pub unsafe fn _mm_loadr_ps(p: *const f32) -> __m128 {
     simd_shuffle4(a, a, [3, 2, 1, 0])
 }
 
+/// Loads unaligned 64-bits of integer data from memory into new vector.
+///
+/// `mem_addr` does not need to be aligned on any particular boundary.
+///
+/// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_loadu_si64)
+#[inline]
+#[target_feature(enable = "sse")]
```
Amanieu, line 1260: This should be `mmx` instead of `sse`.

pickfire: Shouldn't this be `sse`? Intel Intrinsics Guide https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_loadu_si64 shows that it is green (SSE) and not yellow (MMX).

Amanieu: Oh wait nevermind, that's not the issue.

crates/core_arch/src/x86/sse.rs
```diff
@@ -1251,6 +1251,19 @@ pub unsafe fn _mm_loadr_ps(p: *const f32) -> __m128 {
...
+#[cfg_attr(test, assert_instr(movq))]
+#[stable(feature = "simd_x86_mm_loadu_si64", since = "1.46.0")]
+pub unsafe fn _mm_loadu_si64(mem_addr: *const u8) -> __m128i {
+    _mm_set_epi64x(ptr::read_unaligned(mem_addr as *const i64), 0)
```
Amanieu, line 1264: `_mm_set_epi64x` requires the `"sse2"` target feature, but this intrinsic only has `"sse"`. You need to use a different function to build the vector.

Amanieu: The feature mismatch causes the test failure since the compiler can't inline `_mm_set_epi64x`.

pickfire: Ah, I didn't thought about this. I thought only those functions that are available are imported instead of having all functions imported, looks like a potential foot gun for new contributors. I wonder if there is sort of an import for directed acyclic graph with each file contain all the dependencies available. Like `sse2` depends on `sse` so `sse` should not import `sse2`.

But that looks like another issue, I will find out which function this could use.

pickfire: The build still failed, it doesn't seemed like that's the case. My guess is `movq` not being there.

Amanieu: What seems to be happening now is that the 32-bit x86 calling convention is returning `__m128i` by reference, so LLVM is just entirely skipping the use of vector registers here. I would just disable the `assert_instr` on x86 with `#[cfg_attr(all(test, not(target = "x86")), assert_instr(movq))]`.

pickfire: Is that correct? I thought it needs to have `movq`? How do we add it back?

Amanieu: It's fine. The intrinsics don't *have* to generate a particular instruction, it is acceptable for the compiler to generate code using different instructions that has the same effect.

pickfire: But isn't generating the original instruction that it should be faster rather than having multiple instructions to do the same thing?

pickfire:
`error[E0635]: unknown feature `const_transmute`? Weird, is that removed from nightly?

Amanieu:
`const_transmute` is now stable, you can remove the feature flag.

crates/core_arch/src/x86/sse.rs
```diff
@@ -1251,6 +1251,19 @@ pub unsafe fn _mm_loadr_ps(p: *const f32) -> __m128 {
...
 }
 
+/// Loads unaligned 64-bits of integer data from memory into new vector.
+///
+/// `mem_addr` does not need to be aligned on any particular boundary.
+///
+/// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_loadu_si64)
+#[inline]
+#[target_feature(enable = "sse")]
+#[cfg_attr(all(test, not(target = "x86")), assert_instr(movq))]
```
Amanieu, line 1261:
```suggestion
#[cfg_attr(all(test, not(target_arch = "x86")), assert_instr(movq))]
```

Amanieu:
Thanks!

pickfire:
@Amanieu Can you please mark this function done in https://github.com/rust-lang/stdarch/issues/40 first post checklist?

Amanieu:
Done! However I am having second thoughts about exposing this function in SSE instead of SSE2. The Intel documentation seems to be incorrect since it says that this should use the MOVQ instruction but that instruction is only available in SSE2. Also both Clang and GCC only expose this intrinsic in the SSE2 header (`emmintrin.h`) rather than the SSE one (`xmmintrin.h`).

pickfire:
But even though Intel documentation is incorrect we need to follow them right?

Amanieu:
Not really. IMO we should be following GCC and Clang which are the "standard" implementation for the compiler intrinsics. I believe even the Intel compiler puts `_mm_loadu_si64` in `emmintrin.h` which implies it is a SSE2 intrinsic.

PR #871, opened 2020-07-03
Make function signatures consistent

Daniel-B-Smith:
On newer nightlies, the fact that these function signatures conflicted caused warnings that failed CI. I chose to convert unsigned to signed to be consistent with the LLVM signature.

Example failure: https://github.com/rust-lang/stdarch/runs/835662285


Amanieu:
Could you look into fixing the MIPS failures as well?

Daniel-B-Smith:
Fixed the MIPS failures.

Amanieu:
Thanks!

PR #873, opened 2020-07-09
added f32 and f64 unaligned stores and loads from avx512f set

khodzha:


crates/core_arch/src/x86/avx512f.rs
```diff
@@ -1633,6 +1634,97 @@ pub unsafe fn _mm512_mask_cmp_epi64_mask(
...
+
+/// Returns vector of type `__m512` with undefined elements.
+///
+/// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm512_undefined_ps)
+#[inline]
+#[target_feature(enable = "avx512f")]
+// This intrinsic has no corresponding instruction.
+pub unsafe fn _mm512_undefined_ps() -> __m512 {
+    // FIXME: this function should return MaybeUninit<__m512>
+    mem::MaybeUninit::<__m512>::uninit().assume_init()
```
bjorn3, line 1656: This is UB.

crates/core_arch/src/x86/avx512f.rs
```diff
@@ -1633,6 +1634,97 @@ pub unsafe fn _mm512_mask_cmp_epi64_mask(
...
+#[target_feature(enable = "avx512f")]
+#[cfg_attr(test, assert_instr(vmovupd))]
+pub unsafe fn _mm512_loadu_pd(mem_addr: *const f64) -> __m512d {
+    let mut dst = _mm512_undefined_pd();
+    ptr::copy_nonoverlapping(
+        mem_addr as *const u8,
+        &mut dst as *mut __m512d as *mut u8,
+        mem::size_of::<__m512d>(),
+    );
+    dst
```
bjorn3, line 1674: Does `mem::transmute` work?

crates/core_arch/src/x86/avx512f.rs
Amanieu: The `undefined` intrinsics should return zero-initialized vectors like clang does.

crates/core_arch/src/x86/avx512f.rs
Amanieu, line 1674: Simpler version: `ptr::read_unaligned(mem_addr as *const __m512d)`

crates/core_arch/src/x86/avx512f.rs
khodzha: _mm_undefined_ps and _mm_256_undefined_p{d,s} do same thing, should i change them?

crates/core_arch/src/x86/avx512f.rs
khodzha: i used Amanieu version

khodzha:
thanks, updated with feedback addressed

crates/core_arch/src/x86/avx512f.rs
Amanieu: Yes those should be fixed as well.

crates/core_arch/src/x86/avx512f.rs
```diff
@@ -1633,6 +1634,83 @@ pub unsafe fn _mm512_mask_cmp_epi64_mask(
...
 }
 
+/// Returns vector of type `__m512d` with undefined elements.
+///
+/// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm512_undefined_pd)
+#[inline]
+#[target_feature(enable = "avx512f")]
+// This intrinsic has no corresponding instruction.
+pub unsafe fn _mm512_undefined_pd() -> __m512d {
+    _mm512_set1_pd(0.0)
```
bjorn3, line 1644: This is a zero float. I think it should be literal zero bytes. Not sure though.

Amanieu: A zero float happens to be encoded with all bits zeroed.

crates/core_arch/src/x86/avx512f.rs
```diff
@@ -1633,6 +1634,83 @@ pub unsafe fn _mm512_mask_cmp_epi64_mask(
...
+/// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm512_storeu_pd)
+#[inline]
+#[target_feature(enable = "avx512f")]
+#[cfg_attr(test, assert_instr(vmovupd))]
+pub unsafe fn _mm512_storeu_pd(mem_addr: *mut f64, a: __m512d) {
+    ptr::copy_nonoverlapping(
+        &a as *const __m512d as *const u8,
+        mem_addr as *mut u8,
+        mem::size_of::<__m512d>(),
+    );
```
Amanieu, line 1682: You can use `ptr::write_unaligned` here just like the loads.

khodzha: thanks

Amanieu:
For the CI failures just change the `assert_instr` to use `vmovups`. The two instructions are equivalent anyways.

khodzha:
interesting, there are these comments:
https://github.com/rust-lang/stdarch/blob/a3710694f94f13758799996e981dd2cc1069a514/crates/core_arch/src/x86/avx.rs#L1678
https://github.com/rust-lang/stdarch/blob/a3710694f94f13758799996e981dd2cc1069a514/crates/core_arch/src/x86/avx.rs#L1697

Amanieu:
Thanks!

PR #876, opened 2020-07-18
Constify all x86 rustc_args_required_const intrinsics

bjorn3:
Fixes https://github.com/bjorn3/rustc_codegen_cranelift/issues/669

I am not quite sure about how some intrinsics should handle `imm8` values > 255. Where there was no masking before, I have used saturating through `constify_imm8`.

### DIFF
--- a/crates/core_arch/src/x86/avx.rs
+++ b/crates/core_arch/src/x86/avx.rs
@@ -1579,7 +1579,13 @@ pub unsafe fn _mm256_insertf128_si256(a: __m256i, b: __m128i, imm8: i32) -> __m2
 #[rustc_args_required_const(2)]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn _mm256_insert_epi8(a: __m256i, i: i8, index: i32) -> __m256i {
-    transmute(simd_insert(a.as_i8x32(), (index as u32) & 31, i))
+    let a = a.as_i8x32();
+    macro_rules! call {
+        ($index:expr) => {
+            simd_insert(a, $index, i)
+        };
+    }
+    transmute(constify_imm5!(index, call))
 }
 
 /// Copies `a` to result, and inserts the 16-bit integer `i` into result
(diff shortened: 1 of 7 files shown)


bjorn3:
cc @eddyb (https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/stable.20subset.20of.20const.20generics/near/203882415)

Amanieu:
IMO we should panic on out-of-range immediates, with the intention of turning those into compile-time errors once we have enough support for const generics (and `#[rustc_args_required_const]` becomes syntax sugar for const generics).

bjorn3:
All other `#[rustc_args_required_const]` intrinsics either mask or saturate the immediate.

Amanieu:
I reviewed the existing code and it seems that they all use masking via the `constify_*` macros. Let's just stick to that for now.

crates/core_arch/src/x86/avx.rs
```diff
@@ -1579,7 +1579,13 @@ pub unsafe fn _mm256_insertf128_si256(a: __m256i, b: __m128i, imm8: i32) -> __m2
...
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn _mm256_insert_epi8(a: __m256i, i: i8, index: i32) -> __m256i {
-    transmute(simd_insert(a.as_i8x32(), (index as u32) & 31, i))
+    let a = a.as_i8x32();
+    macro_rules! call {
+        ($index:expr) => {
+            simd_insert(a, $index, i)
+        };
+    }
+    transmute(constify_imm5!((index & 31), call))
```
Amanieu, line 1588: There's no need to mask here, it's done for you inside the `constify_imm5` macro. Same with all the other cases.

bjorn3: Missed that the `constify_imm*` macros masked the immediate.

crates/core_arch/src/x86/sse2.rs
```diff
@@ -1752,7 +1806,7 @@ pub unsafe fn _mm_max_sd(a: __m128d, b: __m128d) -> __m128d {
 #[cfg_attr(test, assert_instr(maxpd))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn _mm_max_pd(a: __m128d, b: __m128d) -> __m128d {
-    maxpd(a, b)
+    simd_fmax(a, b)
```
Amanieu, line 1809: Are you sure that this change is correct with regards to edge cases like NaN and negative zero?

bjorn3: No, I am not sure. I have dropped the commit.

PR #877, opened 2020-07-19
[Neon] Absolute Value fns

Lokathor:
This should be the correct version of all the absolute value operations on the integer types.


Lokathor:
Okay, done I think.

PR #879, opened 2020-07-31
Explain the discrepancy in the mask type for _mm_shuffle_ps

georgio:
Hello everyone,
This pull request attempts to solve https://github.com/rust-lang/rust/issues/62490.
I have based my notes on the discussion from #522

Hope this is helpful :-)

### DIFF
--- a/crates/core_arch/src/x86/sse.rs
+++ b/crates/core_arch/src/x86/sse.rs
@@ -997,7 +997,7 @@ pub const fn _MM_SHUFFLE(z: u32, y: u32, x: u32, w: u32) -> i32 {
 /// `_mm_shuffle_ps` is supposed to take an `i32` instead of an `u32`
 /// as is the case for [other shuffle intrinsics](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_shuffle_).
 /// Performing an implicit type conversion between an unsigned integer and a signed integer
-/// does not cause a problem in C, however Rust's commitment to safety does not allow this.
+/// does not cause a problem in C, however Rust's commitment to strong typing does not allow this.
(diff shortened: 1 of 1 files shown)


crates/core_arch/src/x86/sse.rs
```diff
@@ -992,6 +992,12 @@ pub const fn _MM_SHUFFLE(z: u32, y: u32, x: u32, w: u32) -> i32 {
 /// `b`. Mask is split to 2 control bits each to index the element from inputs.
 ///
 /// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_shuffle_ps)
+///
+/// Note that there appears to be a mistake within Intel's Intrinsics Guide.
+/// `_mm_shuffle_ps` is supposed to take an `i32` instead of an `u32`
+/// as is the case for [other shuffle intrinsics](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_shuffle_).
+/// Performing an implicit type conversion between an unsigned integer and a signed integer
+/// does not cause a problem in C, however Rust's commitment to safety does not allow this.
```
Lokathor, line 1000: suggestion:
> however Rust's commitment to strong typing does not allow this.

u32/i32 flipping isn't a safety related issue, at least not in this case.

Amanieu: That's correct, it's more an issue of Rust's strong typing vs C's weak typing which allows much more implicit conversions.

georgio: Thank you for your suggestion! It has been added.

Amanieu:
Thanks!

PR #880, opened 2020-08-03
Fix typo in doc for `_mm256_permute2f128_si256`

samrat:
It should be 128 bits, instead of 258 bits: https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm256_permute2f128_si256


Amanieu:
Thanks!

PR #881, opened 2020-08-08
add more things that do adds

Lokathor:
I'm doing this on an arm32 device so I've already run all those tests and squared that away, but for arm64 we'll have to rely on the CI.

* The last two "add"
* "add across vector" (aka horizontal add)
* "add pairwise"

### DIFF
--- a/crates/core_arch/src/aarch64/neon/mod.rs
+++ b/crates/core_arch/src/aarch64/neon/mod.rs
@@ -82,8 +82,44 @@ extern "C" {
     #[link_name = "llvm.aarch64.neon.pmull64"]
     fn vmull_p64_(a: i64, b: i64) -> int8x16_t;
 
+    #[link_name = "llvm.aarch64.neon.addp.v8i16")]
+    fn vpaddq_s16_(a: int16x8_t, b: int16x8_t) -> int16x8_t;
+    #[link_name = "llvm.aarch64.neon.addp.v4i32")]
+    fn vpaddq_s32_(a: int32x4_t, b: int32x4_t) -> int32x4_t;
     #[link_name = "llvm.aarch64.neon.addp.v16i8"]
-    fn vpaddq_u8_(a: uint8x16_t, b: uint8x16_t) -> uint8x16_t;
+    fn vpaddq_s8_(a: int8x16_t, b: int8x16_t) -> int8x16_t;
+    #[link_name = "llvm.aarch64.neon.uaddv.i64.v2i64"]
+    fn vpaddd_s64_(a: int64x2_t) -> i64;
+    
+    
+    #[link_name = "llvm.aarch64.neon.saddv.i32.v4i16"]
+    fn vaddv_s16_(a: int16x4_t) -> i16;
+    #[link_name = "llvm.aarch64.neon.saddv.i32.v2i32"]
+    fn vaddv_s32_(a: int32x2_t) -> i32;
+    #[link_name = "llvm.aarch64.neon.saddv.i32.v8i8"]
+    fn vaddv_s8_(a: int8x8_t) -> i8;
+    #[link_name = "llvm.aarch64.neon.uaddv.i32.v4i16"]
+    fn vaddv_u16_(a: int16x4_t) -> u16;
+    #[link_name = "llvm.aarch64.neon.uaddv.i32.v2i32"]
+    fn vaddv_u32_(a: int32x2_t) -> u32;
+    #[link_name = "llvm.aarch64.neon.uaddv.i32.v8i8"]
+    fn vaddv_u8_(a: int8x8_t) -> u8;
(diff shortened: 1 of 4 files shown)


Lokathor:
Alright, the current batch of errors should be fixed, except for the `vek_leak` feature no longer being needed.

I _assume_ that we just delete that line now that it's stable, but I'll wait to let someone else confirm that.

crates/core_arch/src/aarch64/neon/mod.rs
```diff
@@ -276,12 +312,160 @@ pub unsafe fn vabsq_s64(a: int64x2_t) -> int64x2_t {
     vabsq_s64_(a)
 }
 
+/// Add pairwise
+#[inline]
+#[target_feature(enable = "neon")]
+#[cfg_attr(test, assert_instr(addp))]
+pub unsafe fn vpaddq_s16(a: int16x8_t, b: int16x8_t) -> int16x8_t {
+    vpaddq_s16_(a, b)
```
bjorn3, line 320: This can use `simd_add`, right?

crates/core_arch/src/aarch64/neon/mod.rs
```diff
@@ -82,8 +82,44 @@ extern "C" {
...
+    #[link_name = "llvm.aarch64.neon.uaddv.i32.v8i16"]
+    fn vaddvq_u16_(a: uint16x8_t) -> u16;
+    #[link_name = "llvm.aarch64.neon.uaddv.i32.v3i32"]
+    fn vaddvq_u32_(a: uint32x4_t) -> u32;
+    #[link_name = "llvm.aarch64.neon.uaddv.i32.v16i8"]
+    fn vaddvq_u8_(a: uint8x16_t) -> u8;
+    #[link_name = "llvm.aarch64.neon.saddv.i32.v2i64"]
+    fn vaddvq_s64_(a: int64x2_t) -> i64;
+    #[link_name = "llvm.aarch64.neon.uaddv.i32.v2i64"]
+    fn vaddvq_u64_(a: uint64x2_t) -> u64;
```
Amanieu, line 122: Are you sure about the `.i32` in these? It seems to me that this should be the return type of the intrinsic.

Amanieu:
Yes, you can just delete the `vec_leak` feature.

crates/core_arch/src/aarch64/neon/mod.rs
Lokathor: No, that does a normal "vertical" add, this is a pairwise add.
```
// this is simd_add
vertical_add([a, b, c, d], [e, f, g, h]) => [a+e, b+f, c+g, d+h]

// this is the padd op family
pairwise_add([a, b, c, d], [e, f, g, h]) => [a+b, c+d, e+f, g+h]
```

crates/core_arch/src/aarch64/neon/mod.rs
Lokathor: Good catch. It uses `i32` as the "out" type for all the 8/16/32 functions and I didn't check ahead of time if it changed when moving up to 64-bit.

Some time next week I hope to get another device which will run an arm64 OS so it'll be easier for me to test these things ahead of time.

Lokathor:
Out of time to work on the PR this morning, I will be back later today. If you could diagnose the ICE in the wasi build @Amanieu that would be quite helpful. I think it's just a flutter in the current nightly because this PR doesn't touch any wasi stuff.

Amanieu:
cc @alexcrichton for the WASI CI failure

alexcrichton:
Oh dear looks like something in rustc changed, feel free to drop it and I will investigate next monday

alexcrichton:
I've fixed the wasi issue at https://github.com/rust-lang/rust/pull/75316, and I'll circle back here to re-enable CI after that lands.

crates/core_arch/src/aarch64/neon/mod.rs
```diff
@@ -82,8 +82,41 @@ extern "C" {
...
+    fn vaddv_u8_(a: uint8x8_t) -> u8;
+    #[link_name = "llvm.aarch64.neon.saddv.i32.v8i16"]
+    fn vaddvq_s16_(a: int16x8_t) -> i16;
+    #[link_name = "llvm.aarch64.neon.saddv.i32.v4i32"]
+    fn vaddvq_s32_(a: int32x4_t) -> i32;
+    #[link_name = "llvm.aarch64.neon.saddv.i32.v16i8"]
+    fn vaddvq_s8_(a: int8x16_t) -> i8;
+    #[link_name = "llvm.aarch64.neon.uaddv.i32.v8i16"]
+    fn vaddvq_u16_(a: uint16x8_t) -> u16;
+    #[link_name = "llvm.aarch64.neon.uaddv.i32.v3i32"]
```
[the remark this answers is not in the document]
Lokathor: Distressing that the tests seemed to work even with that error in there :/

PR #887, opened 2020-08-28
Avx512

minybot:
abs: f32,f64
add: i32,i64,f32,f64
sub: i32,i64,f32,f64
mul: i32,i64,f32,f64,u32,u64
div: f32,f64
max: i32,i64,f32,f64
min: i32,i64,f32,f64
sqrt: f32,f64


Amanieu:
Thanks!

Issue #823, opened 2019-10-22
Remove the MMX intrinsics


gnzlbg:
This will need some motivation and rationale, but the TL;DR is that they are very hard to use correctly, and they cause subtle bugs that take a lot of people a lot of time to debug. They aren't in the path towards stabilization either, so those who really need them can migrate towards using inline assembly instead, which is just as unstable. 

We should probably phase in a deprecation warning first, and remove them some time later.

rkruppe:
I would even argue they are *impossible* to use correctly in several circumstances:

- As discussed in https://github.com/rust-lang/rust/issues/53254#issuecomment-544915569, they break some libm functions on windows-gnu targets and `emm` can't reliably fix that.
- On x86 targets without SSE2, *any* float math will use the x87 stack, and for the same reason as in https://github.com/rust-lang/rust/issues/53254#issuecomment-544915569, `emms` can be reordered across those operations.

So if you're on one of these targets, and use certain operations on floats anywhere in the program, you're basically entirely at the mercy of LLVM not getting too aggressive with reordering code.

mati865:
LLVM developers are wondering about removing MMX: https://www.phoronix.com/scan.php?page=news_item&px=LLVM-Goodbye-MMX-Maybe

Amanieu:
It seems that LLVM's main concern is backwards compatibility with existing code that uses `__m64` and MMX intrinsics. We don't have this concern in Rust because there is no existing (stable) code using MMX.

I propose simply removing the MMX intrinsics and MMX types. I don't think there is much value in emulating them on top of SSE2 since anyone writing new code should not be using MMX intrinsics.

Note that there is still an issue with inline asm in that it can't properly express clobbers of the FP/MMX registers, so you currently can't use it with MMX instructions.

lu-zero:
Sounds a good idea given the potential harm.

mati865:
I could prepare PR but what to do with `is_x86_feature_detected!("mmx")`?
It's currently works on stable: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=41534dce6ad51df68df2e79323adfd44

Amanieu:
It's fine to have feature detection for a feature even if we don't support codegen for it.

PR #890, opened 2020-09-02
Bye bye MMX!

mati865:
Fixes https://github.com/rust-lang/stdarch/issues/823


bjorn3:
There is also special casing in the compiler to give the mmx vector type the right LLVM type. That should also be removed.

mati865:
@bjorn3 sure, after this is merged I'm going to grep `mmx` on Rust repo.

Amanieu:
I'm a bit concerned about the number of functions we're removing from SSE, SSE2, etc. Are we sure that all of these are MMX-only?

crates/std_detect/src/detect/os/x86.rs
```diff
@@ -119,7 +119,6 @@ pub(crate) fn detect_features() -> cache::Initializer {
         enable(extended_features_ebx, 19, Feature::adx);
         enable(extended_features_ebx, 11, Feature::rtm);
         enable(proc_info_edx, 4, Feature::tsc);
-        enable(proc_info_edx, 23, Feature::mmx);
```
Amanieu, line 122: We want to keep feature detection for MMX.

mati865: Test worked without that but sure I'll bring it back:
```
$ cargo test --package std_detect --test cpu-detection -- x86_all --exact --nocapture <

   Compiling libc v0.2.76
   Compiling cfg-if v0.1.10
   Compiling byteorder v1.3.4
   Compiling cupid v0.6.1
   Compiling auxv v0.3.3
   Compiling std_detect v0.1.5 (D:\tmp\stdarch\crates\std_detect)
    Finished test [unoptimized + debuginfo] target(s) in 12.19s  
     Running target\debug\deps\cpu_detection-3a542c0e111d2458.exe

running 1 test
aes: true     
pcmulqdq: true
rdrand: true  
rdseed: true  
tsc: true     
mmx: true
sse: true
sse2: true
sse3: true
ssse3: true
sse4.1: true
sse4.2: true
sse4a: true
sha: true
avx: true
avx2: true
avx512f: false
avx512cd: false
avx512er: false
avx512pf: false
avx512bw: false
avx512dq: false
avx512vl: false
avx512ifma: false
avx512vbmi: false
avx512vpopcntdq: false
avx512vbmi2 false
avx512gfni false
avx512vaes false
avx512vpclmulqdq false
avx512vnni false
avx512bitalg false
avx512bf16 false
avx512vp2intersect false
f16c: true
fma: true
bmi1: true
bmi2: true
abm: true
lzcnt: true
tbm: false
popcnt: true
fxsr: true
xsave: true
xsaveopt: true
xsaves: true
xsavec: true
test x86_all ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out
```

bjorn3: That is because the test only tests that getting the feature compiles and doesn't panic. Because `cfg!(target_feature = "mmx")` returns true, it still shows up as true in the test output, but if the `cfg!(target_feature)` fast-path of `is_x86_feature_detected!` was removed, the test would give false for mmx.

mati865:
> 
> 
> I'm a bit concerned about the number of functions we're removing from SSE, SSE2, etc. Are we sure that all of these are MMX-only?
> 
> EDIT: Apparently just removing anything that uses `__m64` seems to be correct. See MSVC [docs](https://docs.microsoft.com/en-us/cpp/cpp/m64):
> 
> > The __m64 data type is not supported on x64 processors. Applications that use __m64 as part of MMX intrinsics must be rewritten to use equivalent SSE and SSE2 intrinsics.

I wasn't sure whether we want to just remove MMX instructions or all the instructions involving MMX registers.
Went for the latter but I can bring back `__m64` and SSE{,2} instructions using MMX registers.

Amanieu:
Make sure to tag with relnotes when updating the submodule in rust-lang/rust.

Amanieu:
Hmm actually this is only unstable code, so we don't want it in the release notes.

danieldjewell:
It would appear that this change has broken compilation of rust-lang/packed_simd (see referenced issues). I assume that it was this pull that broke things as the "_mm_movemask_pi8" macro appears to have been removed?

Could someone who is more versed in this than I perhaps take a look? 

@Amanieu It would appear that your concerns were well founded.

PR #891, opened 2020-09-03
Avx512

minybot:
fmadd: ps,pd
fnmadd: ps,pd
fmsub: ps,pd
fnmsub: ps,pd
fmaddsub: ps,pd
fmsubadd: ps,pd

add_round: ps,pd
sub_round: ps,pd
mul_round: ps,pd
div_round: ps,pd
sqrt_round: ps,pd

cvt_roundps_epi32
cvt_roundps_epu32


Amanieu:
`assert_instr` can match a prefix, so you can use `assert_instr(vfmadd)`. Please add a comment next to it to explain that the compiler can generate different instructions for this intrinsic.

Amanieu:
For rounding modes can you do the same thing as the `constify_imm5_sae` macro: we want to panic in case of an invalid rounding mode since that could crash the compiler if we let it through to the LLVM intrinsic.

jrmuizel:
It looks like this broke the check-style tests. It wasn't caught before merge because `check-style` passes without checking anything if there's no rustfmt available for the particular nightly.

PR #892, opened 2020-09-05
Add vld1q_f32

jrmuizel:


crates/core_arch/src/arm/neon/mod.rs
```diff
@@ -63,7 +63,7 @@ types! {
     /// ARM-specific 128-bit wide vector of four packed `u32`.
     pub struct uint32x4_t(u32, u32, u32, u32);
     /// ARM-specific 128-bit wide vector of four packed `f32`.
-    pub struct float32x4_t(f32, f32, f32, f32);
+    pub struct float32x4_t(pub f32, pub f32, pub f32, pub f32);
```
Amanieu, line 66: Why did you change these fields to `pub`?

crates/core_arch/src/aarch64/neon/mod.rs
```diff
@@ -1799,6 +1799,14 @@ pub unsafe fn vqtbx4q_p8(a: poly8x16_t, t: poly8x16x4_t, idx: uint8x16_t) -> pol
     ))
 }
 
+#[inline]
+#[target_feature(enable = "neon")]
+#[cfg_attr(test, assert_instr(ldr))]
+pub unsafe fn vld1q_f32(addr: *const f32) -> float32x4_t {
+    use crate::core_arch::simd::f32x4;
+    transmute(f32x4::new(*addr, *addr.offset(1), *addr.offset(2), *addr.offset(3)))
+}
```
Amanieu, line 1808: It would be better to have a single implementation of this in `arm/neon` which is shared by both aarch64 and arm.

crates/core_arch/src/arm/neon/mod.rs
```diff
@@ -1767,6 +1769,16 @@ pub unsafe fn vld1q_u8(addr: *const u8) -> uint8x16_t {
     ptr::read(addr as *const uint8x16_t)
 }
 
+/// Load multiple single-element structures to one, two, three, or four registers
+#[inline]
+#[cfg(target_arch = "arm")]
+#[target_feature(enable = "neon")]
+#[cfg_attr(target_arch = "arm", target_feature(enable = "v7"))]
+#[cfg_attr(test, assert_instr(ldr))]
```
Amanieu, line 1777: This should check for `vldr` in `assert_instr`.

Amanieu (commented):
Are you planning on adding all the other `vld*` intrinsics as well?

crates/core_arch/src/arm/neon/mod.rs
jrmuizel: Originally I was constructing the float32x4_t directly and needed these to be `pub` to do that. I forgot to push the version that had the `pub` removed when I switched to transmuting `f32x4`.

crates/core_arch/src/aarch64/neon/mod.rs
jrmuizel: I agree but went with separate implementations because clang's implementations were separate:

arm:
https://github.com/llvm/llvm-project/blob/d5c45041f1465f4ecc3828efbbb27aa7b4d23d89/clang/lib/CodeGen/CGBuiltin.cpp#L5875

aarch64:
https://github.com/llvm/llvm-project/blob/d5c45041f1465f4ecc3828efbbb27aa7b4d23d89/clang/lib/CodeGen/CGBuiltin.cpp#L10607

I'm happy to unify them if you think that's better.

crates/core_arch/src/arm/neon/mod.rs
jrmuizel: Ok

jrmuizel:
> Are you planning on adding all the other `vld*` intrinsics as well?

My plan was to just do `vld1q_f32` and `vld1q_dup_f32` as those are the ones I need. But once I've got the hang of things I might add the others too.

crates/core_arch/src/aarch64/neon/mod.rs
Amanieu: Ah good point! Indeed ARM's case is different since the `VLD1` intrinsic takes additional alignment information as an operand. In this case it's fine to leave it separate.

crates/core_arch/src/arm/neon/mod.rs
```diff
@@ -1767,6 +1769,16 @@ pub unsafe fn vld1q_u8(addr: *const u8) -> uint8x16_t {
     ptr::read(addr as *const uint8x16_t)
 }
 
+/// Load multiple single-element structures to one, two, three, or four registers
+#[inline]
+#[cfg(target_arch = "arm")]
+#[target_feature(enable = "neon")]
+#[cfg_attr(target_arch = "arm", target_feature(enable = "v7"))]
```
Amanieu, line 1776: You can remove the `cfg_attr` since this is already guarded by `cfg(target_arch = "arm")`.

Amanieu:
LGTM. Do you want to try implementing the other vector load intrinsics as well or do you just want me to merge this?

jrmuizel:
Let's merge this and I'll look at the other intrinsics next.

Issue #889, opened 2020-08-31
Confusion between `target_feature` and runtime feature detection


calebzulawski:
I'm working on a crate that uses SIMD intrinsics for various architectures and ran into 2 seemingly contradictory issues with the `wasm32` and `arm` targets.

If I build for `wasm32` without the `simd128` feature, I still have access to the SIMD types.  When I test I get errors such as:
```
wasm-ld: error: function signature mismatch: _ZN17generic_simd_test21assign_scalar_op_impl17ha42fab674cd3bd98E
>>> defined as (v128) -> void in /target/wasm32-unknown-emscripten/debug/deps/generic_simd_test-5e53622f2d526041.1odaqe15wmy6tg3n.rcgu.o
>>> defined as (i32, i32, i32, i32) -> void in /target/wasm32-unknown-emscripten/debug/deps/generic_simd_test-5e53622f2d526041.2ha4zx2tw8sj5cyc.rcgu.o
```
I'm not too familiar with wasm but it looks to me like a valid ABI issue.  Maybe `v128` and the associated intrinsics should be behind `#[cfg(target_feature = "simd128")]`?

Conversely, I have discovered that the NEON types/intrinsics are only available when built with the `v7` and `neon` features, which is surprising to me since it looks like runtime feature detection should be possible on targets such as `armv7-unknown-linux-gnueabihf`.  Even more confusing, I am not able to simply add `-Ctarget-feature=+v7,+neon`, maybe this is because `core` is already compiled?

I'm aware both `wasm32` and `arm` are unstable and it's possible some of this behavior might be expected. Possible improvements I think could be helpful:
* Document that the `simd128` feature is required and why
* Document that the `v7` and `neon` features are required and why (and why runtime detection isn't possible?)

Thanks for any help!

Amanieu:
cc @alexcrichton for the wasm stuff.

Regarding `neon` I am not exactly sure why the static features are required. It might be some historical reason that no longer applies, in which can you can attempt to fix it with a PR to see if the tests uncover any issue.

alexcrichton:
Can you gist the code you're using for wasm? It definitely looks like a compiler bug we should fix!

sayantn:
`#[target_feature(enable = "simd128")]` was added to WASM intrinsics, and the erroneous `target_feature = "neon"` requirement on ARM was removed

PR #893, opened 2020-09-06
Remove requirement on neon feature for arm

calebzulawski:
As I noted in #889, the `neon` target feature shouldn't be required to use NEON on ARM (it should be possible to detect it at runtime).

### DIFF
--- a/crates/core_arch/src/arm/mod.rs
+++ b/crates/core_arch/src/arm/mod.rs
@@ -19,19 +19,9 @@ mod v7;
 #[cfg(any(target_arch = "aarch64", target_feature = "v7"))]
 pub use self::v7::*;
 
-// NEON is supported on AArch64, and on ARM when built with the v7 and neon
-// features. Building ARM without neon produces incorrect codegen.
-#[cfg(any(
-    target_arch = "aarch64",
-    all(target_feature = "v7", target_feature = "neon"),
-    dox
-))]
+#[cfg(any(target_arch = "aarch64", target_feature = "v7", dox))]
 mod neon;
-#[cfg(any(
-    target_arch = "aarch64",
-    all(target_feature = "v7", target_feature = "neon"),
-    dox
-))]
+#[cfg(any(target_arch = "aarch64", target_feature = "v7", dox))]
 pub use self::neon::*;
 
 #[cfg(any(target_arch = "aarch64", target_feature = "v7"))]


Lokathor:
Ya know that's a good point, we don't gate `avx` intrinsics on `avx` being enabled.

calebzulawski:
The removed comment indicates a codegen issue, which I have no idea how to validate (I couldn't quite figure out what the tests do).  I just removed the `neon` target feature from the CI tests, though.

calebzulawski:
It appears the `vtbl`/`vtbx` intrinsics are generating a few more instructions than allowed:
https://github.com/rust-lang/stdarch/pull/893/checks?check_run_id=1076991757#step:13:1456

calebzulawski:
I've been playing around with this on godbolt, comparing with clang's output.

C++ (optimized): https://godbolt.org/z/babdbh
C++ (unoptimized): https://godbolt.org/z/6KfoTb
Rust: https://rust.godbolt.org/z/dYM93Y

I'm not sure, but my guess is that rustc is generating slightly different IR and LLVM is failing to optimize it.  Both produce sequences of `getelementptr inbounds` and `load`, so the optimizer can clearly handle this.  However, clang simply rewrites the inputs to the stack with `store`, but rustc skips that and does a `bitcast`.  Is is possible the cast is opaque to the optimizer?

Amanieu:
Rust is passing the vector by reference instead of by value because the NEON target feature is not enabled globally.

Specifically this paragraph in the [RFC](https://rust-lang.github.io/rfcs/2325-stable-simd.html):

> The Rust ABI will currently be implemented such that all related packed-SIMD types are passed via memory instead of by-value. This means that regardless of the target features enabled for a function everything should agree on how packed SIMD arguments are passed across boundaries and whatnot.

calebzulawski:
Thanks, that completely clears that up!  Testing a somewhat contrived example (https://rust.godbolt.org/z/f4acxb)
all of those moves do seem purely related to passing by reference and don't appear in actual usage.

My uneducated guess is that this is appearing here because the `int8x8x4_t` type requires a few more instructions to access via reference because it's so big, so intrinsics using it (`vtbl`/`vtbx`) exceed the 22 instruction threshold.  I'm guessing aarch64 didn't have this problem because the larger registers allow fewer instructions for dereferencing.  Not sure why building with the `neon` target feature reduces the number of instructions--would it enable pass-by-value?

Amanieu:
Yes, enabling the `neon` feature globally will change then ABI and cause vectors to be passed by value.

Rather than increasing the instruction limit I think you can just restore `-C target-feature=+neon` in the CI flags. This should allow `as_ptr` and `store` to be inlined which avoids 2 function calls.

calebzulawski:
Ok, that seems reasonable, I removed that change.

PR #901, opened 2020-09-10
Avx512f

minybot:
add constify_imm8_sae to fix rotation

getmant: ps,pd
getmant_round: ps,pd
roundps_pd
cvtps_pd


crates/core_arch/src/x86/macros.rs
```diff
@@ -247,6 +247,491 @@ macro_rules! constify_imm4_sae {
...
+            (15, 3, 8) => $expand!(15, 3, 8),
+            (_, _, _) => panic!("Invalid sae value"),
+        }
+    };
+}
+
+// Constifies 8 bits along with an sae option without rounding control.
+// This macro enforces that.
+#[allow(unused)]
+macro_rules! constify_imm8_sae {
```
Amanieu, line 470: How is this different from just `constify_imm8`?

crates/core_arch/src/x86/avx512f.rs
```diff
@@ -12066,6 +12880,56 @@ mod tests {
...
+            assert_eq_m512i(r, _mm512_setzero_si512());
+            let r = _mm512_maskz_cvt_roundps_ph(
+                0b00000000_11111111,
+                a,
+                _MM_FROUND_TO_NEAREST_INT | _MM_FROUND_NO_EXC,
+            );
+            let e = _mm512_setr_epi32(0, -2, 2, -4, 4, -6, 6, -8, 0, 0, 0, 0, 0, 0, 0, 0);
+            assert_eq_m512i(r, e);
+        }
+    */
```
Amanieu, line 12932: Why are these tests commented out?

crates/core_arch/src/x86/macros.rs
```diff
@@ -247,6 +247,491 @@ macro_rules! constify_imm4_sae {
...
+            (13, 3) => $expand!(13, 3),
+            (14, 0) => $expand!(14, 0),
+            (14, 1) => $expand!(14, 1),
+            (14, 2) => $expand!(14, 2),
+            (14, 3) => $expand!(14, 3),
+            (15, 0) => $expand!(15, 0),
+            (15, 1) => $expand!(15, 1),
+            (15, 2) => $expand!(15, 2),
+            (15, 3) => $expand!(15, 3),
+            (_, _) => panic!("Invalid sae value"),
```
Amanieu, line 321: Should this be invalid mantissa instead?

crates/core_arch/src/x86/macros.rs
minybot: > How is this different from just `constify_imm8`?
        ($imm4:expr, $imm2:expr) => {
...
                $imm2 << 2 | $imm4,
In LLVM function, it merges two parameters to one. Clang accepts (0..15,0..3), but imm4 uses only 2 bits, and imm2 uses 2 bits in the definition. So, we should follow Clang or just use (0..3,0..3) for the function?

crates/core_arch/src/x86/avx512f.rs
minybot: > Why are these tests commented out?

LLVM uses different way to implement float16(ph). I will remove those lines.

crates/core_arch/src/x86/macros.rs
Amanieu: But this macro only has `($imm8:expr, $expand:ident)`, just like `consitfy_imm8`.

minybot: sorry, 
constify_imm8 is [0..255].   
->254 => $expand!(254),
->_ => $expand!(255),

constify_imm8_sae has out of range check.
->255 => $expand!(255),
->_ => panic!("Invalid sae value"),

Do you think this is necessary?

Amanieu: Ah OK. I think the naming could be clearer, or at least a comment explaining the difference.

crates/core_arch/src/x86/macros.rs
minybot: > Should this be invalid mantissa instead?

($imm4:expr, $imm2:expr) => {
...
$imm2 << 2 | $imm4,
In LLVM function, it merges two parameters to one. Clang accepts (0..15,0..3), but imm4 uses only 2 bits, and imm2 uses 2 bits in the definition. So, we should follow Clang or just use (0..3,0..3) for the function?

crates/core_arch/src/x86/macros.rs
minybot: > Ah OK. I think the naming could be clearer, or at least a comment explaining the difference.
ok. I will make it clearer.

crates/core_arch/src/x86/macros.rs
Amanieu: When in doubt you should check what GCC accepts. It is generally closer to the spec than Clang.

minybot: > When in doubt you should check what GCC accepts. It is generally closer to the spec than Clang.
Yes, I tried Intel compiler also. GCC is closer to ICC I guess.
Btw, Clang, GCC, ICC accept (0..15)(0..3).
In Rust, we use enum(0x00..0x03) and enum(0x00..0x02). So, we can do (0..4).(0..3) or just use (0..16)(0..3) in this case?

Amanieu: Just use (0..15)(0..3) like the C compilers.

minybot: Why? compatibility?

Amanieu: Yes

PR #902, opened 2020-09-10
Add vcvtq_u32_f32 and vcvtq_s32_f32

jrmuizel:
These intrinsics are implemented differently for aarch64 and arm
in clang. i.e. aarch64 uses the llvm.aarch64.neon.fcvtzs.v4i32.v4f32
intrinsic. However, there didn't seem to be any advantage to using
that intrinsic instead of just sharing code.

### DIFF
--- a/crates/core_arch/src/aarch64/neon/mod.rs
+++ b/crates/core_arch/src/aarch64/neon/mod.rs
@@ -285,6 +285,11 @@ extern "C" {
         b3: int8x16_t,
         c: uint8x16_t,
     ) -> int8x16_t;
+
+    #[link_name = "llvm.aarch64.neon.fcvtzu.v4i32.v4f32"]
+    fn vcvtq_u32_f32_(a: float32x4_t) -> uint32x4_t;
+    #[link_name = "llvm.aarch64.neon.fcvtzs.v4i32.v4f32"]
+    fn vcvtq_s32_f32_(a: float32x4_t) -> int32x4_t;
 }
 
 /// Absolute Value (wrapping).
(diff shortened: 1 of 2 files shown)


Amanieu:
Can you add tests for out-of-range values? For example passing a negative value to the unsigned conversion should produce a result of 0.

jrmuizel:
It looks like the out of range values will throw a floating point exception: https://developer.arm.com/docs/ddi0596/h/shared-pseudocode-functions/shared-functionsfloat-pseudocode#impl-shared.FPToFixed.5

Amanieu:
That just sets an exception bit in the FP status register and returns a NaN. I'm more worried about LLVM optimizing it to `undef` and returning an arbitrary value.

jrmuizel:
It does seem like there's some badness going on with llvm converting to undef. This shows the difference between the aarch64 and ARM approaches:
https://gcc.godbolt.org/z/5Ws1P8

It feels like the ARM behaviour is not really what you'd when using this intrinsic.

Amanieu:
I checked the behavior with GCC and with the ACLE spec and it seems that the AArch64 behavior is the  correct one. It seems that this is a bug in Clang.

I don't think LLVM currently exposes an intrinsic for this on ARM, but you should definitely open a bug report for Clang on https://bugs.llvm.org/.

In the meantime for this PR you can just keep Clang's behavior but with a clear comment explaining that the behavior isn't 100% correct.

Amanieu:
Can you report the Clang bug on bugs.llvm.org and add a link to the bug report in a comment?

jrmuizel:
It looks like CI is broken because of a missing apple nightly

Amanieu:
It's being worked on, should be fixed for tomorrow's nightlies.

Amanieu:
CI is now fixed but there are errors.

jrmuizel:
https://bugs.llvm.org/show_bug.cgi?id=47510

PR #903, opened 2020-09-10
 Add vgetq_lane_s32

jrmuizel:
The implementation basically matches the existing vgetq_lane_u32

### DIFF
--- a/crates/core_arch/src/arm/neon/mod.rs
+++ b/crates/core_arch/src/arm/neon/mod.rs
@@ -1315,11 +1315,8 @@ pub unsafe fn vtbx4_p8(a: poly8x8_t, b: poly8x8x4_t, c: uint8x8_t) -> poly8x8_t 
 // `mov` seems to be an acceptable intrinsic to compile to
 // #[cfg_attr(all(test, target_arch = "aarch64"), assert_instr(vmov, imm5 = 1))]
 pub unsafe fn vgetq_lane_u64(v: uint64x2_t, imm5: i32) -> u64 {
-    if (imm5) < 0 || (imm5) > 1 {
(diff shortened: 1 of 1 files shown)


crates/core_arch/src/arm/neon/mod.rs
```diff
@@ -1368,6 +1368,21 @@ pub unsafe fn vgetq_lane_u32(v: uint32x4_t, imm5: i32) -> u32 {
...
+/// Move vector element to general-purpose register
+#[inline]
+#[target_feature(enable = "neon")]
+#[cfg_attr(target_arch = "arm", target_feature(enable = "v7"))]
+#[rustc_args_required_const(1)]
+#[cfg_attr(all(test, target_arch = "arm"), assert_instr("vmov.32", imm5 = 2))]
+#[cfg_attr(all(test, target_arch = "aarch64"), assert_instr(mov, imm5 = 2))]
+pub unsafe fn vgetq_lane_s32(v: int32x4_t, imm5: i32) -> i32 {
+    if (imm5) < 0 || (imm5) > 3 {
+        unreachable_unchecked()
```
Amanieu, line 1380: Use a `panic` here instead. This will be optimized out since `imm5` is required to be a constant.

crates/core_arch/src/arm/neon/mod.rs
```diff
@@ -1368,6 +1368,21 @@ pub unsafe fn vgetq_lane_u32(v: uint32x4_t, imm5: i32) -> u32 {
...
+    }
+    let imm5 = (imm5 & 0b11) as u32;
```
Amanieu, line 1382: The masking isn't needed since you've already used a range check.

PR #912, opened 2020-09-17
Avx512f

minybot:
knot, kandn, kxornd, kmov
permute: epi32
extractf32x4_ps not mask and maskz
permute_f32x4
permute_f64x2
permute_i32x4
permute_i64x2 
moveldup_ps
movehdup_ps
movedup_pd


crates/core_arch/src/x86/avx512f.rs
```diff
@@ -9270,6 +10519,77 @@ pub unsafe fn _mm512_kxor(a: __mmask16, b: __mmask16) -> __mmask16 {
     transmute(kxorw(a, b))
 }
 
+/// Compute the bitwise NOT of 16-bit mask a, and store the result in k.
+///
+/// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=knot_mask16&expand=3233)
+#[inline]
+#[target_feature(enable = "avx512f")]
+#[cfg_attr(all(test, not(target_os = "macos")), assert_instr(not))] // generate normal not code instead of knotw
```
Amanieu, line 10527: Why are you special-casing macos here?

minybot: In osX, it does not generate any "not" or "xor" instructions. It generates "vmovaps" tested in CI and osX with clang.

Amanieu: Can you post the fully assembly output you are getting on macOS? I find it extremely strange that it is behaving differently from Linux.

minybot: osX Rust

#[target_feature(enable = "avx512f")]
unsafe fn avx512() {
    let c = _mm512_knot(0b00000000_00000000);
}

fn main() {
    unsafe { avx512(); }
}   
---------------------------------
       .loc    6 10538 0
        .cfi_startproc
        pushq   %rbp
        .cfi_def_cfa_offset 16
        .cfi_offset %rbp, -16
        movq    %rsp, %rbp
        .cfi_def_cfa_register %rbp
        subq    $16, %rsp
        movw    %di, -4(%rbp)
Ltmp21:
        .loc    6 10539 15 prologue_end
        movzwl  %di, %edi
        movl    $65535, %esi
        callq   __ZN9core_arch9core_arch3x867avx512f11_mm512_kxor17hbdf54d0540836621E
        movw    %ax, -6(%rbp)
        .loc    6 0 15 is_stmt 0
        movw    -6(%rbp), %ax
        .loc    6 10539 5
        movw    %ax, -2(%rbp)
        movw    -2(%rbp), %ax
        movw    %ax, -8(%rbp)
        .loc    6 0 5
        movw    -8(%rbp), %ax
        .loc    6 10540 2 is_stmt 1
        addq    $16, %rsp
        popq    %rbp
        retq

Amanieu:
For the `k{not,and,xor, etc}` function I would prefer if we used simple bitwise operations on integers rather than intrinsics. As a bonus it would allow the compiler to optimize these functions better.

We could then remove the mostly useless `assert_instr` on these intrinsics.

Amanieu:
`knot` doesn't use a `not` instruction because it only inverts a 16-bit value. It uses `xor 0xffff` instead.

Amanieu:
You can change it to `xor`. CI will complain if you have an intrinsic without `assert_instr`.

crates/core_arch/src/x86/avx512f.rs
```diff
@@ -10524,7 +10524,7 @@ pub unsafe fn _mm512_kxor(a: __mmask16, b: __mmask16) -> __mmask16 {
 /// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=knot_mask16&expand=3233)
 #[inline]
 #[target_feature(enable = "avx512f")]
-#[cfg_attr(test, assert_instr(not))] // generate normal not code instead of knotw
+//#[cfg_attr(test, assert_instr(xor))] // generate normal not code instead of knotw
```
Amanieu, line 10527: Please don't leave commented-out code. Either keep the line with `xor` or delete it entirely.

Amanieu:
Yes, in that case just remove the `assert_instr`. I think this happens because the `darwin` target optimizes for the `core2` CPU while other targets optimize for the `generic` cpu.

PR #918, opened 2020-09-24
Implement prefetch hints for aarch64

dgbo:
Co-authored-by: Wang Maozhang <wangmaozhang@huawei.com>

This implements prefetch hints for aarch64 via LLVMs `prefetch` instrinsic.

We wintness ~20% performance gain in seal_pre_commit_phase1 of Filecoin with 32GB data input.

Verified with `cargo +nightly test` and `cargo +nightly test --release`.
There are four failures (as shown below), but these are not cause by this PR.
These tests failed both with and without this PR.
```
---- core_arch::aarch64::tme::assert___tcommit_tcommit stdout ----
disassembly for stdarch_test_shim___tcommit_tcommit:
         0: adrp x8, 16f000 <_GLOBAL_OFFSET_TABLE_+0x280>
         1: ldr x8, [x8, #3792]
         2: adrp x9, f1000 <anon.1626de38755f4729e9f2a08512645c0a.14.llvm.15569108828795629504+0x10b87>
         3: add x9, x9, #0xd29
         4: str x9, [x8]
         5: msr s0_3_c3_c0_3, xzr
         6: ret
thread 'core_arch::aarch64::tme::assert___tcommit_tcommit' panicked at 'failed to find instruction `tcommit` in the disassembly', crates/stdarch-test/src/lib.rs:144:9


failures:
    core_arch::aarch64::tme::assert___tcancel_tcancel
    core_arch::aarch64::tme::assert___tcommit_tcommit
    core_arch::aarch64::tme::assert___tstart_tstart
    core_arch::aarch64::tme::assert___ttest_ttest
```

Thanks.

### DIFF
--- a/crates/core_arch/src/aarch64/prefetch.rs
+++ b/crates/core_arch/src/aarch64/prefetch.rs
@@ -63,8 +63,24 @@ pub const _PREFETCH_LOCALITY3: i32 = 3;
 #[cfg_attr(test, assert_instr("prfm pstl3keep", rw = _PREFETCH_WRITE, locality = _PREFETCH_LOCALITY1))]
 #[cfg_attr(test, assert_instr("prfm pstl2keep", rw = _PREFETCH_WRITE, locality = _PREFETCH_LOCALITY2))]
 #[cfg_attr(test, assert_instr("prfm pstl1keep", rw = _PREFETCH_WRITE, locality = _PREFETCH_LOCALITY3))]
+#[rustc_args_required_const(1, 2)]
 pub unsafe fn _prefetch(p: *const i8, rw: i32, locality: i32) {
     // We use the `llvm.prefetch` instrinsic with `cache type` = 1 (data cache).
     // `rw` and `strategy` are based on the function parameters.
-    prefetch(p, rw, locality, 1);
+    macro_rules! pref {
+        ($rdwr:expr, $local:expr) => {
+            match ($rdwr, $local) {
+                (0, 0) => prefetch(p, 0, 0, 1),
+                (0, 1) => prefetch(p, 0, 1, 1),
+                (0, 2) => prefetch(p, 0, 2, 1),
+                (0, 3) => prefetch(p, 0, 3, 1),
+                (1, 0) => prefetch(p, 1, 0, 1),
+                (1, 1) => prefetch(p, 1, 1, 1),
+                (1, 2) => prefetch(p, 1, 2, 1),
+                (1, 3) => prefetch(p, 1, 3, 1),
+                (_, _) => panic!("Illegal (rw, locality) pair in prefetch, value ({}, {}).", $rdwr, $local),
+            }
+        };
+    }
+    pref!(rw, locality);
 }
--- a/crates/core_arch/src/aarch64/prefetch.rs
+++ b/crates/core_arch/src/aarch64/prefetch.rs
@@ -1,0 +1,70 @@
+#[cfg(test)]
+use stdarch_test::assert_instr;
+
+extern {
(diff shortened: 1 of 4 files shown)


dgbo:
The check error is strange:
```
error[E0723]: can only call `transmute` from const items, not `const fn`
   --> crates/core_arch/src/wasm32/simd128.rs:420:5
    |
420 |     transmute(f64x2(a0, a1))
    |     ^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #57563 <https://github.com/rust-lang/rust/issues/57563> for more information
    = help: add `#![feature(const_fn)]` to the crate attributes to enable
```

This PR does not touch any code in wasm32 at all.
And the #![feature(const_fn)] is in  crates/core_arch/src/lib.rs already.

Could anyone give some suggestion here?

Best Regards.

crates/core_arch/src/aarch64/prefetch.rs
```diff
@@ -0,0 +1,70 @@
...
+#[inline(always)]
+#[cfg_attr(test, assert_instr("prfm pldl1strm", rw = _PREFETCH_READ, locality = _PREFETCH_LOCALITY0))]
+#[cfg_attr(test, assert_instr("prfm pldl3keep", rw = _PREFETCH_READ, locality = _PREFETCH_LOCALITY1))]
+#[cfg_attr(test, assert_instr("prfm pldl2keep", rw = _PREFETCH_READ, locality = _PREFETCH_LOCALITY2))]
+#[cfg_attr(test, assert_instr("prfm pldl1keep", rw = _PREFETCH_READ, locality = _PREFETCH_LOCALITY3))]
+#[cfg_attr(test, assert_instr("prfm pstl1strm", rw = _PREFETCH_WRITE, locality = _PREFETCH_LOCALITY0))]
+#[cfg_attr(test, assert_instr("prfm pstl3keep", rw = _PREFETCH_WRITE, locality = _PREFETCH_LOCALITY1))]
+#[cfg_attr(test, assert_instr("prfm pstl2keep", rw = _PREFETCH_WRITE, locality = _PREFETCH_LOCALITY2))]
+#[cfg_attr(test, assert_instr("prfm pstl1keep", rw = _PREFETCH_WRITE, locality = _PREFETCH_LOCALITY3))]
+pub unsafe fn _prefetch(p: *const i8, rw: i32, locality: i32) {
```
Amanieu, line 66: Use `#[rustc_args_required_const(1, 2)]` to ensure that `rw` and `locality` are compile-time constants.

Amanieu:
The wasm failure seems to be due to an issue on the latest Rust nightly. You can fix it by adding the `const_fn_transmute` feature to the list in lib.rs.

crates/core_arch/src/aarch64/prefetch.rs
bjorn3: You also need to match on every valid `(rw, locality)` pair and handle invalid values.

dgbo: Thanks a lot for the quick review.

The `#[rustc_args_required_const(1, 2)]` is added just before the function _prefetch.

Created a macro `pref` to match all valid `(rw, locality)` pairs.
An invalid pair would cause panic now. Suggestions?

dgbo:
> The wasm failure seems to be due to an issue on the latest Rust nightly. You can fix it by adding the `const_fn_transmute` feature to the list in lib.rs.

Added the `const_fn_transmute` feature to the list in lib.rs. The CI is green. Thanks.

Amanieu:
Thanks!

PR #926, opened 2020-10-07
Stabilize WebAssembly atomic intrinsics

alexcrichton:
This commit stabilizes the wasm atomic intrinsics now that the threads
proposal has advanced to Stage 4, and is shipping in browsers.
Unfortunately these intrinsics still aren't available through the
precompiled versions of the standard library because atomics are no

~~This commit also updates the availability and documentation of these
intrinsics to reflect how they're always available, but they probably
won't work as expected unless shared memory is used (which isn't enabled
by default). This reflects a change in the threads proposal from the
last year or so where the instructions validate without requiring a
shared memory.~~

### DIFF
--- a/crates/core_arch/src/wasm32/atomic.rs
+++ b/crates/core_arch/src/wasm32/atomic.rs
@@ -51,6 +51,7 @@ extern "C" {
 /// [instr]: https://webassembly.github.io/threads/syntax/instructions.html#syntax-instr-atomic-memory
 #[inline]
 #[cfg_attr(test, assert_instr("i32.atomic.wait"))]
+#[stable(feature = "wasm_atomics", since = "1.50.0")]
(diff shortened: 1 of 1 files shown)


alexcrichton:
I sort of forget the precise procedure by which these intrinsics are stabilized, but I'd be happy to write up more information if desired! (or start with an issue elsewhere as well)

alexcrichton:
Looks like LLVM isn't as far along as I thought, so I left in the module being disabled without the `atomics` feature.

Amanieu:
Is there a tracking issue for these? It needs an FCP before it can be stabilized. (rfcbot isn't enabled on this repo).

alexcrichton:
Ah yes, indeed! I will follow up with a tracking issue in rust-lang/rust and will cc the libs team there.

alexcrichton:
Ok I opened https://github.com/rust-lang/rust/issues/77839, but upon reflection I don't think stabilization is a good idea at this time, so I'll be leaving it at that.

PR #929, opened 2020-10-16
Support ARM crypto extension on A32/T32

makotokato:
ARM Crypto Extensions supports on aarch32 too. So we should allow this intrinsics on arm.

### DIFF
--- a/crates/core_arch/src/arm/crypto.rs
+++ b/crates/core_arch/src/arm/crypto.rs
@@ -2,35 +2,49 @@ use crate::core_arch::arm::{uint32x4_t, uint8x16_t};
 
 #[allow(improper_ctypes)]
 extern "C" {
-    #[link_name = "llvm.aarch64.crypto.aese"]
+    #[cfg_attr(target_arch = "aarch64", link_name = "llvm.aarch64.crypto.aese")]
+    #[cfg_attr(target_arch = "arm", link_name = "llvm.arm.neon.aese")]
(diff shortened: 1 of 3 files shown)


crates/core_arch/src/arm/crypto.rs
```diff
@@ -40,6 +54,7 @@ use stdarch_test::assert_instr;
 /// AES single round encryption.
 #[inline]
 #[target_feature(enable = "crypto")]
+#[cfg_attr(target_arch = "arm", target_feature(enable = "v8"))]
```
Amanieu, line 57: Is the `v8` feature actually required here? Does it work without this feature?

makotokato: > Is the `v8` feature actually required here? Does it work without this feature?

Yes, `v8` feature is required since this is ARMv8 feature.  If this is `v7` or nothing, this causes `LLVM ERROR: Cannot select: intrinsic %llvm.arm.neon.aese` .

tarcieri:
Per #1003, it looks like this PR potentially broke the crypto extension on `aarch64` although reviewing the changes I'm not quite sure how

CryZe:
It doesn't seem like there's any problem at all with using them: https://rust.godbolt.org/z/r4Yhdn

Maybe it's just a rustdoc bug?

tarcieri:
Huh wow, strange!

Amanieu:
#1055?

tarcieri:
Seems plausible. I guess this can be closed as a dup then?

PR #930, opened 2020-10-17
Updated CONTRIBUTING.MD for new package name and environment variables

DevJPM:
When I tried to execute the instructions in CONTRIBUTING.MD I noticed that they actually don't work out of the box.

First, apparently the `coresimd` package had been renamed at some point but the contributing.md was not updated, I fixed that.

Additionally apparently since this documentation was written there was an update on how tests are invoked now requiring specific environment variables. I have added text to explain which those are and what they need to say to make `cargo` and `rustc` happy.


Amanieu:
I would actually recommend using `TARGET=<target> ci/run.sh` to run the tests instead, it tends to work better in practice.

PR #936, opened 2020-10-26
Add shift-and-insert Arm intrinsics.

AdamHillier:
This PR adds Arm and Aarch64 Neon intrinsics for [shift-left-and-insert](https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics?search=vsli) and [shift-right-and-insert](https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics?search=vsri).

The LLVM codegen for these instructions is quite different on Arm vs Aarch64. In the former case, both `vsli` and `vsri` get transformed to `llvm.arm.neon.vshiftins`, which takes three equal-length vector arguments, where the last argument contains the per-element shift (which can be negative, for a right shift). In the latter case, `vsli` and `vsri` are transformed to `llvm.aarch64.neon.vsli` and `llvm.aarch64.neon.vsri` respectively, which takes two vector inputs and a constant for the shift.


crates/stdarch-verify/tests/arm.rs
```diff
@@ -331,6 +331,46 @@ fn verify_all_signatures() {
                 "_cls_u32",
                 "_cls_u64",
                 "_prefetch",
+                "vsli_n_s8",
+                "vsliq_n_s8",
+                "vsli_n_s16",
+                "vsliq_n_s16",
+                "vsli_n_s32",
+                "vsliq_n_s32",
+                "vsli_n_s64",
```
AdamHillier, line 340: I seem to have to add these for the `verify_all_signatures` tests to pass, I'm guessing because I've put the shift-and-insert tests in their own file (mimicking what's done for the table lookup tests, which also appear in this list of signatures to skip).

crates/core_arch/src/aarch64/neon/mod.rs
```diff
@@ -1858,6 +1893,488 @@ pub unsafe fn vcvtq_u32_f32(a: float32x4_t) -> uint32x4_t {
...
 }
 
+/// Shift Left and Insert (immediate)
+#[inline]
+#[target_feature(enable = "neon")]
+#[cfg_attr(test, assert_instr(sli, n = 1))]
+#[rustc_args_required_const(2)]
+pub unsafe fn vsli_n_s8(a: int8x8_t, b: int8x8_t, n: i32) -> int8x8_t {
+    if n < 0 || n > 7 {
+        unreachable_unchecked();
```
Amanieu, line 1903: Use a panic here for out-of-bounds values. It's optimized away in practice since this function is going to be inlined.

crates/stdarch-verify/tests/arm.rs
Amanieu: I believe that they are not being detected because you are generating the tests within a macro.

crates/core_arch/src/aarch64/neon/mod.rs
AdamHillier: Sounds good, will do :+1:

PR #939, opened 2020-10-27
Add more addition Arm intrinsics

AdamHillier:
This PR adds the following groups of addition Arm intrinsics (72 in total):
* `vaddl_high_[s/u][8/16/32]` - Signed/Unsigned Add Long (vector, high half)
* `vaddw_[s/u][8/16/32]` - Signed/Unsigned Add Wide
* `vaddw_high_[s/u][8/16/32]` - Signed/Unsigned Add Wide (high half)
* `vuqadd[q]_s[8/16/32/64]` - Signed saturating Accumulate of Unsigned value - Aarch64 only
* `vsqadd[q]_u[8/16/32/64]` - Unsigned saturating Accumulate of Signed value - Aarch64 only
* `vaddhn_[s/u][16/32/64]` - Add returning High Narrow
* `vaddhn_high_[s/u][16/32/64]` - Add returning High Narrow (high half)
* `vraddhn_[s/u][16/32/64]` - Rounding Add returning High Narrow
* `vraddhn_high_[s/u][16/32/64]` - Rounding Add returning High Narrow (high half)
* `vpaddl[q]_s[8/16/32]` - Signed Add Long Pairwise
* `vpaddl[q]_u[8/16/32]` - Unsigned Add Long Pairwise
* `vpadal[q]_s[8/16/32]` - Signed Add and Accumulate Long Pairwise - Code-gen on Arm vs Aarch64 is different.
* `vpadal[q]_u[8/16/32]` - Unsigned Add and Accumulate Long Pairwise - Code-gen on Arm vs Aarch64 is different.

### DIFF
--- a/crates/core_arch/src/aarch64/neon/mod.rs
+++ b/crates/core_arch/src/aarch64/neon/mod.rs
@@ -96,6 +96,23 @@ extern "C" {
     #[link_name = "llvm.aarch64.neon.suqadd.v2i64"]
     fn vuqaddq_s64_(a: int64x2_t, b: uint64x2_t) -> int64x2_t;
 
+    #[link_name = "llvm.aarch64.neon.usqadd.v8i8"]
+    fn vsqadd_u8_(a: uint8x8_t, b: int8x8_t) -> uint8x8_t;
+    #[link_name = "llvm.aarch64.neon.usqadd.v16i8"]
+    fn vsqaddq_u8_(a: uint8x16_t, b: int8x16_t) -> uint8x16_t;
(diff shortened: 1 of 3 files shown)


AdamHillier:
The (overall) GitHub diff has become very messed up for some reason -- it makes it look like there are many more lines changed than there actually are, with large sections re-written. If you look at each commit individually you can see that in reality there are very few lines deleted:

image

Which is to say it's probably best to review this commit-by-commit :)

Amanieu (approved):

Issue #344, opened 2018-03-06
f16/half-float support


gnzlbg:
I was implementing the `v{min,max}v` ARM and AArch64 intrinsics and ran into a couple of them that require passing `<4 x half>` vectors to llvm and returning `half` types: 

* `pub unsafe fn vmaxv_f16 (a: f16x4_t a) -> f16`
* `pub unsafe fn vmaxvq_f16 (a: f6x8_t a) -> f16`
* `pub unsafe fn vminv_f16 (a: f16x4_t a) -> f16`
* `pub unsafe fn vminvq_f16 (a: f16x8_t a) -> f16`

For example, `vmaxv_f16` should map to: `call half @llvm.aarch64.neon.fminv.f16.v4f16(<4 x half>)`

alexcrichton:
FWIW I think we've got basically no trans/language support for `f16` right now, so I'd just leave them out

gnzlbg:
The PowerPC 64 architecture also supports a `vector _Float16` type, with 16 16-bit floats.

tgross35:
`f16` and `f128` are now unstably available, so these can be added. This extends to platforms beyond Arm, see https://github.com/rust-lang/rust/issues/125440 for my primitive list

tgross35:
The aarch64 intrinsic mentioned here have been added and are tracked at https://github.com/rust-lang/rust/issues/136306, which currently has a proposed FCP. avx512fp16 is similar, tracked at https://github.com/rust-lang/rust/issues/127213.

Since there is now an unstable `f16` type, any other platforms should be able to add intrinsics that use it relatively easily.

PR #941, opened 2020-10-30
Add most single-register load Arm intrinsics.

AdamHillier:
This PR adds all single-register load Arm intrinsics, with the exception of those for `f16` (blocked by #344), and those (few) which are Aarch64-only. There are 72 new intrinsics in all.


crates/core_arch/src/aarch64/neon/mod.rs
```diff
@@ -8,25 +8,24 @@ pub use self::generated::*;
...
 use crate::{
-    core_arch::{arm::*, simd_llvm::*},
+    core_arch::{arm::*, simd::*, simd_llvm::*},
     mem::{transmute, zeroed},
 };
 #[cfg(test)]
 use stdarch_test::assert_instr;
 
+pub type p64 = u64;
+pub type p128 = u128;
```
Amanieu, line 18: These types should be private.

crates/core_arch/src/aarch64/neon/mod.rs
```diff
@@ -8,25 +8,24 @@ pub use self::generated::*;
...
     /// ARM-specific 64-bit wide vector of one packed `p64`.
-    pub struct poly64_t(i64); // FIXME: check this!
+    pub struct poly64_t(p64); // FIXME: check this!
     /// ARM-specific 64-bit wide vector of two packed `p64`.
-    pub struct poly64x2_t(i64, i64); // FIXME: check this!
-    /// ARM-specific 128-bit wide vector of one packed `p64`.
-    pub struct poly128_t(i128); // FIXME: check this!
+    pub struct poly64x2_t(p64, p64); // FIXME: check this!
+    /// ARM-specific 128-bit wide vector of one packed `p128`.
+    pub struct poly128_t(p128); // FIXME: check this!
```
Amanieu, line 28: These types should also be available on ARM: they are used by the crypto extension which is supported on ARMv8 in both 32-bit and 64-bit modes.

PR #943, opened 2020-11-01
Add VPCLMULQDQ Intrinsics

DevJPM:
This adds support for the two intrinsics covered by VPCLMULQDQ
and extends the tests from PCLMULQDQ as well as compares the results of these two
when the inputs on the lanes are different

Like #942 this depends on rust-lang/rust#78361 to work (though depending on `avx512pclmulqdq`) and was tested with that patch applied.
The test vectors are the same as [pclmulqdq.rs](https://github.com/rust-lang/stdarch/blob/master/crates/core_arch/src/x86/pclmulqdq.rs).
The folder is chosen to match AVX512F / -IFMA.
The name is chosen as [detected in Rust](https://github.com/rust-lang/stdarch/blob/master/crates/std_detect/src/detect/arch/x86.rs#L156).
I have marked this PR as a draft until the required compiler change landed in nightly.

### DIFF
--- a/crates/core_arch/src/x86/avx512vpclmulqdq.rs
+++ b/crates/core_arch/src/x86/avx512vpclmulqdq.rs
@@ -1,0 +1,181 @@
+//! Vectorized Carry-less Multiplication (VCLMUL)
+//!
+//! The reference is [Intel 64 and IA-32 Architectures Software Developer's
+//! Manual Volume 2: Instruction Set Reference, A-Z][intel64_ref] (p. 4-241).
+//!
+//! [intel64_ref]: http://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-instruction-set-reference-manual-325383.pdf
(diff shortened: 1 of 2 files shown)


DevJPM:
OK, I'm confused now. When I ask only for the short names of the intrinsics (i.e. `vpclmulqdq`) it fails not finding the long ones (e.g. `vpclmullqlqdq`). But when I ask for the long ones as well, as the `pclmulqdq.rs` file does, it fails sometimes because the disassembly only contains the long name.

Amanieu:
I think both disassemblies are valid, but you can put a common prefix in `assert_instr` so that it will match both.

DevJPM:
Sorry for all the force-pushed, I majorly screwed up the rebase from master (which was forced by me using the same location in `mod.rs` for VAES and VPCLMUL).

Also thank you very much for the prefix-advice. I hadn't realized that it only checks prefixes / substrings.

PR #947, opened 2020-11-05
Increment instructions limit for vtbl and vtbx.

makotokato:
When running cargo test --release on Ubuntu 20.04/armhf target, instructions limit isn't enough.


Amanieu:
That's strange. Are you adding `-C target-feature=+neon` like CI does?

makotokato:
> That's strange. Are you adding `-C target-feature=+neon` like CI does?

Yep. Thanks. I should add this flags. I withdraw this PR.

PR #948, opened 2020-11-05
Use --no-show-raw-insn to make disassemble parser simple.

makotokato:
Actually, we don't consider thumbv7neon for `stdarch-test`. Actually disassemble doesn't work on thumb/thumb2 now.

`objdump` has `--no-show-raw-insn` argument not to show raw instruction data. We should use it to make disassemble parser simple.


lu-zero (approved):
Looks good to me and since it is surviving the CI seems that works well for our supported platforms.

PR #952, opened 2020-11-11
avx512f example

minybot:
A avx512f example: A game connect5 (Outer-Open Gomoku).
Use avx512f to speed up the pattern matching.


Amanieu:
I think that it's because it is using core_arch from the nightly toolchain instead of from this crate. The submodule was updated 3 days ago so I'm re-running the tests, let's see if it passes.

Amanieu:
Can you force-push to re-run the test?

Amanieu:
You need to rebase your branch on the latest master.

Amanieu:
I believe it's because github actions runs the tests on the tree *after* merging but Cirrus CI (used for FreeBSD) runs the tests on the tree *before* merging.

PR #954, opened 2020-11-19
AVX-512DQ implementations

vertexclique:
* `[t]`: test needed
* `[ ]`: not implemented
* `[x]`: completed


Amanieu:
Ping @vertexclique, please run `cargo fmt` to fix the style so CI passes.

vertexclique:
@Amanieu I don't have that problem as of yet, unless something pressing I will format it when I have finished. I am not relying on the CI atm. Local work is fine as of yet.

PR #955, opened 2020-11-19
Add vmla_* / vmlal_*

makotokato:
Add Vector multiply accumulate intrinsic.

### DIFF
--- a/crates/core_arch/src/aarch64/neon/mod.rs
+++ b/crates/core_arch/src/aarch64/neon/mod.rs
@@ -350,6 +350,15 @@ extern "C" {
     fn vsri_n_s64_(a: int64x1_t, b: int64x1_t, n: i32) -> int64x1_t;
     #[link_name = "llvm.aarch64.neon.vsri.v2i64"]
     fn vsriq_n_s64_(a: int64x2_t, b: int64x2_t, n: i32) -> int64x2_t;
+
+    #[link_name = "llvm.fmuladd.v2f32"]
+    fn vmla_f32_(a: float32x2_t, b: float32x2_t, c: float32x2_t) -> float32x2_t;
+    #[link_name = "llvm.fmuladd.v4f32"]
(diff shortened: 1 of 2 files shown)


Amanieu:
VMLA is *not* a fused multiply-add. It involves an intermediate rounding step just like if you use two separate operations. Only VFMA is a fused multiply-add.

Amanieu:
Ping. The `vmla_*` functions needs to use `simd_add` and `simd_mul`. The `vfma_*` functions need to use `simd_fma`.

crates/core_arch/src/aarch64/neon/mod.rs
```diff
@@ -971,6 +971,40 @@ pub unsafe fn vaddvq_u64(a: uint64x2_t) -> u64 {
...
+    simd_fma(b, c, a)
+}
+
+/* FIXME: simd_fma doesn't support float64x1_t`
+/// Vector multiply accumulate
+#[inline]
+#[target_feature(enable = "neon")]
+#[cfg_attr(test, assert_instr(fmla))]
+pub unsafe fn vmla_f64(a: float64x1_t, b: float64x1_t, c: float64x1_t) -> float64x1_t {
+    simd_fma(b, c, a)
```
Amanieu, line 988: You can use the `mul_add` method on `f64`.

makotokato: Sorry for delay an thank you for review. I will update this PR.

makotokato:
unrelated CI is failure due to OOM...

Amanieu:
CI should be fixed now.

However you still haven't fixed the definitions of the functions: 
- `vfma_f{32,64}` should always do a fused-multiply add. (`vfma` on ARM, `fmadd`/`fmla` on AArch64`)
- `vmla_f{32,64}` should always do a separate multiple and add. (`vmla` on ARM, `fmul` + `fadd` on AArch64).

The `vfma_f{32,64}` functions are also available on 32-bit ARM if the vfpv4 feature is enabled.

Have a look at this: https://godbolt.org/z/34Wo5n

PR #956, opened 2020-11-19
Add reamained vmax and vmin via auto-generated code

makotokato:


Amanieu:
CI seems to be failing

PR #976, opened 2021-01-02
use min_const_generics to avoid incomplete_features

RalfJung:
Cc @lcnr

### DIFF
--- a/crates/core_arch/src/lib.rs
+++ b/crates/core_arch/src/lib.rs
@@ -2,13 +2,11 @@
 #![allow(improper_ctypes_definitions)]
 #![allow(dead_code)]
 #![allow(unused_features)]
-#![allow(incomplete_features)]
 #![feature(
     asm,
     const_fn,
     const_fn_union,
     const_fn_transmute,
-    const_generics,
     custom_inner_attributes,
     link_llvm_intrinsics,
     platform_intrinsics,


RalfJung:
Fixing the build failure would require `cfg(bootstrap)` which, as I understand it, should not be used in stdarch... how is this usually handled?
```
error: the feature `min_const_generics` has been stable since 1.51.0 and no longer requires an attribute to enable
 --> crates/core_arch/src/lib.rs:9:5
  |
9 |     min_const_generics,
  |     ^^^^^^^^^^^^^^^^^^
  |
```

Amanieu:
We try to avoid `cfg(bootstrap)` here if possible, but it's fine to add it if it is really necessary.

In this particular case I think we can just wait until the `min_const_generics` stabilization hits beta so that the bootstrap compiler uses it.

RalfJung:
Then we better hope that nobody adds non-"min" uses of const_generics in the mean time.

pthariensflame:
> In this particular case I think we can just wait until the min_const_generics stabilization hits beta so that the bootstrap compiler uses it.

This has happened now, right?

Amanieu:
@RalfJung Can you rebase so CI passes?

RalfJung:
Okay, done that.

RalfJung:
There's a "Check Style" failure in code that I did not even touch.

Amanieu:
Just run rustfmt from the latest nightly.

RalfJung:
When I do `cargo +nightly fmt`, there is a *huge* diff.

I think it'd be better if someone with more knowledge of this code would fix CI here first (and possibly adjust the CI setup so that PRs are not rejected for unrelated failures...).

Amanieu:
Try rebasing now.

Amanieu:
Since min_const_generics is stable you can just remove the feature.

RalfJung:
Looks like CI passes now, thanks for the help. :)

PR #990, opened 2021-01-30
Clarify _mm256_shuffle_epi8 documentation

jethrogb:
The old documentation seemed to suggest that the low 4 bits of bytes in `b` would index into the 32-element array `a`. That's not quite right and you'd need 5 bits to index 32 elements.


jethrogb:
CI failure seems unrelated:
```
---- core_arch::x86::abm::assert__lzcnt_u32_lzcnt stdout ----
exit code: 1
LLVM ERROR: out of memory
fatal error: /Applications/Xcode_12.3.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/otool: fatal error in /Applications/Xcode_12.3.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/otool-classic
```

PR #992, opened 2021-02-03
fix _mm_loadu_si64

eduardosm:
The value should be loaded into the lowest 64 bits, not the highest.


eduardosm:
I guess this should also be backported to stable.

Amanieu:
I don't think a stable backport is necessary here unless someone is actively encountering this problem when trying to use stable Rust.

PR #994, opened 2021-02-05
Add 'volatile' to cpuid inline asm

thomcc:
`cpuid` is a serializing instruction, and is the only serializing instruction available as an intrinsic AFAIK. Various points of intels documentation suggests to use it as a way to fence other instructions — The typical case is in conjunction with rdtsc/rdtscp for measurement (though on sse2 processors you can also use `mfence`/`lfence` for this purpose), although this is not the only use case.

Without the "volatile" clobber, this intrinsic can't easily be used this way, since you have to fake the use of the result or LLVM will entirely remove it. This fails to reflect the behavior documentated for instruction, so it seems like a clear-cut bug worth fixing to me.

See these for further info if desired:

- https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-system-programming-manual-325384.pdf (in particular sections on serializing instructions)
- https://www.intel.com/content/dam/www/public/us/en/documents/white-papers/ia-32-ia-64-benchmark-code-execution-paper.pdf (which is dated and fails to mention lfence/mfence, but still applies for code in a "no sse2" fallback case)

TLDR: CPUID has side effects in the x86 memory model (including out of order execution) and this patch makes sure the intrinsic works for that.

### DIFF
--- a/crates/core_arch/src/x86/cpuid.rs
+++ b/crates/core_arch/src/x86/cpuid.rs
@@ -55,26 +55,33 @@ pub unsafe fn __cpuid_count(leaf: u32, sub_leaf: u32) -> CpuidResult {
     let ebx;
     let ecx;
     let edx;
+
     #[cfg(target_arch = "x86")]
     {
-        llvm_asm!("cpuid"
-                  : "={eax}"(eax), "={ebx}"(ebx), "={ecx}"(ecx), "={edx}"(edx)
-                  : "{eax}"(leaf), "{ecx}"(sub_leaf)
-                  : : "volatile");
+        asm!(
+            "cpuid",
+            inlateout("eax") leaf => eax,
+            lateout("ebx") ebx,
+            inlateout("ecx") sub_leaf => ecx,
+            lateout("edx") edx,
+            options(nostack, nomem, preserves_flags),
+        );
     }
     #[cfg(target_arch = "x86_64")]
     {
-        // x86-64 uses %rbx as the base register, so preserve it.
+        // x86-64 uses `rbx` as the base register, so preserve it.
         // This works around a bug in LLVM with ASAN enabled:
         // https://bugs.llvm.org/show_bug.cgi?id=17907
-        llvm_asm!(r#"
-                  mov %rbx, %rsi
-                  cpuid
-                  xchg %rbx, %rsi
-                  "#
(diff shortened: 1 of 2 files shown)


Amanieu:
I would prefer if that `llvm_asm!` was rewritten to use `asm!` instead. `llvm_asm!` is going to be deprecated soon and will eventually be removed from the compiler.

thomcc:
Sure, didn't want to do that off the bat since it might have hidden the
change maybe.

Not super sure about the clobbers/options (still new to `asm!`), but this
seems to work and does prevent the removal as I hoped.

crates/core_arch/src/x86/cpuid.rs
```diff
@@ -55,26 +55,33 @@ pub unsafe fn __cpuid_count(leaf: u32, sub_leaf: u32) -> CpuidResult {
...
-                  : "={eax}"(eax), "={ebx}"(ebx), "={ecx}"(ecx), "={edx}"(edx)
-                  : "{eax}"(leaf), "{ecx}"(sub_leaf)
-                  : :);
+        asm!(
+            "cpuid",
+            inlateout("eax") leaf => eax,
+            lateout("ebx") ebx,
+            inlateout("ecx") sub_leaf => ecx,
+            lateout("edx") edx,
+            options(nostack, nomem, preserves_flags),
```
thomcc, line 67: Is an instruction that (among other things) acts as a fence actually `nomem`? IDK. I added it under the logic of: It shouldn't have a \"memory\" clobber.

But I suspect that's not quite what it means.

sunfishcode:
Compilers don't consider out of order execution to be an observable side effect, so volatile won't be enough to prevent compilers from moving most instructions around it.

crates/core_arch/src/x86/cpuid.rs
Amanieu: You need to remove `nomem` for the `asm!` to act as a compiler fence. This will prevent memory accesses from being reordered across the `asm!`.

thomcc:
> Compilers don't consider out of order execution to be an observable side effect, so volatile won't be enough to prevent compilers from moving most instructions around it.

I'm aware. I don't think that should be handled by this intrinsic, as there are other ways of dealing with that already.

crates/core_arch/src/x86/cpuid.rs
thomcc: Yeah, thanks.

PR #996, opened 2021-02-06
Add aarch64-apple-darwin CI

makotokato:


Amanieu:
Thanks!

Incidentally I would appreciate if you could help figure out what went wrong with `x86_64-apple-darwin`. We had to disable it a few weeks ago because `otool` was running out of memory trying to disassemble a binary.

makotokato:
I am looking why failure.

> because otool was running out of memory trying to disassemble a bin

I guess that we should use norun environment at this time.

PR #998, opened 2021-02-09
Implement avx512bf16 intrinsics

kangshan1157:
Implement all avx512bf16 intrinsic APIs.

### DIFF
--- a/crates/core_arch/Cargo.toml
+++ b/crates/core_arch/Cargo.toml
@@ -25,6 +25,3 @@ maintenance = { status = "experimental" }
 [dev-dependencies]
 stdarch-test = { version = "0.*", path = "../stdarch-test" }
 std_detect = { version = "0.*", path = "../std_detect" }
-
-[dependencies]
-cfg-if = "0.1.10"
--- a/crates/core_arch/Cargo.toml
+++ b/crates/core_arch/Cargo.toml
@@ -25,3 +25,6 @@ maintenance = { status = "experimental" }
 [dev-dependencies]
 stdarch-test = { version = "0.*", path = "../stdarch-test" }
 std_detect = { version = "0.*", path = "../std_detect" }
+
+[dependencies]
+cfg-if = "0.1.10"
--- a/crates/core_arch/src/x86/avx512bf16.rs
+++ b/crates/core_arch/src/x86/avx512bf16.rs
@@ -1,0 +1,1122 @@
+//! [AVX512BF16 intrinsics].
+//!
+//! [AVX512BF16 intrinsics]: https://software.intel.com/sites/landingpage/IntrinsicsGuide/#expand=1769&avx512techs=AVX512_BF16
+
+use crate::{
+    core_arch::{simd::*, simd_llvm::*, x86::*},
+    mem::transmute,
+};
+
+#[cfg(test)]
+use stdarch_test::assert_instr;
+
+#[allow(improper_ctypes)]
+extern "C" {
+    #[link_name = "llvm.x86.avx512bf16.cvtne2ps2bf16.128"]
+    fn cvtne2ps2bf16(a: f32x4, b: f32x4) -> i16x8;
+    #[link_name = "llvm.x86.avx512bf16.cvtne2ps2bf16.256"]
+    fn cvtne2ps2bf16_256(a: f32x8, b: f32x8) -> i16x16;
+    #[link_name = "llvm.x86.avx512bf16.cvtne2ps2bf16.512"]
+    fn cvtne2ps2bf16_512(a: f32x16, b: f32x16) -> i16x32;
+    #[link_name = "llvm.x86.avx512bf16.cvtneps2bf16.256"]
+    fn cvtneps2bf16_256(a: f32x8) -> i16x8;
+    #[link_name = "llvm.x86.avx512bf16.cvtneps2bf16.512"]
+    fn cvtneps2bf16_512(a: f32x16) -> i16x16;
+    #[link_name = "llvm.x86.avx512bf16.dpbf16ps.128"]
+    fn dpbf16ps(a: f32x4, b: i32x4, c: i32x4) -> f32x4;
+    #[link_name = "llvm.x86.avx512bf16.dpbf16ps.256"]
+    fn dpbf16ps_256(a: f32x8, b: i32x8, c: i32x8) -> f32x8;
+    #[link_name = "llvm.x86.avx512bf16.dpbf16ps.512"]
+    fn dpbf16ps_512(a: f32x16, b: i32x16, c: i32x16) -> f32x16;
+}
+
+/// Convert packed single-precision (32-bit) floating-point elements in two 128-bit vectors 
+/// a and b to packed BF16 (16-bit) floating-point elements, and store the results in a 
+/// 128-bit wide vector.
+/// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#expand=1769,1651&avx512techs=AVX512_BF16&text=_mm_cvtne2ps_pbh)
+#[inline]
+#[target_feature(enable = "avx512bf16,avx512vl")]
+#[cfg_attr(test, assert_instr("vcvtne2ps2bf16"))]
+pub unsafe fn _mm_cvtne2ps_pbh(a: __m128, b: __m128) -> __m128bh {
+    transmute(cvtne2ps2bf16(a.as_f32x4(), b.as_f32x4()))
+}
+
+/// Convert packed single-precision (32-bit) floating-point elements in two vectors 
+/// a and b to packed BF16 (16-bit) floating-point elements, and store the results 
+/// in single vector dst using writemask k (elements are copied from src when the 
+/// corresponding mask bit is not set).
+/// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#expand=1769,1651&avx512techs=AVX512_BF16&text=_mm_mask_cvtne2ps_pbh)
+#[inline]
+#[target_feature(enable = "avx512bf16,avx512vl")]
+#[cfg_attr(test, assert_instr("vcvtne2ps2bf16"))]
+pub unsafe fn _mm_mask_cvtne2ps_pbh(src: __m128bh, k: __mmask8, a: __m128, b: __m128) -> __m128bh {
+    let cvt = _mm_cvtne2ps_pbh(a, b).as_u16x8();
+    transmute(simd_select_bitmask(k, cvt, src.as_u16x8()))
+}
+
+/// Convert packed single-precision (32-bit) floating-point elements in two vectors 
+/// a and b to packed BF16 (16-bit) floating-point elements, and store the results 
+/// in single vector dst using zeromask k (elements are zeroed out when the corresponding 
+/// mask bit is not set).
+/// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#expand=1769,1651&avx512techs=AVX512_BF16&text=_mm_maskz_cvtne2ps_pbh)
+#[inline]
+#[target_feature(enable = "avx512bf16,avx512vl")]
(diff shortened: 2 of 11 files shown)


crates/core_arch/src/x86/avx512bf16.rs
```diff
@@ -0,0 +1,1122 @@
...
+#[cfg_attr(test, assert_instr("vcvtneps2bf16"))]
+pub unsafe fn _mm_mask_cvtneps_pbh (src: __m128bh, k:__mmask8, a: __m128) -> __m128bh {
+    let mut result:__m128bh;
+    let mask: u32 = k as u32;
+    asm!(
+            "kmovd k1, edi",
+            "vcvtneps2bf16 {1} {{k1}}, {0}",
+            in(xmm_reg) a,
+            inout(xmm_reg) src => result,
+            in("edi") mask,
```
bjorn3, line 176: You need to use at&t syntax for now. Some older LLVM versions that are still supported by rustc don't correctly support intel syntax.

kangshan1157: Thanks @bjorn3 for your suggestion. I have updated it.

kangshan1157:
I tested and debugged my code with launching the command "TARGET=x86_64-unknown-linux-gnu ci/run.sh" on the cooperlake. Are any other tests required?

crates/core_arch/src/x86/avx512bf16.rs
```diff
@@ -0,0 +1,1124 @@
...
+#[target_feature(enable = "avx512bf16,avx512vl")]
+#[cfg_attr(test, assert_instr("vcvtneps2bf16"))]
+pub unsafe fn _mm_cvtneps_pbh (a: __m128) -> __m128bh {
+    let mut result:__m128bh;
+    asm!(
+            "vcvtneps2bf16 {0}, {1}",
+            in(xmm_reg) a,
+            lateout(xmm_reg) result,
+            options(att_syntax)
+        );
```
Amanieu, line 158: Why is this implemented using `asm!` instead of LLVM intrinsics?

Amanieu:
You should fix the CI errors.
- The `asm!` is causing the doc build to fail, but you can fix that by `#[cfg]`ing it out for non-x86.
- The verifier doesn't support the newly added types.

crates/core_arch/src/x86/avx512bf16.rs
```diff
@@ -150,12 +150,11 @@ pub unsafe fn _mm512_maskz_cvtne2ps_pbh (k: __mmask32, a: __m512, b: __m512) ->
...
-            "vcvtneps2bf16 {0}, {1}",
-            in(xmm_reg) a,
-            lateout(xmm_reg) result,
-            options(att_syntax)
-        );
+    llvm_asm!("vcvtneps2bf16 %xmm0, %xmm0"
+    : "={xmm0}"(result) 
+    : "{xmm0}"(a)
+    : "xmm0" 
+    :);
```
Amanieu, line 157: Please don't use `llvm_asm!`, it is going to be deprecated soon.

Instead just wrap the `asm!` in a `cfg_if` and provide a dummy `unreachable!()` implementation on other platforms.

kangshan1157: Thanks for your review. I have updated the code following your suggestions.

crates/core_arch/Cargo.toml
```diff
@@ -25,3 +25,6 @@ maintenance = { status = "experimental" }
 [dev-dependencies]
 stdarch-test = { version = "0.*", path = "../stdarch-test" }
 std_detect = { version = "0.*", path = "../std_detect" }
+
+[dependencies]
+cfg-if = "0.1.10"
```
bjorn3, line 30: core_arch can't have any dependencies as it is included in libcore, which everyone else depends on, including cfg-if.

kangshan1157: OK. Then is there any way to use cfg_if?

crates/core_arch/src/x86/avx512bf16.rs
```diff
@@ -0,0 +1,1140 @@
...
+                    in(xmm_reg) a,
+                    lateout(xmm_reg) result,
+                    options(att_syntax)
+                );
+            result
+        }
+        else {
+            unreachable!()
+        }
+    }
```
bjorn3, line 167:
```suggestion
    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
    {
        let mut result:__m128bh;
        asm!(
                "vcvtneps2bf16 {0}, {1}",
                in(xmm_reg) a,
                lateout(xmm_reg) result,
                options(att_syntax)
        );
        result
    }
    #[cfg(not(any(target_arch = "x86", target_arch = "x86_64")))]
    {
        unreachable!()
    }
```

This should work instead of cfg-if.

kangshan1157: Thank you very much. It works. :)

kangshan1157:
The CI tests for i686-unknown-linux-gnu are still failed. It seems that asm! doesn't support well on that target.  Is it expected or not?

Amanieu:
`xmm_reg` can't be used on `i586` because it doesn't support SSE (enabling it with `#[target_feature]` doesn't help, it needs to be globally enabled).

We really should be using LLVM intrinsics for this instead of inline asm. From what I understand, the main issue is that we can't represent a `i1x4` with Rust code. Support for this needs to be added to rustc.

Amanieu:
I would recommend asking the compiler team on [Zulip](https://rust-lang.zulipchat.com).

I think this would require defining a wrapper struct like this:
```rust
#[repr(simd_mask(4))]
struct i1x4(u8);
```

Then you would need to adjust the calling convention code to handle this type.

But I'm sure the compiler team will have a clearer idea of what would be required.

kangshan1157:
Thanks for all of your advices. You are right and we should call LLVM intrinsics to avoid the issues. Actually I asked the question in rust [github ](https://github.com/rust-lang/rust/issues/81552) several days ago. But they didn't give me the final answer. I will try to ask  it on  [Zulip](https://rust-lang.zulipchat.com/).
Moreover I will remove those asm implemented functions in this PR and recover them once i1x4 is supported by rustc.

kangshan1157:
> > Thanks for all of your advices. You are right and we should call LLVM intrinsics to avoid the issues. Actually I asked the question in rust [github ](https://github.com/rust-lang/rust/issues/81552) several days ago. But they didn't give me the final answer. I will try to ask it on [Zulip](https://rust-lang.zulipchat.com/).
> > Moreover I will remove those asm implemented functions in this PR and recover them once i1x4 is supported by rustc.
> > In some of my case to implement avx512f_avx512vl. LLVM use i1x2. I think rustc need to support i1 to solve those issues.

OK. I will also mention this to them.
Moreover CI tests for x86_64-unknown-linux-gnu and i586-unkonwn-linux-gnu failed because of the error "thread 'core_arch::x86::avx512bf16::assert__mm512_maskz_cvtneps_pbh_vcvtneps2bf16' panicked at 'failed to find instruction `vcvtneps2bf16` in the disassembly', crates/stdarch-test/src/lib.rs:148:9".  How can I resolve this? I tried them on my test machine and they work normally.

kangshan1157:
> > Moreover CI tests for x86_64-unknown-linux-gnu and i586-unkonwn-linux-gnu failed because of the error "thread 'core_arch::x86::avx512bf16::assert__mm512_maskz_cvtneps_pbh_vcvtneps2bf16' panicked at 'failed to find instruction `vcvtneps2bf16` in the disassembly', crates/stdarch-test/src/lib.rs:148:9". How can I resolve this? I tried them on my test machine and they work normally.
> 
> thread 'core_arch::x86::avx512bf16::assert__mm256_mask_dpbf16_ps_vdpbf16ps' panicked at 'failed to find instruction `vdpbf16ps` in the disassembly', crates/stdarch-test/src/lib.rs:148:9
> 
> When rustc tried to generate your code, it did not use vcvtne2ps2bf16 instruction.
> ---- core_arch::x86::avx512bf16::assert__mm512_cvtne2ps_pbh_vcvtne2ps2bf16 stdout ----
> disassembly for stdarch_test_shim__mm512_cvtne2ps_pbh_vcvtne2ps2bf16:
> 0: lea 0x1fdfe8(%rip),%rax # 2edcdf <anon.e2d3528a6cfe9ff3fa80755a098d527b.13.llvm.18001095552927935278+0x34>
> 1: lea 0x611312(%rip),%rcx # 701010 <_ZN12stdarch_test11_DONT_DEDUP17h0d83dd6e7a766ff3E>
> 2: mov %rax,(%rcx)
> 3: (bad)
> 4: rol $0xf,%ebx
> 5: (bad)
> 6: test %al,(%rax)
> 7: add %al,(%rax)
> 8:
> Usually, it is relating to your code error.
> Do you try to compile your code to assembly?

Yes, I tried it on my test machine. They are generated correctly. It seems that vcvtneps2bf16 is not supported by the compiler in the CI environment.

kangshan1157:
> I would recommend asking the compiler team on [Zulip](https://rust-lang.zulipchat.com).
> 
> I think this would require defining a wrapper struct like this:
> 
> ```rust
> #[repr(simd_mask(4))]
> struct i1x4(u8);
> ```
> 
> Then you would need to adjust the calling convention code to handle this type.
> 
> But I'm sure the compiler team will have a clearer idea of what would be required.

Hi 

> I would recommend asking the compiler team on [Zulip](https://rust-lang.zulipchat.com).
> 
> I think this would require defining a wrapper struct like this:
> 
> ```rust
> #[repr(simd_mask(4))]
> struct i1x4(u8);
> ```
> 
> Then you would need to adjust the calling convention code to handle this type.
> 
> But I'm sure the compiler team will have a clearer idea of what would be required.

Hi Amanieu, does bf16 related feature supported by the rust compiler in CI environment?  The CI tests failed because failed to find instruction vdpbf16ps in the disassembly', crates/stdarch-test/src/lib.rs:148:9

Amanieu:
The `(bad)` in the disassembly is actually because the disassembler is too old and doesn't recognize the instruction. LLVM is still generating the correct code.

Amanieu:
Maybe try upgrading the `Dockerfile` from `ubuntu:18.04` to `ubuntu:20.04`.

kangshan1157:
> Maybe try upgrading the `Dockerfile` from `ubuntu:18.04` to `ubuntu:20.04`.

How to upgrade it? Do I have the permission to do it?

Amanieu:
Just modify the `Dockerfile` in `ci/docker` for all platforms where CI fails.

Amanieu:
If the MSVC targets still give you trouble then you can disable the test for those targets in the source code by modifying the `cfg_attr` to exclude `target_env = "msvc"`.

kangshan1157:
> Just modify the `Dockerfile` in `ci/docker` for all platforms where CI fails.

Thank you very much for your help. 
For x86_64-unknown-linux-gnu-emulated target, as it doesn't allow skipping tests, those avx512bf16 related tests will fail. So I force to skip those tests through adding "SKIP_TESTS". Now all of the CI tests can pass.

If we want to run those bf16 tests in CI, another emulator needs to be set up. From what I learned, only cooper lake supports avx512bf16. If we start the emulator through "sde64 -cpx" to simulate cooper lake, then the other tests such as vnni will fail. So I suggest to start a separate emulator for bf16 tests only if CI doesn't allow to skip any tests.

Amanieu:
> If we start the emulator through "sde64 -cpx" to simulate cooper lake, then the other tests such as vnni will fail.

I thought cooper lake did support vnni?

In any case, I'm happy to merge this if you don't think you can get the emulator working.

kangshan1157:
> > If we start the emulator through "sde64 -cpx" to simulate cooper lake, then the other tests such as vnni will fail.
> 
> I thought cooper lake did support vnni?
> 
> In any case, I'm happy to merge this if you don't think you can get the emulator working.

Sorry for my typo, cooper lake doesn't support vbmi,vbmi2, so those tests will fail. Could you please help to merge this PR? I will add a new emulator target for bf16 tests later.

PR #999, opened 2021-02-09
Avx512 avx512vl

minybot:
permute_ps,pd: mm256,mm; permutevar_ps,pd: mm256,mm
permutex_epi64,pd: mm256
permutexvar_epi32,epi64,ps,pd: mm256
permutex2var_epi32,epi64,ps,pd: mm256,mm
i32gather_epi32,epi64,ps,pd: mm256,mm
i64gather_epi32,epi64,ps,pd: mm256,mm


crates/core_arch/src/x86/avx512f.rs
```diff
@@ -13441,6 +13587,54 @@ pub unsafe fn _mm512_mask_i32gather_ps(
...
+    scale: i32,
+) -> __m256 {
+    let slice = slice as *const f32;
+    macro_rules! call {
+        ($imm8:expr) => {
+            _mm256_i32gather_ps(slice, offsets, $imm8)
+        };
+    }
+    let r = constify_imm8_gather!(scale, call);
+    transmute(simd_select_bitmask(mask, r.as_f32x8(), src.as_f32x8()))
```
Amanieu, line 13611: `simd_select_bitmask` is incorrect in this case. We need to use the special LLVM intrinsic `llvm.x86.avx512.mask.gather3siv4.sf`.

The difference here is that the LLVM intrinsic will only access memory addresses which are not masked out. However your version will access memory addresses that are masked out, which could cause a crash if a masked value is an invalid offset.

minybot: > `simd_select_bitmask` is incorrect in this case. We need to use the special LLVM intrinsic `llvm.x86.avx512.mask.gather3siv4.sf`.
> 
> The difference here is that the LLVM intrinsic will only access memory addresses which are not masked out. However your version will access memory addresses that are masked out, which could cause a crash if a masked value is an invalid offset.

True, Thanks for your remind.
llvm.x86.avx512.mask.gather3siv4.sf will requires i1, so I need to remove those mask gather functions

Issue #989, opened 2021-01-30
Is there any way to call a llvm intrinisc API which contains type like llvm_v4i1_ty?


kangshan1157:
I'm studying rust and would like to contribute the intrinsic APIs implementation. 
During the investigation, I find some llvm intrinsic APIs may contain types like llvm_v4i1_ty (4 bits), llvm_v2i1_ty (2 bits) for which I fail to find the corresponding types in Rust.  What I learned Rust's type has 8 bits at least.
I'm a beginner of Rust and I have no idea how to solve such limitation. Does anyone have some suggestions?

Amanieu:
It isn't possible with special support in the compiler.

bjorn3:
Which LLVM intrinsics do you want to use?

kangshan1157:
I want to use this one.
def int_x86_avx512bf16_mask_cvtneps2bf16_128:
              Intrinsic<[llvm_v8i16_ty],
                        [llvm_v4f32_ty, llvm_v8i16_ty, llvm_v4i1_ty],
                        [IntrNoMem]>;

vertexclique:
All avx512 vp2intersect intrinsics are falling into this group.

PR #1000, opened 2021-02-10
AVX512-VP2INTERSECT implementations

vertexclique:
Currently, these implementations aren't finished. Link representations for `N x i1` for llvm needs to be incorporated into the compiler.
The best way to do the i1 linking is using the proc macro attribute on extern to make linking explicit to `i1` representations.

Note that `bool` is still `i8` representation from the rust's side with zero extend. That hasn't been solved for linking to llvm.

Relates to #989 and rust-lang/rust#81552

### DIFF
--- a/crates/core_arch/src/x86/avx512vp2intersect.rs
+++ b/crates/core_arch/src/x86/avx512vp2intersect.rs
@@ -1,0 +1,77 @@
+//! Vector Pair Intersection to a Pair of Mask Registers (VP2INTERSECT)
+//!
+//! The intrinsics here correspond to those in the `immintrin.h` C header.
+//!
+//! The reference is [Intel® Architecture
+//! Instruction Set Extensions and Future Features
+//! Programming Reference][intel64_ext_ref].
+//!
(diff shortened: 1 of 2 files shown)


vertexclique:
we have talked about with @Amanieu about it. This can wait until we land the `i1` linking in compiler.

Gab-Menezes:
Hey @Amanieu, first I would like to thank you for your work.

Any plans on continue development to support this instrinsics ? 
I tried using this instructions with inline asm, but without much success, since they "clobber" an undermined register (the clobbed kreg depends on the the first operand, since the masks are written in a even/odd pair). AFAIK to use this with inline asm the first operand needs to be fixed and can't be generated dynamically by the compiler, since if the compiler choose the register kX the register kX+1 will also be written, but you can't specify kX+1 as an output.
I would like to help if needed, but I have never contributed to core_arch, so if you can point me to some resources.

folkertdev:
Implemented in https://github.com/rust-lang/stdarch/pull/2081

PR #1004, opened 2021-02-14
Uncomment wasm32 bitmask instruction

Daniel-Liu-c0deb0t:
#874 introduced some `iNNxMM_bitmask` instructions that were supposed to be supported in LLVM 11, but they were commented out. Since Rust is already on LLVM 11 (https://blog.rust-lang.org/2020/10/08/Rust-1.47.html), this PR uncomments those instructions so they can be used.

### DIFF
--- a/ci/docker/wasm32-wasi/Dockerfile
+++ b/ci/docker/wasm32-wasi/Dockerfile
@@ -14,8 +14,7 @@ ENV PATH="$PATH:/root/.wasmer:/root/.wasmer/bin"
 ENV CARGO_TARGET_WASM32_WASI_RUNNER="wasmer run \
   --llvm \
   --enable-simd \
-  --mapdir .::/checkout/target/wasm32-wasi/release/deps \
-  --"
+  --mapdir .::/checkout/target/wasm32-wasi/release/deps"
 
 #RUN curl -L https://github.com/bytecodealliance/wasmtime/releases/download/v0.22.1/wasmtime-v0.22.1-x86_64-linux.tar.xz | tar xJf -
 #ENV PATH=$PATH:/wasmtime-v0.22.1-x86_64-linux
--- a/ci/docker/wasm32-wasi/Dockerfile
+++ b/ci/docker/wasm32-wasi/Dockerfile
@@ -9,6 +9,8 @@ RUN apt-get update -y && apt-get install -y --no-install-recommends \
 
 RUN curl https://get.wasmer.io -sSfL | sh
 
+RUN source /root/.wasmer/wasmer.sh
+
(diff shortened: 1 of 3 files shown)


Amanieu:
cc @alexcrichton: This looks good to me but I want to run it by you first since it's WASM-related.

Daniel-Liu-c0deb0t:
Saw some other commented out lines for the `abs` instructions and `i64x2_mul`, so I uncommented those as well.

Daniel-Liu-c0deb0t:
The tests for WASM don't seem to be running. Is there a problem with them?

Amanieu:
I believe it was commented it in the CI configuration because it was failing and never got re-enabled. You can try re-enabling it.

Daniel-Liu-c0deb0t:
I'll have to play around with enabling and updating the CI config and I'll likely open another PR for that.

Amanieu:
Can you rebase to see if it passes WASM CI?

alexcrichton (approved):
Looks great to me!

FWIW wasm simd and LLVM are still sort of constantly moving targets. The simd proposal for wasm is nearing the next phase of stabilization, however, so I hope that it will settle down in the near future. Once everything settles down again on the wasm side I plan to make another pass through this module so sync with the current state of the spec (which has renamed, added, and tweaked various instructions)

For now though looks great!

Daniel-Liu-c0deb0t:
The reason why I want `bitmask` right now is because I want to use it in a project I'm working on.

Running the CI results in weird cranelift errors that look like:
```
thread '<unnamed>' panicked at 'ABI argument v4 should be in a register', cranelift/codegen/src/regalloc/coloring.rs:409:21
```
Not sure what to do, as I'm not familiar with this.

Daniel-Liu-c0deb0t:
In the latest wasmtime dev release, I'm getting this error:
```
not implemented: Currently only SIMD instructions are mapped to their return type; the following instruction is not mapped: V128AnyTrue', cranelift/wasm/src/code_translator.rs:2688:14
```
@alexcrichton any idea what is going on?

alexcrichton:
The SIMD support in Wasmtime is still experimental and not fully fleshed out, so you're probably hitting a bug of a case that isn't implemented yet or just a bug in the implementation. If possible could you minimize the case and report it to rust-lang/rust?

Daniel-Liu-c0deb0t:
Another potential thing that should be fixed: missing target feature attributes for some WASM SIMD functions. Though this shouldn't be a problem once https://github.com/rust-lang/rust/issues/74320 is fixed.

alexcrichton:
I believe that wasm simd stuff is all up to date by this point, so I think this can be closed now?

PR #1005, opened 2021-02-14
Turn std_detect into a no_std crate

Amanieu:
The goal of this PR is to turn `std_detect` into a proper `no_std` crate so that we can avoid dealing with the mess of including it as a submodule of `std`.

### DIFF
--- a/crates/std_detect/src/detect/cache.rs
+++ b/crates/std_detect/src/detect/cache.rs
@@ -3,9 +3,9 @@
 
 #![allow(dead_code)] // not used on all platforms
 
-use crate::sync::atomic::Ordering;
+use core::sync::atomic::Ordering;
 
-use crate::sync::atomic::AtomicUsize;
+use core::sync::atomic::AtomicUsize;
 
 /// Sets the `bit` of `x`.
 #[inline]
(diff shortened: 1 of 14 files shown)


lu-zero (approved):
Seems good to me.

Amanieu:
It's not finished yet, I still need to remove dependencies on `std::*` and use `libc` instead.

lu-zero:
Poke me and I'll look into it again when ready :)

Amanieu:
@lu-zero Should be ready for review now.

crates/std_detect/src/detect/cache.rs
```diff
@@ -121,31 +121,11 @@ impl Cache {
     }
 }
 
-cfg_if::cfg_if! {
```
lu-zero, line 124: why do you want to remove it?

Amanieu: - It's never used.
- It's never tested.
- In fact the feature flag is even missing from `Cargo.toml`.

Amanieu: I changed `std_detect_env_override` to use libc instead of std.

lu-zero: Thank you a lot!

lu-zero (approved):
Looks fine to me, thank you :)

crates/std_detect/src/detect/cache.rs
bjorn3: Does this work on Windows? I did guess not.

PR #1016, opened 2021-02-26
Fix integer overflow in vadd aarch64 test cases

SparrowLii:
When I tested on my aarch64, I found that some of the vadd test cases would panic due to the overflow of x+y. Use wrapping_add (discard the overflow part) to avoid them. This will be consistent with the behavior of the instruction itself.


Amanieu:
Thanks!

Issue #248, opened 2017-12-22
Improve handling of immediate mode arguments


gnzlbg:
Many intrinsics take immediate more argument (that is, some arguments must be compile-time constants). 

We currently implement them by taking these arguments as run-time values, and "converting" them to compile-time values using a match statement with a match arm for each particular case:

```rust
fn foo(imm8: u8) {
    match imm8 {
        0 => foo_(0),
        1 => foo_(1), 
        ... 256 match arms later ...
        u8::MAX => foo_(u8::MAX),
    }
}
```

If a user pass them a compile-time value the optimizer can select the appropriate match arm and remove the rest of the code. 

Passing them a run-time value should generate an error, but with this approach the compiler silently accepts this code, and generates code for all match arms, potentially making binary sizes / compile-times explode.

Even when the user passes a compile-time constant to the immediate arguments, some intrinsics (AMD loves these) require match statements with >= 65k arms that destroy compile-times in all cases (from 2x to 3x increase in compile-times of the whole `stdsimd` test suite if one of this functions is used somewhere). The following intrinsics have such extremely large match arms: 

- `sse4a::_mm_extracti_si64`
- `sse4a::_mm_inserti_si64`
- `tbm::bextri`

This issue blocks completing SSE4a in #40 and fixing #26 . It was previously discussed in #190 but never got an issue of its own. 

Amongst the proposed solutions were to either implement these in the compilers or wait till const generics + extensions arrive (e.g. `foo(x: const i8)` to require `x` to be a compile-time constant).

gnzlbg:
@alexcrichton I have lifted this issue to the rust-lang/rust repo.

I think that handling immediate mode arguments properly is something we should attempt to fix before stabilization somehow. 

That is, either we get `foo(const imm8: T)`, or we should enumerate the intrinsics with immediate mode arguments and special-case them in trans somehow.

alexcrichton:
Thanks! I'm not sure I 100% agree we should block stabilization, but we can always sort out these details later :)

gnzlbg:
I guess my thoughts are more aligned with either:

* don't stabilizing these intrinsics at first (some of them are very useful, so I'd rather not do this)
* use the compiler to enforce that the arguments passed to these are compile-time constants somehow (not necessarily a language feature, but either an unstable attribute or just implementing these as compiler-intrinsics and exposing them via `stdsimd`).

[...]

PR #1018, opened 2021-02-26
Initial conversion to const generics

Amanieu:
This PR builds on top https://github.com/rust-lang/rust/pull/82447 to start transition intrinsics to use const generics. This is the first step towards solving #248.

- Updated the testing infrastructure to support const generics.
- Updated the intrinsic verification to support const generics.
- Ported `_mm_shuffle_ps` and `_mm_prefetch` to use const generics.

### DIFF
--- a/crates/core_arch/src/x86/sse.rs
+++ b/crates/core_arch/src/x86/sse.rs
@@ -1007,52 +1007,20 @@ pub const fn _MM_SHUFFLE(z: u32, y: u32, x: u32, w: u32) -> i32 {
 #[inline]
 #[target_feature(enable = "sse")]
 #[cfg_attr(test, assert_instr(shufps, mask = 3))]
-#[rustc_args_required_const(2)]
+#[rustc_legacy_const_generics(2)]
 #[stable(feature = "simd_x86", since = "1.27.0")]
-pub unsafe fn _mm_shuffle_ps(a: __m128, b: __m128, mask: i32) -> __m128 {
-    let mask = (mask & 0xFF) as u8;
-
-    macro_rules! shuffle_done {
-        ($x01:expr, $x23:expr, $x45:expr, $x67:expr) => {
-            simd_shuffle4(a, b, [$x01, $x23, $x45, $x67])
-        };
-    }
-    macro_rules! shuffle_x67 {
-        ($x01:expr, $x23:expr, $x45:expr) => {
-            match (mask >> 6) & 0b11 {
-                0b00 => shuffle_done!($x01, $x23, $x45, 4),
-                0b01 => shuffle_done!($x01, $x23, $x45, 5),
-                0b10 => shuffle_done!($x01, $x23, $x45, 6),
-                _ => shuffle_done!($x01, $x23, $x45, 7),
-            }
-        };
-    }
-    macro_rules! shuffle_x45 {
-        ($x01:expr, $x23:expr) => {
-            match (mask >> 4) & 0b11 {
-                0b00 => shuffle_x67!($x01, $x23, 4),
-                0b01 => shuffle_x67!($x01, $x23, 5),
-                0b10 => shuffle_x67!($x01, $x23, 6),
-                _ => shuffle_x67!($x01, $x23, 7),
-            }
-        };
-    }
-    macro_rules! shuffle_x23 {
-        ($x01:expr) => {
-            match (mask >> 2) & 0b11 {
-                0b00 => shuffle_x45!($x01, 0),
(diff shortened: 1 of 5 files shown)


Lokathor:
I can help reviewing this later today.

crates/core_arch/src/x86/sse.rs
```diff
@@ -1007,52 +1007,20 @@ pub const fn _MM_SHUFFLE(z: u32, y: u32, x: u32, w: u32) -> i32 {
...
-    match mask & 0b11 {
-        0b00 => shuffle_x23!(0),
-        0b01 => shuffle_x23!(1),
-        0b10 => shuffle_x23!(2),
-        _ => shuffle_x23!(3),
-    }
+#[rustc_legacy_const_generics(2)]
+#[stable(feature = "simd_x86", since = "1.27.0")]
+pub unsafe fn _mm_shuffle_ps<const mask: i32>(a: __m128, b: __m128) -> __m128 {
+    assert!(mask >= 0 && mask <= 255);
```
Lokathor, line 1013: Technically this is a breaking change compared to before, where you just discarded bits outside the lowest 8.

crates/core_arch/src/x86/sse.rs
```diff
@@ -1007,52 +1007,20 @@ pub const fn _MM_SHUFFLE(z: u32, y: u32, x: u32, w: u32) -> i32 {
...
+    simd_shuffle4(
+        a,
+        b,
+        [
+            mask as u32 & 0b11,
+            (mask as u32 >> 2) & 0b11,
+            ((mask as u32 >> 4) & 0b11) + 4,
+            ((mask as u32 >> 6) & 0b11) + 4,
```
Lokathor, line 1021: I'm not clear why two of these get a +4 on the end, and two do not.

Lokathor (commented):
Seems solid so far, though I don't really know about `assert-instr-macro` or `srdarch-verify` at all.

crates/core_arch/src/x86/sse.rs
Amanieu: This causes those elements to be taken from the second vector instead of the first.

crates/core_arch/src/x86/sse.rs
Amanieu: True, but C compilers already reject out-of-range immediates at compile-time. I would argue that this is a bug fix since we model the vendor's API more closely.

Lokathor: Checking the Intel Intrinsics Guide again, I see that it's supposed to be `unsigned int imm8`, so you're right.

Why did we ever accept an `i32` to begin with? I guess this should have been `u32` the whole time, and arguably `u8`.

Amanieu: #522

Lokathor: oof, i'm actually already subscribed to this one and i'd forgotten about it. okay then.

Actually since `_MM_SHUFFLE` is still nightly, maybe we should change the shuffle to correctly take `u32` and then change `_MM_SHUFFLE` (again) to make it also emit `u32` values.

But even then, if we're going to panic on values outside of `u8`, it's hard to justify not just taking a `u8` to begin with. I guess, are these conversions to const generics supposed to be otherwise completely non-breaking?

Amanieu: rust-lang/rust#82447 will rewrite an expression at the AST level to use const generics, but won't do anything to change the type of that expression. So I don't think we can change the type here, unless we decide to accept the breaking change anyways.

Lokathor: Well if the ship has sailed that's ultimately fine I guess.

crates/stdarch-verify/src/lib.rs
```diff
@@ -101,7 +109,23 @@ fn functions(input: TokenStream, dirs: &[&str]) -> TokenStream {
             } else {
                 quote! { None }
             };
-            let required_const = find_required_const(&f.attrs);
+
+            let required_const = find_required_const("rustc_args_required_const", &f.attrs);
```
lqd, line 113: Won't this ignore the ` #[rustc_legacy_const_generics]` attributes when the verifier checks the required consts, since this empty list will be passed to it [here](https://github.com/rust-lang/stdarch/blob/aba9587a3064024aa2ffc2f7d96ba576f11b267a/crates/stdarch-verify/src/lib.rs#L147) ?

PR #1020, opened 2021-02-27
Convert _mm_slli_epi{16,32,64} & _mm_srli_epi{16,32,64} to const generics

tmiasko:
* `_mm_slli_epi64`
* `_mm_slli_epi32`
* `_mm_slli_epi16`
* `_mm_srli_epi64`
* `_mm_srli_epi32`
* `_mm_srli_epi16`

### DIFF
--- a/crates/core_arch/src/x86/sse2.rs
+++ b/crates/core_arch/src/x86/sse2.rs
@@ -503,16 +503,10 @@ pub unsafe fn _mm_bsrli_si128(a: __m128i, imm8: i32) -> __m128i {
 #[inline]
 #[target_feature(enable = "sse2")]
 #[cfg_attr(test, assert_instr(psllw, imm8 = 7))]
-#[rustc_args_required_const(1)]
+#[rustc_legacy_const_generics(1)]
 #[stable(feature = "simd_x86", since = "1.27.0")]
-pub unsafe fn _mm_slli_epi16(a: __m128i, imm8: i32) -> __m128i {
-    let a = a.as_i16x8();
-    macro_rules! call {
-        ($imm8:expr) => {
-            transmute(pslliw(a, $imm8))
-        };
-    }
-    constify_imm8!(imm8, call)
+pub unsafe fn _mm_slli_epi16<const imm8: i32>(a: __m128i) -> __m128i {
+    transmute(pslliw(a.as_i16x8(), imm8))
 }
 
 /// Shifts packed 16-bit integers in `a` left by `count` while shifting in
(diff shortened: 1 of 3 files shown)


crates/core_arch/src/x86/sse2.rs
```diff
@@ -503,16 +503,10 @@ pub unsafe fn _mm_bsrli_si128(a: __m128i, imm8: i32) -> __m128i {
...
-pub unsafe fn _mm_slli_epi16(a: __m128i, imm8: i32) -> __m128i {
-    let a = a.as_i16x8();
-    macro_rules! call {
-        ($imm8:expr) => {
-            transmute(pslliw(a, $imm8))
-        };
-    }
-    constify_imm8!(imm8, call)
+pub unsafe fn _mm_slli_epi16<const imm8: i32>(a: __m128i) -> __m128i {
+    transmute(pslliw(a.as_i16x8(), imm8))
```
Amanieu, line 509: This should have a `static_assert!` to ensure the immediate is between 0 and 255.

Amanieu: In fact use `static_assert_imm8` that was added in #1021.

tmiasko: Added range checks using `static_assert_imm8`.

What do you use as a reference point, to make decision whether it should be hard error or not? The fact that it is documented as imm8? clang, gcc, and icc, as far as I can see accept arbitrary values there (in fact run-time ones as well).

Amanieu: We follow the intel intrinsics data which specifies that the argument is an 8-bit immediate. There is a separate intrinsic (`_mm_sll_epi32`) which allows run-time values.

```
<intrinsic tech="SSE2" vexEq="TRUE" name="_mm_slli_epi32">
	<type>Integer</type>
	<CPUID>SSE2</CPUID>
	<category>Shift</category>
	<return type="__m128i" varname="dst" etype="UI32"/>
	<parameter type="__m128i" varname="a" etype="UI32"/>
	<parameter type="int" varname="imm8" etype="IMM" immwidth="8"/>
	<description>Shift packed 32-bit integers in "a" left by "imm8" while shifting in zeros, and store the results in "dst".</description>
	<operation>
FOR j := 0 to 3
	i := j*32
	IF imm8[7:0] &gt; 31
		dst[i+31:i] := 0
	ELSE
		dst[i+31:i] := ZeroExtend32(a[i+31:i] &lt;&lt; imm8[7:0])
	FI
ENDFOR
	</operation>
	<instruction name="PSLLD" form="xmm, imm8" xed="PSLLD_XMMdq_IMMb"/>
	<header>emmintrin.h</header>
</intrinsic>
```

The presence of an appropriate `static_assert!` should probably be enforced by `stdarch-verify`, I'll look into implementing that.

PR #1021, opened 2021-02-27
Convert some SSE2 intrinsics to const generics

lqd:
This PR switches:
- `_mm_shuffle_epi32`
- `_mm_shufflehi_epi16`
- `_mm_shufflelo_epi16`
- `_mm_srai_epi16`
- `_mm_srai_epi32`

to const generics using `#[rustc_legacy_const_generics]`

### DIFF
--- a/crates/core_arch/src/macros.rs
+++ b/crates/core_arch/src/macros.rs
@@ -320,48 +320,6 @@ macro_rules! constify_imm5 {
     };
 }
 
-//immediate value: -16:15
-#[allow(unused)]
-macro_rules! constify_imm5 {
-    ($imm8:expr, $expand:ident) => {
-        #[allow(overflowing_literals)]
-        match ($imm8) & 0b1_1111 {
-            0 => $expand!(0),
-            1 => $expand!(1),
-            2 => $expand!(2),
-            3 => $expand!(3),
-            4 => $expand!(4),
-            5 => $expand!(5),
-            6 => $expand!(6),
-            7 => $expand!(7),
-            8 => $expand!(8),
-            9 => $expand!(9),
-            10 => $expand!(10),
-            11 => $expand!(11),
-            12 => $expand!(12),
-            13 => $expand!(13),
-            14 => $expand!(14),
-            15 => $expand!(15),
-            16 => $expand!(16),
-            17 => $expand!(17),
-            18 => $expand!(18),
-            19 => $expand!(19),
-            20 => $expand!(20),
-            21 => $expand!(21),
-            22 => $expand!(22),
-            23 => $expand!(23),
-            24 => $expand!(24),
-            25 => $expand!(25),
-            26 => $expand!(26),
-            27 => $expand!(27),
-            28 => $expand!(28),
-            29 => $expand!(29),
-            30 => $expand!(30),
-            _ => $expand!(31),
-        }
-    };
-}
-
 //immediate value: 0:16
 #[allow(unused)]
 macro_rules! constify_imm4 {
--- a/crates/core_arch/src/macros.rs
+++ b/crates/core_arch/src/macros.rs
@@ -1,5 +1,21 @@
 //! Utility macros.
 
+// Helper struct used to trigger const eval errors when a const generic immediate value is
+// out of range.
+pub(crate) struct ValidateConstImm8<const imm8: i32>();
+impl<const imm8: i32> ValidateConstImm8<imm8> {
+    pub(crate) const VALID: () = {
+        let _ = 1 / ((imm8 >= 0 && imm8 <= 255) as usize);
+    };
+}
+
+#[allow(unused)]
+macro_rules! static_assert_imm8 {
+    ($imm:ident) => {
+        let _ = $crate::core_arch::macros::ValidateConstImm8::<$imm>::VALID;
+    };
+}
+
 #[allow(unused)]
 macro_rules! static_assert {
     ($imm:ident : $ty:ty where $e:expr) => {
--- a/crates/core_arch/src/x86/sse2.rs
+++ b/crates/core_arch/src/x86/sse2.rs
@@ -1454,7 +1454,7 @@ pub unsafe fn _mm_movemask_epi8(a: __m128i) -> i32 {
 #[rustc_legacy_const_generics(1)]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub unsafe fn _mm_shuffle_epi32<const imm8: i32>(a: __m128i) -> __m128i {
-    static_assert!(imm8: i32 where imm8 >= 0 && imm8 <= 255);
+    static_assert_imm8!(imm8);
     let a = a.as_i32x4();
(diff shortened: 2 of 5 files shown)


crates/core_arch/src/x86/sse2.rs
```diff
@@ -1461,60 +1461,21 @@ pub unsafe fn _mm_movemask_epi8(a: __m128i) -> i32 {
...
-pub unsafe fn _mm_shuffle_epi32(a: __m128i, imm8: i32) -> __m128i {
-    // simd_shuffleX requires that its selector parameter be made up of
-    // constant values, but we can't enforce that here. In spirit, we need
-    // to write a `match` on all possible values of a byte, and for each value,
-    // hard-code the correct `simd_shuffleX` call using only constants. We
-    // then hope for LLVM to do the rest.
-    //
-    // Of course, that's... awful. So we try to use macros to do it for us.
-    let imm8 = (imm8 & 0xFF) as u8;
+pub unsafe fn _mm_shuffle_epi32<const imm8: i32>(a: __m128i) -> __m128i {
```
BurntSushi, line 1466: Is this a breaking change?

Lokathor: Yep, same thing happened in a previous PR. We're _kinda_ allowed to file it as "bugfix" though, because the api shouldn't accept a value outside 0..=255

lqd: yes, similarly to `_mm_shuffle_ps` discussed here https://github.com/rust-lang/stdarch/pull/1018#discussion_r584005413 which was acceptable as a bug fix if I understand correctly ?

Lokathor: Well, we could make it still accept any value and do `&0xFF` if we wanted to preserve the bug.

lqd: sure, whichever you all prefer. my understanding is that a crater run will ultimately help decide whether the bug should be preserved if need be ?

Lokathor: Yeah. I think it's fine for now, my comment was more to BurntSushi in case they were worried about breaks in practice.

lqd: (interestingly this time the immediate is not an `unsigned int` but an `int` in [the intrinsics guide](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_shuffle_epi32&expand=5144) 🤔 )

Lokathor: yeah it's just wrong sometimes, or it's right and the headers were specified wrong to begin with.

BurntSushi: But this isn't just about what the immediate value is right? Doesn't this break all uses since it changes the number of parameters to the function?

tmiasko: The old form is rewritten to const generics, see rust-lang/rust#82447 for details.

Lokathor: Oh, that, I believe that the new attribute it's getting will let rustc magically rewrite calls using the old style to be the new style instead.

Lokathor: Though because of the limits of const generics, some old uses might be broken by this rewriting.

there's a thread about this on Zulip in the consy eval stream. though I don't have the link at hand on mobile, I'll ping you.

BurntSushi: Ah I see. If this is only going to break things where imm is not a constant, then I'm okay with that and agree it is a bug fix. I am only concerned about function params being changed.

I guess i would also be concerned with how this renders in rustdoc, but whatever the case, that seems fixable.

lqd: Some initial discussion about this feature can be also be seen in [this other zulip link](https://rust-lang.zulipchat.com/#narrow/stream/268952-edition-2021/topic/Replacing.20rustc_args_required_const.20with.20const.20generics), leading to the rustc PR linked above.

crates/core_arch/src/x86/sse2.rs
```diff
@@ -594,16 +594,10 @@ pub unsafe fn _mm_sll_epi64(a: __m128i, count: __m128i) -> __m128i {
...
-pub unsafe fn _mm_srai_epi16(a: __m128i, imm8: i32) -> __m128i {
-    let a = a.as_i16x8();
-    macro_rules! call {
-        ($imm8:expr) => {
-            transmute(psraiw(a, $imm8))
-        };
-    }
-    constify_imm8!(imm8, call)
+pub unsafe fn _mm_srai_epi16<const imm8: i32>(a: __m128i) -> __m128i {
+    transmute(psraiw(a.as_i16x8(), imm8))
```
Amanieu, line 600: This should have a `static_assert!` to ensure the immediate is between 0 and 255.

crates/core_arch/src/x86/sse2.rs
Amanieu: Note that `#[rustc_args_required_const]` already required the value to be a constant.

BurntSushi: Does that mean this change cannot break anything?

crates/core_arch/src/x86/sse2.rs
lqd: Looks like `static_assert!(imm8: i32 where imm8 >= 0 && imm8 <= 255);` is going to be used a quite a bit, we could have a `static_assert_imm8!` or in general static asserts analogues to the various `constify_imm*` ?

Amanieu: That sounds like a good idea. Also we should use a single shared `Validate` struct for those to reduce the MIR size (rather than instantiating a new one in each function).

crates/core_arch/src/x86/sse2.rs
Amanieu: There is one specific situation where it can break: you can't use an expression derived from generic arguments in a const item (e.g. `const` or a const generic parameter). But `#[rustc_args_required_const]` accepted this.

Example
```rust
fn foo<const X: i32>(a: __m128) {
	// This was previously accepted but will now fail to compile.
	stdarch::_mm_shuffle_ps(a, a, X + 1);

	// This still works.
	stdarch::_mm_shuffle_ps(a, a, X);
}
```

crates/core_arch/src/macros.rs
```diff
@@ -1,5 +1,22 @@
 //! Utility macros.
 
+// Helper struct used to trigger const eval errors when a const generic immediate value is
+// out of range.
+#[doc(hidden)]
```
Amanieu, line 5: `#[doc(hidden)]` is not needed when this is already private to the crate.

lqd: Right, I've removed it now

PR #1024, opened 2021-02-28
Convert _mm_slli_si128 & _mm_srli_si128 to const generics

tmiasko:


crates/core_arch/src/x86/sse2.rs
```diff
@@ -412,65 +412,51 @@ pub unsafe fn _mm_subs_epu16(a: __m128i, b: __m128i) -> __m128i {
...
+    struct Mask<const shift: i32, const i: u32>();
+    impl<const shift: i32, const i: u32> Mask<shift, i> {
+        const MASK: u32 = {
+            if (shift as u32) > 15 {
+                i
+            } else {
+                16 - (shift as u32) + i
+            }
         };
     }
```
Amanieu, line 436: Does using a `const fn` work here?

tmiasko: Yes, `const fn` works here.

PR #1025, opened 2021-02-28
Add stricter validation of const arguments on x86 intrinsics

Amanieu:


Lokathor:
What exactly does the imm8 = 7 part mean?

Amanieu:
It's just an arbitrary value for the codegen test to check that the correct instruction is emitted.

Lokathor:
ah i thought it was somehow indicating 7 bits used.

Lokathor (approved):

PR #1026, opened 2021-02-28
Convert `x86/sse41.rs` intrinsics to const generics

lqd:
This PR updates the x86 SSE 4.1 intrinsics to use const generics.

Of note, some of these made an effort to test the masking/clamping of the constify immediates macros, testing with both in-range and out-of-range values in:
- `_mm_extract_ps`: [_mm_extract_ps(a, 5)](https://github.com/rust-lang/stdarch/blob/dab4581ced4f0dc82ef9ec67b4da19da707d88fb/crates/core_arch/src/x86/sse41.rs#L1301) on a 2-bit immediate
- `_mm_extract_epi8`: [_mm_extract_epi8(a, 19)](https://github.com/rust-lang/stdarch/blob/dab4581ced4f0dc82ef9ec67b4da19da707d88fb/crates/core_arch/src/x86/sse41.rs#L1313) on a 4-bit immediate
- `_mm_extract_epi32`: [_mm_extract_epi32(a, 5)](https://github.com/rust-lang/stdarch/blob/dab4581ced4f0dc82ef9ec67b4da19da707d88fb/crates/core_arch/src/x86/sse41.rs#L1323) on a 2-bit immediate
- `_mm_insert_epi8`: [_mm_insert_epi8(a, 32, 17)](https://github.com/rust-lang/stdarch/blob/dab4581ced4f0dc82ef9ec67b4da19da707d88fb/crates/core_arch/src/x86/sse41.rs#L1342) on a 4-bit immediate
- `_mm_insert_epi32`: [_mm_insert_epi32(a, 32, 5)](https://github.com/rust-lang/stdarch/blob/dab4581ced4f0dc82ef9ec67b4da19da707d88fb/crates/core_arch/src/x86/sse41.rs#L1352) on a 2-bit immediate

As discussed with Amanieu on zulip, I've made such out-of-range immediates errors with static assertions, and changed the tests in these cases. Opinions are welcome.

I've noticed our previous PRs were squashed when merged, so I didn't make a commit per intrinsic this time. Of course, I can do so if it would be helpful for the review ?

### DIFF
--- a/crates/core_arch/src/macros.rs
+++ b/crates/core_arch/src/macros.rs
@@ -1,18 +1,39 @@
 //! Utility macros.
 
-// Helper struct used to trigger const eval errors when a const generic immediate value is
-// out of range.
-pub(crate) struct ValidateConstImm8<const imm8: i32>();
-impl<const imm8: i32> ValidateConstImm8<imm8> {
+// Helper struct used to trigger const eval errors when the const generic immediate value `imm` is
+// out of `bits`-bit range.
+pub(crate) struct ValidateConstImm<const imm: i32, const bits: i32>;
+impl<const imm: i32, const bits: i32> ValidateConstImm<imm, bits> {
     pub(crate) const VALID: () = {
-        let _ = 1 / ((imm8 >= 0 && imm8 <= 255) as usize);
+        let _ = 1 / ((imm >= 0 && imm < (1 << bits)) as usize);
+    };
+}
+
+#[allow(unused)]
+macro_rules! static_assert_imm2 {
+    ($imm:ident) => {
+        let _ = $crate::core_arch::macros::ValidateConstImm::<$imm, 2>::VALID;
+    };
+}
+
+#[allow(unused)]
+macro_rules! static_assert_imm3 {
+    ($imm:ident) => {
+        let _ = $crate::core_arch::macros::ValidateConstImm::<$imm, 3>::VALID;
+    };
+}
+
+#[allow(unused)]
+macro_rules! static_assert_imm4 {
+    ($imm:ident) => {
+        let _ = $crate::core_arch::macros::ValidateConstImm::<$imm, 4>::VALID;
     };
 }
 
 #[allow(unused)]
 macro_rules! static_assert_imm8 {
     ($imm:ident) => {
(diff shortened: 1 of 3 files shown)


crates/core_arch/src/macros.rs
```diff
@@ -1,8 +1,8 @@
 //! Utility macros.
 
 // Helper struct used to trigger const eval errors when a const generic immediate value is
-// out of range.
-pub(crate) struct ValidateConstImm8<const imm8: i32>();
+// out of 8-bit range.
+pub(crate) struct ValidateConstImm8<const imm8: i32>;
```
Amanieu, line 5: It might be worth making this generic over the bit width:

```rust
struct ValidateConstImm8<const imm i32, const bits: i32>
```

The check would then be:
```rust
let _ = 1 / ((imm >= 0 && imm < (1 << bits)) as usize);
```

Then we can simply define macros referring to a single instance of this struct. These macros should ideally be in the generic `macros.rs` instead of the arch-specific one.

crates/core_arch/src/x86/macros.rs
```diff
@@ -1,5 +1,51 @@
...
+// out of 2-bit range.
+pub(crate) struct ValidateConstImm2<const imm2: i32>;
+impl<const imm2: i32> ValidateConstImm2<imm2> {
+    pub(crate) const VALID: () = {
+        let _ = 1 / ((imm2 >= 0 && imm2 <= 3) as usize);
+    };
+}
+
+#[allow(unused)]
+macro_rules! static_assert_imm2 {
```
Amanieu, line 12: Move these to the generic `macros.rs` instead of the x86-specific one.

lqd: I wasn't exactly sure why some `constify_imm*` are duplicated here and there, so I added these new ones at the same places as the old ones. I'll combine this with the suggestion above.

crates/core_arch/src/macros.rs
lqd: good idea, that should be enough for all other `constify_imm`s analogues. (the other design has the advantage of encoding the immediate width in its name, and the expression shows up in the const error message, but a single struct will be better)

Issue #1022, opened 2021-02-27
Remove all uses of `#[rustc_args_required_const]`


Amanieu:
To solve #248 we are moving away from `#[rustc_args_required_const]` and switching to using the recently stabilized const generics instead. The goal is to massively reduce the binary size of `libcore.rlib` which is currently bloated due to the massive `match` statements needed to properly handle constant arguments for intrinsics.

## Instructions
1. Move constant arguments to const generics.

2. Change the `#[rustc_args_required_const(N)]` attribute to `#[rustc_legacy_const_generics(N)]`. The `N` arguments to the attribute remains the same. This attribute allows legacy code to continue calling this function by passing constants as a parameter. The legacy support will likely be deprecated in Rust 2021.

3. Use the `static_assert!` macro to perform bounds checking on the value. This is necessary because passing invalid value to intrinsics tends to trigger assert failures in LLVM if they reach codegen. Simply using `assert!` or `match` is not enough since they don't prevent the intrinsic from reaching codegen.

4. Eliminate unneeded `constify_*`/`shuffle` macros and pass the constant directly to the underlying LLVM intrinsic.

5. Update tests to call the function using const generics. Note that `#[rustc_legacy_const_generics]` does not work for functions defined in the same crate.

### Before

```rust
#[inline]
#[target_feature(enable = "sse")]
#[cfg_attr(test, assert_instr(shufps, mask = 3))]
#[rustc_args_required_const(2)]
#[stable(feature = "simd_x86", since = "1.27.0")]
pub unsafe fn _mm_shuffle_ps(a: __m128, b: __m128, mask: i32) -> __m128 {
    let mask = (mask & 0xFF) as u8;

    macro_rules! shuffle_done {
        ($x01:expr, $x23:expr, $x45:expr, $x67:expr) => {
            simd_shuffle4(a, b, [$x01, $x23, $x45, $x67])
        };
    }
    macro_rules! shuffle_x67 {
        ($x01:expr, $x23:expr, $x45:expr) => {
            match (mask >> 6) & 0b11 {
                0b00 => shuffle_done!($x01, $x23, $x45, 4),
                0b01 => shuffle_done!($x01, $x23, $x45, 5),
                0b10 => shuffle_done!($x01, $x23, $x45, 6),
                _ => shuffle_done!($x01, $x23, $x45, 7),
            }
        };
    }
    macro_rules! shuffle_x45 {
        ($x01:expr, $x23:expr) => {
[...]

PR #1028, opened 2021-03-01
1/3 mips: Convert rustc_args_required_const(0) functions to const generics

tesuji:
I intended to work on mips in many small PRs for easier reviewing.

cc #1022

### DIFF
--- a/crates/core_arch/src/mips/msa/macros.rs
+++ b/crates/core_arch/src/mips/msa/macros.rs
@@ -1,5 +1,11 @@
 //! Utility macros.
 
+macro_rules! static_assert_imm_s10 {
+    ($imm:ident) => {
+        let _ = $crate::core_arch::macros::ValidateConstImm::<$imm, -512, 511>::VALID;
+    };
+}
+
 //immediate value: -4096:4088
 macro_rules! constify_imm_s13 {
     ($imm_s13:expr, $expand:ident) => {
--- a/crates/core_arch/src/mips/msa.rs
+++ b/crates/core_arch/src/mips/msa.rs
@@ -2870,14 +2870,10 @@ pub unsafe fn __msa_ceqi_d(a: v2i64, imm_s5: i32) -> v2i64 {
 #[inline]
 #[target_feature(enable = "msa")]
 #[cfg_attr(test, assert_instr(cfcmsa, imm5 = 0b11111))]
-#[rustc_args_required_const(0)]
-pub unsafe fn __msa_cfcmsa(imm5: i32) -> i32 {
-    macro_rules! call {
-        ($imm5:expr) => {
-            msa_cfcmsa($imm5)
-        };
-    }
-    constify_imm5!(imm5, call)
+#[rustc_legacy_const_generics(0)]
+pub unsafe fn __msa_cfcmsa<const IMM5: i32>() -> i32 {
+    static_assert_imm5!(IMM5);
+    msa_cfcmsa(IMM5)
 }
 
 /// Vector Compare Signed Less Than or Equal
@@ -3578,14 +3574,10 @@ pub unsafe fn __msa_copy_u_d(a: v2i64, imm1: i32) -> u64 {
 #[inline]
 #[target_feature(enable = "msa")]
 #[cfg_attr(test, assert_instr(ctcmsa, imm1 = 0b1))]
-#[rustc_args_required_const(0)]
-pub unsafe fn __msa_ctcmsa(imm5: i32, a: i32) -> () {
-    macro_rules! call {
-        ($imm5:expr) => {
-            msa_ctcmsa($imm5, a)
-        };
-    }
-    constify_imm5!(imm5, call)
+#[rustc_legacy_const_generics(0)]
+pub unsafe fn __msa_ctcmsa<const IMM5: i32>(a: i32) -> () {
+    static_assert_imm5!(IMM5);
+    msa_ctcmsa(IMM5, a)
 }
 
 /// Vector Signed Divide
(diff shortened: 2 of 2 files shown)


crates/core_arch/src/macros.rs
```diff
@@ -9,34 +9,98 @@ impl<const imm: i32, const bits: i32> ValidateConstImm<imm, bits> {
...
 }
 
-#[allow(unused)]
+#[allow(non_camel_case_types)]
+pub(crate) struct ValidateConstImm_S<const imm_s5: i32, const min: i32, const max: i32>();
+impl<const imm: i32, const min: i32, const max: i32> ValidateConstImm_S<imm, min, max> {
+    pub(crate) const VALID: () = {
+        let _ = 1 / ((imm >= min && imm <= max) as usize);
+    };
+}
```
Amanieu, line 18: I would prefer modifying the main `ValidateConstImm` to support both a min and max value.

crates/core_arch/src/macros.rs
```diff
@@ -9,34 +9,98 @@ impl<const imm: i32, const bits: i32> ValidateConstImm<imm, bits> {
...
+macro_rules! static_assert_imm_s10 {
+    ($imm:ident) => {
+        let _ = $crate::core_arch::macros::ValidateConstImm_S::<$imm, -512, 511>::VALID;
+    };
+}
+
+#[allow(unused_macros)]
+macro_rules! static_assert_imm_s11 {
+    ($imm:ident) => {
+        let _ = $crate::core_arch::macros::ValidateConstImm_S::<$imm, -1024, 1022>::VALID;
```
Amanieu, line 86: I would expect this to be have 1023 as the upper bound here. Is this a MIPS limitation? Is the value required to be a multiple of 2?

Same with the others below.

Amanieu: If these are MIPS-specific then they should be moved to the MIPS `macro.rs`.

tesuji: Reasonable. Will do!

Amanieu:
Can you change your code to use uppercase names for const generics? I am changing this for existing code in #1035.

crates/core_arch/src/macros.rs
```diff
@@ -2,38 +2,59 @@
...
+    ($imm:ident) => {
+        let _ = $crate::core_arch::macros::ValidateConstImm::<$imm, 0, {1 << 1}>::VALID;
+    };
+}
+
+#[allow(unused_macros)]
 macro_rules! static_assert_imm2 {
     ($imm:ident) => {
-        let _ = $crate::core_arch::macros::ValidateConstImm::<$imm, 2>::VALID;
+        let _ = $crate::core_arch::macros::ValidateConstImm::<$imm, 0, {1 << 3}>::VALID;
```
Amanieu, line 22:
```suggestion
        let _ = $crate::core_arch::macros::ValidateConstImm::<$imm, 0, {1 << 2}>::VALID;
```

crates/core_arch/src/mips/msa/macros.rs
```diff
@@ -1,5 +1,35 @@
 //! Utility macros.
 
+macro_rules! static_assert_imm_s5 {
+    ($imm:ident) => {
+        let _ = $crate::core_arch::macros::ValidateConstImm_S::<$imm, -16, 16>::VALID;
```
Amanieu, line 5:
```suggestion
        let _ = $crate::core_arch::macros::ValidateConstImm::<$imm, -16, 16>::VALID;
```

crates/core_arch/src/macros.rs
```diff
@@ -2,38 +2,59 @@
 
 // Helper struct used to trigger const eval errors when the const generic immediate value `imm` is
 // out of `bits`-bit range.
-pub(crate) struct ValidateConstImm<const IMM: i32, const BITS: i32>;
-impl<const IMM: i32, const BITS: i32> ValidateConstImm<IMM, BITS> {
+pub(crate) struct ValidateConstImm<const IMM: i32, const MIN: i32, const MAX: i32>;
+impl<const IMM: i32, const MIN: i32, const MAX: i32> ValidateConstImm<IMM, MIN, MAX> {
     pub(crate) const VALID: () = {
-        let _ = 1 / ((IMM >= 0 && IMM < (1 << BITS)) as usize);
+        let _ = 1 / ((IMM >= MIN && IMM < MAX) as usize);
```
Amanieu, line 8:
```suggestion
        let _ = 1 / ((IMM >= MIN && IMM <= MAX) as usize);
```

I think it will be clearer if `MAX` refers to the inclusive upper bound of the value.

crates/core_arch/src/mips/msa/macros.rs
```diff
@@ -1,5 +1,35 @@
...
+    };
+}
+
+macro_rules! static_assert_imm_s12 {
+    ($imm:ident) => {
+        let _ = $crate::core_arch::macros::ValidateConstImm::<$imm, -2048, 2045>::VALID;
+    };
+}
+
+macro_rules! static_assert_imm_s13 {
```
Amanieu, line 27: For the intrinsics using these there are 2 conditions:
- Range is `-4096` to `4095`.
- Immediate is a multiple of 8.

It may be better to split these 2 conditions into separate `static_assert`s.

tesuji: This macro `*_s{5,11,12,13}` are unused in this PR, but they will be used in patch 2 and 3.
I will add `static_assert_multiply_of` macro for asserting multiple issue.

Amanieu: Just use a normal `static_assert!` for that. The specific `static_assert_imm*` macros are mainly used to avoid duplication.

tesuji: Moved *imm_s to subsequent patches.

Amanieu:
CI should be fixed, you need to rebase.

tesuji:
I'm away from computer in a few weeks. If anybody wants to take off this PR, feels free.

PR #1030, opened 2021-03-01
blocked: mips: Convert *const(1) to const generics

tesuji:
Blocked on  #1028


tesuji:
Sorry, intended to open against my own fork.

PR #1037, opened 2021-03-01
Convert shuffle_ps and shuffle_pd to const generics

minybot:
shuffle_ps,pd
roundscale_ps,pd
fixupimm_ps,pd
ternarylogic_epi32,epi64

### DIFF
--- a/crates/core_arch/src/macros.rs
+++ b/crates/core_arch/src/macros.rs
@@ -45,6 +45,13 @@ macro_rules! static_assert_imm8 {
 }
 
 #[allow(unused)]
+macro_rules! static_assert_rounding {
+    ($imm:ident) => {
+        let _ = $imm == 4 || $imm == 8 || $imm == 9 || $imm == 10 || $imm == 11;
+    };
+}
+
+#[allow(unused)]
 macro_rules! static_assert {
     ($imm:ident : $ty:ty where $e:expr) => {
         struct Validate<const $imm: $ty>();
--- a/crates/core_arch/src/macros.rs
+++ b/crates/core_arch/src/macros.rs
@@ -45,13 +45,6 @@ macro_rules! static_assert_imm8 {
 }
 
 #[allow(unused)]
-macro_rules! static_assert_rounding {
-    ($imm:ident) => {
-        let _ = $imm == 4 || $imm == 8 || $imm == 9 || $imm == 10 || $imm == 11;
-    };
-}
-
-#[allow(unused)]
 macro_rules! static_assert {
     ($imm:ident : $ty:ty where $e:expr) => {
         struct Validate<const $imm: $ty>();
--- a/crates/core_arch/src/x86/avx512f.rs
+++ b/crates/core_arch/src/x86/avx512f.rs
@@ -6863,17 +6863,13 @@ pub unsafe fn _mm_maskz_getmant_pd(
 /// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm512_add_round_ps&expand=145)
 #[inline]
 #[target_feature(enable = "avx512f")]
-#[cfg_attr(test, assert_instr(vaddps, rounding = 8))]
-#[rustc_args_required_const(2)]
-pub unsafe fn _mm512_add_round_ps(a: __m512, b: __m512, rounding: i32) -> __m512 {
+#[cfg_attr(test, assert_instr(vaddps, IMM4 = 8))]
+#[rustc_legacy_const_generics(2)]
+pub unsafe fn _mm512_add_round_ps<const IMM4: i32>(a: __m512, b: __m512) -> __m512 {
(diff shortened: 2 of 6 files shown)


crates/core_arch/src/x86/sse2.rs
```diff
@@ -4852,7 +4848,8 @@ mod tests {
         let a = _mm_setr_pd(1., 2.);
         let b = _mm_setr_pd(3., 4.);
         let expected = _mm_setr_pd(1., 3.);
-        let r = _mm_shuffle_pd(a, b, 0);
+        //let r = _mm_shuffle_pd(a, b, 0);
```
Amanieu, line 4851: Please don't leave commented out code.

crates/core_arch/src/x86/sse2.rs
```diff
@@ -2653,21 +2653,17 @@ pub unsafe fn _mm_loadu_pd(mem_addr: *const f64) -> __m128d {
 #[target_feature(enable = "sse2")]
 #[cfg_attr(
     all(test, any(not(target_os = "windows"), target_arch = "x86")),
-    assert_instr(shufps, imm8 = 1)
+    cfg_attr(test, assert_instr(shufps, MASK = 2)) // FIXME shufpd expected
 )]
 #[cfg_attr(
     all(test, all(target_os = "windows", target_arch = "x86_64")),
-    assert_instr(shufpd, imm8 = 1)
+    cfg_attr(test, assert_instr(movaps, MASK = 2)) // FIXME shufpd expected
```
Amanieu, line 2660: Check which instruction is actually generated in the assembly output for the shuffle. Not `movaps` which has nothing to do with the shuffle.

Amanieu: Try using a different `imm8` value so it generates a shuffle instruction.

minybot: Yes, it works now.

crates/core_arch/src/x86/sse2.rs
Amanieu: > Please don't leave commented out code.

minybot: Sorry, I will correct it now.

crates/core_arch/src/x86/avx512f.rs
```diff
@@ -6863,17 +6863,13 @@ pub unsafe fn _mm_maskz_getmant_pd(
 /// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm512_add_round_ps&expand=145)
 #[inline]
 #[target_feature(enable = "avx512f")]
-#[cfg_attr(test, assert_instr(vaddps, rounding = 8))]
-#[rustc_args_required_const(2)]
-pub unsafe fn _mm512_add_round_ps(a: __m512, b: __m512, rounding: i32) -> __m512 {
+#[cfg_attr(test, assert_instr(vaddps, IMM4 = 8))]
+#[rustc_legacy_const_generics(2)]
+pub unsafe fn _mm512_add_round_ps<const IMM4: i32>(a: __m512, b: __m512) -> __m512 {
```
Amanieu, line 6868: There's no reason to change the change from `rounding` to `IMM4`. Using `ROUNDING` is better.

minybot: OK. I will change it.
static_assert!(IMM4: i32 where IMM4 == 4 || IMM4 == 8 || IMM4 == 9 || IMM4 == 10 || IMM4 == 11); is this OK?

Amanieu: It might be better to create a specific macro just for this condition since it is used quite often.

minybot: Ok. I will create a special one.
Btw. Can the const generics handle 2 constants in a function now?
For example.
#[rustc_args_required_const(2, 3)]
_mm_maskz_getmant_pd

crates/core_arch/src/macros.rs
```diff
@@ -44,6 +44,13 @@ macro_rules! static_assert_imm8 {
     };
 }
 
+#[allow(unused)]
+macro_rules! static_assert_rounding {
+    ($imm:ident) => {
+        let _ = $imm == 4 || $imm == 8 || $imm == 9 || $imm == 10 || $imm == 11;
```
Amanieu, line 50: This doesn't actually assert anything. You need to either call `static_assert` within this macro or create a custom `ValidateConst` struct like the other macros.

Also this is x86-specific, please move it to `x86/macros.rs`.

crates/core_arch/src/x86/avx512f.rs
Amanieu: Yes, you can have multiple constant arguments. You need to use 2 separate `static_assert` for them.

PR #1038, opened 2021-03-01
More trivial x86 intrinsics conversions to const generics

lqd:
This PR converts these trivial x86 intrinsics to const generics.

#### x86/sse2.rs:
- `_mm_extract_epi16`
- `_mm_insert_epi16`

#### x86/avx.rs:
- `_mm256_round_pd`
- `_mm256_round_ps`
- `_mm256_dp_ps`
- `_mm256_permute2f128_ps`
- `_mm256_permute2f128_pd`
- `_mm256_permute2f128_si256`
- `_mm256_insert_epi16`
- `_mm256_insert_epi32`

#### x86/avx2.rs:
- `_mm256_mpsadbw_epu8`
- `_mm256_permute2x128_si256`
- `_mm256_extract_epi16`
- `_mm256_extract_epi32`

I limited to these at first, as the other intrinsics I've done require different static assertions and `ValidateConstImm` changes that are being done in other open PRs (the MIPS ones). I wouldn't want to create conflicts and bitrot these other WIP PRs.

### DIFF
--- a/crates/core_arch/src/x86/avx.rs
+++ b/crates/core_arch/src/x86/avx.rs
@@ -438,7 +438,7 @@ pub unsafe fn _mm256_floor_pd(a: __m256d) -> __m256d {
 }
 
 /// Rounds packed single-precision (32-bit) floating point elements in `a`
-/// according to the flag `b`. The value of `b` may be as follows:
+/// according to the flag `ROUNDING`. The value of `ROUNDING` may be as follows:
 ///
 /// - `0x00`: Round to the nearest whole number.
 /// - `0x01`: Round down, toward negative infinity.
@@ -452,16 +452,12 @@ pub unsafe fn _mm256_floor_pd(a: __m256d) -> __m256d {
 /// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm256_round_ps)
 #[inline]
 #[target_feature(enable = "avx")]
-#[cfg_attr(test, assert_instr(vroundps, b = 0x00))]
-#[rustc_args_required_const(1)]
+#[cfg_attr(test, assert_instr(vroundps, ROUNDING = 0x00))]
+#[rustc_legacy_const_generics(1)]
 #[stable(feature = "simd_x86", since = "1.27.0")]
-pub unsafe fn _mm256_round_ps(a: __m256, b: i32) -> __m256 {
-    macro_rules! call {
-        ($imm8:expr) => {
-            roundps256(a, $imm8)
-        };
-    }
-    constify_imm8!(b, call)
+pub unsafe fn _mm256_round_ps<const ROUNDING: i32>(a: __m256) -> __m256 {
+    static_assert_imm4!(ROUNDING);
+    roundps256(a, ROUNDING)
 }
 
 /// Rounds packed single-precision (32-bit) floating point elements in `a`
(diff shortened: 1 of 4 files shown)


crates/core_arch/src/x86/avx.rs
```diff
@@ -405,16 +405,12 @@ pub unsafe fn _mm256_div_pd(a: __m256d, b: __m256d) -> __m256d {
...
 #[stable(feature = "simd_x86", since = "1.27.0")]
-pub unsafe fn _mm256_round_pd(a: __m256d, b: i32) -> __m256d {
-    macro_rules! call {
-        ($imm8:expr) => {
-            roundpd256(a, $imm8)
-        };
-    }
-    constify_imm8!(b, call)
+pub unsafe fn _mm256_round_pd<const B: i32>(a: __m256d) -> __m256d {
+    static_assert_imm8!(B);
```
Amanieu, line 412: This should be `imm4`. Not your fault, the original code was invalid in the first place.

crates/core_arch/src/x86/avx.rs
```diff
@@ -456,16 +452,12 @@ pub unsafe fn _mm256_floor_pd(a: __m256d) -> __m256d {
...
 #[stable(feature = "simd_x86", since = "1.27.0")]
-pub unsafe fn _mm256_round_ps(a: __m256, b: i32) -> __m256 {
-    macro_rules! call {
-        ($imm8:expr) => {
-            roundps256(a, $imm8)
-        };
-    }
-    constify_imm8!(b, call)
+pub unsafe fn _mm256_round_ps<const B: i32>(a: __m256) -> __m256 {
+    static_assert_imm8!(B);
```
Amanieu, line 459: Same here.

crates/core_arch/src/x86/avx.rs
lqd: Do we also want to rename the parameter to match the intel intrinsics guide ? (it should be `rounding` here)

Amanieu: Yes that's probably a good idea.

lqd: Ok. I've pushed both changes

lqd:
By the way, when working on this yesterday, there was a problem that I only saw on the CI tests with the obscure const errors. I was able to diagnose it via bisecting, since there was one commit per intrinsic (and I assume it does help reviews). So it makes me wonder if we should still "squash and merge" the const generics conversion PRs if it hinders automated bisecting in the future ? The granularity will be coarser, per-squashed-PR rather than per-commit.

Amanieu:
Yea I'm not too happy about the situation with const errors, but I don't really see what else we could do about it. I wonder if `static_assert!` gives better errors since the struct definition is within the macro?

Amanieu:
I didn't squash it this time. But for some reason this repo is configured to disallow merge commits and I don't have permissions to change that.

lqd:
`static_assert!` seems to show the macro invocation within the error message https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=1a2a7f7d6d7dd1f7c7e6aca5fae242cb so it is a possible fallback.

I was hoping `-Z macro-backtrace` would show more info in our `static_assert_imm` cases but unfortunately not. 

Having a low number of structs for validation seems better than the thousand we'd need with `static_assert`s, of course. Probably good enough to stay with the `static_assert_imm`s for now, until at least the crater run. We could take a look at the const error diagnostics in rustc as well 🤔

Amanieu:
We'll have to evaluate the situation when we get around to updating stdarch in rustc. Specifically to measure the perf/size impact of `static_assert`.

PR #1040, opened 2021-03-02
Other trivial x86 intrinsics conversions to const generics

lqd:
This PR converts these trivial x86 intrinsics to const generics.

#### x86/avx.rs:
- `_mm_cmp_pd`
- `_mm256_cmp_pd`
- `_mm_cmp_ps`
- `_mm256_cmp_ps`
- `_mm_cmp_sd`
- `_mm_cmp_ss`
- `_mm256_insert_epi8`

#### x86/avx2.rs:
- `_mm256_extract_epi8`

Of note, all those but the last two had an "incorrect" width compared to the intrinsics guide so I changed that at the same time.

This PR also cherry-picks a commit from #1028 as the author mentioned they will be away for some weeks, and that PR requires a rebase anyway. It's adding new widths to the static assertions that these intrinsics use.


lqd:
Ah but it's not that the widths were wrong per se, it's that the [x86 `constify_imm6` macro](https://github.com/rust-lang/stdarch/blob/47339280b9f68d4e4eb892dceb3c9cbeb23b9d9c/crates/core_arch/src/x86/macros.rs#L35-L73) is likely misnamed and made for imm5s. I'll try removing the unused x86 macros.

crates/core_arch/src/x86/avx.rs
```diff
@@ -792,16 +792,12 @@ pub const _CMP_TRUE_US: i32 = 0x1f;
...
+#[rustc_legacy_const_generics(2)]
 #[stable(feature = "simd_x86", since = "1.27.0")]
-pub unsafe fn _mm_cmp_pd(a: __m128d, b: __m128d, imm8: i32) -> __m128d {
-    macro_rules! call {
-        ($imm8:expr) => {
-            vcmppd(a, b, $imm8)
-        };
-    }
-    constify_imm6!(imm8, call)
+pub unsafe fn _mm_cmp_pd<const IMM8: i32>(a: __m128d, b: __m128d) -> __m128d {
```
Amanieu, line 798: The name `IMM8` is misleading here since the immediate isn't actually 8 bits.

lqd: I can change that. I thought the old `imm8` name was going for consistency with the Intel intrinsics guide.

Amanieu: Oh hmm, I didn't notice that the intrinsics guide actually used `imm8`. That's just misleading...

I'll leave it up to your discretion as to whether to rename the argument or not.

lqd: Yeah it seems it does so more often than not, misleading indeed.

We do deviate in the naming here and there already, so it's probably ok for us to be more precise in the naming and have `IMM5` here ? We could automate renaming everything to match the guide if we needed to some day. 

I'll do a quick pass for the intrinsics in this PR, and we'll see if everyone likes it :)

lqd: I've pushed 210cf93 to address this, for these new converted intrinsincs:
- whenever Intel has an `imm8` of width different from 8, I went with the width, like `IMM5`, so that we're more precise
- whenever they have another name not mentioning the width, for example we're already using `INDEX` in a couple places, I went with that (even though I wish we could still mention the width)

what do you think ?

Amanieu: Looks great!

PR #1045, opened 2021-03-04
Modify stdarch-gen to generate instructions with a single parameter and add vceqz instructions

SparrowLii:
This pr dose the following: 
1. Modify the automatic generation program of neon instructions so that it can generate neon instructions that contain a single parameter (`vceqz`, `vdup`, `vmov_n`, etc.)
2. Add `vecqz` instructions to test 1. 

In addition, this two lines in stdarch-gen/src/main.rs do the opposite of what they should do:

        "MAX" => min_val(t),
        "MIN" => max_val(t),

I corrected them so arm/neon/generated.rs changed accordingly.


SparrowLii:
Adding `pub` to `type!` is not a good way, I will make the corresponding changes.

PR #1056, opened 2021-03-06
More AVX512F intrinsics conversions to const generics

lqd:
This PR converts the following `x86/avx512f.rs` "roundscale" intrinsics to const generics, since they were using huge macros and contributed a lot of to the rlib size and bootstrap times:

- `_mm512_roundscale_round_ps` 
- `_mm512_mask_roundscale_round_ps` 
- `_mm512_maskz_roundscale_round_ps` 
- `_mm512_roundscale_round_pd` 
- `_mm512_mask_roundscale_round_pd` 
- `_mm512_maskz_roundscale_round_pd` 
- `_mm512_fixupimm_round_ps` 
- `_mm512_mask_fixupimm_round_ps` 
- `_mm512_maskz_fixupimm_round_ps` 
- `_mm512_fixupimm_round_pd` 
- `_mm512_mask_fixupimm_round_pd` 
- `_mm512_maskz_fixupimm_round_pd` 
- `_mm_roundscale_round_ss` 
- `_mm_mask_roundscale_round_ss` 
- `_mm_maskz_roundscale_round_ss` 
- `_mm_roundscale_round_sd` 
- `_mm_mask_roundscale_round_sd` 
- `_mm_maskz_roundscale_round_sd` 
- `_mm_maskz_fixupimm_round_sd` 
- `_mm_fixupimm_round_ss` 
- `_mm_mask_fixupimm_round_ss` 
- `_mm_maskz_fixupimm_round_ss` 
- `_mm_fixupimm_round_sd` 
- `_mm_mask_fixupimm_round_sd`  

Converting these reduces the size of `libcore_arch.rlib` by 7MB for me.

Something stands out in the existing handling of the SAE parameter (in general, but there was a dedicated macro for the roundscale operations so I'll only talk about this one).

The intel intrinsics guide, for example https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm512_roundscale_round_ps, mentions using the `sae` parameter to expectedly suppress exceptions. IIRC, since these bits can be set in the immediate, the parameter could be seen as redundant. 

However in the existing [`constify_imm8_roundscale`](https://github.com/rust-lang/stdarch/blob/ad47bd1f75584c8779301e4231a9713e1088d544/crates/core_arch/src/x86/macros.rs#L758-L1536) macro, it seems the values tested are for immediates with SAE bits set (much like clang does, but I don't find their technique directly applicable to our use case) rather than an SAE flag per se. Similarly our tests seem to pass rounding constants to the SAE parameter, for example `_mm512_roundscale_round_ps`:

```rust
#[simd_test(enable = "avx512f")]
unsafe fn test_mm512_roundscale_round_ps() {
    let a = _mm512_set1_ps(1.1);
    let r = _mm512_roundscale_round_ps(a, 0, _MM_FROUND_CUR_DIRECTION);
    let e = _mm512_set1_ps(1.0);
    assert_eq_m512(r, e);
}
```

I'm not super familiar with avx512f, and don't have the hardware to verify these assumptions, but the documentation in the intel intrinsics guide makes me wonder whether the last parameter should be either `_MM_FROUND_RAISE_EXC` (the "default" of 0) or  `_MM_FROUND_NO_EXC` (whose value is 8) rather than the 4/8/12 currently used.

Does anyone know whether the current intrinsics and tests are indeed correct, and I'm misunderstanding here ?


lqd:
I think so, if we're to believe https://github.com/llvm/llvm-project/blob/bd50cf905fa7c0c7caa134301c6ca0658c81eeb1/clang/lib/Sema/SemaChecking.cpp#L3499 (but am trying to find a more version of these checks and cannot find them as of yet). 

But they could be mixing the immediate and the SAE parameters into their single `result`, to simplify the checking. Since we don't do that and instead verify the two parameters separately, maybe those values are not applicable in our case, but I don't know :)

lqd:
heh true, and the pseudocode even uses `imm8[3:0]`.  Sounds like they are indeed allowing (or expecting ?) the `imm8` to possibly be mixed with the exceptions suppressed bit.
I'm not exactly sure who to follow here, the differences between intel docs and clang are a bit confusing to me. 

Maybe the others will have opinions about this ?

Amanieu:
I did a bit of reading in the GCC and Clang source code and I'm still confused as to how this is supposed to work. I think we should just mirror Clang here: only assert the basic `imm8` constraint on the first argument and a stricter SAE constraint on the second. TBH this is just a case of AVX-512 being poorly designed.

Amanieu:
In the end, both GCC and Clang pass the `imm8` and `sae` parameters straight through to the assembler, which then combines the values.

lqd:
Turns out all these were merged in #1054, so I'll close this PR.

PR #1060, opened 2021-03-07
More avx512f conversions to const generics

lqd:
This PR converts some more trivial avx512f intrinsics to const generics.

There's now only around 50 `rustc_args_required` x86 intrinsics left (and around 20 in other targets). Locally libcore_arch.rlib is now around 20MB. We all managed to divide its size by 3 since last week 🎉


lqd:
@minybot thanks to you too as well, you've done a huge amount of work !

Happy to help. 

Hopefully there won't be many problems when we update stdarch in the rust repo in the future (I'm looking forward to the improved bootstrap times), or the crater run to validate our assumptions about static assertions and the code using the intrinsics. 

We'll get it done :)

PR #1061, opened 2021-03-07
Convert _mm_alignr_epi8 to const generics

tmiasko:


crates/core_arch/src/x86/ssse3.rs
```diff
@@ -86,71 +86,55 @@ pub unsafe fn _mm_shuffle_epi8(a: __m128i, b: __m128i) -> __m128i {
...
 #[target_feature(enable = "ssse3")]
-#[cfg_attr(test, assert_instr(palignr, n = 15))]
-#[rustc_args_required_const(2)]
+#[cfg_attr(test, assert_instr(palignr, IMM8 = 15))]
+#[rustc_legacy_const_generics(2)]
 #[stable(feature = "simd_x86", since = "1.27.0")]
-pub unsafe fn _mm_alignr_epi8(a: __m128i, b: __m128i, n: i32) -> __m128i {
-    let n = n as u32;
+pub unsafe fn _mm_alignr_epi8<const IMM8: i32>(a: __m128i, b: __m128i) -> __m128i {
+    static_assert_imm8!(IMM8);
```
tmiasko, line 93: Currently with 8-bit immediate, instead 5-bit as specified, so it can be used by _mm_maskz_alignr_epi8 & _mm_mask_alignr_epi8, both of which are specified to accept 8-bit immediate.

Amanieu: `_mm_maskz_alignr_epi8` and `_mm_mask_alignr_epi8` should be changed to accept a 5-bit immediate instead.

Amanieu: Hmm, actually `_mm_alignr_epi8` *does* have a well-defined behavior if the immediate is greater than 32, which is to return 0. So on second thought setting the constraint to 8-bit is fine.

PR #1062, opened 2021-03-08
add neon instruction vbsl

surechen:
add neon instruction vbsl
vbsl:   Bitwise Select. This instruction sets each bit in the destination SIMD&FP register to the corresponding bit from the first source SIMD&FP register when the original destination bit was 1, otherwise from the second source SIMD&FP register.

### DIFF
--- a/crates/core_arch/src/aarch64/neon/mod.rs
+++ b/crates/core_arch/src/aarch64/neon/mod.rs
@@ -703,6 +703,23 @@ pub unsafe fn vabsq_s64(a: int64x2_t) -> int64x2_t {
     vabsq_s64_(a)
 }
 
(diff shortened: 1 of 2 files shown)


Amanieu:
Looking at the GCC/Clang headers it seems the p64 and f64 variants should also be available on ARM.

surechen:
> Looking at the GCC/Clang headers it seems the p64 and f64 variants should also be available on ARM.

Hi. Thanks for reviewing.
I use https://godbolt.org/ with clang to check this instruction.
https://godbolt.org/z/P7x1dv
https://godbolt.org/z/r3T68b
The result is <Compilation failed>.
If there is any problem with my verification method, please help me. Thank you very much.

Amanieu:
Ah sorry you are correct, I misread the headers.

PR #1069, opened 2021-03-09
add vcgez, vcgtz, vclez, vcltz neon instructions

SparrowLii:
All are automatically generated single-parameter comparison instructions. In order to be consistent with the implementation in Clang, some changes have been made to stdarch-gen.

### DIFF
--- a/crates/stdarch-gen/neon.spec
+++ b/crates/stdarch-gen/neon.spec
@@ -426,13 +426,9 @@ generate float32x2_t:uint32x2_t, float32x4_t:uint32x4_t
 
 /// Compare signed greater than or equal to zero
 name = vcgez
-multi_fn = fixed, c:in_t
-multi_fn = fixed_2, d:in_t
-multi_fn = simd_shr, e:, a, transmute(c)
-multi_fn = simd_xor, transmute(e), transmute(d)
+fn = simd_ge
 a = MIN, -1, 0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B, 0x0C, MAX
-fixed = BITS_M1, BITS_M1, BITS_M1, BITS_M1, BITS_M1, BITS_M1, BITS_M1, BITS_M1, BITS_M1, BITS_M1, BITS_M1, BITS_M1, BITS_M1, BITS_M1, BITS_M1, BITS_M1
-fixed_2 = -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1
+fixed = 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
 validate FALSE, FALSE, TRUE, TRUE, TRUE, TRUE, TRUE, TRUE, TRUE, TRUE, TRUE, TRUE, TRUE, TRUE, TRUE, TRUE
 
 aarch64 = cmge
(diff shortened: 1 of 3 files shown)


crates/stdarch-gen/neon.spec
```diff
@@ -424,6 +424,92 @@ arm = vcge.s
 // we are missing float16x4_t:uint16x4_t, float16x8_t:uint16x8_t
 generate float32x2_t:uint32x2_t, float32x4_t:uint32x4_t
 
+/// Compare signed greater than or equal to zero
+name = vcgez
+multi_fn = fixed, c:in_t
+multi_fn = fixed_2, d:in_t
+multi_fn = simd_shr, e:, a, transmute(c)
+multi_fn = simd_xor, transmute(e), transmute(d)
```
Amanieu, line 432: Why not just use `simd_ge` here?

crates/stdarch-gen/neon.spec
```diff
@@ -424,6 +424,92 @@ arm = vcge.s
...
+fixed = 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0
+validate TRUE, TRUE, FALSE, FALSE, FALSE, FALSE, FALSE, FALSE
+
+aarch64 = fcmle
+generate float32x2_t:uint32x2_t, float32x4_t:uint32x4_t, float64x1_t:uint64x1_t, float64x2_t:uint64x2_t
+
+/// Compare signed less than zero
+name = vcltz
+multi_fn = fixed, b:in_t
+multi_fn = simd_shr, c:in_t, a, transmute(b)
```
Amanieu, line 494: And `simd_lt` here?

crates/stdarch-gen/neon.spec
SparrowLii: This is to be consistent with Clang's implementation. The following is the test I did in https://godbolt.org/:

    #include <arm_neon.h>
    int test() {
      return (int) vcgez_s32;
    }        

And the Output:

    define dso_local i32 @test() local_unnamed_addr #0 {
        ret i32 ptrtoint (<2 x i32> (<2 x i32>)* @vcgez_s32 to i32)
    }
    
    define internal <2 x i32> @vcgez_s32(<2 x i32> %0) #1 {
        %2 = ashr <2 x i32> %0, <i32 31, i32 31>
        %3 = xor <2 x i32> %2, <i32 -1, i32 -1>
        ret <2 x i32> %3
    }
    
    attributes #0 = { norecurse nounwind readnone uwtable willreturn "frame-pointer"="non-leaf" "no-trapping-math"="true" "stack-protector-buffer-size"="8" "target-cpu"="generic" "target-features"="+neon" }
    attributes #1 = { alwaysinline norecurse nounwind readnone uwtable willreturn "frame-pointer"="non-leaf" "min-legal-vector-width"="64" "no-trapping-math"="true" "stack-protector-buffer-size"="8" "target-cpu"="generic" "target-features"="+neon" }

Amanieu: If you compile with `-O0` you will see that Clang actually emits an `icmp sge`. LLVM optimizations are then turning this into a shift + xor.

crates/stdarch-gen/neon.spec
SparrowLii: Same as above, the following is my test in https://godbolt.org/:

    #include <arm_neon.h>
    int test() {
    return (int) vcltz_s32;
    }                                     

And the Output:

    define dso_local i32 @test() local_unnamed_addr #0 {
        ret i32 ptrtoint (<2 x i32> (<2 x i32>)* @vcltz_s32 to i32)
    }
    
    define internal <2 x i32> @vcltz_s32(<2 x i32> %0) #1 {
        %2 = ashr <2 x i32> %0, <i32 31, i32 31>
        ret <2 x i32> %2
    }
    
    attributes #0 = { norecurse nounwind readnone uwtable willreturn "frame-pointer"="non-leaf" "no-trapping-math"="true" "stack-protector-buffer-size"="8" "target-cpu"="generic" "target-features"="+neon" }
    attributes #1 = { alwaysinline norecurse nounwind readnone uwtable willreturn "frame-pointer"="non-leaf" "min-legal-vector-width"="64" "no-trapping-math"="true" "stack-protector-buffer-size"="8" "target-cpu"="generic" "target-features"="+neon" }

crates/stdarch-gen/neon.spec
SparrowLii: The url of godbolt is from here: https://github.com/rust-lang/stdarch/issues/148

Amanieu: That usually works, but in this particular case it gives a different result because the IR is not the one generated by Clang directly: it is the IR after LLVM has run optimization passes that expand the `icmp eq` into shift and xor.

You can use `simd_ge` in Rust and it will produce the same IR as Clang.

SparrowLii: Umm.. That's right. If we use an implementation consistent with -O0, can we ensure that LLVM achieves the same optimization? If so, we should indeed use simd_ge IMO

Amanieu: We run the same LLVM passes as Clang (mostly) so rustc will also transform `simd_ge` into a shift + xor.

SparrowLii: Thanks for explanation!

Amanieu:
Can you add ARM versions of these functions?

SparrowLii:
> Can you add ARM versions of these functions?

It seems that these instructions are unique to aarch64 and only accept Signed parameters

PR #1071, opened 2021-03-10
Add neon instruction vdup

surechen:
add neon instruction   vdup_n_*    vdupq_n_*

vdup_n: Duplicate vector element to vector or scalar. This instruction duplicates the vector element at the specified element index in the source SIMD&FP register into a scalar or each element in a vector, and writes the result to the destination SIMD&FP register.

### DIFF
--- a/crates/core_arch/src/arm/neon/mod.rs
+++ b/crates/core_arch/src/arm/neon/mod.rs
@@ -3833,7 +3833,7 @@ pub unsafe fn vdupq_n_u64(value: u64) -> uint64x2_t {
 #[cfg_attr(target_arch = "arm", target_feature(enable = "v7"))]
 #[cfg_attr(all(test, target_arch = "arm"), assert_instr("vdup.8"))]
 #[cfg_attr(all(test, target_arch = "aarch64"), assert_instr(dup))]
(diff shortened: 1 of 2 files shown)


crates/core_arch/src/arm/neon/mod.rs
```diff
@@ -3736,6 +3736,11 @@ pub unsafe fn vget_lane_u8<const IMM5: i32>(v: uint8x8_t) -> u8 {
     simd_extract(v, IMM5 as u32)
 }
 
+/// Duplicate vector element to vector or scalar.
+/// This instruction duplicates the vector element at the specified element index
+/// in the source SIMD&FP register into a scalar or each element in a vector,
+/// and writes the result to the destination SIMD&FP register.
```
Amanieu, line 3742: The doc comment should be duplicated for each function so that it appears properly in rustdoc.

surechen: Done. Thanks. Delete redundant comments to keep consistent with the previous concise comment style.

surechen: > The doc comment should be duplicated for each function so that it appears properly in rustdoc.

Done. Thanks. 
Remove redundant comments, consistent with the previous concise comment style

PR #1072, opened 2021-03-11
add vcls, vclz, vcagt, vcage, vcalt, vcale neon instructions

SparrowLii:
This PR does the following:
1. Add count leading sign bits, count leading zero bits and floating-bits compare instructions.
2. Enhance the fuction of multi_fn in code generator . It can now generate function calls recursively and can call functions with different prefixes and suffixes, so these instructions can be generated automatically.


Amanieu:
Can you change the code generator to include the transmutes? Maybe explicitly specify a subset of types in the spec that need a transmute?

SparrowLii:
> Can you change the code generator to include the transmutes? Maybe explicitly specify a subset of types in the spec that need a transmute?

Yes, we can regard transmute as a function or just a parameter like `transmute(a)`. I enhanced the syntax in multi_fn, it can now produce function calls recursively and can call functions with different suffixes

crates/core_arch/src/arm/neon/mod.rs
```diff
@@ -292,6 +292,25 @@ extern "C" {
     #[cfg_attr(target_arch = "arm", link_name = "llvm.ctpop.v16i8")]
     #[cfg_attr(target_arch = "aarch64", link_name = "llvm.ctpop.v16i8")]
     fn vcntq_s8_(a: int8x16_t) -> int8x16_t;
+
+    #[cfg_attr(target_arch = "arm", link_name = "llvm.ctlz.v8i8")]
+    #[cfg_attr(target_arch = "aarch64", link_name = "llvm.ctlz.v8i8")]
+    fn vclz_s8_(a: int8x8_t) -> int8x8_t;
```
Amanieu, line 298: Why are these defined here instead of in the generated code?

SparrowLii: If you use automatic generation, the function will be defined inside functions such as vcls_s8, so vcls_u8 must call transmute(vcls_s8(a)) instead of transmute(vcls_s8_(a))

Amanieu: I don't see the problem. It's fine if we have duplicate `extern` definitions.

SparrowLii: On the other hand, the link format of these functions(which is llvm.ctlz.v8i8 ) is different from other functions( which is llvm.arm.neon.*).

SparrowLii: Maybe I should modify the syntax of the code generator to make it more flexible to generate link information?

Amanieu: I think this is good enough.

PR #1073, opened 2021-03-11
Update WebAssembly SIMD/Atomics

alexcrichton:
This commit syncs the atomics/simd intrinsics used on WebAssembly with
LLVM 12 and the current specifications. LLVM 12 uses new names for
atomic intrinsics and the SIMD specification has added a lot of
intrinsics and renamed a few as well.

I was hoping to hold off on this until more of SIMD had landed since
there are some opcode renumberings that have happened at the spec level
but haven't happened in LLVM. Additionally there's a small handful of
instructions that have yet to be implemented in LLVM. This means that
many tests for the simd128 feature are ignored right now and/or are
known to not pass. The breakage in the name of the atomic intrinsics,
however, has prompted me to want to update this and land ahead of time.

For now I've disabled the SIMD testing and I'll get back to it once
things have settled a bit more with LLVM and runtimes.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -77,7 +77,7 @@ jobs:
         - mips64-unknown-linux-gnuabi64
         - mips64el-unknown-linux-gnuabi64
         - s390x-unknown-linux-gnu
-        # - wasm32-wasi
+        - wasm32-wasi
         - i586-unknown-linux-gnu
         - x86_64-linux-android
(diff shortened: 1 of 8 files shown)


Amanieu:
I had a quick look and it still looks very much like a WIP, so I won't do an in-depth review. Still merging it since it fixes WASM intrinsics on LLVM 12.

PR #1074, opened 2021-03-12
add neon instruction vget_high_*

surechen:
add neon instruction vget_high_*
vget_high_* : 
Duplicate vector element to vector or scalar. This instruction duplicates the vector element at the specified element index in the source SIMD&FP register into a scalar or each element in a vector, and writes the result to the destination SIMD&FP register.


Amanieu:
This is missing `vget_high_p64` on both arm & aarch64, and `vget_high_f64` on aarch64-only.

surechen:
> This is missing `vget_high_p64` on both arm & aarch64, and `vget_high_f64` on aarch64-only.

Done. Thanks.

PR #1077, opened 2021-03-12
Add vdiv neon instructions

CryZe:
I'm not sure if that's all there is to do. Seemed like an odd omission in the first place. Maybe I'm missing something?

### DIFF
--- a/crates/stdarch-gen/neon.spec
+++ b/crates/stdarch-gen/neon.spec
@@ -601,6 +601,15 @@ generate float64x*_t
 arm = vmul.
 generate float*_t
 
+/// Divide
+name = vdiv
+fn = simd_div
+a = 2.0, 6.0, 4.0, 10.0
+b = 1.0, 2.0, 1.0, 2.0
+validate 2.0, 3.0, 4.0, 5.0
+
+aarch64 = fdiv
+generate float*_t, float64x*_t
 
 /// Subtract
 name = vsub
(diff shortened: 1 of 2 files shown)


crates/stdarch-gen/neon.spec
```diff
@@ -601,6 +601,15 @@ generate float64x*_t
...
 generate float*_t
 
+/// Divide
+name = vdiv
+fn = simd_div
+a = 2.0, 6.0, 4.0, 10.0
+b = 1.0, 2.0, 1.0, 2.0
+validate 2.0, 3.0, 4.0, 5.0
+
+aarch64 = fdiv
```
Amanieu, line 611: Can you generate arm versions of this which use the `vdiv` instruction?

Amanieu: Note that arm only supports `f32` variants: `f64` is aarch64-only.

CryZe: According to the docs the supported architectures are only A64: https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics?search=vdiv

> Supported architectures
> A64

Maybe the docs are wrong?

Amanieu: Ah you are correct, it seems the ARM `vdiv` instruction only supports scalar division, not vector division.

PR #1052, opened 2021-03-06
Convert all uses of llvm_asm! to asm!

Amanieu:
Fixes #904

PR #1081, opened 2021-03-13
make is_{arch}_feature_detected! available for std

the8472:
The feature detection macros currently can't be used in std itself due to these errors.

```
error: macro-expanded `macro_export` macros from the current crate cannot be referred to by absolute paths
   --> library/std/src/time.rs:30:5
    |
30  | use super::is_x86_feature_detected;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[deny(macro_expanded_macro_exports_accessed_by_absolute_paths)]` on by default
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #52234 <https://github.com/rust-lang/rust/issues/52234>
note: the macro is defined here
   --> library/std/src/../../stdarch/crates/std_detect/src/detect/macros.rs:14:9
```

The workaround in this PR was suggested in this [zulip discussion](https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/using.20is_x86_feature_detected.20in.20std)

### DIFF
--- a/crates/std_detect/src/detect/macros.rs
+++ b/crates/std_detect/src/detect/macros.rs
@@ -43,6 +43,9 @@ macro_rules! features {
             };
         }
 
+        #[cfg(not(test))]
+        pub use $macro_name;
+
         /// Each variant denotes a position in a bitset for a particular feature.
         ///
         /// PLEASE: do not use this, it is an implementation detail subject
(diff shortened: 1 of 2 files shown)


Amanieu:
Note that with #1005, `std_detect` is a separate crate instead of a sub-module of `std`.

the8472:
Will std be updated to use it that way in the near future?

Amanieu:
Yes, I plan on updating the stdarch submodule in std as soon as #1052 lands, which is blocked on https://github.com/rust-lang/rust/pull/82838.

Amanieu:
Actually there might be a libs team FCP on the merge due to the const generics changes, which may take a while to resolve.

the8472:
Well, this only for a small optimization. I'll just submit a rust PR and note that it's blocked.

PR #1084, opened 2021-03-16
add vcvt, vcvta, vcvtn, vcvtm, vcvtp neon instructions

SparrowLii:
Add vcvt, vcvta, vcvtn, vcvtm, vcvtp neon instructions(Floating-point convert to fixed-point with different rounding methods).

### DIFF
--- a/crates/core_arch/src/arm/neon/generated.rs
+++ b/crates/core_arch/src/arm/neon/generated.rs
@@ -5523,7 +5523,7 @@ mod test {
 
     #[simd_test(enable = "neon")]
     unsafe fn test_vcvt_s32_f32() {
-        let a: f32x2 = f32x2::new(-1.0, 2.0);
+        let a: f32x2 = f32x2::new(-1.1, 2.1);
         let e: i32x2 = i32x2::new(-1, 2);
         let r: i32x2 = transmute(vcvt_s32_f32(transmute(a)));
         assert_eq!(r, e);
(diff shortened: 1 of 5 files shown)


crates/stdarch-gen/neon.spec
```diff
@@ -652,6 +652,126 @@ validate -1.0, 2.0, -3.0, 4.0
 aarch64 = fcvtxn
 generate float32x2_t:float64x2_t:float32x4_t
 
+/// Floating-point convert to signed fixed-point, rounding toward zero
+name = vcvt
+double-suffixes
+fn = simd_cast
+a = -1.0, 2.0, -3.0, 4.0
```
CryZe, line 659: Wouldn't it make more sense to use .5 and co. to actually test the rounding (for all of these instructions)?

SparrowLii: Yes, but I don’t have a good way to test the results. How can we simulate rounding towards to even?

crates/core_arch/src/arm/neon/generated.rs
```diff
@@ -1941,6 +1941,42 @@ pub unsafe fn vcaleq_f32(a: float32x4_t, b: float32x4_t) -> uint32x4_t {
     vcageq_f32(b, a)
 }
 
+/// Floating-point convert to signed fixed-point, rounding toward zero
+#[inline]
+#[target_feature(enable = "neon")]
+#[cfg_attr(target_arch = "arm", target_feature(enable = "v7"))]
+#[cfg_attr(all(test, target_arch = "arm"), assert_instr(vcvt))]
+pub unsafe fn vcvt_s32_f32(a: float32x2_t) -> int32x2_t {
```
CryZe, line 1949: It doesn't seem like there's any assert_instr for aarch64 for these?!

SparrowLii: There are different implementations for arm and aarch64. You can use https://godbolt.org/ to verify

crates/stdarch-gen/neon.spec
CryZe: e.g.:

a = -1.5, 2.5, -3.5, 4.5
validate = -2, 2, -4, 4

I'm confused what the problem is?

SparrowLii: Sorry I forgot that this is rounding to an integer. I will correct accordingly

crates/core_arch/src/arm/neon/generated.rs
CryZe: Well it's fcvtzs and co., so idk why those aren't being asserted for here.

SparrowLii: https://godbolt.org/z/GTrMP6 In this version of the test, the arm uses the fptosi command. The test link comes from the top link [here](https://github.com/rust-lang/stdarch/issues/148)

SparrowLii: The arm version does not need to introduce external links #[link_name = ...]. which aarch64 needs

SparrowLii: I think you are right, thanks a lot

PR #1085, opened 2021-03-16
add neon instruction vmov_n_* and vmovq_n_*

surechen:
add neon instruction vmov_n_* and vmovq_n_*
vmov_n_* : Duplicate vector element to vector or scalar. This instruction duplicates the vector element at the specified element index in the source SIMD&FP register into a scalar or each element in a vector, and writes the result to the destination SIMD&FP register. 

Without vmov_n_p64 and vmovq_n_p64 because they are not in arm-intrinsics.html(get missing arm definition in Automatic intrinsic verification).

### DIFF
--- a/crates/stdarch-verify/tests/arm.rs
+++ b/crates/stdarch-verify/tests/arm.rs
@@ -394,8 +394,8 @@ fn verify_all_signatures() {
             "brk" | "__breakpoint" | "udf" | "_prefetch" => continue,
             _ => {}
         }
-        // Skip some intrinsics that are present in GCC and Clang but are missing
-        // from the official documentation(arm-intrinsics.html).
+        // Skip some intrinsics that are present in GCC and Clang but
+        // are missing from the official documentation.
         let skip_intrinsic_verify = ["vmov_n_p64", "vmovq_n_p64"];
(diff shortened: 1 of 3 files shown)


Amanieu:
> Without vmov_n_p64 and vmovq_n_p64 because they are not in arm-intrinsics.html(get missing arm definition in Automatic intrinsic verification).

It seems that both GCC and Clang have these, so this is most likely a bug in the spec. You should add an exception in `stdarch-verify` for these intrinsics.

crates/stdarch-verify/tests/arm.rs
```diff
@@ -394,6 +394,7 @@ fn verify_all_signatures() {
             "brk" | "__breakpoint" | "udf" | "_prefetch" => continue,
             _ => {}
         }
+        let skip_intrinsic_verify = ["vmov_n_p64", "vmovq_n_p64"];
```
Amanieu, line 397: Add a small comment explaining that these are present in GCC and Clang but are missing from the official documentation.

surechen: > Add a small comment explaining that these are present in GCC and Clang but are missing from the official documentation.

Done. Thanks.

PR #1086, opened 2021-03-16
Add vrndn neon instructions

CryZe:
This adds the neon instructions for lane-wise rounding without actually converting the lanes to integers.

### DIFF
--- a/crates/core_arch/src/aarch64/neon/generated.rs
+++ b/crates/core_arch/src/aarch64/neon/generated.rs
@@ -2522,32 +2522,6 @@ pub unsafe fn vrndaq_f64(a: float64x2_t) -> float64x2_t {
 #[inline]
 #[target_feature(enable = "neon")]
 #[cfg_attr(test, assert_instr(frintn))]
-pub unsafe fn vrndn_f32(a: float32x2_t) -> float32x2_t {
-    #[allow(improper_ctypes)]
-    extern "C" {
-        #[cfg_attr(target_arch = "aarch64", link_name = "llvm.aarch64.neon.frintn.v2f32")]
-        fn vrndn_f32_(a: float32x2_t) -> float32x2_t;
-    }
-    vrndn_f32_(a)
-}
-
-/// Floating-point round to integral, to nearest with ties to even
-#[inline]
-#[target_feature(enable = "neon")]
-#[cfg_attr(test, assert_instr(frintn))]
-pub unsafe fn vrndnq_f32(a: float32x4_t) -> float32x4_t {
-    #[allow(improper_ctypes)]
-    extern "C" {
-        #[cfg_attr(target_arch = "aarch64", link_name = "llvm.aarch64.neon.frintn.v4f32")]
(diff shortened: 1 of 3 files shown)


Amanieu:
These methods are also available on 32-bit ARM when the "v8" feature is enabled.

CryZe:
It doesn't seem like Rust ships any target for AArch32. I guess this is meant for when you build std yourself then?

crates/core_arch/src/arm/mod.rs
```diff
@@ -14,24 +14,16 @@ pub use self::armclang::*;
...
 pub use self::neon::*;
 
-#[cfg(any(target_arch = "aarch64", target_feature = "v7"))]
 mod crc;
-#[cfg(any(target_arch = "aarch64", target_feature = "v7"))]
 pub use self::crc::*;
 
-#[cfg(any(target_arch = "aarch64", target_feature = "v7"))]
 mod crypto;
-#[cfg(any(target_arch = "aarch64", target_feature = "v7"))]
```
CryZe, line 34: The latest commit is mostly to raise some discussion maybe. I removed a bunch of cfg and target_feature(enable = "v7") bounds that don't seem necessary to me.

Amanieu:
> The more I look at it, the more I feel like stdarch doesn't support 32-bit v8 properly anyway. float64x2_t isn't even defined for the ARM target for example (update: even if it did, LLVM can't even handle it, it seems?! Is this an LLVM bug or a NEON docs bug?).

AArch32 NEON still only supports a subset of AArch64 NEON. In particular `f64` SIMD is only supported on AArch64.

AArch32 and AArch64 should be treated as distinct architectures, they have completely different instruction sets. They just happen to share the same intrinsics for NEON.

CryZe:
> In particular f64 SIMD is only supported on AArch64.

Yeah seems like the docs are just wrong then.

crates/core_arch/src/arm/mod.rs
CryZe: Alright so it seems that runtime detection for the ARM version isn't a thing anyway, so the former doesn't quite apply here. Though that now begs the additional question, how I even write tests for these then? The tests seem to use runtime detection (or do they?)

Amanieu: The `is_arm_feature_detected` macro in `std_detect` needs to be updated to add the target features defined in https://github.com/rust-lang/rust/blob/36f1f04f18b89ba4a999bcfd6584663fd6fc1c5d/compiler/rustc_codegen_ssa/src/target_features.rs#L12.

Amanieu:
I think the ARM/AArch64 code needs to be reorganized into separate `arm`/`aarch64`/`arm_common` directories where `arm_common` is included by both `arm` and `aarch64`.

CryZe:
Seems like most of this PR has actually already been done in the meantime, even the instruction got added. Though when it was added AArch32 support was forgotten about, so I reduced this PR to that change. It's ready for review again.

PR #1087, opened 2021-03-16
add vneg neon instructions

SparrowLii:
Fixed-point and Floating-point negate instructions


crates/core_arch/src/aarch64/neon/generated.rs
```diff
@@ -1601,6 +1601,42 @@ pub unsafe fn vcvtpq_u64_f64(a: float64x2_t) -> uint64x2_t {
     vcvtpq_u64_f64_(a)
 }
 
+/// Negate
+#[inline]
+#[target_feature(enable = "neon")]
+#[cfg_attr(test, assert_instr(sub))]
+pub unsafe fn vneg_s64(a: int64x1_t) -> int64x1_t {
+    let b: i64x1 = i64x1::new(0);
+    simd_sub(transmute(b), a)
```
CryZe, line 1610: Nice, I was actually looking into this one as well earlier and "gave up on it", because there didn't seem to be a way other than 0  - the_vec. Good to see I didn't miss anything xD

SparrowLii: It seems that core_arch does not have an instruction such as simd_neg, I don’t know how to add it, because I don’t know the compiler very well. :(

bjorn3: You will need to add to the function at https://github.com/rust-lang/rust/blob/f24ce9b0140d9be5a336954e878d0c1522966bb8/compiler/rustc_codegen_llvm/src/intrinsic.rs#L730 It is probably similar to https://github.com/rust-lang/rust/blob/f24ce9b0140d9be5a336954e878d0c1522966bb8/compiler/rustc_codegen_llvm/src/intrinsic.rs#L1631-L1661 except a single argument instead of two arguments.

SparrowLii: Thanks for guiding. I will try later. I think the implementation of sub(0, a) is sufficient for now. Since the implementation in Clang is the [same](https://godbolt.org/z/e956hK)

Amanieu: They are different if you look at the code for floating-point numbers. `0.0 - x` and `-x` have different behavior for floats with regards to negative zeros, NaNs, etc.

SparrowLii: It looks like we do need to add simd_neg. I suggest to close this PR, and then raise another issue to track its implementation.

SparrowLii:
It looks like we do need to add simd_neg. I suggest to close this PR, and then raise an issue to track its implementation.

Amanieu:
I'll close the PR for now.

`simd_neg` should be fairly simple to implement in the compiler, so I don't think a tracking issue is necessary.

SparrowLii:
@Amanieu I submitted the PR [#rust-lang/rust/83271](https://github.com/rust-lang/rust/pull/83271) for adding simd_neg to the compiler. Could you help to review it?

PR #1092, opened 2021-03-19
More wasm SIMD updates

alexcrichton:
* Sync with the latest LLVM which has a few new intrinsic names
* Move explicit tests back to `assert_instr` since `assert_instr` now
  supports specifying const-generic arguments inline.
* Enable tests where wasmtime implements the instruction as well as LLVM.
* Ensure there are tests for all functions that can be tested at this
  time (those that aren't unimplemented in wasmtime).

There's still a number of `assert_instr` tests that are commented out.
These are either because they're unimplemented in wasmtime at the moment
or LLVM doesn't have an implementation for the instruction yet.

### DIFF
--- a/ci/docker/wasm32-wasi/Dockerfile
+++ b/ci/docker/wasm32-wasi/Dockerfile
@@ -20,5 +20,7 @@ COPY --from=0 /usr/local/cargo/bin/wasmtime /usr/local/bin/wasmtime
 
 ENV CARGO_TARGET_WASM32_WASI_RUNNER="wasmtime \
   --enable-simd \
+  --enable-threads \
+  --opt-level 0 \
   --mapdir .::/checkout/target/wasm32-wasi/release/deps \
   --"
--- a/ci/docker/wasm32-wasi/Dockerfile
+++ b/ci/docker/wasm32-wasi/Dockerfile
@@ -1,14 +1,22 @@
+FROM rust:1.50.0
+
+# Install wasmtime from source for now while the `experimental_x64` feature is
+# not yet the default. (it's not actually that experimental at the time of this
+# writing, wasmtime should switch defaults soon and the backend this enables has
+# better support for simd instructions)
+RUN \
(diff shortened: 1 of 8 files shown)


alexcrichton:
I'll note that this is blocked on https://github.com/rust-lang/rust/pull/83285 for landing, but I wanted to go ahead and put this up for review if others have time.

Amanieu:
I had a quick look through and the changes LTGM.

alexcrichton:
I was struck with inspiration after reading over https://github.com/rust-lang/stdarch/pull/1093 and I've updated this to also improve support for atomic intrinsics by always including them. Updates to the upstream proposal mean that this is now feasible (yay!) which means we are now also testing them on CI that they create the right instruction (and should catch bugs like https://github.com/rust-lang/stdarch/pull/1093 earlier)

This is still, however, waiting on a nightly to get fully enabled. I realized that I was testing with a build of Wasmtime, however, that had a compile-time feature enabled to get more SIMD things working so I updated the docker container for wasm tests to compile a local version of Wasmtime. This does add ~5min to the wasm builder, although Wasmtime will hopefully change in the near future so we can use the precompiled binaries again.

alexcrichton:
Ok should be good to go now!

Amanieu:
As part of #1066 I changed the intrinsics in `memory.rs` to use const generics. However I left the existing check for invalid values as a runtime check.

What do you think of changing this to use `static_assert!` like all the other intrinsics that were converted to use const generics?

alexcrichton:
Oh perfect! I was going to bring that up on the stabilization issue because I agree that these should all have static assertions that only valid constants can be provided. I'll update all wasm bits to have a static assertion.

alexcrichton:
Ok I've added a `static_assert!` for all const-generics used for wasm simd functions, and I've also updated the memory intrinsics to have a static assert that the memory is 0.

Amanieu:
Great work!

PR #1094, opened 2021-03-22
add vtrn1 vtrn2, vzip1, vzip2 neon instructions

SparrowLii:
This PR adds transpose vectors and Zip vectors neon instructions which are only avaliable in aarch64.


SparrowLii:
There seems to be some problems with the comparison program of github. In fact, this PR did not make any changes to the original generated code.

PR #1096, opened 2021-03-24
Tweak names of wasm SIMD intrinsics

alexcrichton:
The original intention of wasm SIMD support was to very closely match
the spec 1:1 to provide low-level primitives. Over time though this has
become less appealing and "niceties" such as `i64x2_shuffle` have been
added instead of only giving users `i8x16_shuffle` (same thing for
`*_const` functions).

This commit goes through and gives special treatment to all SIMD
intrinsics and instructions for wasm. This performs changes such as:

* All `*_u` and `*_s` suffixes are now baked into the intrinsic type,
  e.g. changing `i16x8` to `u16x8` where appropriate.
* `*_const` functions have been added for unsigned types
* `*_replace_lane` functions have been added for unsigned types
* `*_extract_lane` functions have been added for unsigned types

While I think it makes sense at the wasm simd spec layer that all these
instructions have the names they do, I've concluded that at the Rust
layer we'll want more understandable names since they're so easy to bind
and are easier to understand when reading/writing code.

### DIFF
--- a/crates/core_arch/src/wasm32/simd128.rs
+++ b/crates/core_arch/src/wasm32/simd128.rs
@@ -305,7 +305,7 @@ pub unsafe fn i16x8_load_extend_i8x8(m: *const i8) -> v128 {
 #[inline]
 #[cfg_attr(test, assert_instr(v128.load8x8_u))]
 #[target_feature(enable = "simd128")]
-pub unsafe fn u16x8_load_extend_u8x8(m: *const u8) -> v128 {
+pub unsafe fn i16x8_load_extend_u8x8(m: *const u8) -> v128 {
     transmute(simd_cast::<_, simd::u16x8>(*(m as *const simd::u8x8)))
 }
 
@@ -321,7 +321,7 @@ pub unsafe fn i32x4_load_extend_i16x4(m: *const i16) -> v128 {
 #[inline]
 #[cfg_attr(test, assert_instr(v128.load16x4_u))]
 #[target_feature(enable = "simd128")]
-pub unsafe fn u32x4_load_extend_u16x4(m: *const u16) -> v128 {
+pub unsafe fn i32x4_load_extend_u16x4(m: *const u16) -> v128 {
     transmute(simd_cast::<_, simd::u32x4>(*(m as *const simd::u16x4)))
 }
 
@@ -337,7 +337,7 @@ pub unsafe fn i64x2_load_extend_i32x2(m: *const i32) -> v128 {
 #[inline]
 #[cfg_attr(test, assert_instr(v128.load32x2_u))]
 #[target_feature(enable = "simd128")]
-pub unsafe fn u64x2_load_extend_u32x2(m: *const u32) -> v128 {
+pub unsafe fn i64x2_load_extend_u32x2(m: *const u32) -> v128 {
     transmute(simd_cast::<_, simd::u64x2>(*(m as *const simd::u32x2)))
 }
 
@@ -3218,19 +3218,19 @@ pub mod tests {
             let arr: [i8; 8] = [-3, -2, -1, 0, 1, 2, 3, 4];
             let vec = i16x8_load_extend_i8x8(arr.as_ptr());
             compare_bytes(vec, i16x8(-3, -2, -1, 0, 1, 2, 3, 4));
-            let vec = u16x8_load_extend_u8x8(arr.as_ptr() as *const u8);
+            let vec = i16x8_load_extend_u8x8(arr.as_ptr() as *const u8);
             compare_bytes(vec, i16x8(253, 254, 255, 0, 1, 2, 3, 4));
 
             let arr: [i16; 4] = [-1, 0, 1, 2];
             let vec = i32x4_load_extend_i16x4(arr.as_ptr());
             compare_bytes(vec, i32x4(-1, 0, 1, 2));
-            let vec = u32x4_load_extend_u16x4(arr.as_ptr() as *const u16);
+            let vec = i32x4_load_extend_u16x4(arr.as_ptr() as *const u16);
             compare_bytes(vec, i32x4(65535, 0, 1, 2));
 
             let arr: [i32; 2] = [-1, 1];
             let vec = i64x2_load_extend_i32x2(arr.as_ptr());
             compare_bytes(vec, i64x2(-1, 1));
-            let vec = u64x2_load_extend_u32x2(arr.as_ptr() as *const u32);
+            let vec = i64x2_load_extend_u32x2(arr.as_ptr() as *const u32);
             compare_bytes(vec, i64x2(u32::max_value().into(), 1));
         }
     }
--- a/crates/core_arch/src/wasm32/simd128.rs
+++ b/crates/core_arch/src/wasm32/simd128.rs
@@ -297,7 +297,7 @@ pub unsafe fn v128_load(m: *const v128) -> v128 {
 #[inline]
 #[cfg_attr(test, assert_instr(v128.load8x8_s))]
 #[target_feature(enable = "simd128")]
-pub unsafe fn i16x8_load8x8(m: *const i8) -> v128 {
+pub unsafe fn i16x8_load_extend_i8x8(m: *const i8) -> v128 {
     transmute(simd_cast::<_, i16x8>(*(m as *const i8x8)))
 }
 
@@ -305,7 +305,7 @@ pub unsafe fn i16x8_load8x8(m: *const i8) -> v128 {
 #[inline]
 #[cfg_attr(test, assert_instr(v128.load8x8_u))]
 #[target_feature(enable = "simd128")]
-pub unsafe fn u16x8_load8x8(m: *const u8) -> v128 {
+pub unsafe fn u16x8_load_extend_u8x8(m: *const u8) -> v128 {
     transmute(simd_cast::<_, u16x8>(*(m as *const u8x8)))
 }
 
@@ -313,7 +313,7 @@ pub unsafe fn u16x8_load8x8(m: *const u8) -> v128 {
 #[inline]
 #[cfg_attr(test, assert_instr(v128.load16x4_s))]
 #[target_feature(enable = "simd128")]
-pub unsafe fn i32x4_load16x4(m: *const i16) -> v128 {
+pub unsafe fn i32x4_load_extend_i16x4(m: *const i16) -> v128 {
     transmute(simd_cast::<_, i32x4>(*(m as *const i16x4)))
 }
 
@@ -321,7 +321,7 @@ pub unsafe fn i32x4_load16x4(m: *const i16) -> v128 {
 #[inline]
 #[cfg_attr(test, assert_instr(v128.load16x4_u))]
 #[target_feature(enable = "simd128")]
-pub unsafe fn u32x4_load16x4(m: *const u16) -> v128 {
+pub unsafe fn u32x4_load_extend_u16x4(m: *const u16) -> v128 {
     transmute(simd_cast::<_, u32x4>(*(m as *const u16x4)))
 }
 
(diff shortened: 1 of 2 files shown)


CryZe:
Are these only being renamed in Rust or C / C++ as well? Cause I find it a bit odd for Rust to diverge from C's intrinsics when this was something that was explicitly decided against for x86.

alexcrichton:
Personally I would hope that each architecture inside of the `core::arch` module can determine its own conventions as appropriate. For example x86 has *thousands* of intrinsics and the libs team didn't want to review them all individually so it made sense to use the preexisting standard of the intrinsic header files. In isolation, though, the x86 names don't make a ton of sense in Rust (e.g. they don't need `mm` prefixes and all that) and ideally we would have thought of better names for them (given infinite time)

For wasm we don't have the convenience of a preexisting standard to copy from, and additionally we're working with the scale of hundreds of intrinsics instead of thousands. These two combined lead me to believe that wasm should reserve the right to blaze it's own trail here and set the precedent. Functions like `memory_grow` and `memory_size` are trying to match the name of the instruction they're matching, but such a convention is not necessarily the best here given the simd intrinsics.

All that to say that while I think we can take inspiration from what clang is doing for C I do not think we should be bound to follow it exactly. For example if you look at clang's [header file](https://github.com/llvm/llvm-project/blob/main/clang/lib/Headers/wasm_simd128.h) for `wasm_simd128.h` you'll notice that all functions are prefixed with `wasm_*` which I don't think makes sense in Rust. Furthermore the header defines a number of vector types for each lane/type pair, but I don't think that makes sense in Rust because the implicit conversions in C aren't available in Rust (and would make the Rust API more verbose).

To reiterate, though, there's no point in intentionally being different from clang. It's not like clang made bad decisions or anything like that, my point is just that the decisions for C-the-language are not always the right decisions for Rust-the-language as well. As such I'm of the opinion that we should do what's right for Rust in this module, not just copy C. Language differences largely come up when handling types/conversions in this instance, but if you look at the header file for C you'll find that this PR actually brings most intrinsics to match names with the C header file.

crates/core_arch/src/wasm32/simd128.rs
```diff
@@ -297,47 +297,47 @@ pub unsafe fn v128_load(m: *const v128) -> v128 {
...
+pub unsafe fn i16x8_load8x8(m: *const i8) -> v128 {
     transmute(simd_cast::<_, i16x8>(*(m as *const i8x8)))
 }
 
 /// Load eight 8-bit integers and zero extend each one to a 16-bit lane
 #[inline]
 #[cfg_attr(test, assert_instr(v128.load8x8_u))]
 #[target_feature(enable = "simd128")]
-pub unsafe fn v128_load8x8_u(m: *const u8) -> v128 {
+pub unsafe fn u16x8_load8x8(m: *const u8) -> v128 {
```
Amanieu, line 308: Slight nit: I really feel that these intrinsics need an `ext` *somewhere* in the name to indicate what is going on. Perhaps `u16x8_loadext8x8`?

crates/core_arch/src/wasm32/simd128.rs
```diff
@@ -499,7 +499,7 @@ pub unsafe fn v128_store64_lane<const L: usize>(v: v128, m: *mut u64) {
         a15 = 15,
     )
 )]
-pub const unsafe fn v128_const(
+pub const unsafe fn i8x16_const(
```
Amanieu, line 502: The `const` suffix here feels a bit strange since the intrinsic works with runtime values as well. Have you considered just removing the suffix and naming the function just `i8x16`? That way it looks somewhat like a constructor in the code.

Amanieu:
Looking at the Clang header, it uses `vNxM` instead of `iNxM` when the signedness doesn't matter for the operation. Perhaps we should do the same? It feels a bit strange that some `iNxM` function have a `uNxM` counterpart while others don't.

crates/core_arch/src/wasm32/simd128.rs
alexcrichton: That seems reasonable to me, yeah. There's sort of two conventions in wasm simd, one with `extadd` and `extmul` but another with `extend`. Perhaps `u16x8_load8x8_extend`?

I don't really mind one way or another, I just figured it was better to use the `u16x8` name here instead of `v128` to represent the output type lanes.

crates/core_arch/src/wasm32/simd128.rs
alexcrichton: How about `i8x16_new`? I'm a bit hesitant to take the small name of `i8x16`, but only slightly, I could imagine it working as well.

alexcrichton:
I was just wondering that myself this morning about the `v*` prefix. It's worth noting that the `wasm_simd128.h` header has not been updated for the latest batch of renamings in the wasm simd proposal. Some functions there mirror old instruction names which have since been renamed. For example `wasm_v8x16_shuffle` mirrors the old `v8x16.shuffle` instruction, but it was decided to rename the wasm instruction in https://github.com/WebAssembly/simd/issues/316. From reading that issue it appears that the main motivation is that by renaming shuffles [and loads](https://github.com/WebAssembly/simd/issues/297) it was possible to remove the `vNNxMM` syntax entirely from the wasm spec.

I think that this naming only really applies to `*_load_splat` and `*_shuffle`, right? I'm personally fine keeping the `v*` prefix there since I do think it helps indicate that signedness doesn't matter, it's just about lanes of that width.

crates/core_arch/src/wasm32/simd128.rs
```diff
@@ -1413,7 +1610,7 @@ pub unsafe fn i64x2_gt_s(a: v128, b: v128) -> v128 {
 #[inline]
 // #[cfg_attr(test, assert_instr(i64x2.le_s))] // FIXME llvm
 #[target_feature(enable = "simd128")]
-pub unsafe fn i64x2_le_s(a: v128, b: v128) -> v128 {
+pub unsafe fn i64x2_le(a: v128, b: v128) -> v128 {
     transmute(simd_le::<_, i64x2>(a.as_i64x2(), b.as_i64x2()))
 }
```
Amanieu, line 1615: `u64x2_{gt,lt,ge,le}` seem to be missing.

Amanieu:
The `vNxM` naming could also apply to `vNxM_eq`/`vNxM_ne`, but that might be confusing since we do care about the type a bit. I'm not feeling too strongly about `vNxM` so let's just keep it as `iNxM` to match the spec.

crates/core_arch/src/wasm32/simd128.rs
alexcrichton: Oh that's intentional, they're not actually [part of the spec](https://webassembly.github.io/simd/core/binary/instructions.html#simd-instructions)

alexcrichton:
For comparisons we it would also apply only to integers because `f64x2_eq` naturally has different semantics than `i64x2_eq`.

I do find this an oddity though where if you use `u32x4_const` to create a vector you'd then execute `i32x4_add` to add two together. That originally made me hesitant to make this change but in the end I figured that at least for the sign-suffixed cases it just seems to universally make more sense to factor the sign into the type at a programming-language level for readability. We could always add `u32x4_add` in the future if necessary.

crates/core_arch/src/wasm32/simd128.rs
Amanieu: That seems like a glaring hole in the spec, is there a reason for it?

CryZe:
In general I'm definitely in favor of cleaning up the API, in fact I was about to write a clean wrapper API myself (probably as part of safe_arch which does this already for x86). So if this is happening in std, that would be even better. Though it's sad that this did not happen for x86.

Though I would then also be in favor of adding proper f32x4, f64x2, ... types. Most of the transmuting will come from iterating e.g. [f32] slices as f32x4 chunks. For this you will need to have some transmute solution anyway. I'd argue that e.g. using f64x2 intrinsics on f32x4 vectors is so rare that you also probably should be transmuting, just so the compiler can catch the likely mistake that this is otherwise. Also it just straight up seems weird to willingly introduce a weakly typed API into Rust which otherwise is very strongly typed (neither x86 nor ARM SIMD is this weakly typed either). It probably makes more sense to just introduce functions / methods on those types for safely converting between them as part of the std::arch::wasm32 API if the transmutation is considered too much of a problem.

crates/core_arch/src/wasm32/simd128.rs
alexcrichton: I was not part of the discussions myself, but some digging reveals https://github.com/WebAssembly/simd/pull/414.

At the very least it's not accidental that they're left out!

alexcrichton:
This is also one of the major reasons x86 *didn't* do anything more than the header file, it allows libstd to entirely side-step the question of "what if it were actually nice to use simd?" and instead just provides the tools to build something nice externally (hoping that someone would do so). Now that being said I think that building something nice externally for x86 is a lot more difficult than it is for wasm (at least as wasm simd stands today).

I'm personally not entirely sure what the best direction to go in wasm is. The wav project you pointed out I think is the exact API we'd want (more-or-less) in Rust if the goal of `std` was to have an ergonomic API for exposing all these intrinsics. A wav-like API built on what `std` currently has, however, could certainly also exist on crates.io.

The cost of doing something wav-like in this crate (and libstd) is basically just effort for stabilization. It would require much more scrutiny. Additionally I don't think it's obvious/clear that such an abstraction would be able to live on into the future. For example I couldn't say with certainty that all future wasm simd functionality could live in a well-typed API like that.

I think that the current state of the `core::arch::wasm32` module is a bit of a weird middle-ground. It doesn't exactly match the "specification" (which in this case would be the online specification, not C, so all we have is instructions and type signatures), but it also doesn't go the full 9 yards of being a nice and safe API to use. Instead it only provides some mild conveniences in a few locations. This feels weird to me because it's trying to be more ergonomic but it's unclear where the line is drawn and why at this point in time.

I don't really know what to do about any of this myself. I want wasm simd on a track to stable in a reasonable time frame, especially given that engines are picking up steam implementing the proposal. Beyond that though I don't know what the best long-term option is for this module.

crates/core_arch/src/wasm32/simd128.rs
Amanieu: I'd say try `i8x16` at first and then revisit this question when stabilizing to see how it worked out.

crates/core_arch/src/wasm32/simd128.rs
Amanieu: I'd keep the naming close to `i16x8_extend_low_i8x16` since the behavior is very similar:
- The destination type is always `i`. The result is always a valid signed integer even if we zero-extended from an unsigned value.
- The source type is either `i8x8` or `u8x8` depending on whether we are sign or zero extending.

So perhaps something like `i16x8_load_extend_i8x8`/`i16x8_load_extend_u8x8`?

alexcrichton:
Ok, should be green now!

I initially tried to get the `*_unaligned` changes here too but that led to [odd link errors](https://github.com/rust-lang/stdarch/runs/2187635060?check_suite_focus=true#step:14:1426) which I don't feel like debugging right now.

crates/core_arch/src/wasm32/simd128.rs
```diff
@@ -297,80 +297,80 @@ pub unsafe fn v128_load(m: *const v128) -> v128 {
...
+    transmute(simd_cast::<_, simd::i16x8>(*(m as *const simd::i8x8)))
 }
 
 /// Load eight 8-bit integers and zero extend each one to a 16-bit lane
 #[inline]
 #[cfg_attr(test, assert_instr(v128.load8x8_u))]
 #[target_feature(enable = "simd128")]
-pub unsafe fn v128_load8x8_u(m: *const u8) -> v128 {
-    transmute(simd_cast::<_, u16x8>(*(m as *const u8x8)))
+pub unsafe fn u16x8_load_extend_u8x8(m: *const u8) -> v128 {
```
Amanieu, line 308:
```suggestion
pub unsafe fn i16x8_load_extend_u8x8(m: *const u8) -> v128 {
```
This makes the naming consistent with `i16x8_extend_low_u8x16`. Same with the other 3 below.

alexcrichton: Makes sense to me, done now!

PR #1097, opened 2021-03-26
support for neon instructions vabal_* and vabal_high_*

surechen:
vabal_s*:
Signed Absolute difference and Accumulate Long. This instruction subtracts the vector elements in the lower or upper half of the second source SIMD&FP register from the corresponding vector elements of the first source SIMD&FP register, and accumulates the absolute values of the results into the vector elements of the destination SIMD&FP register. The destination vector elements are twice as long as the source vector elements.
vabal_u*:
Unsigned Absolute difference and Accumulate Long. This instruction subtracts the vector elements in the lower or upper half of the second source SIMD&FP register from the corresponding vector elements of the first source SIMD&FP register, and accumulates the absolute values of the results into the vector elements of the destination SIMD&FP register. The destination vector elements are twice as long as the source vector elements. All the values in this instruction are unsigned integer values.

### DIFF
--- a/crates/core_arch/src/aarch64/neon/generated.rs
+++ b/crates/core_arch/src/aarch64/neon/generated.rs
@@ -2727,6 +2727,66 @@ pub unsafe fn vzip2q_f64(a: float64x2_t, b: float64x2_t) -> float64x2_t {
     simd_shuffle2(a, b, [1, 3])
 }
 
+/// Unsigned Absolute difference and Accumulate Long
+#[inline]
+#[target_feature(enable = "neon")]
+#[cfg_attr(test, assert_instr(add))]
+pub unsafe fn vabal_high_u8(a: uint16x8_t, b: uint8x16_t, c: uint8x16_t) -> uint16x8_t {
+    let d: uint8x16_t = vabdq_u8(b, c);
+    let e: uint8x8_t = simd_shuffle8(d, d, [8, 9, 10, 11, 12, 13, 14, 15]);
+    simd_add(a, simd_cast(e))
+}
+
+/// Unsigned Absolute difference and Accumulate Long
+#[inline]
+#[target_feature(enable = "neon")]
+#[cfg_attr(test, assert_instr(add))]
+pub unsafe fn vabal_high_u16(a: uint32x4_t, b: uint16x8_t, c: uint16x8_t) -> uint32x4_t {
+    let d: uint16x8_t = vabdq_u16(b, c);
+    let e: uint16x4_t = simd_shuffle4(d, d, [4, 5, 6, 7]);
+    simd_add(a, simd_cast(e))
+}
+
+/// Unsigned Absolute difference and Accumulate Long
+#[inline]
+#[target_feature(enable = "neon")]
+#[cfg_attr(test, assert_instr(add))]
+pub unsafe fn vabal_high_u32(a: uint64x2_t, b: uint32x4_t, c: uint32x4_t) -> uint64x2_t {
+    let d: uint32x4_t = vabdq_u32(b, c);
+    let e: uint32x2_t = simd_shuffle2(d, d, [2, 3]);
+    simd_add(a, simd_cast(e))
+}
+
+/// Signed Absolute difference and Accumulate Long
+#[inline]
+#[target_feature(enable = "neon")]
+#[cfg_attr(test, assert_instr(add))]
+pub unsafe fn vabal_high_s8(a: int16x8_t, b: int8x16_t, c: int8x16_t) -> int16x8_t {
+    let d: int8x16_t = vabdq_s8(b, c);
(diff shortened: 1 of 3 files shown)


crates/stdarch-gen/neon.spec
```diff
@@ -1339,3 +1339,121 @@ validate 8., 9., 10., 11., 12., 13., 14., 15.
...
+name = vabal
+multi_fn = vabd-unsigned-noext, b, c, d:in_t
+multi_fn = simd_add, a, {simd_cast, d}
+a = 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16
+b = 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16
+c = 10, 10, 10, 10, 10, 10, 10, 10, 20, 0, 2, 4, 6, 8, 10, 12
+validate 10, 10, 10, 10, 10, 10, 10, 10, 20, 20, 20, 20, 20, 20, 20, 20
+
+arm = add
+aarch64 = add
```
Amanieu, line 1357: This looks suspicious. Why are you checking for an `add` instruction instead of the SIMD `sabal`/`uabal` (AArch64) or `vabal` (ARM) instruction.

crates/core_arch/src/aarch64/neon/generated.rs
```diff
@@ -2727,6 +2727,66 @@ pub unsafe fn vzip2q_f64(a: float64x2_t, b: float64x2_t) -> float64x2_t {
...
 }
 
+/// Unsigned Absolute difference and Accumulate Long
+#[inline]
+#[target_feature(enable = "neon")]
+#[cfg_attr(test, assert_instr(uabd))]
+pub unsafe fn vabal_high_u8(a: uint16x8_t, b: uint8x16_t, c: uint8x16_t) -> uint16x8_t {
+    let d: uint8x16_t = vabdq_u8(b, c);
+    let e: uint8x8_t = simd_shuffle8(d, d, [8, 9, 10, 11, 12, 13, 14, 15]);
+    simd_add(a, simd_cast(e))
```
Amanieu, line 2737: This implementation is incorrect. You need to extract the high part of each input and then call `vabal_u8`. See the clang implementation in `arm_neon.h`.

crates/core_arch/src/arm/neon/generated.rs
```diff
@@ -4413,6 +4413,72 @@ pub unsafe fn vrecpeq_f32(a: float32x4_t) -> float32x4_t {
...
+
+/// Signed Absolute difference and Accumulate Long
+#[inline]
+#[target_feature(enable = "neon")]
+#[cfg_attr(target_arch = "arm", target_feature(enable = "v7"))]
+#[cfg_attr(all(test, target_arch = "arm"), assert_instr("vabd.s8"))]
+#[cfg_attr(all(test, target_arch = "aarch64"), assert_instr(sabd))]
+pub unsafe fn vabal_s8(a: int16x8_t, b: int8x8_t, c: int8x8_t) -> int16x8_t {
+    let d: int8x8_t = vabd_s8(b, c);
+    simd_add(a, simd_cast(d))
```
Amanieu, line 4457: `simd_cast` is wrong here. You need to zero-extend each element of `d`, not sign-extend them. This is why you are not generating the correct instruction.

crates/core_arch/src/aarch64/neon/generated.rs
Amanieu: Actually I think the issue here is the same as the other one. The `simd_cast` should be changed to do zero-extension instead of sign-extension.

crates/core_arch/src/arm/neon/generated.rs
surechen: > `simd_cast` is wrong here. You need to zero-extend each element of `d`, not sign-extend them. This is why you are not generating the correct instruction.

Hi, thanks for reviewing. Very helpful to me. I havn't find simd_zext in simd_llvm.rs. Should I edit both compiler(use LLVMBuildZExt or others) and stdarch to supprot this ? I am very interested in this, but due to the lack of relevant knowledge, it may take a few days. 
I'd welcome any suggestions. Thank you very much.

Amanieu: You need to cast to `uint8x8_t` first before casting to `int16x8_t`.

crates/core_arch/src/aarch64/neon/generated.rs
surechen: > Actually I think the issue here is the same as the other one. The `simd_cast` should be changed to do zero-extension instead of sign-extension.

Done. Thanks

crates/stdarch-gen/neon.spec
surechen: > This looks suspicious. Why are you checking for an `add` instruction instead of the SIMD `sabal`/`uabal` (AArch64) or `vabal` (ARM) instruction.

Done. Thanks

PR #1105, opened 2021-04-02
add neon instruction vmaxnm_f* vpmaxnm_f* vminnm_f* vpminnm_f*

surechen:
vmaxnm_f* :  Floating-point Maximum Number (vector). This instruction compares corresponding vector elements in the two source SIMD&FP registers, writes the larger of the two floating-point values into a vector, and writes the vector to the destination SIMD&FP register.
vpmaxnm_f* :  Floating-point Maximum Number Pairwise (vector). This instruction creates a vector by concatenating the vector elements of the first source SIMD&FP register after the vector elements of the second source SIMD&FP register, reads each pair of adjacent vector elements in the two source SIMD&FP registers, writes the largest of each pair of values into a vector, and writes the vector to the destination SIMD&FP register. All the values in this instruction are floating-point values.
vminnm_f* :  Floating-point Minimum Number (vector). This instruction compares corresponding vector elements in the two source SIMD&FP registers, writes the smaller of the two floating-point values into a vector, and writes the vector to the destination SIMD&FP register.
vpminnm_f* :  Floating-point Minimum Number Pairwise (vector). This instruction creates a vector by concatenating the vector elements of the first source SIMD&FP register after the vector elements of the second source SIMD&FP register, reads each pair of adjacent vector elements in the two source SIMD&FP registers, writes the smallest of each pair of floating-point values into a vector, and writes the vector to the destination SIMD&FP register. All the values in this instruction are floating-point values.

### DIFF
--- a/crates/core_arch/src/aarch64/neon/generated.rs
+++ b/crates/core_arch/src/aarch64/neon/generated.rs
@@ -2536,32 +2536,6 @@ pub unsafe fn vmaxnmq_f64(a: float64x2_t, b: float64x2_t) -> float64x2_t {
     vmaxnmq_f64_(a, b)
 }
 
-/// Floating-point Maximun Number (vector)
-#[inline]
-#[target_feature(enable = "neon")]
-#[cfg_attr(test, assert_instr(fmaxnm))]
-pub unsafe fn vmaxnm_f32(a: float32x2_t, b: float32x2_t) -> float32x2_t {
-    #[allow(improper_ctypes)]
-    extern "C" {
-        #[cfg_attr(target_arch = "aarch64", link_name = "llvm.aarch64.neon.fmaxnm.v2f32")]
-        fn vmaxnm_f32_(a: float32x2_t, b: float32x2_t) -> float32x2_t;
-    }
-    vmaxnm_f32_(a, b)
-}
-
-/// Floating-point Maximun Number (vector)
-#[inline]
-#[target_feature(enable = "neon")]
-#[cfg_attr(test, assert_instr(fmaxnm))]
-pub unsafe fn vmaxnmq_f32(a: float32x4_t, b: float32x4_t) -> float32x4_t {
-    #[allow(improper_ctypes)]
-    extern "C" {
-        #[cfg_attr(target_arch = "aarch64", link_name = "llvm.aarch64.neon.fmaxnm.v4f32")]
-        fn vmaxnmq_f32_(a: float32x4_t, b: float32x4_t) -> float32x4_t;
-    }
-    vmaxnmq_f32_(a, b)
-}
-
 /// Floating-point Maximum Number Pairwise (vector).
 #[inline]
 #[target_feature(enable = "neon")]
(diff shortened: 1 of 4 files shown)


Amanieu:
`vmaxnm_f32`/`vmaxnmq_f32` (and the min equivalents) are also available on ARM.

surechen:
> vmaxnmq_f32

Hi, thanks for reviewing. I check these instructions in https://godbolt.org/ and get <Compilation failed>.

https://godbolt.org/z/1nM7eYvrz
https://godbolt.org/z/nMo63G89W

surechen:
> `vmaxnm_f32`/`vmaxnmq_f32` (and the min equivalents) are also available on ARM.

Hi, thanks for reviewing. I check these instructions in https://godbolt.org/ and get .

https://godbolt.org/z/1nM7eYvrz
https://godbolt.org/z/nMo63G89W

Amanieu:
These instruction were added in ARMv8 for both 32-bit and 64-bit mode. If you change `armv7` to `armv8` in godbolt then it will be accepted.

surechen:
> These instruction were added in ARMv8 for both 32-bit and 64-bit mode. If you change `armv7` to `armv8` in godbolt then it will be accepted.
Hi,  Thank you for your guidance. 
If I remove line #[cfg_attr(target_arch = "arm", target_feature(enable = "v7"))] . Is the following code right？

#[inline]
#[cfg_attr(all(test, target_arch = "arm"), assert_instr("vmaxnm"))]
#[cfg_attr(all(test, target_arch = "aarch64"), assert_instr(fmaxnm))]
pub unsafe fn vmaxnm_f32(a: float32x2_t, b: float32x2_t) -> float32x2_t {
    #[allow(improper_ctypes)]
    extern "C" {
        #[cfg_attr(target_arch = "arm", link_name = "llvm.arm.neon.vmaxnm.v2f32")]
        #[cfg_attr(target_arch = "aarch64", link_name = "llvm.aarch64.neon.fmaxnm.v2f32")]
        fn vmaxnm_f32_(a: float32x2_t, b: float32x2_t) -> float32x2_t;
    }
    vmaxnm_f32_(a, b)
}

Amanieu:
I think you need to enable the `v8` target feature for this instruction.

surechen:
> I think you need to enable the `v8` target feature for this instruction.

Ok, Thank you very much.

Amanieu:
I looked into the compiler crash. You need to enable the `fp-armv8` feature. But first you need to update rustc to expose this feature on ARM. This can be done by adding it to `compiler/rustc_codegen_ssa/src/target_features.rs`.

surechen:
> I looked into the compiler crash. You need to enable the `fp-armv8` feature. But first you need to update rustc to expose this feature on ARM. This can be done by adding it to `compiler/rustc_codegen_ssa/src/target_features.rs`.

Hi, Thank you very much. I'll try

Amanieu:
You need both `v8` and `fp-armv8`.

PR #1111, opened 2021-04-08
add vshl and vshr neon instructions

SparrowLii:
This PR adds vshl, vshl_n, vshll, vshll_high, vshr_n, vshrn, vshrn_high neon instructions.
It makes some modifications to the code generator to generate the correct suffixes.

### DIFF
--- a/crates/core_arch/src/aarch64/neon/generated.rs
+++ b/crates/core_arch/src/aarch64/neon/generated.rs
@@ -3640,6 +3640,132 @@ pub unsafe fn vreinterpretq_f32_f64(a: float64x2_t) -> float32x4_t {
     transmute(a)
 }
 
+/// Signed shift left long
+#[inline]
+#[target_feature(enable = "neon")]
+#[cfg_attr(test, assert_instr(sshll2, N = 2))]
+#[rustc_legacy_const_generics(1)]
+pub unsafe fn vshll_high_n_s8<const N: i32>(a: int8x16_t) -> int16x8_t {
+    static_assert!(N : i32 where N >= 0 && N <= 8);
+    let b: int8x8_t = simd_shuffle8(a, a, [8, 9, 10, 11, 12, 13, 14, 15]);
+    vshll_n_s8::<N>(b)
+}
+
+/// Signed shift left long
+#[inline]
+#[target_feature(enable = "neon")]
+#[cfg_attr(test, assert_instr(sshll2, N = 2))]
+#[rustc_legacy_const_generics(1)]
+pub unsafe fn vshll_high_n_s16<const N: i32>(a: int16x8_t) -> int32x4_t {
+    static_assert!(N : i32 where N >= 0 && N <= 16);
+    let b: int16x4_t = simd_shuffle4(a, a, [4, 5, 6, 7]);
+    vshll_n_s16::<N>(b)
+}
+
+/// Signed shift left long
+#[inline]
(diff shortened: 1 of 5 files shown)


SparrowLii:
I used manual implementation in vshrq_n_s8 and vshrq_n_u8, because on the aarch64 architecture, calling `simd_shr` on the `i8x16` type seems to have some problems.
Actually, when I test it like this:

    let a = i8x16::new(4, 8, 12, 16, 20, 24, 28, 32, 36, 40, 44, 48, 52, 56, 60, 64);
    let b = i8x16(2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2);
    unsafe {
        let c = simd_shr(a, b);
        println!("{:?}", c);
    }

On my x86_64, it's output is OK:

    i8x16(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16)

But on my aarch64, it will get the wrong result:

    i8x16(1, 1, 3, 2, 5, 3, 7, 4, 9, 5, 11, 6, 13, 7, 15, 8)

The strange part is that in this [commit](https://github.com/rust-lang/stdarch/pull/1111/commits/b8bdeb5d58c9a96d2b15c7b10e532e51a3d16863), no such error occurred in the CI test. 
I am not sure under what circumstances this error will occur. To be safe, I used manual implementation on both vshrq_n_s8 and vshrq_n_u8.

Amanieu:
Does the incorrect result with `simd_shr` only happen on your aarch64 machine? Does it happen when running on CI?

SparrowLii:
It only happened on my aarch64. It is all right when run the CI.

Amanieu:
If you can give me a small self-contained test program I can look into this issue. I would prefer if we use `simd_shr` here like we do for the other types.

SparrowLii:
[Here](https://github.com/SparrowLii/stdarch/commit/e195aa2db6d7cbe66996eff31219e2cbf89c5e1d) is all my test program. 
It's OK if we use `simd_shr` uniformly, as well as we can see if others have the same problem.

Amanieu:
This is a bug in LLVM:
```
# Debug
0000000000401a38 <simd_shr>:
  401a38:       3dc00000        ldr     q0, [x0]
  401a3c:       3dc00021        ldr     q1, [x1]
  401a40:       6e60b821        neg     v1.8h, v1.8h <---- THIS IS WRONG
  401a44:       4e214400        sshl    v0.16b, v0.16b, v1.16b
  401a48:       3d800100        str     q0, [x8]
  401a4c:       d65f03c0        ret

# Release
000000000040181c <simd_shr>:
  40181c:       3dc00020        ldr     q0, [x1]
  401820:       3dc00001        ldr     q1, [x0]
  401824:       6e20b800        neg     v0.16b, v0.16b
  401828:       4e204420        sshl    v0.16b, v1.16b, v0.16b
  40182c:       3d800100        str     q0, [x8]
  401830:       d65f03c0        ret
```

It doesn't happen in CI because we build with optimizations enabled.

I will open a bug in LLVM for this.

SparrowLii:
Thanks for helping!

Amanieu:
Let's just merge it with `simd_shr`, we can fix the LLVM bug in rustc separately.

SparrowLii:
That's all right

Amanieu:
https://bugs.llvm.org/show_bug.cgi?id=49904

PR #1104, opened 2021-04-01
[DRAFT] intrinsics for all architectures appear in rustdoc

Byron:
I will keep pushing changes until all files mentioned in #1055 have been adjusted, and would be glad if you could let me know early if I generally get it right 😅.

This PR starts out with a sample on how I would go about this.
Thanks for your help.

PR #1114, opened 2021-04-12
Allow primitive types in the code generator and add vdup instructions

SparrowLii:
This PR does the following:
1. Enhance the code generator so that primitive types can be used
2. Enhance code generator to generate more flexible suffixes
3. Add vdup neon instructions to test 1 and 2.

### DIFF
--- a/crates/stdarch-gen/neon.spec
+++ b/crates/stdarch-gen/neon.spec
@@ -890,6 +890,133 @@ aarch64 = fcvtpu
 link-aarch64 = fcvtpu._EXT2_._EXT_
 generate float32x2_t:uint32x2_t, float32x4_t:uint32x4_t, float64x1_t:uint64x1_t, float64x2_t:uint64x2_t
 
+/// Set all vector lanes to the same value
+name = vdup
+lane-suffixes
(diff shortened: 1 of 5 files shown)


crates/stdarch-gen/neon.spec
```diff
@@ -890,6 +890,148 @@ aarch64 = fcvtpu
...
+name = vdup
+lane-suffixes
+constn = N
+multi_fn = static_assert_imm-in_exp_len-N
+multi_fn = a
+a = 0
+n = HFLEN
+validate 0
+
+aarch64 = str
```
Amanieu, line 959: Rather than using a "fake" instruction like this, could you instead somehow mark this intrinsic as not requiring an instruction test?

Amanieu: This can cause issues like https://github.com/rust-lang/stdarch/pull/1104#issuecomment-817437987 when running the tests in release mode.

SparrowLii: It seems that the result of llvm is different in release mode? I will try to correct the test.

PR #1116, opened 2021-04-13
add neon instruction vfma

surechen:
vfma：Floating-point fused Multiply-Add to accumulator (vector). This instruction multiplies corresponding floating-point values in the vectors in the two source SIMD&FP registers, adds the product to the corresponding vector element of the destination SIMD&FP register, and writes the result to the destination SIMD&FP register.

### DIFF
--- a/crates/stdarch-gen/neon.spec
+++ b/crates/stdarch-gen/neon.spec
@@ -1546,16 +1546,12 @@ generate poly8x16_t:poly8x16_t:poly16x8_t
 
 /// Floating-point fused Multiply-Add to accumulator(vector)
 name = vfma
-no-q
 a = 2.0, 3.0
 b = 6.0, 4.0
 c = 8.0, 18.0
 validate 20.0, 30.0
 
-target = fp-armv8
-arm = vfma
 aarch64 = fma
-link-arm = llvm.fma._EXT_
 link-aarch64 = llvm.fma._EXT_
 generate float64x1_t
 
--- a/crates/stdarch-gen/neon.spec
+++ b/crates/stdarch-gen/neon.spec
@@ -1544,6 +1544,28 @@ validate 9, 30, 11, 20, 13, 18, 15, 48
 aarch64 = pmull
 generate poly8x16_t:poly8x16_t:poly16x8_t
 
+/// Floating-point fused Multiply-Add to accumulator(vector)
+name = vfma
+no-q
+a = 2.0, 3.0
+b = 6.0, 4.0
+c = 8.0, 18.0
+validate 20.0, 30.0
+
+target = fp-armv8
+arm = vfma
+aarch64 = fma
+link-arm = llvm.fma._EXT_
+link-aarch64 = llvm.fma._EXT_
+generate float64x1_t
+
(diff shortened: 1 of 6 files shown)


crates/stdarch-gen/neon.spec
```diff
@@ -1544,6 +1544,28 @@ validate 9, 30, 11, 20, 13, 18, 15, 48
 aarch64 = pmull
 generate poly8x16_t:poly8x16_t:poly16x8_t
 
+/// Floating-point fused Multiply-Add to accumulator(vector)
+name = vfma
+no-q
```
Amanieu, line 1549: There are q variants of this intrinsic.

crates/stdarch-gen/neon.spec
```diff
@@ -1544,6 +1544,28 @@ validate 9, 30, 11, 20, 13, 18, 15, 48
...
+b = 6.0, 4.0
+c = 8.0, 18.0
+validate 20.0, 30.0
+
+target = fp-armv8
+arm = vfma
+aarch64 = fma
+link-arm = llvm.fma._EXT_
+link-aarch64 = llvm.fma._EXT_
+generate float64x1_t
```
Amanieu, line 1560: The `f64` versions of this intrinsic are only available on aarch64 according to clang and gcc. The `float64x1_t` type is also aarch64-only.

surechen: Done. Thanks.

surechen: > The `f64` versions of this intrinsic are only available on aarch64 according to clang and gcc. The `float64x1_t` type is also aarch64-only.

Done. Thanks.

Amanieu:
Can you add the `f32x4` and `f64x2` variants as well?

surechen:
> Can you add the `f32x4` and `f64x2` variants as well?

Done. Thanks.

crates/stdarch-gen/neon.spec
```diff
@@ -1544,6 +1544,28 @@ validate 9, 30, 11, 20, 13, 18, 15, 48
...
 generate poly8x16_t:poly8x16_t:poly16x8_t
 
+/// Floating-point fused Multiply-Add to accumulator(vector)
+name = vfma
+a = 2.0, 3.0, 4.0, 5.0
+b = 6.0, 4.0, 7.0, 8.0
+c = 8.0, 18.0, 12.0, 10.0
+validate 20.0, 30.0, 40.0, 50.0
+
+aarch64 = fma
```
Amanieu, line 1554: This should be `fmadd`?

surechen: > This should be `fmadd`?

Hi, thanks for reviewing. the way I judge at present is mainly to look at the instruction name, and then according to the result of the disassembly check, fmadd can also pass the check. Is there any other verification method?

Amanieu: You can usually find the instruction here: https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics

It only shows the AArch64 instruction though.

PR #1122, opened 2021-04-17
add neon instruction vfma_n_*

surechen:
vfma_n_* :  Floating-point fused Multiply-Add to accumulator (vector). This instruction multiplies corresponding floating-point values in the vectors in the two source SIMD&FP registers, adds the product to the corresponding vector element of the destination SIMD&FP register, and writes the result to the destination SIMD&FP register.


crates/core_arch/src/aarch64/neon/generated.rs
```diff
@@ -2810,6 +2810,27 @@ pub unsafe fn vfmaq_f64(a: float64x2_t, b: float64x2_t, c: float64x2_t) -> float
...
+
+/// Floating-point fused Multiply-Add to accumulator(vector)
+#[inline]
+#[target_feature(enable = "neon")]
+#[cfg_attr(test, assert_instr(fmla))]
+pub unsafe fn vfmaq_n_f64(a: float64x2_t, b: float64x2_t, c: f64) -> float64x2_t {
+    let d: f64x2 = f64x2::new(0.0, 0.0);
+    let e: float64x2_t = simd_insert(transmute(d), 0, c);
+    let f: float64x2_t = simd_shuffle2(e, e, [0, 0]);
+    vfmaq_f64(b, f, a)
```
Amanieu, line 2831: This could be much simpler:
```suggestion
    let d: f64x2 = f64x2::new(c, c);
    vfmaq_f64(b, d, a)
```

surechen: Done. Thanks. 

I get a new check error : vfmaq_n_f32_vfma' panicked at 'instruction found, but the disassembly contains too many instructions: #instructions = 26 >= 22 (limit). 

 I avoided checking this instruction.

PR #1130, opened 2021-04-20
Add vqdmul* neon instructions

SparrowLii:
This PR does the following:
1. Add vqdmull, vqdmlal, vqdmlsl, vqdmulh neon instructions.
2. Correct the links of vqadd and vqsub to be consistent with Clang.

### DIFF
--- a/crates/core_arch/src/aarch64/neon/generated.rs
+++ b/crates/core_arch/src/aarch64/neon/generated.rs
@@ -3357,7 +3357,7 @@ pub unsafe fn vqdmullh_s16(a: i16, b: i16) -> i32 {
 /// Signed saturating doubling multiply long
 #[inline]
 #[target_feature(enable = "neon")]
-#[cfg_attr(test, assert_instr(sqdmulls))]
+#[cfg_attr(test, assert_instr(sqdmull))]
 pub unsafe fn vqdmulls_s32(a: i32, b: i32) -> i64 {
     #[allow(improper_ctypes)]
     extern "C" {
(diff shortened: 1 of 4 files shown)


SparrowLii:
There seems to be bugs in the llvm.aarch64.neon.*.i32 (.i64) link types, so I currently cancel the instruction to add parameters using primitive types

Amanieu:
What is the bug? Does the same issue happen in Clang?

SparrowLii:
Look this [CI fail](https://github.com/rust-lang/stdarch/runs/2378839348) in #1127 , `llvm.aarch64.neon.srshl.i64` went to error `LLVM ERROR: Cannot select: 0x7f88213835b0: i64 = AArch64ISD::VSHL Constant:i64<1>, Constant:i32<2>` under the target `aarch64-linux-android`. But it did not fail on other aarch64 targets. I don't know the reason for it.

SparrowLii:
So I now avoid using primitive type suffixes in linking (for example, using .v1i64 instead of .i64), but this will cause the compiler to not be able to optimize it into the instructions we want.

Amanieu:
I can reproduce the crash in Clang ([godbolt](https://godbolt.org/z/aebM19oqs)), I will open a bug on LLVM.

Amanieu:
https://bugs.llvm.org/show_bug.cgi?id=50057

Amanieu:
We can add back the missing intrinsics once the LLVM bug is fixed.

PR #1139, opened 2021-04-23
add vcopy neon instructions

SparrowLii:
This PR adds vcopy neon instructions which are only available on aarch64


Amanieu:
Do you think you could add ARM versions of these intrinsics as well?

SparrowLii:
These instructions are dedicated to aarch64, but we can use the same implementation for arm in stdarch. In llvm, the related implementations are only provided for aarch64, and they can be optimized to `mov` instructions. I am not sure whether they can be optimized in arm.

PR #1140, opened 2021-04-23
remove unused const_fn feature

RalfJung:
This does not seem to be needed any more (and the feature is about to be removed and replaced by more fine-grained features).


RalfJung:
What does it take to get this updated in the rustc repo? Can I just send a PR updating the submodule?

bjorn3:
https://github.com/rust-lang/rust/pull/83278 bumps stdarch though to a version before this PR. There was a change on master of stdarch that required a crater run and some fixes in several crates using stdarch.

RalfJung:
Ah I see, thanks!

PR #1144, opened 2021-04-28
Add vld1_dup and vld1_lane neon instructions

SparrowLii:
This PR add vld1_dup and vld1_lane instructions.


SparrowLii:
There are already implementations of these instructions in arm_shared/neon/mod.rs, so I discarded this PR

Issue #993, opened 2021-02-03
The AArch64 feature list is incomplete.


jacobbramley:
The set of features in `arch/aarch64.rs` is rather behind the architecture (and what's described by Linux's `auxv`).

I went to bring the list up to date, and noticed that a few are actually listed (in `os/linux/aarch64.rs`), but commented out. Are they deliberately omitted? If so, what criteria are used for inclusion here?

Amanieu:
It's just because these features are not supported [here](https://github.com/rust-lang/rust/blob/master/compiler/rustc_codegen_ssa/src/target_features.rs#L35) in the compiler. You are welcome to update both to add the latest features.

jacobbramley:
Oh, thanks for that pointer! I will do that; it makes sense to keep these in agreement.

Actually _making use_ of them in the compiler will be rather more difficult, of course. Can I assume that all that is required is awareness (e.g. for `asm!`)?

Amanieu:
The main concern is that the compiler and stdarch (and LLVM to some extent) agree on the same name for a particular feature. You should check against the feature names that LLVM uses in https://github.com/llvm/llvm-project/blob/main/llvm/lib/Target/AArch64/AArch64.td, though rustc has some renaming logic for some feature names.

jacobbramley:
Yes, I saw reference to that in a comment in rustc (in your first link).
I think there may be a missing translation there, actually: "fp16" is called "fphp" in Linux and the tests don't detect it (when other tools do). I've not followed the logic through the whole stack yet; consider it a strong hunch :-) Since finding hardware with _all_ of the features is difficult, I intend to test what I can, and audit the the whole list to check that there aren't similar issues elsewhere.

Amanieu:
The feature detection code was fixed in #1005. However the submodule in rust-lang/rust hasn't been updated yet.

PR #1146, opened 2021-04-28
Update aarch64 linux feature detection

adamgemmell:
All features supported by auxv (see https://github.com/torvalds/linux/blob/master/arch/arm64/include/uapi/asm/hwcap.h) / cpuinfo (https://github.com/torvalds/linux/blob/master/arch/arm64/kernel/cpuinfo.c) and in LLVM 12 are added.

See `llvm-project/llvm/lib/Target/AArch64/AArch64.td` for the LLVM feature names and dependencies. We've done a few passes for silly cross-referencing mistakes ourselves so I'm reasonably confident in its accuracy.

I had a couple notes after working on this:

- Some features, though detection is easy, are still not added based on @Amanieu 's wording [here](https://github.com/rust-lang/stdarch/issues/993#issuecomment-772893003). Is it possible to add detection for a few of the remaining features but not add them to the allowed list in rustc? I don't see why users shouldn't be able to check for these features themselves.
- usize::BITS is [stabilised](https://github.com/rust-lang/rust/issues/76904) but is not yet in stable Rust. I don't THINK this is a problem but wanted to highlight it.

See #993 
See also: https://github.com/rust-lang/rust/pull/84665

### DIFF
--- a/crates/core_arch/src/arm_shared/crypto.rs
+++ b/crates/core_arch/src/arm_shared/crypto.rs
@@ -57,7 +57,10 @@ use stdarch_test::assert_instr;
 /// AES single round encryption.
 #[inline]
 #[cfg_attr(any(bootstrap, target_arch = "arm"), target_feature(enable = "crypto"))]
-#[cfg_attr(not(any(bootstrap, target_arch = "arm")), target_feature(enable = "aes"))]
+#[cfg_attr(
+    not(any(bootstrap, target_arch = "arm")),
+    target_feature(enable = "aes")
+)]
 #[cfg_attr(target_arch = "arm", target_feature(enable = "v8"))]
 #[cfg_attr(test, assert_instr(aese))]
 pub unsafe fn vaeseq_u8(data: uint8x16_t, key: uint8x16_t) -> uint8x16_t {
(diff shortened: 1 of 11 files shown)


SparrowLii:
This [change](https://github.com/rust-lang/rust/pull/84216) makes your PR failed to compile. You can rebase to the latest master branch to avoid this error.

adamgemmell:
These changes will only work when my changes to rustc make it to the nightly compiler, see https://github.com/rust-lang/rust/pull/84665

Hence no reason to run the CI yet

adamgemmell:
I missed this in my testing but armv7 still requires crypto for the intrinsics. This was fixed in https://github.com/llvm/llvm-project/commit/b8baa2a9132498ea286dbb0d03f005760ecc6fdb#diff-e65dd2f7194df68bc6e9b754449ae241a6717f77ca1a28ee9479990ffe292049R7230 but that commit needs to make it to the minimum LLVM version before it can be removed fully

Amanieu:
We can cherry-pick that commit into rust-lang/llvm-project and then update the llvm-project submodule in rust-lang/rust.

Amanieu:
Hmm on second thought that might be a breaking change that would prevent cross-lang LTO with Clang 12. So in the end we will most likely have to wait for LLVM 13.

PR #1147, opened 2021-04-29
Add vmul_n, vmul_lane, vmulx neon instructions

SparrowLii:
This PR completes the vmul and vmulx instructions

### DIFF
--- a/crates/stdarch-gen/neon.spec
+++ b/crates/stdarch-gen/neon.spec
@@ -338,7 +338,7 @@ generate float32x2_t:uint32x2_t, float32x4_t:uint32x4_t, float64x1_t:uint64x1_t,
 
 /// Signed compare bitwise Test bits nonzero
 name = vtst
-multi_fn = simd_and, c:in_t
+multi_fn = simd_and, c:in_t, a, b
 multi_fn = fixed, d:in_t
 multi_fn = simd_ne, c, transmute(d)
 a = MIN, 0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B, 0x0C, 0x0D, MAX
@@ -354,7 +354,7 @@ generate int8x8_t:uint8x8_t, int8x16_t:uint8x16_t, int16x4_t:uint16x4_t, int16x8
 
 /// Unsigned compare bitwise Test bits nonzero
 name = vtst
-multi_fn = simd_and, c:in_t
+multi_fn = simd_and, c:in_t, a, b
 multi_fn = fixed, d:in_t
 multi_fn = simd_ne, c, transmute(d)
 a = MIN, 0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B, 0x0C, 0x0D, MAX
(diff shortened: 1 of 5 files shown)


crates/stdarch-gen/neon.spec
```diff
@@ -1955,6 +2069,143 @@ validate 9, 30, 11, 20, 13, 18, 15, 48
...
 
+/// Polynomial multiply long
+name = vmull_high
+no-q
+multi_fn = vmull-noqself-noext, {simd_extract, a, 1}, {simd_extract, b, 1}
+a = 1, 15
+b = 1, 3
+validate 17
+
+aarch64 = pmull2
```
Amanieu, line 2080: `pmull` with the `p128` type should depend on the `crypto` feature.

Amanieu: While you're at it, could you move `vmull_p64` into the code generator as well? It should be available on both arm and aarch64 with the `crypto` feature enabled.

SparrowLii: So should we use `#[target_feature(enable = "neon, crypto")]`?

Amanieu: Yes. Although keep in mind that in the future we will split the `crypto` feature into separate `aes` and `sha2` features to match LLVM.

SparrowLii: Ok. But aes and sha2feature do not seem to affect the implementation of vmull

SparrowLii: I did not find the implementation of vmull_p64 on arm in llvm, so I don’t know what link should be used

Amanieu: I checked Clang and it seems to be missing this intrinsic on ARM. However I checked the LLVM source code and you should be able to make it generate the correct instruction with `llvm.arm.neon.vmullp.v2i64`. However I have not tested it so it may not work.

SparrowLii: It looks like it cannot compile successfully:

```
#[cfg(target_arch = "arm")]
#[target_feature(enable = "neon,crypto,v8")]
pub unsafe fn vmull_p64(a: p64, b: p64) -> p128 {
    #[allow(improper_ctypes)]
    extern "C" {
        #[cfg_attr(target_arch = "arm", link_name = "llvm.arm.neon.vmullp.v2i64")]
        fn vmull_p64_(a: int64x1_t, b: int64x1_t) -> int64x2_t;
    }
transmute(vmull_p64_(transmute(a), transmute(b)))
}
```
```
Intrinsic has incorrect argument type!
<2 x i64> (<1 x i64>, <1 x i64>)* @llvm.arm.neon.vmullp.v2i64
in function _ZN7example9vmull_p6417hc3e4f087b1fb350cE
LLVM ERROR: Broken function found, compilation aborted!
```
I tried to find a solution in the source code of llvm, but it seems that it should use v2i64. I don't know where the problem is
```
  def  VMULLp8   :  N3VLInt<0, 1, 0b00, 0b1110, 0, IIC_VMULi16D, "vmull", "p8",
                            v8i16, v8i8, int_arm_neon_vmullp, 1>;
  def  VMULLp64  : N3VLIntnp<0b00101, 0b10, 0b1110, 0, 0, NoItinerary,
                          "vmull", "p64", v2i64, v1i64, int_arm_neon_vmullp, 1>,
                    Requires<[HasV8, HasCrypto]>;
```

PR #1157, opened 2021-05-10
Add vmull_p64 and vmull_high_p64 for aarch64

### LINKED ISSUE #1156: Why `vmull_p64 ` was removed?
the `aarch64` [vmull_p64](https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics?search=vmull_p64)  instruction seems to have been deleted in `nightly rust`. what happened?

Stable Rust: https://doc.rust-lang.org/core/arch/aarch64/fn.vmull_p64.html

SparrowLii:
FIXES #1156 
This PR serves as a supplement to #1147 
Although vmull_p64 is temporarily unavailable on Arm32 due to the llvm‘s unsupporting, we can still make it available on aarch64.

### DIFF
--- a/crates/core_arch/src/aarch64/neon/generated.rs
+++ b/crates/core_arch/src/aarch64/neon/generated.rs
@@ -4648,12 +4648,33 @@ pub unsafe fn vmull_high_u32(a: uint32x4_t, b: uint32x4_t) -> uint64x2_t {
 
 /// Polynomial multiply long
 #[inline]
+#[target_feature(enable = "neon,crypto")]
+#[cfg_attr(test, assert_instr(pmull))]
+pub unsafe fn vmull_p64(a: p64, b: p64) -> p128 {
+    #[allow(improper_ctypes)]
+    extern "C" {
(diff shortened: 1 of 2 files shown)


LuoZijun:
Why they have different type signatures ？

stdarch:

```rust
pub type p64  = u64;
pub type p128 = u128;

pub unsafe fn vmull_p64(a: p64, b: p64) -> p128 {
     ...
}

```

AARCH64 Neon Intrinsics: https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics?search=vmull_p64

```c
poly128_t  vmull_p64 (poly64_t a, poly64_t b)
poly128_t  vmull_high_p64 (poly64x2_t a, poly64x2_t b)
```

SparrowLii:
This is to be consistent with other primitive variables(`u64`, `s64`). `float64_t` will also be replaced with `f64` in the function signature.
You can look at https://github.com/rust-lang/stdarch/blob/master/crates/stdarch-verify/src/lib.rs#L252 and https://github.com/rust-lang/stdarch/blob/master/crates/stdarch-verify/tests/arm.rs#L824. Both `p64` and `poly64_t` are regarded as `P64` types

PR #1164, opened 2021-05-14
Add support for the remaining vget(q)_lane functions.

JamieCunliffe:


Amanieu:
Have you considered implementing these using `stdarch-gen` instead of writing each one manually?

Amanieu:
I feel that using code generation makes maintenance easier since there is much less code to write and review. I would prefer if this was implemented using stdarch-gen.

Amanieu:
I'm just going to accept these as they are.

PR #1167, opened 2021-05-17
Modify the implementation of d_s64 suffix instructions

SparrowLii:
This PR revised the implementation of the d_s64 suffix neon instructions to be consistent with Clang

### DIFF
--- a/crates/core_arch/src/aarch64/neon/generated.rs
+++ b/crates/core_arch/src/aarch64/neon/generated.rs
@@ -7866,12 +7866,7 @@ pub unsafe fn vsetq_lane_f64<const LANE: i32>(a: f64, b: float64x2_t) -> float64
 #[target_feature(enable = "neon")]
 #[cfg_attr(test, assert_instr(sshl))]
 pub unsafe fn vshld_s64(a: i64, b: i64) -> i64 {
-    #[allow(improper_ctypes)]
-    extern "C" {
-        #[cfg_attr(target_arch = "aarch64", link_name = "llvm.aarch64.neon.sshl.i64")]
-        fn vshld_s64_(a: i64, b: i64) -> i64;
-    }
-    vshld_s64_(a, b)
+    transmute(vshl_s64(transmute(a), transmute(b)))
 }
 
 /// Unsigned Shift left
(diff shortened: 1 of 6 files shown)


SparrowLii:
I think `vshl` may have a bug in llvm. It is compiled correctly in [godbolt](https://rust.godbolt.org/z/naP96vK4j) but it failed in CI:
`LLVM ERROR: Cannot select: 0x7fb987899d00: i64 = AArch64ISD::VSHL Constant:i64<1>, Constant:i32<2>, crates/core_arch/src/aarch64/neon/generated.rs:7874:5 @[ crates/core_arch/src/aarch64/neon/generated.rs:15577:32 @[ crates/core_arch/src/aarch64/neon/generated.rs:15572:5 @[ crates/core_arch/src/aarch64/neon/generated.rs:15572:5 @[ /rustc/fe72845f7bb6a77b9e671e6a4f32fe714962cec4/library/core/src/ops/function.rs:227:5 ] ] ] ]
  0x7fb985815750: i64 = Constant<1>
  0x7fb9d0d1e410: i32 = Constant<2>
In function: _ZN4core3ops8function6FnOnce9call_once17h1592c9ba5b3050ecE`

Amanieu:
You can reproduce the error in godbolt by adding a test case which uses constant inputs: https://rust.godbolt.org/z/nPq57xxso

I can think of 2 ways to fix this:
- Just use normal shift instructions instead of intrinsics.
- Use `int64x1_t` for the intrinsic arguments instead of `i64`.

SparrowLii:
Yes, it seems that both i64 and int64x1_t will get the same assembly instructions. Why I changed the suffix. My idea was to use the same link as Clang, so that llvm can be optimized to get the desired result. (look `vrsrad_n_s64`) But it doesn’t seem to affect.

SparrowLii:
It seems that there is something wrong with Cirrus CI, can we fix it?

    error: error: 'sysinfo not supported on this platform'
    info: installing component 'clippy'
    error: error: 'sysinfo not supported on this platform'
    info: installing component 'rust-std'
    error: error: 'sysinfo not supported on this platform'

PR #1171, opened 2021-05-24
Add vst1_* neon intrinsics.

hasenbanck:
I took the vld1_* implementation as an example while writing the vst1_* implementation.

While the arm intrinsics use the LLVM provided intrinsics, I had to copy the approach taken by the SSE unaligned store intrinsics for the aarch64 implementation.

I also ported the SSE tests for the unaligned stores.

I'm not really sure about the `#[cfg_attr(test, assert_instr(*))]` asserts, since I couldn't find a away how to verify these tests. It would be great if somebody could hint me how to tests these.


crates/core_arch/src/aarch64/neon/mod.rs
```diff
@@ -795,6 +796,292 @@ pub unsafe fn vld1q_f64(ptr: *const f64) -> float64x2_t {
...
+}
+
+/// Store multiple single-element structures from one, two, three, or four registers.
+#[inline]
+#[target_feature(enable = "neon")]
+#[cfg_attr(test, assert_instr(str))]
+#[allow(clippy::cast_ptr_alignment)]
+pub unsafe fn vst1q_s16(ptr: *mut i16, a: int16x8_t) {
+    copy_nonoverlapping(
+        &a as *const int16x8_t as *const u8,
```
Amanieu, line 845: I checked with Clang and these aren't actually fully unaligned stores. Instead the pointers are required to be aligned to the element size (`i16`, `i32`, `i64`, etc). You can do the same here by using that type as the element for `copy_nonoverlapping` instead of `u8`.

hasenbanck: Fixed as adviced.

crates/core_arch/src/aarch64/neon/mod.rs
```diff
@@ -795,6 +796,292 @@ pub unsafe fn vld1q_f64(ptr: *const f64) -> float64x2_t {
...
+#[cfg_attr(test, assert_instr(str))]
+#[allow(clippy::cast_ptr_alignment)]
+pub unsafe fn vst1q_f32(ptr: *mut f32, a: float32x4_t) {
+    copy_nonoverlapping(
+        &a as *const float32x4_t as *const f32,
+        ptr as *mut f32,
+        size_of::<float32x4_t>(),
+    )
+}
+
```
SparrowLii, line 1084: Can you add `vst1_f64` and `vst1_p64` instructions at the same time?:)

hasenbanck: Oh snap, I forgot those? Sure, I can add them too.

hasenbanck: I added the missing intrinsics for aarch64.

PR #1177, opened 2021-05-28
wasm: Mark most simd intrinsics as safe

alexcrichton:
This updates all simd intrinsics to be safe to account for
rust-lang/rust#84988. In the spirit of more safety it also removes the
pervasive usage of `transmute` in the intrinsics in favor of more
targeted casts.

### DIFF
--- a/crates/core_arch/src/lib.rs
+++ b/crates/core_arch/src/lib.rs
@@ -34,7 +34,6 @@
     adx_target_feature,
     rtm_target_feature,
     f16c_target_feature,
-    external_doc,
     allow_internal_unstable,
     decl_macro,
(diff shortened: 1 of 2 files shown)


alexcrichton:
Ok safe support has landed in nightly and at least that test is green. The documentation is failing to build, though, since rustdoc is still requiring target_feature functions to be `unsafe`. I've updated that behavior in https://github.com/rust-lang/rust/pull/85982 so this is now blocked on that.

I plan on having a follow-up after this lands which actually applies stabilization to all intrinsics.

alexcrichton:
Ok with the latest nightly this is all green and should be good to land now!

PR #1179, opened 2021-06-10
wasm: Mark simd intrinsics as stable

alexcrichton:
This is a follow-up from rust-lang/rust#74372 which has finished FCP for
the stabilization of wasm intrinsics. This marks them all stable, as-is
and additionally marks the functions which create integer vectors as
`const`-stable as well. The only remaining unstable bits are that
`f32x4` and `f64x2` are `const`-unstable. Mostly just because I couldn't
figure out how to make them `const`-stable.


Amanieu:
What is the problem with const `f32x4`/`f64x2` and what does it have to do with https://github.com/rust-lang/rust/issues/73288?

PR #1183, opened 2021-07-14
Move `asm!` and `global_asm!` to `core::arch`

bstrie:
Implements the first part of the libs team decision given at https://github.com/rust-lang/rust/issues/84019#issuecomment-875880229 .

The first step is to define these items in stdarch. Once this PR is accepted, I have another patch prepared for the Rust repo that will simultaneously remove the definitions in libcore and bump the stdarch submodule.

cc @Amanieu

### DIFF
--- a/crates/core_arch/src/mod.rs
+++ b/crates/core_arch/src/mod.rs
@@ -11,6 +11,32 @@ mod simd;
 #[doc = include_str!("core_arch_docs.md")]
 #[stable(feature = "simd_arch", since = "1.27.0")]
 pub mod arch {
+    /// Inline assembly.
+    ///
+    /// Read the [unstable book] for the usage.
+    ///
+    /// [unstable book]: ../unstable-book/library-features/asm.html
+    #[unstable(
+        feature = "asm",
+        issue = "72016",
+        reason = "inline assembly is not stable enough for use and is subject to change"
+    )]
+    #[rustc_builtin_macro]
+    pub macro asm("assembly template", $(operands,)* $(options($(option),*))?) {
+        /* compiler built-in */
+    }
+
+    /// Module-level inline assembly.
+    #[unstable(
+        feature = "global_asm",
+        issue = "35119",
+        reason = "`global_asm!` is not stable enough for use and is subject to change"
+    )]
+    #[rustc_builtin_macro]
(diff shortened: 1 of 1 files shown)


bstrie:
> error[E0773]: attempted to define built-in macro more than once

This error is expected. Because this is a submodule, there's no way to atomically remove the definitions in libcore and add the definitions here. One way or the other we have to either temporarily break stdarch or temporarily break the Rust repo, so we'll just have to live with HEAD being broken until the Rust repo PR (which I already have written, and passes the tests) can be merged.

crates/core_arch/src/mod.rs
```diff
@@ -11,6 +11,32 @@ mod simd;
...
+    /// Read the [unstable book] for the usage.
+    ///
+    /// [unstable book]: ../unstable-book/library-features/asm.html
+    #[unstable(
+        feature = "asm",
+        issue = "72016",
+        reason = "inline assembly is not stable enough for use and is subject to change"
+    )]
+    #[rustc_builtin_macro]
+    pub macro asm("assembly template", $(operands,)* $(options($(option),*))?) {
```
Amanieu, line 25: `asm!` is currently defined using `macro_rules!`. What are the differences of using `pub macro` compared to `macro_rules!`?

bstrie: To export a `macro_rules` macro you have to use `#macro_export`, which both makes the macro public and declares the macro in the crate root. Being declared in the crate root is undesirable because it means the macro is effectively "in the prelude" whether you want it to be or not. It also means that documentation lists the macro as being defined in the crate root, which is why the stdlib API docs display a giant wall of random macros (https://doc.rust-lang.org/std/index.html#macros).

In contrast, `pub macro` makes the macro public without defining it in the crate root. This is strictly more flexible, since we can add it to the prelude if we choose (which I will be doing with my aforementioned Rust repo PR, so as to not break nightly users). It also means we can have the documentation live somewhere sensible and avoid further cluttering the main stdlib API index.

As far as the implementation goes, all the actual macro machinery gets ignored since it's ultimately just a builtin. So the proper namespacing should be the only thing different here AFAIK.

Amanieu:
Great! I don't mind stdarch being broken for a bit, we don't update the submodule in rust-lang/rust very often.

PR #1194, opened 2021-08-10
Update arm vcvt intrinsics to use llvm.fpto(su)i.sat

JamieCunliffe:
Those intrinsics have the correct semantics for the desired fcvtz instruction,
without any undefined behaviour. The previous simd_cast was undefined for
infinite and NaN which could cause issues.


bjorn3:
I think `simd_cast` should be changed instead, probably keeping the old behavior as `simd_cast_unchecked`.

Amanieu:
I'll defer the `simd_cast` question to @rust-lang/project-portable-simd. For now this is good enough to fix stdarch.

PR #1196, opened 2021-08-11
mark core::arch::wasm32::unreachable as safe

matklad:
This intrinsic doesn't have any preconditions and is always safe to
call, so it can be safe.

This function is already stable, but dropping `unsafe` is a backwards
compatible change.

Note tha we already have a precedent for wasm intrinsics being safe --
wasm simd is safe.

It is relatively practically important to mark this safe --
wasm32::unreachable is directly useful in practice as more codesize
efficient `panic!()`.

### DIFF
--- a/crates/core_arch/src/wasm32/mod.rs
+++ b/crates/core_arch/src/wasm32/mod.rs
@@ -12,10 +12,15 @@ pub use self::simd128::*;
 mod memory;
 pub use self::memory::*;
 
-/// Generates the trap instruction `UNREACHABLE`
+/// Generates the [`unreachable`] instruction, which causes an unconditional [trap].
+///
+/// This function is safe to call and immediately aborts the execution.
+///
+/// [`unreachable`]: https://webassembly.github.io/spec/core/syntax/instructions.html#syntax-instr-control
+/// [trap]: https://webassembly.github.io/spec/core/intro/overview.html#trap
 #[cfg_attr(test, assert_instr(unreachable))]
 #[inline]
 #[stable(feature = "unreachable_wasm32", since = "1.37.0")]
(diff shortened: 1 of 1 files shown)


bjorn3:
> This function is already stable, but dropping unsafe is a backwards compatible change.

What if a function expects `unsafe fn() -> !` as argument and you pass it `unreachable`? Or a different trait impl is used depending on the safety of a function?

This is technically a breaking change, but it may be acceptable.

Amanieu:
@alexcrichton I'm leaving the decision to you. This is very likely to not break anything, so I don't mind making the change after a crater run.

alexcrichton:
Oh thanks for the heads up!

I personally agree that, in retrospect, this is the right signature for the intrinsic (safe, not unsafe). It might be worth updating the documentation to indicate that despite the name being "unreachable" it will deterministically and safely raise a wasm trap which is likely to abort the program (or be caught by whatever embedder there is).

In terms of breakage, while a crater run would be ideal here I'm not sure if crater has the ability to cross-compile to wasm. I suspect that there will be little-to-no breakage of this in the wild, despite theoretical breakage being possible. I don't think this is used nearly heavily enough to have it exercised in so many weird ways.

Overall, I think this is ok to land. This would want to be tagged in the release notes of Rust though and we'd want to keep a close eye on the issue tracker if this causes any regressions.

Amanieu:
Sounds good!

@matklad could you add a quick sentence in the doc comment explaining that why this is safe to call?

matklad:
Adjusted the docs: I didn't go out of my way to explain that this can be caught by embedder and instead just linked the upstream docs explaining what the "trap" is (which we really should do, using terms-of-art without defining them hurts approachableness).

PR #1197, opened 2021-08-11
Remove references to the crates on crates.io.

m-ou-se:
They haven't been published in years. This removes the suggestion that the crates on crates.io are actively updated/maintained.


Amanieu:
We could even add `publish = false` to `Cargo.toml`, but it seems that we don't currently do this in rust-lang/rust so I don't think here would be the right place to start.

PR #1198, opened 2021-08-14
Rewording in Wasm all_true SIMD instruction docs

osa1:
- Use "true" and "false" instead of "1" and "0"
- "nonzero" -> "non-zero"
- "returns true if ... or false if ..." -> "returns true ..., false otherwise"

https://github.com/rust-lang/rust/issues/87684


Amanieu:
Thanks!

PR #1200, opened 2021-08-20
add vldx neon instructions

### LINKED ISSUE #1143: Can't get the correct return type of llvm link
Look the following implementation of vld1_s16_x2 neon instruction: 
[godbolt](https://godbolt.org/z/qnvera3sh)
the return type of `vld1x2.v4i16.p0i16` should be `{<4 x i16>, <4 x i16>}`
But if we do a similar implementation in rust:
[gotbolt](https://rust.godbolt.org/z/bf3Ts6cjE)
llvm will change the return type to the wrong `[ 2 x <i8 x 8> ]`. As a result, we cannot implement the `vld1_s16_x2` instruction.

SparrowLii:
This PR adds vld1 neon intruscions which read multiple vectors at once.
Fixes #1143


crates/stdarch-gen/src/main.rs
```diff
@@ -1437,15 +1737,21 @@ fn gen_arm(
             link.push_str(&link_aarch64);
             link.replace("_EXT_", ext).replace("_EXT2_", ext2)
         };
+        let abi = if test_fn == "load_test" {
+            "unadjusted"
+        } else {
+            "C"
+        };
```
Amanieu, line 1744: I think we would want to always use `"unadjusted"` when linking with LLVM intrinsics, unless I'm missing something?

SparrowLii: I'm not sure if these links depend on "C" Abi, especially other architectures, like x86. In view of the fact that most of the structures are marked by `#[repr(simd)]`, the difference in Abi should only affect the structure containing multiple vectors IMO. Let's try it.

SparrowLii: Using "unadjusted" looks fine, at least for arm/aarch64. It also slightly simplifies the compilation process, so I think it makes sense. 
A few `assert_instr`s of some instructions using `simd_extract` need to be correct, I think this may be due to some updates to the compiler

crates/core_arch/src/arm_shared/neon/mod.rs
```diff
@@ -2867,7 +3117,7 @@ pub unsafe fn vpmax_f32(a: float32x2_t, b: float32x2_t) -> float32x2_t {
 #[target_feature(enable = "neon")]
 #[cfg_attr(target_arch = "arm", target_feature(enable = "v7"))]
 #[rustc_legacy_const_generics(1)]
-#[cfg_attr(all(test, target_arch = "arm"), assert_instr("vmov.32", IMM5 = 1))]
+#[cfg_attr(all(test, target_arch = "arm"), assert_instr("vmov", IMM5 = 1))]
 #[cfg_attr(all(test, target_arch = "aarch64"), assert_instr(mov, IMM5 = 1))]
```
Amanieu, line 3121: I think for these we should just use `assert_instr(nop, ...)`. `nop` is handled specially and accepts any instruction.

SparrowLii: It makes sense. I replaced some obvious places where `"nop"` should be used

crates/core_arch/src/arm_shared/neon/mod.rs
```diff
@@ -2867,11 +3117,8 @@ pub unsafe fn vpmax_f32(a: float32x2_t, b: float32x2_t) -> float32x2_t {
 #[target_feature(enable = "neon")]
 #[cfg_attr(target_arch = "arm", target_feature(enable = "v7"))]
 #[rustc_legacy_const_generics(1)]
-#[cfg_attr(all(test, target_arch = "arm"), assert_instr("vmov.32", IMM5 = 1))]
-#[cfg_attr(all(test, target_arch = "aarch64"), assert_instr(mov, IMM5 = 1))]
-// Based on the discussion in https://github.com/rust-lang/stdarch/pull/792
-// `mov` seems to be an acceptable intrinsic to compile to
-// #[cfg_attr(all(test, target_arch = "aarch64"), assert_instr(vmov, IMM5 = 1))]
+#[cfg_attr(all(test, target_arch = "arm"), assert_instr("nop", IMM5 = 1))]
+#[cfg_attr(all(test, target_arch = "aarch64"), assert_instr(nop, IMM5 = 1))]
```
Amanieu, line 3121:
```suggestion
#[cfg_attr(test, assert_instr(nop, IMM5 = 1))]
```

SparrowLii: OK, I have made the corresponding changes.
I think I can submit a new PR later to correct all `assert_instr` in arm/aarch64 mod.

PR #1201, opened 2021-08-22
Fix typos “a”→“an”

steffahn:
See rust-lang/rust#88230


steffahn:
The CI failure seems unrelated, right?

Amanieu:
Yea it's unrelated.

PR #1207, opened 2021-09-07
Change aarch64 vld1* instructions to not cause individual loads

### LINKED ISSUE #1148: Aarch64 performance: `vld1q_u8` intrinsic can cause single-byte loads
While adding aarch64 support to [simdutf8](https://github.com/rusticstuff/simdutf8) I encountered an unexpected eight times slowdown when hand-unrolling a loop. This slowdown was the result of the compiler deciding to suddenly load 128-bit `uint8x16_t` values with single-byte load instructions instead of 128-bit loads.

It turns out, that the `vld1q_u8` intrinsic is at fault. The code generator thinks it can "optimize" loads by loading bytes individually if a SIMD shuffle instruction follows. According to [the ARM docs](https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics?search=vld1q_u8) this intrinsic should always be coded as one instruction. I fixed it by doing the load similar to how it is [currently done for SSE2](https://github.com/rust-lang/stdarch/blob/6495bb0e33578443c21764655a4dd8b55399c008/crates/core_arch/src/x86/sse2.rs#L1193).

[Testcase and proposed fix on Godbolt](https://godbolt.org/z/T1h8145Tq)

The same issue likely applies is to the other `vld1q` intrinsics.

hkratz:
vld1* instructions are required to always compile to a single load instruction (see [ARM developer documentation](https://developer.arm.com/architectures/instruction-sets/intrinsics/)). The current implementation causes individual loads to be emitted in LLVM-IR which are not always combined to a single load instruction during LLVM optimization passes. This change causes a single load to be emitted in all cases.

cc @SparrowLii 

TODO:
- [ ] Add test
- [x] Open question: Really change all impls for consistency or just where it matters?

Fixes #1148.

### DIFF
--- a/crates/core_arch/src/aarch64/neon/mod.rs
+++ b/crates/core_arch/src/aarch64/neon/mod.rs
@@ -12,8 +12,7 @@ pub use self::generated::*;
 use crate::{
     core_arch::{arm_shared::*, simd::*, simd_llvm::*},
     hint::unreachable_unchecked,
-    mem::{size_of, transmute, zeroed},
-    ptr::copy_nonoverlapping,
+    mem::{transmute, zeroed},
 };
 #[cfg(test)]
 use stdarch_test::assert_instr;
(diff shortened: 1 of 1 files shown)


SparrowLii:
> Really change all impls for consistency or just where it matters?
I think at least the vst1* instruction should also use `write_unaligned` for consistency.

Amanieu:
@hkratz Thanks! Could you update the vst1 intrinsics as well like @SparrowLii suggested.

Amanieu:
LGTM. Are you planning on implementing something else (the PR is still a draft).

I don't think a test is needed for this.

hkratz:
I was thinking of a regression test for #1148, but it is not strictly necessary and afaics there is no easy way to do it.

PR #1208, opened 2021-09-07
Use LLVM objdump on Macos instead of otool for CI tests

hkratz:
Enabling target features is not possible with otool. Setting a target CPU is possible but the Transactional Memory Extension (TME) is not supported by any CPU yet so that does not help. LLVM objdump is installed with the developer tools out of the box and can be used instead.

Edit: Now uses LLVM objdump for x86_64 as well.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -138,9 +138,8 @@ jobs:
           norun: true
         - target: aarch64-unknown-linux-gnu
           os: ubuntu-latest
-        # Temporarily disabled because otool crashes with "Out of memory", seems Github CI issue
-        #- target: x86_64-apple-darwin
-        #  os: macos-latest
+        - target: x86_64-apple-darwin
+          os: macos-11
         - target: x86_64-pc-windows-msvc
           os: windows-latest
         - target: i686-pc-windows-msvc
--- a/crates/stdarch-test/src/disassembly.rs
+++ b/crates/stdarch-test/src/disassembly.rs
@@ -125,16 +125,7 @@ fn parse(output: &str) -> HashSet<Function> {
                 cached_header = None;
                 break;
             }
-            let parts = if cfg!(target_os = "macos") {
-                // Each line of instructions should look like:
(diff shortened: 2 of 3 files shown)


lu-zero (approved):
It looks fine, trying to not duplicate the two similar codepaths probably is not worth it. I wonder if just unifying to use objdump on macos also on x86_64 wouldn't make the code simpler though.

hkratz:
> It looks fine, trying to not duplicate the two similar codepaths probably is not worth it. I wonder if just unifying to use objdump on macos also on x86_64 wouldn't make the code simpler though.

Good idea! I added a commit to use objdump for x86_64 as well and it actually works better on a M1 Mac with Rosetta2 emulation than otool which just hangs.

hkratz:
Objdump seems to work properly with Big Sur only (don't have an older Mac to test). ARM64 support came with Big Sur so that is not a problem, not sure if it is an issue for x86_64.

lu-zero:
If the one provided from [llvm-tools-preview](https://github.com/rust-lang/rust/issues/85658) we could just suggest to use it.

lu-zero:
I like the set @Amanieu is it good for you as well? I guess it could land with the `TEST` commit removed/squashed away.

hkratz:
> If the one provided from [llvm-tools-preview](https://github.com/rust-lang/rust/issues/85658) works, we could just suggest to use it.

It works but exposes another bug in the testing framework, namely that we currently check instructions for the string "call" to find out non-inlined intrinsics (despite that not being the instruction for subroutine call on aarch64) and the disassembly crated from the new llvm-objdump contains relative addresses like this:
```
2: adrp x9, 0x10011f000 <__ZN4core3ops8function6FnOnce9call_once17h00643be011955a98E+0x10>
```
which thus cause incorrect test failures.

lu-zero:
You may squash the starts_with fixup with 0cb9065, then it would be good to go IMHO.

Thank you a lot :)

Issue #1206, opened 2021-09-07
Mutliple tests fails on Macos aarch64


hkratz:
```
failures:

---- core_arch::aarch64::neon::assert_vdup_n_f64_ldr stdout ----
disassembly for stdarch_test_shim_vdup_n_f64_ldr:
	 0: adrp x8, 505 ; 0x100260000
	 1: nop
	 2: adrp x9, 275 ; 0x10017a000
	 3: add x9, x9, #0x433
	 4: str x9, [x8, #0x398]
	 5: ret
thread 'core_arch::aarch64::neon::assert_vdup_n_f64_ldr' panicked at 'failed to find instruction `ldr` in the disassembly', crates/stdarch-test/src/lib.rs:157:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- core_arch::aarch64::neon::assert_vget_low_f64_ldr stdout ----
disassembly for stdarch_test_shim_vget_low_f64_ldr:
	 0: adrp x8, 505 ; 0x100260000
	 1: nop
	 2: adrp x9, 275 ; 0x10017a000
	 3: add x9, x9, #0x756
	 4: str x9, [x8, #0x398]
	 5: ret
thread 'core_arch::aarch64::neon::assert_vget_low_f64_ldr' panicked at 'failed to find instruction `ldr` in the disassembly', crates/stdarch-test/src/lib.rs:157:9

---- core_arch::aarch64::neon::assert_vget_high_p64_ldr stdout ----
disassembly for stdarch_test_shim_vget_high_p64_ldr:
	 0: adrp x8, 505 ; 0x100260000
	 1: nop
	 2: adrp x9, 275 ; 0x10017a000
	 3: add x9, x9, #0x6f9
	 4: str x9, [x8, #0x398]
	 5: ext.16b v0, v0, v0, #0x8
	 6: ret
thread 'core_arch::aarch64::neon::assert_vget_high_p64_ldr' panicked at 'failed to find instruction `ldr` in the disassembly', crates/stdarch-test/src/lib.rs:157:9

---- core_arch::aarch64::neon::assert_vget_low_p64_ldr stdout ----
disassembly for stdarch_test_shim_vget_low_p64_ldr:
	 0: adrp x8, 505 ; 0x100260000
	 1: nop
	 2: adrp x9, 275 ; 0x10017a000
	 3: add x9, x9, #0x7b1
	 4: str x9, [x8, #0x398]
	 5: ret
[...]

PR #1213, opened 2021-09-08
Normalize [us]shll.* ..., #0 aarch64 disassembly to the preferred [us]xtl.*

hkratz:
This is a workaround for an LLVM objdump shortcoming. `uxtl`and `sxtl` are the preferred disassembly for `ushll` and `sshll` if the shift is zero as documented here:
* https://developer.arm.com/documentation/ddi0602/latest/SIMD-FP-Instructions/UXTL--UXTL2--Unsigned-extend-Long--an-alias-of-USHLL--USHLL2-
* https://developer.arm.com/documentation/ddi0602/latest/SIMD-FP-Instructions/SXTL--SXTL2--Signed-extend-Long--an-alias-of-SSHLL--SSHLL2-

This PR fixes some failing tests on Macos ARM64 (see #1206).

### DIFF
--- a/crates/stdarch-test/src/disassembly.rs
+++ b/crates/stdarch-test/src/disassembly.rs
@@ -125,7 +125,7 @@ fn parse(output: &str) -> HashSet<Function> {
                 cached_header = None;
                 break;
             }
-            let parts = if cfg!(target_env = "msvc") {
+            let mut parts = if cfg!(target_env = "msvc") {
                 // Each line looks like:
                 //
                 // >  $addr: ab cd ef     $instr..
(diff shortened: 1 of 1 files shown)


lu-zero:
Did you report the problem to llvm?

hkratz:
> Did you report the problem to llvm?

Not yet. It's on my todo list.

hkratz:
Actually no need to report, it is already in the bug tracker: https://bugs.llvm.org/show_bug.cgi?id=39721 and there is patch since 2018: https://reviews.llvm.org/D54791

PR #1214, opened 2021-09-09
Fix non-inlined fn calls for some arm vfma intrinsics

hkratz:
Some VFMA functions have `target_feature(enable = "fp-armv8,v8")` while the called functions `vdup_n_f32` and `vdupq_n_f32` are `target_feature(enable = "v7")`. LLVM does not inline the functions due to the different feature flags. Using private _v8 variants of those functions causes them to be inlined.

### DIFF
--- a/crates/core_arch/src/arm_shared/neon/mod.rs
+++ b/crates/core_arch/src/arm_shared/neon/mod.rs
@@ -3707,6 +3707,16 @@ pub unsafe fn vdupq_n_f32(value: f32) -> float32x4_t {
 /// Duplicate vector element to vector or scalar
 #[inline]
 #[target_feature(enable = "neon")]
+#[cfg_attr(target_arch = "arm", target_feature(enable = "fp-armv8,v8"))]
+#[cfg_attr(all(test, target_arch = "arm"), assert_instr("vdup.32"))]
+#[cfg_attr(all(test, target_arch = "aarch64"), assert_instr(dup))]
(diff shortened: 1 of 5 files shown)


crates/core_arch/src/arm_shared/neon/mod.rs
```diff
@@ -3704,6 +3704,16 @@ pub unsafe fn vdupq_n_f32(value: f32) -> float32x4_t {
     float32x4_t(value, value, value, value)
 }
 
+/// Duplicate vector element to vector or scalar
+#[inline]
+#[target_feature(enable = "neon")]
+#[cfg_attr(target_arch = "arm", target_feature(enable = "fp-armv8,v8"))]
+#[cfg_attr(all(test, target_arch = "arm"), assert_instr("vdup.32"))]
+#[cfg_attr(all(test, target_arch = "aarch64"), assert_instr(dup))]
+unsafe fn vdupq_n_f32_v8(value: f32) -> float32x4_t {
```
Amanieu, line 3713: Can you add a comment explaining why this is needed? Also this should probably be under `#[cfg(target_arch = "arm")]`.

hkratz: @Amanieu Will do, but first I will fix arm fused multiply-add to only require vfp4 instead of v8 (PR forthcoming) and then the fn name will change to `vdupq_n_f32_vfp4`.

hkratz:
Superseded by #1219.

Issue #1212, opened 2021-09-08
The assembly emitted for some arm neon fma intrinsics contains function calls


hkratz:
`vfma_n_f32`, `vfmaq_n_f32`, `vfms_n_f32` and `vfms_n_f32`are affected.

From a CI with inlining check enabled (https://github.com/rust-lang/stdarch/runs/3547875318):
```
---- core_arch::arm_shared::neon::generated::assert_vfma_n_f32_vfma stdout ----
disassembly for stdarch_test_shim_vfma_n_f32_vfma: 
	 0: push {fp, lr}
	 1: vpush {d8-d9}
	 2: sub sp, sp, #8
	 3: ldr r0, [pc, #56] ; 395e0 <stdarch_test_shim_vfma_n_f32_vfma+0x4c>
	 4: vorr d9, d0, d0
	 5: ldr r1, [pc, #52] ; 395e4 <stdarch_test_shim_vfma_n_f32_vfma+0x50>
	 6: vmov.f32 s0, s4
	 7: add r0, pc, r0
	 8: vorr d8, d1, d1
	 9: ldr r1, [pc, r1]
	10: str r0, [r1]
	11: mov r0, sp
	12: bl 34030 <_ZN9core_arch9core_arch10arm_shared4neon10vdup_n_f3217h6b11480dbcf9ce6fE>
	13: vldr d16, [sp]
	14: vfma.f32 d9, d8, d16
	15: vorr d0, d9, d9
	16: add sp, sp, #8
	17: vpop {d8-d9}
	18: pop {fp, pc}
	19: .word 0x0010780f
	20: .word 0x001be9d8
thread 'core_arch::arm_shared::neon::generated::assert_vfma_n_f32_vfma' panicked at 'instruction found, but the disassembly contains subroutine call instructions, which hint that inlining failed', crates/stdarch-test/src/lib.rs:177:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- core_arch::arm_shared::neon::generated::assert_vfmaq_n_f32_vfma stdout ----
disassembly for stdarch_test_shim_vfmaq_n_f32_vfma: 
	 0: push {r4, lr}
	 1: vpush {d8-d11}
	 2: sub sp, sp, #16
	 3: ldr r0, [pc, #60] ; 3958c <stdarch_test_shim_vfmaq_n_f32_vfma+0x50>
	 4: vorr q5, q0, q0
[...]

PR #1216, opened 2021-09-10
Complete vld1 instructions with some corrections

SparrowLii:
This PR does the following:
(1) completes` vld1` neon instructions, mainly for `*_p64`s
(2) replaces `crypto` feature with `aes` for arm 
(3) adds `"core_arch/src/arm_shared/neon"` dir in `stdarch-verify`. Based on this, the function names of `vmull_n_s16`, `vmull_n_s32`, `vmull_n_u16`, `vmull_n_u32`, `vqdmulhq_n_s16`, `vqdmulhq_n_s32` have been corrected
(4) optimizes `stdarch-gen` so that `vld2*` (and other) instructions can be generated more clearly(For the convenience of reviewing, they are not generated in this PR)


crates/stdarch-gen/neon.spec
```diff
@@ -2429,8 +2594,8 @@ aarch64 = pmull
 generate poly64x2_t:poly64x2_t:p128
 
 /// Vector long multiply with scalar
-name = vmull
-n-suffix
+name = vmull_n
+no-q
```
hkratz, line 2598: Maybe this change should be documented in the PR or in an extra commit as they change the names of public intrinsics (to their correct names).

crates/stdarch-gen/neon.spec
```diff
@@ -3568,7 +3733,7 @@ generate int16x4_t:i16:int16x4_t, int32x2_t:i32:int32x2_t
 
 /// Vector saturating doubling multiply high with scalar
 name = vqdmulhq_n
-out-suffix
+no-q
```
hkratz, line 3736: See above.

crates/stdarch-gen/neon.spec
SparrowLii: Thanks for pointing it out! I added relevant documents to the description of the PR.

crates/core_arch/src/aarch64/neon/mod.rs
```diff
@@ -749,6 +749,22 @@ pub unsafe fn vld1q_p16(ptr: *const p16) -> poly16x8_t {
...
+pub unsafe fn vld1_p64(ptr: *const p64) -> poly64x1_t {
+    transmute(u64x1::new(*ptr))
+}
+
+/// Load multiple single-element structures to one, two, three, or four registers.
+#[inline]
+#[target_feature(enable = "neon,aes")]
+#[cfg_attr(test, assert_instr(ldr))]
+pub unsafe fn vld1q_p64(ptr: *const p64) -> poly64x2_t {
+    transmute(u64x2::new(*ptr, *ptr.offset(1)))
```
Amanieu, line 765: There was a concern raised in #1148 that this doesn't get optimized down to a single instruction.

hkratz: I agree, all others have been already changed to use `read_unaligned()` instead which cannot be seen in the diff as it this PR is based on an older commit.

SparrowLii: That makes sense.

hkratz:
@Amanieu Can this be merged? Fixing #1212 and #1217 requires changes to stdarch-gen and this PR contains a big refactoring of it. I would rather like to build on this instead of running into rebase conflicts.

Issue #1210, opened 2021-09-08
Inlining checks are broken for arm


hkratz:
There are some arm neon intrinsics which contain calls to other intrinsics. 

Also the deduplication guard code is to heavy for arm and is emitted as function calls for some tests.

hkratz:
When #1211 has landed and #1212 is fixed the inlining checks can be enabled for arm.

Nugine:
Hi, is there any progress?

PR #1224, opened 2021-09-26
Complete vld* and vst* neon instructions

SparrowLii:
This PR completes vld* and vst* neon instructions
One commit will contain one type of them.


SparrowLii:
I think all vld* and vst* instructions have been implemented.

hkratz:
A lot of the new intrinsics are apparently not properly inlined on arm (thus the necessary bumps to the instructions limits and the nop expected instructions) but we can probably take care of that later as it is tracked in #1210.

PR #1230, opened 2021-10-13
Complete the remaining neon instructions

SparrowLii:
This PR completes the remaining neon instructions, except for the following instructions:

1. The following instructions are only available in aarch64 now, because the corresponding `target_feature` cannot be found in the available features of arm:
`vcadd_rot`、`vcmla`、`vdot`

2. The feature `i8mm` is not valid:
`vmmla`、`vusmmla`: https://rust.godbolt.org/z/8GbKW5ef4

3. LLVM ERROR(Can be reproduced in godbolt):
`vsm4e`: https://rust.godbolt.org/z/xhT1xvGTP

4. LLVM ERROR(Normal in gotbolt, but `LLVM ERROR: Cannot select: intrinsic` raises at runtime)
`vsudot`、`vusdot`: https://rust.godbolt.org/z/aMnEvab3n
`vqshlu`: https://rust.godbolt.org/z/hvGhrhdMT

5. Not implmented in LLVM and cannot be implemented manually:
`vmull_p64`（for arm）、`vsm3`、`vrax1q_u64`、`vxarq_u64`、`vrnd32`、`vrnd64`、`vsha512`

Update #148

### DIFF
--- a/crates/core_arch/src/aarch64/neon/generated.rs
+++ b/crates/core_arch/src/aarch64/neon/generated.rs
@@ -245,7 +245,7 @@ pub unsafe fn vceqq_f64(a: float64x2_t, b: float64x2_t) -> uint64x2_t {
 /// Compare bitwise equal
 #[inline]
 #[target_feature(enable = "neon")]
-#[cfg_attr(test, assert_instr(cmeq))]
+#[cfg_attr(test, assert_instr(cmp))]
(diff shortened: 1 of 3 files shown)


Amanieu:
#1231 is now merged, can you rebase?

SparrowLii:
It seems that the intrinsics test tool needs some updates to support multi-vector (such as `float32x2x2_t`). I will try to see if I can fix it

PR #1237, opened 2021-10-22
Handle intrinsics with constraints in the test tool.

JamieCunliffe:
Rather than using the csv export from the tracking spreadsheet we now
use the csv from the ACLE repository. This data contains any
constraints that the intrinsic has.


JamieCunliffe:
@pthariensflame  No reason for using that version of the ACLE, it was just when I added the submodule. From looking at the changes they wouldn't impact this tool as we are only testing on the aarch64-unknown-linux-gnu target so A32 isn't touched, and we don't actually use the column that the changes were in.

Amanieu:
LGTM but there still seems to be some failing tests.

JamieCunliffe:
I just opened https://github.com/rust-lang/stdarch/pull/1246 once that has been merged I'll rebase this and then all the tests should pass.

PR #1243, opened 2021-11-01
Remove unnecessary stability attribute

jhpratt:
This was caught when creating an internal rustc lint: rust-lang/rust#90356. As this method is not `pub`, it does not need a stability attribute.


jhpratt:
Hmm...@oli-obk's understanding of the attributes was correct. Time to update the lint.

PR #1245, opened 2021-11-03
Remove documentation for std_detect error_macros

adamgemmell:
These comments show up when there's proper documentation for the runtime feature detection macros, such as here: https://doc.rust-lang.org/nightly/std/macro.is_aarch64_feature_detected.html

The messages also make no sense from an API user's perspective.

### DIFF
--- a/crates/std_detect/src/detect/error_macros.rs
+++ b/crates/std_detect/src/detect/error_macros.rs
@@ -2,8 +2,8 @@
 //! architecture. These macros provide a better error messages when the user
 //! attempts to call them in a different architecture.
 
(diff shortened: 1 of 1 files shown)


Amanieu:
I think the proper solution here would be to use `cfg(doc)` to show the proper documentation for all arch-specific macros when rustdoc is used.

adamgemmell:
That does seem useful, thanks, I'll update the PR. To me it still makes sense to remove this error documentation - do you agree?

Amanieu:
Yes, it's fine to change these to normal comments instead of doc comments.

Amanieu:
Superseded by #1276.

PR #1246, opened 2021-11-04
Correct the vqrdmlah intrinsics.

JamieCunliffe:
The `sqrdmulh` followed by a `sqadd` is only optimized into `sqrdmlah` when the rdm target feature is enabled. I believe this is actually an incorrect optimization in LLVM though. However this patch will cause the correct instruction for the intrinsic, and I couldn't see another way to generate this instruction.

For instance using `a = 0x80000000, b = 0x80000000, c = 0x80000000`  `sqrdmlah` would be `0` whereas the `sqrdmulh` `sqadd` sequence would be `-1`


Amanieu:
By the way, could you send a bug report to LLVM about the incorrect optimization?

PR #1249, opened 2021-11-07
Work-around buggy Intel chips erroneously reporting BMI1/BMI2 support

thomcc:
Apparently on some chips, `cpuid` is not to be fully trusted. According to http://www.intel.com/content/dam/www/public/us/en/documents/specification-updates/desktop-6th-gen-core-family-spec-update.pdf, some Skylake chips (notably Celerons) will report support for BMI1 and BMI2 without actual support for those operations.

Quoting that document's Errata section:

> | SKL052 | CPUID Incorrectly Reports Bit Manipulation Instructions Support |
> | :-- | :-- |
> | Problem | Executing CPUID with EAX = 7 and ECX = 0 may return EBX with bits [3] and [8] set, incorrectly indicating the presence of BMI1 and BMI2 instruction set extensions. |
> | Implication | Attempting to use instructions from the BMI1 or BMI2 instruction set extensions will result in a #UD exception. |
> | Workaround | It is possible for the BIOS to contain a workaround for this erratum. |

IIUC, the problematic chips are fairly widespread[^wide], and searches online do seem to bring up people hitting this crash in the real world... so it would be nice if a better workaround existed than "ask the user to update their BIOS" (I think the OS could work around it too — if nothing else, by performing a similar hack to what's in this patch).

That said, one seems to exist[^exist]. From what I was able to tell[^tell], all current Intel chips with working support for either BMI instruction set also have AVX, and the chips impacted by this bug here do not, so we can just assume any Intel chip without AVX also has no BMI1/BMI2. If this is correct, we can work around this with no downside to any current chip, so even if a more ideal place for a hack like this would be in the OS[^os], it's worth doing.

This is what's implemented in this patch, which also contains a big comment describing the issue, since it's pretty WTF.

[^exist]: It's not surprising that Intel wouldn't include this as the suggested workaround, since they might want to keep the ability to ship chips with BMI and not AVX in the future.

[^tell]: Unfortunately, this isn't definitive, since I'm not sure where to find such information concretely, and just went by what I could find on Wikipedia and Agner. That said, it makes sense that this would be the case both now and in the future, and it's safe even if I'm wrong or things change in the future.

[^os]: And ofc we can't rely on the user's OS or BIOS being up to date, beyond what libstd supports.

[^wide]: I am pretty sure I know someone with one. This would be great because it would allow me to verify the issue/fix, but the Intel documentation is pretty clear either way.

### DIFF
--- a/crates/std_detect/src/detect/os/x86.rs
+++ b/crates/std_detect/src/detect/os/x86.rs
@@ -249,5 +249,25 @@ pub(crate) fn detect_features() -> cache::Initializer {
         }
     }
 
+    // Unfortunately, some Skylake chips erroneously report support for BMI1 and
+    // BMI2 without actual support. These chips don't support AVX, and it seems
+    // that all Intel chips with non-erroneous support BMI do (I didn't check
+    // other vendors), so we can disable these flags for chips that don't also
+    // report support for AVX.
+    //
+    // It's possible this will pessimize future chips that do support BMI and
+    // not AVX, but this seems minor compared to a hard crash you get when
+    // executing an unsupported instruction (to put it another way, it's safe
+    // for us to under-report CPU features, but not to over-report them). Still,
+    // to limit any impact this may have in the future, we only do this for
+    // Intel chips, as it's a bug only present in their chips.
+    //
+    // This bug is documented as `SKL052` in the errata section of this document:
+    // http://www.intel.com/content/dam/www/public/us/en/documents/specification-updates/desktop-6th-gen-core-family-spec-update.pdf
+    if vendor_id == *b"GenuineIntel" && !value.test(Feature::avx as u32) {
+        value.unset(Feature::bmi1 as u32);
+        value.unset(Feature::bmi2 as u32);
+    }
+
     value
 }


crates/std_detect/src/detect/os/x86.rs
```diff
@@ -249,5 +249,25 @@ pub(crate) fn detect_features() -> cache::Initializer {
...
+    // for us to under-report CPU features, but not to over-report them). Still,
+    // to limit any impact this may have in the future, we only do this for
+    // Intel chips, as it's a bug only present in their chips.
+    //
+    // This bug is documented as `SKL052` in the errata section of this document:
+    // http://www.intel.com/content/dam/www/public/us/en/documents/specification-updates/desktop-6th-gen-core-family-spec-update.pdf
+    if vendor_id == *b"GenuineIntel" && !value.test(Feature::avx as u32) {
+        value.unset(Feature::bmi1 as u32);
+        value.unset(Feature::bmi2 as u32);
+    }
```
bjorn3, line 270: Could you limit this to the specific cpu family that has the bug?

thomcc: Probably in theory, I mean there's model/family info in the bits of __cpuid_count(1, 0).eax...

But it's not clear to me at all which values for these correspond to the impacted chip :(

workingjubilee: Affected Steppings for this appear to be...
| Number | Y  | U | - | H | - | S | - | 
|--|--|--|--|--|--|--|--|
| - | D-1 | D-1 | K-1 | N-0 | R-0 | R-0 | S-0 |
| SKL052 | X | X | X | X | X | X | X |

So that looks like effectively "all Skylake (non-server)", according to Intel, therefore we can constrain this logic to all Skylake processors. We don't know which of the processors this is actually true for, since it's a "may", so we must apply it to all Skylake "client" chips. The possible values for family, model, etc., are specified on page 21 of this specification update, so we just add another constraint.

New Skylake chips are not being shipped so we can assume this won't change.

PR #1250, opened 2021-11-09
[WIP]Add remaining insturctions

SparrowLii:
This PR add `vmmla` `vusmmla` `vsm4e` `vsm3` `vrax1` `vxar` `vsha512` `vbcax`  `veor3` neon instructions


crates/stdarch-gen/neon.spec
```diff
@@ -6862,6 +6846,163 @@ aarch64 = usra
...
+
+/// Floating-point round to 32-bit integer, using current rounding mode
+name = vrnd32x
+a = 1.1, 1.9, -1.7, -2.3
+validate 1.0, 2.0, -2.0, -2.0
+target = v8.5a
+
+aarch64 = frint32x
+link-aarch64 = frint32x._EXT_
+// v8.5a is no_runtime feature
```
Amanieu, line 6973: I think this should be changed to frintts as discussed in #148.

Amanieu:
I think the C tests are failing because they are using clang-12 but some of these intrinsics were only added in clang-13.

PR #1251, opened 2021-11-13
Fix unused link_name attribute.

ehuss:
It looks like there was a mistake commenting out this function, which left an unused attribute. 

Unblocks https://github.com/rust-lang/rust/pull/88681 which is extending the unused_attributes lint to catch more situations.


ehuss:
@Amanieu Is it OK for me to bump the stdarch module on rust-lang/rust?

Amanieu:
Yes, that's fine.

PR #1252, opened 2021-11-14
Fix i8mm feature with bootstrap compiler.

ehuss:
The beta compiler (1.57) does not support the i8mm feature. This fixes a build failure by not enabling it when bootstrapping.


ehuss:
I'm not certain that this will work, as I do not have an arm system to test on.

SparrowLii:
If we accept this change, does it mean that we need to change it again when the beta compiler supports the `i8mm` feature?

ehuss:
Yes, I believe that is what is usually done (#1150 for example).

PR #1253, opened 2021-11-14
Exclude intel SVML functions from missing intrinsics report

jhorstmann:
Intel's Short Vector Math Library (SVML) defines optimized math functions which do not directly map to single cpu instructions but rather to a sequence of instructions. Exclude these from the missing intrinsics report so that it is much easier to see which real instructions are still missing for example for full AVX512F support.

This also updates the intrinsics data file to version 3.6

### DIFF
--- a/crates/stdarch-verify/tests/x86-intel.rs
+++ b/crates/stdarch-verify/tests/x86-intel.rs
@@ -116,6 +116,7 @@ struct Intrinsic {
     #[serde(rename = "return")]
     return_: Return,
     name: String,
+    tech: String,
     #[serde(rename = "CPUID", default)]
     cpuid: Vec<String>,
     #[serde(rename = "parameter", default)]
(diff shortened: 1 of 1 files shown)


crates/stdarch-verify/tests/x86-intel.rs
```diff
@@ -155,7 +156,7 @@ fn verify_all_signatures() {
     //   https://software.intel.com/sites/landingpage/IntrinsicsGuide/#
     //
     // Open up the network console and you'll see an xml file was downloaded
-    // (currently called data-3.4.xml). That's the file we downloaded
+    // (currently called data-3.6.xml). That's the file we downloaded
```
bjorn3, line 159: Ignoring pack files this update will grow the repo size by 6.74 MB. With pack files, it will likely be a bit smaller, though possibly not by much.

jhorstmann: Good point, I did not consider the size. If that is a concern I could revert and rebase adding this file. It could be added once there are actual plans to implement to implement for example the VNNI or FP16 extensions. My goal is to work on avx512f and avx512vl for now, especially masked loads/stores.

jhorstmann:
#1594 contains an updated and much extended version of this idea

PR #1254, opened 2021-11-15
Implement avx512 masked load and store intrinsics

jhorstmann:
Implement avx512 masked load and store intrinsics using inline assembly.

- [x] aligned/unaligned masked/zero-masked loads
- [x] aligned/unaligned stores
- [x] avx512vl (_mm256 and _mm) variants of the above
- [x] avx512bw (byte and word) variants of the above
- [x] formatting
- [x] tests for all of the above
- [x] updated avx512f.md and avx512bw.md 

The same approach also works for masked gather/scatter and compress/expand intrinsics. Probably makes sense to split these into their own PR.


Amanieu:
LGTM!

Just a small style nit: please indent the contents of the `asm!` macro, like rustfmt does for function calls.

Amanieu:
I believe in the past we avoided defining functions inside macros because it interacts poorly with our intrinsic checking tools.

jhorstmann:
> I believe in the past we avoided defining functions inside macros because it interacts poorly with our intrinsic checking tools.

Was just about to update the description to mention this. I saw the `assert_instr` tests are running, but are probably not very useful since the same parameter gets used for the assertion as in the inline asm. Tests for all functions are still on my todo list.

I think the macro approach is worth it since it reduces code for the load intrinsics by about 30x and reduces chances of copy-paste mistakes.

Amanieu:
Specifically the `stdarch_verify` crate will parse every `.rs` file to find the names and signatures of all intrinsic functions. This parsing does *not* perform macro expansion (that would require running the full rust compiler).

I think it would be better to avoid using macros for now. The ARM code avoids this issue by using a code generator, but it is probably not worth the effort in this case since AVX512 is mostly complete already.

Amanieu:
Could you also mark the intrinsics as implemented in crates/core_arch/avx512f.md. We should be able to start stabilizing avx512 once it is complete.

jhorstmann:
Should be ready for review now. The github diff view looks confusing, individual commits might be clearer.

I ended up using a "poor man's" code generator by expanding the [macros from the earlier commit](https://github.com/rust-lang/stdarch/pull/1254/commits/c902d9096a4beac621d3777deb7ec655e7f06272#diff-914d6b7878c6b1e20336674bd2bb4aa30a80faa94cd573015501b05b6ed992a9) and postprocessing the output with some small regular expressions. It's a bit manual and probably not worth checking in. More time was spent in writing all the tests.

The avx512vl functions required adding `avx` to `target_feature` in order to use `ymm` registers. It seems the CI run for i586 also required the `sse` feature to use `xmm` registers. I could not reproduce that locally. Would be nice if `avx512vl` already allowed using those registers.

Amanieu:
LGTM! I'm just waiting on https://github.com/rust-lang/rust/pull/91381 which is causing the Android CI to fail.

luojia65:
Did this commit break rollup merge? :)
Ref: https://github.com/rust-lang/rust/pull/91548#issuecomment-988854839

crates/core_arch/src/x86/avx512f.rs
```diff
@@ -30323,2842 +30323,4138 @@ pub unsafe fn _mm512_store_pd(mem_addr: *mut f64, a: __m512d) {
...
-    e6: f64,
-    e7: f64,
-) -> __m512d {
-    let r = f64x8::new(e0, e1, e2, e3, e4, e5, e6, e7);
-    transmute(r)
+pub unsafe fn _mm512_mask_loadu_epi32(src: __m512i, k: __mmask16, mem_addr: *const i32) -> __m512i {
+    let mut dst: __m512i = src;
+    asm!(
+         "vmovdqu32 {2}{{{1}}}, [{0}]",
+         in(reg) mem_addr,
```
luojia65, line 30337:
```
Compiling std v0.0.0 (/checkout/library/std)
error: formatting may not be suitable for sub-register argument
     --> library/core/src/../../stdarch/crates/core_arch/src/x86/avx512f.rs:30336:34
      |
30336 |          "vmovdqu32 {2}{{{1}}}, [{0}]",
      |                                  ^^^
30337 |          in(reg) mem_addr,
      |                  -------- for this argument
      |
      = note: `-D asm-sub-register` implied by `-D warnings`
      = help: use the `e` modifier to have the register formatted as `eax`
      = help: or use the `r` modifier to keep the default formatting of `rax`
```
(from rollup CI Result)

Amanieu:
Yes. The issue is that x32 (x86_64 with 32-bit pointers) the address operand is inserted into the asm as `rax` instead of `eax`. The fix is to use the `:e` modifier on x86 and x32 (but not x86_64). Have a look at `bt.rs` for a similar case.

jhorstmann:
Oh, my bad. I'll keep this in mind when I start working on remaining intrinsics.

luojia65:
Awesome! I'd update submodule version in my fork soon for the pull request at rust-lang/rust.

luojia65:
@jhorstmann I submitted fix at: https://github.com/rust-lang/stdarch/pull/1264

PR #1259, opened 2021-11-26
Split aarch64 `pauth` feature into `paca` and `pacg`

adamgemmell:
As per discussion on https://github.com/rust-lang/rust/issues/86941, this is consistent with Linux and in the future there may well be environments with only one of address authentication or generic code authentication available.


Amanieu:
LGTM, but the feature names in rustc also need to be adjusted since the macro also checks for `cfg!(target_feature = "<name>")` which is set by rustc.

adamgemmell:
Thanks, I didn't realise the two were coupled like that

Amanieu:
This is still blocked on getting these features split on the rustc side.

Amanieu:
This needs a rebase due to a conflict with #1239.

PR #1261, opened 2021-11-29
Fix _mm_extract_ps example.

ehuss:
The example fails when run with rustdoc since this crate denies warnings. The test has an unneeded `unsafe` block which triggers `unused_unsafe`.

This normally doesn't show up in CI since it requires testing with a `-C target-cpu` that has sse4.1.

### DIFF
--- a/crates/core_arch/src/x86/sse41.rs
+++ b/crates/core_arch/src/x86/sse41.rs
@@ -155,11 +155,9 @@ pub unsafe fn _mm_blend_ps<const IMM4: i32>(a: __m128, b: __m128) -> __m128 {
 /// #       #[target_feature(enable = "sse4.1")]
 /// #       unsafe fn worker() {
 /// let mut float_store = vec![1.0, 1.0, 2.0, 3.0];
-/// unsafe {
-///     let simd_floats = _mm_set_ps(2.5, 5.0, 7.5, 10.0);
-///     let x: i32 = _mm_extract_ps::<2>(simd_floats);
-///     float_store.push(f32::from_bits(x as u32));
-/// }
+/// let simd_floats = _mm_set_ps(2.5, 5.0, 7.5, 10.0);
+/// let x: i32 = _mm_extract_ps::<2>(simd_floats);
+/// float_store.push(f32::from_bits(x as u32));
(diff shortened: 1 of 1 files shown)


crates/core_arch/src/x86/sse41.rs
```diff
@@ -155,11 +155,9 @@ pub unsafe fn _mm_blend_ps<const IMM4: i32>(a: __m128, b: __m128) -> __m128 {
...
 /// #       unsafe fn worker() {
 /// let mut float_store = vec![1.0, 1.0, 2.0, 3.0];
-/// unsafe {
 ///     let simd_floats = _mm_set_ps(2.5, 5.0, 7.5, 10.0);
 ///     let x: i32 = _mm_extract_ps::<2>(simd_floats);
 ///     float_store.push(f32::from_bits(x as u32));
-/// }
 /// assert_eq!(float_store, vec![1.0, 1.0, 2.0, 3.0, 5.0]);
 /// #       }
 /// #       unsafe { worker() }
```
workingjubilee, line 163: This appears "unused" from inside the source, but not to the user reading the docs, due to the unsafe "worker" function (that is required to use `target_feature(enable)`) it is nested inside, so if a user copies this code from the rendered docs with this change, they will need to readd the `unsafe { }`. That is to say: this makes the documentation less correct. This is an instance of `unused_unsafe` only because `fn worker` is not annotated `#[warn(unsafe_op_in_unsafe_fn)]`, which should also eliminate the warning, and without altering the correctness of the documentation.

Amanieu:
CI is blocked on https://github.com/rust-lang/rust/pull/91381.

moxian:
Pardon me bumming in like this, but should this PR even be needed in face of https://rust-lang.github.io/rfcs/2585-unsafe-block-in-unsafe-fn.html ? Wouldn't the better fix be adding `#[warn(unsafe_op_in_unsafe_fn)]` on the `unsafe worker()` instead (that changes from "`unsafe` blocks inside `unsafe fn` are redundant and are warned about" to "`unsafe` blocks inside `unsafe fn` are required, and their *absence* is warned about") ?

PR #1262, opened 2021-12-02
Add RISC-V platform and PAUSE instruction

luojia65:
This pull request contains two parts: new RISC-V platform `core::arch` package, and a first platform specific instruction wrapper `pause()` function on RISC-V platform.

As is defined in RISC-V Unprivileged Specification version 20191214-draft (December 1, 2021, with Zihintpause 2.0 Ratified), the PAUSE instruction is defined as follows:

> Chapter 4
> 
> “Zihintpause” Pause Hint, Version 2.0
> 
> The PAUSE instruction is a HINT that indicates the current hart’s rate of instruction retirement
> should be temporarily reduced or paused. The duration of its effect must be bounded and may be
> zero. No architectural state is changed.
> 
> ...
> 
> PAUSE is encoded as a FENCE instruction with pred=W, succ=0, fm=0, rd=x0, and rs1=x0.

This instruction is useful when we need to implement `core::hint::spin_loop()` function for RISC-V platforms.

The function `pause()` is defined unsafe because caller must ensure it's called on RISC-V platforms. As is defined in the Specification, implementation of the corresponding pause instruction varies between platforms, it may be spin loop hint, a no-op or others, but will never throw illegal instruction exception. This instruction does not perform as a memory barrier.

For using `.word`: the `pause` pseudoinstruction would be supported after this LLVM review: [D93019](https://reviews.llvm.org/D93019), by now (if I am correct) I can't write this instruction in current version Rust's `asm!` macro. In case if I'm incorrect or this way is inappropriate, I'd submit this pull request as a draft.

### DIFF
--- a/crates/core_arch/src/riscv/mod.rs
+++ b/crates/core_arch/src/riscv/mod.rs
@@ -1,1 +1,10 @@
 //! RISC-V intrinsics
+
+/// Generates the `PAUSE` instruction
+///
+/// The PAUSE instruction is a HINT that indicates the current hart's rate of instruction retirement
+/// should be temporarily reduced or paused. The duration of its effect must be bounded and may be zero.
+#[inline]
+pub fn pause() {
+    unsafe { asm!(".word 0x0100000F", options(nomem, nostack)) }
+}
--- a/crates/core_arch/src/riscv/mod.rs
+++ b/crates/core_arch/src/riscv/mod.rs
@@ -1,0 +1,1 @@
+//! RISC-V intrinsics
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -72,6 +72,7 @@ jobs:
         - arm-unknown-linux-gnueabihf
         - armv7-unknown-linux-gnueabihf
         - aarch64-unknown-linux-gnu
+        - riscv64gc-unknown-linux-gnu
         - powerpc64le-unknown-linux-gnu
         - mips-unknown-linux-gnu
         - mips64-unknown-linux-gnuabi64
--- a/crates/core_arch/src/core_arch_docs.md
+++ b/crates/core_arch/src/core_arch_docs.md
@@ -185,6 +185,7 @@ others at:
 * [`x86_64`]
 * [`arm`]
 * [`aarch64`]
+* [`riscv`]
 * [`mips`]
 * [`mips64`]
 * [`powerpc`]
@@ -196,6 +197,7 @@ others at:
 [`x86_64`]: x86_64/index.html
 [`arm`]: arm/index.html
 [`aarch64`]: aarch64/index.html
+[`riscv`]: riscv/index.html
 [`mips`]: mips/index.html
 [`mips64`]: mips64/index.html
 [`powerpc`]: powerpc/index.html
--- a/crates/core_arch/src/mod.rs
+++ b/crates/core_arch/src/mod.rs
@@ -55,6 +55,16 @@ pub mod arch {
         pub use crate::core_arch::aarch64::*;
     }
 
+    /// Platform-specific intrinsics for the `riscv` platform.
+    ///
+    /// See the [module documentation](../index.html) for more details.
+    #[cfg(any(target_arch = "riscv32", target_arch = "riscv64", doc))]
+    #[doc(cfg(any(target_arch = "riscv32", target_arch = "riscv64")))]
+    #[unstable(feature = "stdsimd", issue = "27731")]
+    pub mod riscv {
+        pub use crate::core_arch::riscv::*;
+    }
+
     /// Platform-specific intrinsics for the `wasm32` platform.
     ///
     /// This module provides intrinsics specific to the WebAssembly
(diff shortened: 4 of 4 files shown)


lu-zero:
Thank you for your contribution!

> This pull request contains two parts: new RISC-V platform `core::arch` package, and a first platform specific instruction wrapper `pause()` function on RISC-V platform.

Please split it in two commits.

> 
> As is defined in RISC-V Unprivileged Specification version 20191214-draft (December 1, 2021, with Zihintpause 2.0 Ratified), the PAUSE instruction is defined as follows:
> 
> > Chapter 4
> > “Zihintpause” Pause Hint, Version 2.0
> > The PAUSE instruction is a HINT that indicates the current hart’s rate of instruction retirement
> > should be temporarily reduced or paused. The duration of its effect must be bounded and may be
> > zero. No architectural state is changed.
> > ...
> > PAUSE is encoded as a FENCE instruction with pred=W, succ=0, fm=0, rd=x0, and rs1=x0.
> 
> This instruction is useful when we need to implement `core::hint::spin_loop()` function for RISC-V platforms.
> 
> The function `pause()` is defined unsafe because caller must ensure it's called on RISC-V platforms. As is defined in the Specification, implementation of the corresponding pause instruction varies between platforms, it may be spin loop hint, a no-op or others, but will never throw illegal instruction exception. This instruction does not perform as a memory barrier.

It is not unsafe for that reason and probably it can be considered safe (see the reasoning of safe/unsafe in wasm).

> For using `.word`: the `pause` pseudoinstruction would be supported after this LLVM review: [D93019](https://reviews.llvm.org/D93019), by now (if I am correct) I can't write this instruction in current version Rust's `asm!` macro. In case if I'm incorrect or this way is inappropriate, I'd submit this pull request as a draft.

Once we use the right version of llvm it should work with `asm!`, I think.

crates/core_arch/src/riscv/mod.rs
```diff
@@ -0,0 +1,10 @@
+//! RISC-V intrinsics
+
+/// Generates the `PAUSE` instruction
+///
+/// The PAUSE instruction is a HINT that indicates the current hart's rate of instruction retirement
+/// should be temporarily reduced or paused. The duration of its effect must be bounded and may be zero.
+#[inline(always)]
+pub unsafe fn pause() {
```
lu-zero, line 8: You should add a feature that maps to the extension, see what it is done for the other arch extensions.

luojia65: Yes, adding feature gate to extension specific instructions are necessary. For pause instruction, this instruction is in `HINT` space (in `I` set which must be implemented by RISC-V cores), whether there is Zihintpause or not does not change architectural state after instruction is retired. It will always run on all RISC-V cores. I don't know if it's still required to add feature gate to this function, or should we write it down in documentation for clarification. :)

crates/core_arch/src/riscv/mod.rs
```diff
@@ -1 +1,10 @@
 //! RISC-V intrinsics
+
+/// Generates the `PAUSE` instruction
+///
+/// The PAUSE instruction is a HINT that indicates the current hart's rate of instruction retirement
+/// should be temporarily reduced or paused. The duration of its effect must be bounded and may be zero.
+#[inline(always)]
+pub fn pause() {
+    unsafe { asm!(".word 0x0100000F") }
```
bjorn3, line 9: Can you set the nomem and nostack options? Also does LLVM have an intrinsic for this instruction?

luojia65: Hello! I'll perform changes soon. BTW, I checked the LLVM source here (https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/IR/IntrinsicsRISCV.td), but I didn't find LLVM intrinsic for RISC-V pause instruction.

bjorn3: There is a PR for adding the pause instruction in inline asm. It seems to have been accepted but not yet merged for over a year: https://reviews.llvm.org/D93019

crates/core_arch/src/riscv/mod.rs
lu-zero: You may add a comment about it.

I assume nothing else would use the same value so either it is ignored or it is used as intended and would not change in unexpected ways the behavior in the future.

Can you confirm that I understood correctly?

crates/core_arch/src/riscv/mod.rs
```diff
@@ -0,0 +1,10 @@
+//! RISC-V intrinsics
+
+/// Generates the `PAUSE` instruction
+///
+/// The PAUSE instruction is a HINT that indicates the current hart's rate of instruction retirement
+/// should be temporarily reduced or paused. The duration of its effect must be bounded and may be zero.
+#[inline(always)]
```
Amanieu, line 7: Is there a specific reason for using `#[inline(always)]`. A normal `#[inline]` should be sufficient.

Amanieu:
Other than that, LGTM.

crates/core_arch/src/riscv/mod.rs
luojia65: Hello! I suspect if a single instruction wrap would gain better performance if it inlines. Yes, for some debug purposes we should not inline these functions. and `#[inline]` only would behave better in these scenearios.

Amanieu:
This PR is still marked as a draft, is there anything else to do?

luojia65:
@Amanieu The instruction is encoded as `.word` by now instead of a normal instruction name, other than that that's nothing else to do :)

PR #1266, opened 2021-12-10
Stabilize armv8 neon instruction set on aarch64

SparrowLii:
According to the description in [#90972](https://github.com/rust-lang/rust/issues/90972) and the fcp has entered the final comment period stage, I think we can stabilize the armv8 neon instruction set on aarch64.

### DIFF
--- a/crates/core_arch/src/aarch64/crc.rs
+++ b/crates/core_arch/src/aarch64/crc.rs
@@ -13,6 +13,7 @@ use stdarch_test::assert_instr;
 #[inline]
 #[target_feature(enable = "crc")]
 #[cfg_attr(test, assert_instr(crc32x))]
+#[stable(feature = "neon_intrinsics", since = "1.59.0")]
 pub unsafe fn __crc32d(crc: u32, data: u64) -> u32 {
     crc32x_(crc, data)
 }
@@ -21,6 +22,7 @@ pub unsafe fn __crc32d(crc: u32, data: u64) -> u32 {
 #[inline]
 #[target_feature(enable = "crc")]
 #[cfg_attr(test, assert_instr(crc32cx))]
+#[stable(feature = "neon_intrinsics", since = "1.59.0")]
 pub unsafe fn __crc32cd(crc: u32, data: u64) -> u32 {
     crc32cx_(crc, data)
 }
--- a/crates/core_arch/src/aarch64/crc.rs
+++ b/crates/core_arch/src/aarch64/crc.rs
@@ -13,7 +13,6 @@ use stdarch_test::assert_instr;
 #[inline]
 #[target_feature(enable = "crc")]
 #[cfg_attr(test, assert_instr(crc32x))]
-#[stable(feature = "neon_intrinsics", since = "1.59.0")]
 pub unsafe fn __crc32d(crc: u32, data: u64) -> u32 {
     crc32x_(crc, data)
 }
@@ -22,7 +21,6 @@ pub unsafe fn __crc32d(crc: u32, data: u64) -> u32 {
 #[inline]
 #[target_feature(enable = "crc")]
 #[cfg_attr(test, assert_instr(crc32cx))]
-#[stable(feature = "neon_intrinsics", since = "1.59.0")]
 pub unsafe fn __crc32cd(crc: u32, data: u64) -> u32 {
     crc32cx_(crc, data)
 }
--- a/crates/core_arch/src/aarch64/tme.rs
+++ b/crates/core_arch/src/aarch64/tme.rs
@@ -71,7 +71,7 @@ pub const _TMFAILURE_TRIVIAL: u64 = 1 << 24;
 /// [ARM TME Intrinsics](https://developer.arm.com/docs/101028/0010/transactional-memory-extension-tme-intrinsics).
 #[inline]
 #[target_feature(enable = "tme")]
-#[cfg_attr(test, assert_instr(tstart))]
+#[cfg_attr(test, assert_instr(nop))]
 pub unsafe fn __tstart() -> u64 {
     aarch64_tstart()
 }
@@ -83,7 +83,7 @@ pub unsafe fn __tstart() -> u64 {
 /// [ARM TME Intrinsics](https://developer.arm.com/docs/101028/0010/transactional-memory-extension-tme-intrinsics).
 #[inline]
 #[target_feature(enable = "tme")]
-#[cfg_attr(test, assert_instr(tcommit))]
+#[cfg_attr(test, assert_instr(nop))]
 pub unsafe fn __tcommit() {
     aarch64_tcommit()
 }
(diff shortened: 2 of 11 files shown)


crates/core_arch/src/aarch64/crc.rs
```diff
@@ -13,6 +13,7 @@ use stdarch_test::assert_instr;
 #[inline]
 #[target_feature(enable = "crc")]
 #[cfg_attr(test, assert_instr(crc32x))]
+#[stable(feature = "neon_intrinsics", since = "1.59.0")]
 pub unsafe fn __crc32d(crc: u32, data: u64) -> u32 {
```
Amanieu, line 17: The FCP only covers the NEON (SIMD) intrinsics, not crc intrinsics.

crates/core_arch/src/aarch64/v8.rs
```diff
@@ -11,20 +11,23 @@ use stdarch_test::assert_instr;
 /// Reverse the order of the bytes.
 #[inline]
 #[cfg_attr(test, assert_instr(rev))]
+#[stable(feature = "neon_intrinsics", since = "1.59.0")]
 pub unsafe fn _rev_u64(x: u64) -> u64 {
```
Amanieu, line 15: Same here, there intrinsics are not covered by the FCP.

crates/core_arch/src/aarch64/tme.rs
```diff
@@ -71,7 +71,7 @@ pub const _TMFAILURE_TRIVIAL: u64 = 1 << 24;
 /// [ARM TME Intrinsics](https://developer.arm.com/docs/101028/0010/transactional-memory-extension-tme-intrinsics).
 #[inline]
 #[target_feature(enable = "tme")]
-#[cfg_attr(test, assert_instr(tstart))]
+#[cfg_attr(test, assert_instr(nop))]
```
Amanieu, line 74: Why were these changed to nop? This is incorrect.

crates/stdarch-gen/neon.spec
```diff
@@ -7044,7 +7044,7 @@ a = 1.1, 1.9, -1.7, -2.3
 validate 1.0, 2.0, -2.0, -2.0
 target = frintts
 
-aarch64 = frint32x
+aarch64 = nop
```
Amanieu, line 7047: Why is this changed to nop? This is incorrect.

crates/core_arch/src/arm_shared/neon/mod.rs
```diff
@@ -4811,7 +6687,7 @@ pub unsafe fn vpadalq_u32(a: uint64x2_t, b: uint32x4_t) -> uint64x2_t {
 #[target_feature(enable = "neon")]
 #[cfg_attr(target_arch = "arm", target_feature(enable = "v8"))]
 #[cfg_attr(all(test, target_arch = "arm"), assert_instr(nop))]
-#[cfg_attr(all(test, target_arch = "aarch64"), assert_instr(smmla))]
+#[cfg_attr(all(test, target_arch = "aarch64"), assert_instr(nop))]
```
Amanieu, line 6690: Why is this changed to nop? This is incorrect.

Also why is this function not marked as stable?

crates/stdarch-gen/src/main.rs
```diff
@@ -1064,9 +1064,9 @@ fn gen_aarch64(
     };
     let current_target = match target {
         Default => "neon",
-        ArmV7 => "v7",
-        Vfp4 => "vfp4",
-        FPArmV8 => "fp-armv8,v8",
+        ArmV7 => "neon",
+        Vfp4 => "neon",
+        FPArmV8 => "neon",
```
Amanieu, line 1069: Why did you change this? The previous target features were correct.

Amanieu:
We still need to wait ~1 week for the FCP to complete. Once it is complete, this PR can be merged.

crates/core_arch/src/aarch64/crc.rs
SparrowLii: OK. Unmarked it.

crates/core_arch/src/aarch64/v8.rs
SparrowLii: OK. Unmarked it.

crates/core_arch/src/aarch64/tme.rs
SparrowLii: The `assert_instr` of these instructions failed on my local aarch64. But since these instructions are not included in this stabilization, I think we can keep them unchanged.

crates/stdarch-gen/neon.spec
SparrowLii: Same as above. Changes have been cancelled.

crates/core_arch/src/arm_shared/neon/mod.rs
SparrowLii: Same as above. And vmmla instructions require `v8.6a` architecture, so they are not included in this stabilization.

crates/stdarch-gen/src/main.rs
SparrowLii: There are two functions used to generate instructions: `gen_arm` and `gen_aarch64`. The part I changed is in `gen_aarch64`. The values of these targets do not affect the `target_feature` on aarch64, so unified to `neon` I think it is more appropriate

crates/core_arch/src/arm_shared/neon/mod.rs
```diff
@@ -4179,6 +5835,10 @@ pub unsafe fn vext_s64<const N: i32>(a: int64x1_t, _b: int64x1_t) -> int64x1_t {
 #[cfg_attr(all(test, target_arch = "arm"), assert_instr("nop", N = 0))]
 #[cfg_attr(all(test, target_arch = "aarch64"), assert_instr("nop", N = 0))]
 #[rustc_legacy_const_generics(2)]
+#[cfg_attr(
+    target_arch = "aarch64",
+    stable(feature = "neon_intrinsics", since = "1.59.0")
+)]
 pub unsafe fn vext_u64<const N: i32>(a: uint64x1_t, _b: uint64x1_t) -> uint64x1_t {
     if N != 0 {
         unreachable_unchecked()
```
Amanieu, line 5844: I just noticed this: we should be using a static assert here, not `unreachable_unchecked`. This applies to all `vext_{s,u,p,f}64`.

crates/stdarch-gen/src/main.rs
Amanieu: That's fine if it's only on aarch64.

crates/core_arch/src/aarch64/tme.rs
Amanieu: It might have failed because the binutils version on your system is too old and objdump is not able to disassemble the instruction. We use a recent version of ubuntu in the docker image used in CI for this reason.

glandium:
Is it an oversight that many of these now stable functions take arguments with types that still aren't stable? (I say many, but I haven't checked, it could just as well be all of them)

thomcc:
It’s not an oversight, it’s a rustdoc bug. Those are stable on aarch64, but not arm32, so it gets mislabeled (theres an open issue for it, I can’t provide it at the moment though)

PR #1271, opened 2022-01-01
More RISC-V instructions in `core::arch`

luojia65:
This pull request includes:

- Hypervisor extension, including memory management insturctions and virtual machine memory load and store instructions
- Supervisor memory management, including address translation fence insturctions (important for Rust kernel development) and cache invalidation insturctions
- Non privileged instructions FENCE.I, NOP and WFI

Module `riscv64` is added to include RV64 only virtual machine memory instructions. RV64 supports all RV32 virtual machine instructions, it's implemented like:

```rust
    pub mod riscv64 {
        pub use crate::core_arch::riscv::*;
        pub use crate::core_arch::riscv64::*;
    }
```

(Edit: implemented like `pub use crate::core_arch::riscv32::*;` etc)

Most of implemented functions are marked as unsafe. That's because these functions either load or store effectively into raw pointer (hlv, hlvx and hsv instructions), or would effect how the RISC-V core handles page memory translation (memory management fences and cache invalidations).

Specifically, here's a list of implemented instruction functions from updated RISC-V core arch module:

<details>

图片
</details>

~~Marked as draft: more polishing on documents~~(done)~~, and decide whether #[must_use] is necessary on virtual machine memory loads.~~(ref: ptr::read & ptr::write, does not put #[must_use] by now)

### DIFF
--- a/crates/core_arch/src/mod.rs
+++ b/crates/core_arch/src/mod.rs
@@ -66,6 +66,17 @@ pub mod arch {
         pub use crate::core_arch::riscv::*;
     }
 
+    /// Platform-specific intrinsics for the `riscv64` platform.
+    ///
+    /// See the [module documentation](../index.html) for more details.
+    #[cfg(any(target_arch = "riscv64", doc))]
+    #[doc(cfg(any(target_arch = "riscv64")))]
+    #[unstable(feature = "stdsimd", issue = "27731")]
+    pub mod riscv64 {
+        pub use crate::core_arch::riscv::*;
+        pub use crate::core_arch::riscv64::*;
+    }
+
     /// Platform-specific intrinsics for the `wasm32` platform.
     ///
     /// This module provides intrinsics specific to the WebAssembly
@@ -265,6 +276,10 @@ mod arm;
 #[cfg(any(target_arch = "riscv32", target_arch = "riscv64", doc))]
 #[doc(cfg(any(target_arch = "riscv32", target_arch = "riscv64")))]
 mod riscv;
+
+#[cfg(any(target_arch = "riscv64", doc))]
+#[doc(cfg(any(target_arch = "riscv64")))]
+mod riscv64;
 
 #[cfg(any(target_family = "wasm", doc))]
 #[doc(cfg(target_family = "wasm"))]
--- a/crates/core_arch/src/mod.rs
+++ b/crates/core_arch/src/mod.rs
@@ -56,14 +56,14 @@ pub mod arch {
         pub use crate::core_arch::aarch64::*;
     }
 
-    /// Platform-specific intrinsics for the `riscv` platform.
+    /// Platform-specific intrinsics for the `riscv32` platform.
     ///
     /// See the [module documentation](../index.html) for more details.
-    #[cfg(any(target_arch = "riscv32", target_arch = "riscv64", doc))]
-    #[doc(cfg(any(target_arch = "riscv32", target_arch = "riscv64")))]
+    #[cfg(any(target_arch = "riscv32", doc))]
+    #[doc(cfg(any(target_arch = "riscv32")))]
     #[unstable(feature = "stdsimd", issue = "27731")]
-    pub mod riscv {
-        pub use crate::core_arch::riscv::*;
+    pub mod riscv32 {
+        pub use crate::core_arch::riscv_shared::*;
     }
 
     /// Platform-specific intrinsics for the `riscv64` platform.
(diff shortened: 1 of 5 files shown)


crates/core_arch/src/mod.rs
```diff
@@ -66,6 +66,17 @@ pub mod arch {
         pub use crate::core_arch::riscv::*;
```
Amanieu, line 66: If we're going to have separate intrinsics for riscv32 and riscv64 then we should probably rename the existing `riscv` module to `riscv32`. This makes it consistent with the other architectures.

Amanieu (commented):
LGTM other than the module name issue.

crates/core_arch/src/mod.rs
luojia65: Yes, I'd accept this suggestion. RISC-V RV64 only contains more instructions in each set than RISC-V RV32, no instructions are removed from RISC-V RV32 yet. RV32 is not a legacy extension, it's better to make it explict. This way provides more readable rust docs as well.
After this change, I'd modify the implementation of `core::hint::spin_loop` in standard library.

luojia65:
I modified `riscv` to `riscv32`.

r? @Amanieu

crates/core_arch/src/riscv32/mod.rs
```diff
@@ -0,0 +1,481 @@
...
+    unsafe { asm!(".insn i 0x0F, 0, x0, x0, 0x010", options(nomem, nostack)) }
+}
+
+/// Generates the `NOP` instruction
+///
+/// The NOP instruction does not change any architecturally visible state, except for
+/// advancing the `pc` and incrementing any applicable performance counters.
+#[inline]
+pub fn nop() {
+    unsafe { asm!("nop") }
```
Amanieu, line 20: All of the `asm!` blocks here should use `nostack` since they don't push anything to the stack.

Additionally:
- Most should use `nomem`, except for loads which should use `readonly` and stores which should use neither.

crates/core_arch/src/mod.rs
```diff
@@ -264,7 +276,11 @@ mod arm;
 
 #[cfg(any(target_arch = "riscv32", target_arch = "riscv64", doc))]
 #[doc(cfg(any(target_arch = "riscv32", target_arch = "riscv64")))]
-mod riscv;
+mod riscv32;
```
Amanieu, line 279: Perhaps calling this `riscv_shared` (like `arm_shared`) makes more sense?

crates/core_arch/src/riscv32/mod.rs
luojia65: Thanks, I'll check it out.
There are many privileged memory fences as well, I'd also figure it out in these intrinsic functions.

crates/core_arch/src/mod.rs
luojia65: That makes sense, it would be a more common module system design.

crates/core_arch/src/riscv32/mod.rs
Amanieu: Good point about the fences, those should not use `nomem`/`readonly` because they need to act as a compiler fence.

luojia65:
Hello! I renamed internal shared riscv module to `riscv_shared`, and modified asm! options under following rules:

- Trivial no-op instruction (NOP), wait-for-interrupt (WFI): `options(nomem, nostack)`
- Unprivileged fences (FENCE.I), privileged fences (SFENCE.*, SINVAL.*, HFENCE.*, HINVAL.*): `options(nostack)`
- Hypervisor loads (HLV.*, HLVX.*): `options(readonly, nostack)`
- Hypervisor stores (HSV.*): `options(nostack)`

r? @Amanieu

crates/core_arch/src/riscv64/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+/// Loads virtual machine memory by unsigned word integer
+///
+/// This instruction performs an explicit memory access as though `V=1`;
+/// i.e., with the address translation and protection, and the endianness, that apply to memory
+/// accesses in either VS-mode or VU-mode.
+///
+/// This operation is not available under RV32 base instruction set.
+///
+/// This function is unsafe for it accesses the virtual supervisor or user via a `HLV.WU`
+/// instruction which is effectively an unreference to any memory address.
```
luojia65, line 13: note: in following pull requests the word 'unreference' should be replaced by 'dereference' for consistency

luojia65: fixed in https://github.com/rust-lang/stdarch/pull/1278

PR #1275, opened 2022-01-24
fixed documentation crc32 -> crc32-c (with x86(_64))

post-rex:
x86(-64) actually implements crc32-c and not crc32, the intel documentation fails to mention this properly.
The only indication is the polynomial they use (0x11EDC6F41) in the `Operation` section.
(See: https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_crc32_&ig_expand=1495,1494,1493,1493)
Cross-referencing with this site, the polynomial is mentioned to be of crc32-c (https://sci.math.narkive.com/7MUM37mh/when-is-regular-crc32-better-or-worse-than-crc32-c).
And finally on Wikipedia it is mentioned in the description of `CRC32` that it is actually `CRC32C` (https://en.wikipedia.org/wiki/SSE4#SSE4.2).

Finally, I have tested this experimental with the `crc32c` crate:
```rs
fn main() {
    let intel = unsafe {
        std::arch::x86_64::_mm_crc32_u32(0xFFFF_FFFF, 0xDEADBEEF) ^ 0xFFFF_FFFF
    };
    assert_eq!(intel, crc32c::crc32c(&[0xEF, 0xBE, 0xAD, 0xDE]));
}
```
Note that _mm_crc32_u32 updates a crc32 state, the initial state is `0xFFFF_FFFF` and gets finalized by XORing it with `0xFFFF_FFFF`.

### DIFF
--- a/crates/core_arch/src/x86/sse42.rs
+++ b/crates/core_arch/src/x86/sse42.rs
@@ -520,7 +520,7 @@ pub unsafe fn _mm_cmpestra<const IMM8: i32>(a: __m128i, la: i32, b: __m128i, lb:
 }
 
 /// Starting with the initial value in `crc`, return the accumulated
-/// CRC32 value for unsigned 8-bit integer `v`.
+/// CRC32-C value for unsigned 8-bit integer `v`.
 ///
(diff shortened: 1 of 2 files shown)


Amanieu:
The descriptions that we use are taken straight from Intel's intrinsic documentation. But I don't mind making this change since it makes things clearer.

PR #1277, opened 2022-01-28
Change core_arch_docs.md so it works from both std and core

Amanieu:


lu-zero (approved):
Seems fine to me.

Amanieu:
There will be broken links if you try to build the docs for the `core_arch` crate itself, but nobody reads those anyways.

PR #1279, opened 2022-01-30
Complete avx512vbmi2

minybot:
expandloadu_epi16,epi8
mask_compressstoreu_epi16,epi8

### DIFF
--- a/crates/core_arch/avx512vbmi2.md
+++ b/crates/core_arch/avx512vbmi2.md
@@ -12,12 +12,12 @@
   * [x] [`_mm256_maskz_compress_epi8`]
   * [x] [`_mm512_mask_compress_epi8`]
   * [x] [`_mm512_maskz_compress_epi8`]
-  * [_] [`_mm_mask_compressstoreu_epi16`] //need i1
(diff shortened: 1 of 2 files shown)


jhorstmann:
Nice, I also had those on my todo list but did not get around to implementing them yet. Note that there are also llvm intrinsics for expand load (examples [1](https://github.com/llvm/llvm-project/blob/dc6a344637a6417aedf85e5fd1236e7fcd43c3f1/llvm/test/CodeGen/X86/avx512-intrinsics-upgrade.ll#L8977), [2](https://github.com/llvm/llvm-project/blob/dc6a344637a6417aedf85e5fd1236e7fcd43c3f1/llvm/test/CodeGen/X86/avx512vbmi2-intrinsics-upgrade.ll#L18)) which might lead to a bit simpler code and automatically support 32-bit pointers.

jhorstmann:
There are two variants, the architecture independent ones with `i1` masks and the avx512 specific ones, like `llvm.x86.avx512.mask.expand.load.w.512(i8* %addr, <32 x i16> %data, i32 %mask)`. I don't know whether the avx512 specific ones are deprecated, that would probably make `asm` the better option.

PR #1283, opened 2022-02-16
Avoid the use of #[cfg(doc)] in std_detect

Amanieu:
This doesn't work when building the standard library docs since that
cfg flag is not passed to dependencies.


adamgemmell:
Much more clever than my solution with the macro use! I see this also marks the required error macros as stable.

Could we put in `#[doc(cfg(...))]` attributes too?

PR #1284, opened 2022-02-18
Stabilize vget_low_s8 on AArch64

Amanieu:
This was missed  when most of the AArch64 intrinsics were stabilized.

cc https://github.com/rust-lang/rust/issues/94083

### DIFF
--- a/crates/core_arch/src/arm_shared/neon/mod.rs
+++ b/crates/core_arch/src/arm_shared/neon/mod.rs
@@ -4989,6 +4989,10 @@ pub unsafe fn vget_high_f32(a: float32x4_t) -> float32x2_t {
 #[target_feature(enable = "neon")]
 #[cfg_attr(target_arch = "arm", target_feature(enable = "v7"))]
 #[cfg_attr(test, assert_instr(nop))]
+#[cfg_attr(
+    target_arch = "aarch64",
+    stable(feature = "vget_low_s8", since = "1.60.0")
+)]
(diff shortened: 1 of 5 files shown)


crates/core_arch/src/arm_shared/neon/mod.rs
```diff
@@ -4989,6 +4989,10 @@ pub unsafe fn vget_high_f32(a: float32x4_t) -> float32x2_t {
 #[target_feature(enable = "neon")]
 #[cfg_attr(target_arch = "arm", target_feature(enable = "v7"))]
 #[cfg_attr(test, assert_instr(nop))]
+#[cfg_attr(
+    target_arch = "aarch64",
+    stable(feature = "neon_intrinsics", since = "1.59.0")
```
xfix, line 4994: This probably shouldn't be 1.59.0 as it's probably not going to be backported.

Amanieu: I tried that but rustc doesn't like it when I put different version numbers on the same feature name.

xfix: I think you can use a different feature name such as `vget_low_s8`, then rustc won't complain. Feature names for stable features are mostly irrelevant anyway. This was done before, see https://github.com/rust-lang/rust/pull/86344 for instance (in this case, this stabilized `MaybeUninit::write` which back then used `maybe_uninit_extra` feature without stabilizing everything else that used `maybe_uninit_extra` feature.

PR #1285, opened 2022-02-28
Add the rdm target feature to the sqrdmlsh intrinsic.

JamieCunliffe:
I noticed the `sqrdmlsh` intrinsic was asserting for the wrong instruction and didn't have the rdm target feature.

This has removed the stabilization of this intrinsic, I'm not sure what the policy is here as the other rdm intrinsics haven't yet been stabilized.


Amanieu:
We previously held back on the stabilization of the RDM intrinsics due to issues with LLVM 13, however these have been resolved now that rustc uses LLVM 14. I think we should just go ahead and stabilize all the RDM intrinsics.

@rfcbot fcp merge

yaahc:
I'm a bit confused. This remove's the `stable` attribute without adding back an `unstable` attribute. Is this a difference between the way `rust-lang/stdarch` works vs  how `rust-lang/rust` works? I would expect this to produce errors about missing stability attributes.

Amanieu:
This PR hasn't yet been updated to use the approach I am proposing, which is to mark all AArch64 intrinsics under the "rdm" target feature as stable. They have been tested to work correctly on AArch64 and produce the same results as the equivalent Clang intrinsics.

yaahc:
> This PR hasn't yet been updated to use the approach I am proposing, which is to mark all AArch64 intrinsics under the "rdm" target feature as stable. They have been tested to work correctly on AArch64 and produce the same results as the equivalent Clang intrinsics.

Aah, I see, so this was submitted to revert a broken stabilization and you're requesting that we instead go ahead with a proper stabilization to fix it.

Amanieu:
> Aah, I see, so this was submitted to revert a broken stabilization and you're requesting that we instead go ahead with a proper stabilization to fix it.

That's correct.

Amanieu:
ping @rust-lang/libs-api for checkboxes

yaahc:
> ping @rust-lang/libs-api for checkboxes

so to be clear, the current PR won't be merged, and this is just seeking approval for stabilizing the intrinsics before spending time updating the PR in case it gets denied? I hadn't submitted my checkmark because I was waiting for the PR to get updated. Is it also fair to assume that in the case that we decide against stabilization, the plan is to update the PR to add back the missing `unstable` attributes?

PR #1288, opened 2022-03-04
Use asm macro to rewrite Avx512 mask intrinsics

minybot:
kunpack: 16,32,64;
kand,kandn_mask: 16,32,64; 512_kand,kandn
kor_mask: 16,32,64; 512_kor
kxor_mask: 16,32,64; 512_kxor
kxnor_mask: 16,32,64; 512_kxnor
knot_mask: 16,32,64; 512_knot
512_kmov, int2mask, mask2int
_cvtmasku32,u64_mask32,64; _cvtmask32,64_u32,u64
kadd_mask32,64


Amanieu:
I don't think this is the right approach. I would prefer keeping mask operations on `k*` registers as normal integer operations and simply removing the instruction constraints we have (just set them to `nop`).

Amanieu:
The are useful if the value is already in a `k*` register instead of a general purpose register. But this is a decision for the register allocator to make. At the Rust level we treat them as equivalent.

PR #1299, opened 2022-03-24
Change remaining _undefined_ functions to zero-init

saethlin:
There are three functions which return uninitialized vectors by doing `MaybeUninit::uninit().assume_init()`:
https://github.com/rust-lang/stdarch/blob/bcbe010614f398ec86f3a9274d22e33e5f2ee60b/crates/core_arch/src/x86/sse2.rs#L2738-L2741
https://github.com/rust-lang/stdarch/blob/bcbe010614f398ec86f3a9274d22e33e5f2ee60b/crates/core_arch/src/x86/sse2.rs#L2749-L2752
https://github.com/rust-lang/stdarch/blob/bcbe010614f398ec86f3a9274d22e33e5f2ee60b/crates/core_arch/src/x86/avx.rs#L2721-L2724

Miri flags any code that calls these as UB<sup>1</sup>. This may or may not be too conservative of Miri, but the use of `MaybeUninit::uninit().assume_init()` seems unnecessary in these cases. There are other `_undefined_` intrinsics which just return a zero-init vector, so this is just completing that pattern.

There's also a FIXME comment on all of them that says they should return `MaybeUninit`, but I don't see how that's possible, because it looks like the functions have been stabilized. So this PR just deletes the comment :shrug: 

---

<sup>1</sup> Run `MIRIFLAGS=-Zmiri-check-number-validity cargo miri test` on any of: `aes==0.8.1`, `twox-hash==1.6.2`, `chacha20==0.8.1`, `safe_arch==0.6.0`, `grep-printer==0.1.6`, `cmac==0.6.0`, `sw-composite==0.7.14`, `xoodyak==0.7.3`, `sliceslice==0.3.1`, `jomini==0.16.4`, `debruijn==0.3.4`, `kuznyechik==0.7.2`, `typos==0.8.5`, `xoodoo==0.1.0`.


saethlin:
Ah, I see CI is kicking this back with the same errors that I got locally. Is this PR the right place to make the changes that the errors are suggesting?

Amanieu:
Sure, just update the nightly features in this PR.

PR #1305, opened 2022-05-06
Allow unused macro rules for two macros

est31:
The unused macro rules lint is an upcoming lint
for the rust compiler: https://github.com/rust-lang/rust/pull/96150. This PR prepares
for the introduction [so that the CI passes](https://github.com/rust-lang/rust/pull/96150#issuecomment-1119861986).

I have not tested whether this is all of the changes needed. If any more pop up I guess one can just add an `allow` to lib.rs.

### DIFF
--- a/crates/core_arch/src/macros.rs
+++ b/crates/core_arch/src/macros.rs
@@ -93,7 +93,7 @@ macro_rules! types {
     )*)
 }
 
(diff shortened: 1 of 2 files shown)


est31:
I've tested it now with rust-lang's CI and it works great. r? @Amanieu

PR #1306, opened 2022-05-10
Allow unused macro rules for simd_shuffle16 too

est31:
Follow-up of #1305


lu-zero (approved):

est31:
The `test-various` job is fine now as well:
Screenshot_20220510_190036

PR #1308, opened 2022-06-09
Remove useless conditional compilation

Urgau:
Context: `rustc` added a few month ago [`--check-cfg`](https://doc.rust-lang.org/nightly/unstable-book/compiler-flags/check-cfg.html) support and enable it in the `rust-lang/rust` repo for every build.

I found out that `stdarch` had condition that were triggering the unstable `unexpected_cfgs` lint:
 - `stdarch_intel_sde`: Internal to `stdarch`, status unknown, never set as far as I can tell. [bootstrap exception](https://github.com/rust-lang/rust/blob/1a97162cb245b5e2c7458c28859e3df779908c02/src/bootstrap/lib.rs#L201)
 - `dont_compile_me`: Used on the `_rbit_u32` test for armv7, with an outdated comment. [bootstrap exception](https://github.com/rust-lang/rust/blob/1a97162cb245b5e2c7458c28859e3df779908c02/src/bootstrap/lib.rs#L218-L219)
 - `core_arch_docs`: Set in `dox.sh` and in the `build.rs` of `core_arch`, but never used in the codebase. *no exception because never used*

This PR removes every cfg usage of `dont_compile_me` and `core_arch_docs` because they are useless. I tested the modifications and nothing failed on my side, so I think it's good to go.

### DIFF
--- a/ci/dox.sh
+++ b/ci/dox.sh
@@ -22,9 +22,6 @@ dox() {
   rm -rf "target/doc/${arch}"
   mkdir "target/doc/${arch}"
 
-  export RUSTFLAGS="--cfg core_arch_docs"
-  export RUSTDOCFLAGS="--cfg core_arch_docs"
-
   cargo build --verbose --target "${target}" --manifest-path crates/core_arch/Cargo.toml
   cargo build --verbose --target "${target}" --manifest-path crates/std_detect/Cargo.toml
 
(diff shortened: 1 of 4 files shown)


Urgau:
I would also like to remove `stdarch_intel_sde`. As far as I can tell it's only used with negative (`#[cfg(not(stdarch_intel_sde))]`) cfg and is never set by anything. Should I update the PR and also include it's removal ?

Amanieu:
`stdarch_intel_sde` is set in `.github/workflows/main.yml` when running x86 tests under Intel's x86 emulator. This is used to test features that may not be available on the CPU used in CI.

Amanieu:
The rest look fine to me.

PR #1309, opened 2022-06-12
Ensure the neon vector aggregates like `float32x4x4_t` are `#[repr(C)]`

thomcc:
These are C structs from `arm_neon.h`, so having them be `#[repr(Rust)]` is wrong. Also they are used by certain intrinsics, and I suspect if they had a weird layout, it would be very broken.

Moreover, it seems reasonable for someone to want to do things like transmute a `float32x4x4_t` into a `[float32x4_t; 4]` and such, which requires `#[repr(C)]`.

### DIFF
--- a/crates/core_arch/src/aarch64/neon/mod.rs
+++ b/crates/core_arch/src/aarch64/neon/mod.rs
@@ -28,14 +28,17 @@ types! {
 }
 
 /// ARM-specific type containing two `float64x1_t` vectors.
+#[repr(C)]
 #[derive(Copy, Clone, Debug)]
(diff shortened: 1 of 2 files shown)


infinity0:
@thomcc could this be related to rust-lang/rust#96983 ?

infinity0:
That bug is for 1.59 which predates this PR, so I was wondering if backporting this PR to 1.59 would fix this issue for us in Debian. However this doesn't match your explanation either, so I'll have to continue investigating. Anyway, thanks for the comment.

thomcc:
Ah, hm, yeah, I don't think it would do that, but could be mistaken.

PR #1310, opened 2022-06-20
Fully qualify recursive macro calls

adamgemmell:
Allows these edge-cases to be used without `use`ing the `std::arch` module.

This doesn't happen on x86 due to the macro existing in the `std` root but all other platforms seem to see a cryptic "macro not found" error.


Amanieu:
Nice catch!

PR #1315, opened 2022-07-18
Remove restrictions on compare-exchange memory ordering.

tmiasko:
Update cmpxchg16b to match changes from rust-lang/rust#98383.


tmiasko:
cc @m-ou-se

PR #1316, opened 2022-08-09
Adding documentation links for arm crc32 intrinsics

psvri:
Hello,

I have added documentation links for arms crc32 intrinsics . If this is okay , I will slowly add the links for other intrinsics in other PR's.

PR #1317, opened 2022-08-10
Adding doc links for arm crypto and aes intrinsics

psvri:
Hello,

This is followup of #1316 .
This PR has doc links for arms aes and crypto intrinsics. This should cover most of the non neon intrinsics.

Hopefully, for the neon intrinsics I shall work on it later since it would need changes in stdarch-gen .


Amanieu:
Thanks!

PR #1321, opened 2022-08-13
Allow internal use of stdsimd from detect_feature

tmiasko:
This allows using feature detection macros, without placing a
requirement of enabled stdsimd feature gate from end users.

A follow-up to changes from #1311, which introduced the new macro.

Noticed while preparing a bump of stdarch submodule in the standard library.

### DIFF
--- a/crates/std_detect/src/detect/macros.rs
+++ b/crates/std_detect/src/detect/macros.rs
@@ -1,4 +1,5 @@
 #[macro_export]
+#[allow_internal_unstable(stdsimd)]
 macro_rules! detect_feature {
     ($feature:tt, $feature_lit:tt) => {
         $crate::detect_feature!($feature, $feature_lit : $feature_lit)


Amanieu:
What exactly is this enabling? I'm concerned about accidentally stabilizing something we don't want to stabilize here.

tmiasko:
This allows the macro to internally use `stdsimd` feature, without requiring
end users to enable it. This is necessary to use already stable macros, without
requiring `stdsimd` feature gate.

Note that this implementation strategy is in place since #1303, and this pull
request merely extends the existing approach to a new macro. I presumed this
implementation strategy was already "blessed".

This change surfaces the same stability concerns as #1303. It makes `stdsimd`
feature gate ineffective from within macros. The only reason this seems fine to
me, is that things happen to be organized such that the generated macro itself
requires `stdsimd` feature gate whenever more fine grained target features
require it.

PR #1323, opened 2022-08-19
Replace some calls to `pointer::offset` with `add` and `sub`

WaffleLapkin:
See https://github.com/rust-lang/rust/pull/100746


WaffleLapkin:
Seeing https://github.com/rust-lang/stdarch/runs/7917316301?check_suite_focus=true (doc only change) I think CI fails not because of my changes 🤔

lu-zero:
It looks fairly strange since nobody touched ppc64le in ages. I'll try to see if the problem is in qemu by running on real hardware ...

lu-zero (commented):
The change looks fine though.

Amanieu:
CI should be fixed by #1326, can you rebase?

WaffleLapkin:
Rebased 👍🏻

PR #1335, opened 2022-09-13
Add HWCAP2 support for AArch64 Linux.

jacobbramley:
This doesn't extend the set of recognised features, but corrects some existing ones to refer to `HWCAP2`.

Note that this changes the behaviour when `HWCAP2` is absent on any platform. In particular:
* A zero result from `getauxval()` could just mean that none of those features are enabled.
* If `HWCAP2` is known to be truly absent (e.g. from reading `/proc/self/auxv`), but we have `HWCAP`, then it's likely that the kernel is too old to support the corresponding features, and looking at a fallback isn't likely to be productive.


jacobbramley:
I'm going to force-push to drop those Gerrit "Change-Id" tags. I apologise for the disruption.

jacobbramley:
Fixing the feature detection enabled a test that's apparently broken. (I didn't see it here because my hardware genuinely doesn't have that feature, so the test didn't run at all.) I'm investigating!

PR #1337, opened 2022-09-28
Update the Android NDK script to r25b

chriswailes:
Updating the Android NDK in accordance with new platform support policy.


chriswailes:
This is related to https://github.com/rust-lang/rust/pull/102332

Amanieu:
```
+ mv android-ndk-r25b-linux /android/ndk-x86_64
mv: cannot stat 'android-ndk-r25b-linux': No such file or directory
```
Perhaps the path within the zip is incorrect?

chriswailes:
> Perhaps the path within the zip is incorrect?

Sure was.  Needed to drop the `-linux` suffix.

Amanieu:
`bench_black_box` is now stable, you need to remove it from the unstable feature list.

Amanieu:
ping @chriswailes, is this PR still relevant?

Amanieu:
Actually it seems we've removed our Android CI. This is fine since stdarch doesn't really need OS-specific CI now that `std::detect` has been moved to the standard library.

PR #1338, opened 2022-09-30
Update the Android Docker files to Ubuntu 22.04

chriswailes:


Amanieu:
Can you include the `bench_black_box` fix here? Then I should be able to merge it.

Amanieu:
Thanks!

PR #1344, opened 2022-10-25
Fix dupe word typos

Rageking8:


Rageking8:
Hmm, seems like the CI failure is not caused by my PR since the rest of the other PRs also has the failure.

Amanieu:
Thanks!

PR #1349, opened 2022-10-29
Don't require AVX512 for 128/256-bit GFNI & VPCLMULQDQ intrinsics

### LINKED ISSUE #1325: avx512gfni and avx512vaes feature names might be misleading
I already opened rust-lang/rust#100752, but this might be the more relevant location.  More details are in the attached issue, but there exist CPUs that have GFNI and VAES without AVX-512, and Intel doesn't seem to classify them as part of AVX-512.  It might be better to name them simply `gfni` and `vaes`.

Amanieu:
Fixes #1325

### DIFF
--- a/crates/core_arch/src/x86/avx512gfni.rs
+++ b/crates/core_arch/src/x86/avx512gfni.rs
@@ -121,7 +121,7 @@ pub unsafe fn _mm512_maskz_gf2p8mul_epi8(k: __mmask64, a: __m512i, b: __m512i) -
 ///
 /// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm256_gf2p8mul_epi8)
 #[inline]
-#[target_feature(enable = "avx512gfni,avx512bw,avx512vl")]
+#[target_feature(enable = "avx512gfni,avx")]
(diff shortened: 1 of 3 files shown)


calebzulawski:
I think this requires a bit more changes to each individual intrinsic, including the EVEX ones.  For example, from the Software Developer's Manual:
Screen Shot 2022-10-29 at 2 58 31 PM

Note that the 512-bit EVEX encoded instructions only require `AVX512F`, but the smaller width EVEX instructions require `AVX512VL` (and neither require `AVX512BW`, which is what the intrinsics currently require).  VAES and VPCLMULQDQ have similar requirements.

Amanieu:
I left the EVEX-encoded 128/256-bit variant as they were (`"avx512gfni,avx512bw,avx512vl"`).

The AVX512BW requirement comes from LLVM, which will refuse to emit the correct instruction if we don't add that target feature.

calebzulawski:
That's unfortunate, sure enough I see the same issue.  I actually see an ICE when removing it.  Do you know if there's an issue reported to LLVM?

calebzulawski:
I opened llvm/llvm-project#58687

PR #1355, opened 2022-11-17
Rename misleading features

calebzulawski:
Renamed features according to rust-lang/rust#100752 (and #1325)

~This works on master, but rust-lang/rust#103750 hasn't hit nightly yet.~ landed on nightly

An open question: in `is_x86_feature_detected`, should `avx512gfni` be an alias for `gfni` (even though it doesn't actually check for `avx512f`), or should it be a separate feature that checks the multiple feature flags required to enable the AVX-512 GFNI instructions? (and the same goes for `avx512vaes` and `avx512vpclmulqdq`)

### DIFF
--- a/crates/core_arch/src/x86/gfni.rs
+++ b/crates/core_arch/src/x86/gfni.rs
@@ -65,7 +65,7 @@ extern "C" {
 ///
 /// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm512_gf2p8mul_epi8)
 #[inline]
-#[target_feature(enable = "avx512gfni,avx512bw,avx512f")]
+#[target_feature(enable = "gfni,avx512bw,avx512f")]
 #[cfg_attr(test, assert_instr(vgf2p8mulb))]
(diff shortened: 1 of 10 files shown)


Amanieu:
I think `avx512gfni` can just be an alias for `gfni`, however it would be nice to have a deprecation warning trigger when you use it.

calebzulawski:
I added a deprecation warning, but the way it's emitted is admittedly a hack.  It emits this reasonable message:
```
warning: use of deprecated macro `deprecated_feature`: the `avx512gfni` feature has been renamed to `gfni`
  --> crates/std_detect/tests/x86-specific.rs:40:33
   |
40 |     println!("avx512gfni {:?}", is_x86_feature_detected!("avx512gfni"));
   |                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(deprecated)]` on by default
   = note: this warning originates in the macro `is_x86_feature_detected` (in Nightly builds, run with -Z macro-backtrace for more info)

PR #1356, opened 2022-11-17
Detect MOVBE

calebzulawski:


thomcc:
Since this is insta-stable, does this need an FCP?

(Seems like it might be overkill, but worth calling out perhaps?)

calebzulawski:
I'm not really sure how about how an FCP or the attribute works in this case.  MOVBE itself was already stabilized, detection was omitted for whatever reason.

thomcc:
It might just be an oversight and not need FCP then. I suspect the `since` should still be the actual version, though. I think it can be the reviewer's call either way, but just felt it's worth mentioning.

dtolnay:
I'm marking @yaahc's box because she has stepped down from T-libs-api after the point that this feature got proposed for FCP.

Amanieu:
Note that `movbe_target_feature` is still unstable in rustc, it should probably be stabilized there as well.

PR #1364, opened 2022-12-28
Resolve old FIXME comments

Nugine:


crates/core_arch/src/wasm32/simd128.rs
```diff
@@ -2349,10 +2349,6 @@ pub use i8x16_all_true as u8x16_all_true;
 #[doc(alias("i8x16.bitmask"))]
 #[stable(feature = "wasm_simd", since = "1.54.0")]
 pub fn i8x16_bitmask(a: v128) -> u16 {
-    // FIXME(https://bugs.llvm.org/show_bug.cgi?id=50507) - this produces an
-    // extraneous `i32.and` instruction against a mask of 65535 when converting
-    // from the native intrinsic's i32 return value to our desired u16. This
-    // shouldn't be necessary, though, but requires upstream LLVM changes.
```
lu-zero, line 2355: Maybe keep the note and mention since which version it got fixed?

Nugine: Actually I cannot find when it was fixed (before 1.54?)

PR #1366, opened 2022-12-31
Stabilize x86 F16C

### LINKED ISSUE #1234: Why is `F16C` still unstable?
I noticed that most `x86_64` intrinsics were stable, yet `F16C` isn't. I'd like to know what is keeping it from hitting stable, and what can be done to make it stable. According to another [tracking issue comment](https://github.com/rust-lang/rust/issues/48556#issuecomment-743190548) there doesn't seem to be anything blocking it.

starkat99:
Resolves #1234

Compiler PR for stabilizing corresponding target_feature: https://github.com/rust-lang/rust/pull/106323

### DIFF
--- a/crates/core_arch/src/x86/f16c.rs
+++ b/crates/core_arch/src/x86/f16c.rs
@@ -29,6 +29,7 @@ extern "unadjusted" {
 #[inline]
 #[target_feature(enable = "f16c")]
 #[cfg_attr(test, assert_instr("vcvtph2ps"))]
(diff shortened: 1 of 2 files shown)


Amanieu:
@rfcbot fcp merge

starkat99:
https://github.com/rust-lang/rust/pull/106323 was merged, which stabilized the target feature

Shnatsel:
Github shows that this was merged after the release of Rust 1.68, but the attributes still indicate `#[stable(since = "1.68.0")]`. Is that intentional? If not, perhaps the attributes should be updated?

PR #1358, opened 2022-11-26
Stabilize cmpxchg16b instrinsic

Nugine:
Resolves #827

PR #1367, opened 2023-01-06
Fix generated Neon simd_test features.

jacobbramley:
The tests can run if and only if the target_features for the corresponding intrinsics are detected at run-time, so make sure that the tests have an appropriate `simd_test()`.

This fixes some failures due to tests running when they shouldn't. For example, some tests would fail on hardware that lacks "fcma".

This decouples the feature names in the generator, since "arm" and "arm64" use different names, and for "arm" the dynamic `simd_test` features don't match the static `target_feature` names.

There are several changes to the tests, but changes to the generated intrinsics themselves are minimal, and cosmetic.

### DIFF
--- a/crates/core_arch/src/aarch64/neon/generated.rs
+++ b/crates/core_arch/src/aarch64/neon/generated.rs
@@ -22346,7 +22346,7 @@ mod test {
         assert_eq!(r, e);
     }
 
-    #[simd_test(enable = "neon")]
+    #[simd_test(enable = "neon,aes")]
     unsafe fn test_vmull_p64() {
         let a: p64 = 15;
         let b: p64 = 3;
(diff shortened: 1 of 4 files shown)


jacobbramley:
Hmm... `ci/style.sh` is (now) failing on a file I didn't modify. (It fails on `master` for me too.)

Nugine:
> Hmm... `ci/style.sh` is (now) failing on a file I didn't modify. (It fails on `master` for me too.)

I accidentally added an invisble space in [this line](https://github.com/rust-lang/stdarch/blob/f151316993a9c72cd60b771d3267748482057a92/crates/core_arch/src/x86_64/cmpxchg16b.rs#L79). CI passed in [the PR](https://github.com/rust-lang/stdarch/pull/1358) because rustfmt-preview was missing.

https://github.com/rust-lang/stdarch/blob/368ba1a204f3cc1bb245769fd1d655ab47fe92c9/ci/style.sh#L5-L9

Is the single space worth a PR? :joy:

Amanieu:
As @Nugine said, just remove the extra space in this PR.

jacobbramley:
Done. Sorry about the force-push; the whitespace fix wouldn't apply otherwise.

PR #1370, opened 2023-01-13
Declare and check licences in crates/intrinsics-test.

jacobbramley:
This doesn't change any licences, but simply declares them.

Notably, this checks the ACLE (submodule) licence, to make sure that we are alerted if it ever changes.


Amanieu:
LGTM! Can you rebase now that #1371 is merged?

PR #1371, opened 2023-01-16
Don't declare f16c_target_feature.

jacobbramley:
f16c_target_feature was recently stabilised.

This fixes a warning that results in a CI failure (due to `-D stable-features`) with current nightly Rust.


jacobbramley:
This fixes the CI failure in #1370. There's related work going on in #1366, but this seems different.

I'm not certain if this is the right process to fix this failure (e.g. as opposed to a fix included with some other PR), so please let me know if another approach would be better.

PR #1373, opened 2023-01-23
std_detect: Support detecting lse/dotprod/jsconv/rcpc on aarch64 Windows

taiki-e:
IsProcessorFeaturePresent currently supports the detection of more CPU features.

winnt

([IsProcessorFeaturePresent docs](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-isprocessorfeaturepresent) currently don't mention most of them but it appears there is [a PR to update it](https://github.com/MicrosoftDocs/sdk-api/pull/1409/files))


taiki-e:
Hmm, I ran the test on a windows VM on a mac m1 that supports all these features and it looks like the detection of FEAT_RCPC is not yet supported.
```
asimd: true
crc: true
lse: true
dotprod: true
jsconv: true
rcpc: false
aes: true
pmull: true
sha2: true
test aarch64_windows ... ok
```

PR #1374, opened 2023-01-25
std_detect: Support run-time detection on aarch64 OpenBSD

taiki-e:
OpenBSD doesn't trap the mrs instruction, but exposes the system registers through sysctl.
https://github.com/openbsd/src/commit/d335af936b9d7dd9cf655cae1ce19560c45de6c8

So, we can support run-time detection using the register parsing code in [os/aarch64.rs](https://github.com/rust-lang/stdarch/blob/f325623f26c7ed7eb81e2fbd8d71216bc7c54337/crates/std_detect/src/detect/os/aarch64.rs) (currently used on aarch64 FreeBSD) and sysctl. (See also go's patch that does the same thing as this patch: https://github.com/golang/go/commit/cd54ef1f61945459486e9eea2f016d99ef1da925. See also OpenBSD libcrypto patch: https://github.com/openbsd/src/commit/be689f1071529b27fdf00b1407302592c71d1c1a.)

Tested on aarch64 OpenBSD VM 7.2-current/7.2-stable/7.1-stable.
The following screenshot is from a test on 7.2-current:

openbsd7 2-current

On 7.2-stable/7.1-stable, the features referring to ID_AA64PFR0_EL1 returns false because the patch that supports ID_AA64PFR0_EL1 (https://github.com/openbsd/src/commit/c7654cd65262d532212f65123ee3905ba200365c) is not included:

openbsd7 1-stable

### DIFF
--- a/crates/std_detect/src/detect/mod.rs
+++ b/crates/std_detect/src/detect/mod.rs
@@ -56,6 +56,12 @@ cfg_if! {
         mod aarch64;
         #[path = "os/freebsd/mod.rs"]
         mod os;
+    } else if #[cfg(all(target_os = "openbsd", target_arch = "aarch64", feature = "libc"))] {
+        #[allow(dead_code)] // we don't use code that calls the mrs instruction.
+        #[path = "os/aarch64.rs"]
+        mod aarch64;
+        #[path = "os/openbsd/aarch64.rs"]
+        mod os;
     } else if #[cfg(all(target_os = "windows", target_arch = "aarch64"))] {
         #[path = "os/windows/aarch64.rs"]
         mod os;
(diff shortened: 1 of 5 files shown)


crates/std_detect/src/detect/mod.rs
```diff
@@ -56,6 +56,12 @@ cfg_if! {
         mod aarch64;
         #[path = "os/freebsd/mod.rs"]
         mod os;
+    } else if #[cfg(all(target_os = "openbsd", target_arch = "aarch64", feature = "libc"))] {
+        #[allow(dead_code)]
```
lu-zero, line 60: Why this?

lu-zero (approved):
It looks fine to me.

crates/std_detect/src/detect/mod.rs
taiki-e: That module has code that calls the mrs instruction, but it is not used on OpenBSD.

Added a comment mentioning it.

PR #1377, opened 2023-02-07
attempt to resolve #852

### LINKED ISSUE #852: x86_64 MXCSR denormals are zero bit: add constant
On x86_64, the SSE floating point control register (MXCSR) has two bits concerned with denormal floating point values.  The purpose of these bits is to enable a mode that avoids slowdowns from calculations with denormal numbers at the cost of getting incorrect results from underflows.

The more important one is the *flush to zero* bit (bit 15 in the register).  When that bit is set, when a floating-point arithmetic instruction would output a denormal number (and would not raise an unmasked exception), then it instead outputs zero (and the denormal exception that it would normally flag is suppressed).  The consts for this bit in the core::arch::x86_64 module are _MM_FLUSH_ZERO_ON, _MM_FLUSH_ZERO_OFF, _MM_FLUSH_ZERO_MASK. 

The less important bit is the *denormals are zero* bit (bit 6 in the register).  That bit affects the input arguments of floating-point arithmetic instructions, rather than the outputs.  When the bit is set, when a floating-point arithmetic instruction has a number in a source argument that is a denormal number, the instruction behaves as if that number was zero instead.  On x86_32, setting this bit is only conditionally supported, because old CPUs didn't have this mode.  Testing and clearing the bit is always supported if the MXCSR register exists.  

The crate does not have consts for the denormals are zero bit.  This is probably an oversight, and this ticket asks to correct them.  I suggest the following names, based on Intel's C interface, but I don't insist on them.

```rust
pub const _MM_DENORMALS_ZERO_MASK: u32 = 0x0040;
pub const _MM_DENORMALS_ZERO_ON: u32 = 0x0040;
pub const _MM_DENORMALS_ZERO_OFF: u32 = 0x0000;
```

The C interface also has convenience macros for getting and setting the bit, so you may add those too.  Personally I think they're superfluous, because functions that access this bit will most likely set or clear the bits together with the flush to zero bits, eg.

```rust
_mm_setcsr(_mm_getcsr() | _MM_FLUSH_ZERO_ON | _MM_DENORMALS_ZERO_ON);
// XMM floating-point arithmetic computations here
_mm_setcsr(_mm_getcsr() & !_MM_FLUSH_ZERO_MASK & !_MM_DENORMALS_ZERO_MASK);
```

matthiascy:
This PR introduces `_MM_{GET,SET}_DENORMALS_ZERO_MODE`, `MM_DENORMALS_ZERO_ON`, `MM_DENORMALS_ZERO_OFF`, `MM_DENORMALS_ZERO_MASK`. One remaining question is that in which file these should be placed? `sse.rs` or `sse3.rs`?

### DIFF
--- a/crates/core_arch/src/x86/sse.rs
+++ b/crates/core_arch/src/x86/sse.rs
@@ -1490,7 +1490,20 @@ pub unsafe fn _mm_getcsr() -> u32 {
 /// _MM_SET_ROUNDING_MODE(_MM_ROUND_DOWN)
 /// ```
 ///
-/// ## Denormals-are-zero/Flush-to-zero Mode
+/// ## Denormals-are-Zero Mode
+///
+/// If this bit is set, denormal values used as input to floating-point instructions
+/// wil be treated as zero.
+///
+/// You can read and enable/disable this mode via the helper functions
+/// `_MM_GET_DENORMALS_ZERO_MODE()` and `_MM_SET_DENORMALS_ZERO_MODE()`.
+///
+/// ```rust,ignore
+/// _MM_SET_DENORMALS_ZERO_MODE(_MM_DENORMALS_ZERO_OFF); // turn off (default)
+/// _MM_SET_DENORMALS_ZERO_MODE(_MM_DENORMALS_ZERO_ON); // turn on
+/// ```
+///
+/// ## Flush-to-zero Mode
 ///
 /// If this bit is set, values that would be denormalized will be set to zero
 /// instead. This is turned off by default.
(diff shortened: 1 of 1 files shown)


matthiascy:
`_MM_{SET, GET}_DENORMALS_ZERO_MODE` is not defined in [Intel Intrinsics Guid](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html) but can be found in [emmintrin.h](https://github.com/llvm/llvm-project/blob/97fbbd657f9afd68f7dd03ac5349fbd5ac56be02/clang/lib/Headers/emmintrin.h#L4766) of llvm, [pmmintrin.h](https://github.com/gcc-mirror/gcc/blob/c300e251f5b22d15b126f3c643cd55a119994e48/gcc/config/i386/pmmintrin.h#L41) of gcc, and [pmmintrin.h](https://www.intel.com/content/www/us/en/develop/documentation/cpp-compiler-developer-guide-and-reference/top/compiler-reference/floating-point-operations/set-the-ftz-and-daz-flags.html) of Intel C++ Compiler.

Amanieu:
LGTM! For the CI failure, you can add an exception in `crates/stdarch-verify/tests/x86-intel.rs` to ignore the check for this function.

This requires an FCP since it is adding a stable API.

@rfcbot fcp merge

BurntSushi:
My understanding is that we should try to land new APIs as unstable, and only after some time has passed, then we can mark them stable.

@rfcbot concern unstable for a time

Amanieu:
That's a good point.

@matthiascy Could you change your PR to make these functions unstable and open a tracking issue in rust-lang/rust?

Amanieu:
@rfcbot fcp cancel

PR #1378, opened 2023-02-10
std_detect: Workaround Exynos 9810 bug on aarch64 Android

taiki-e:
[Samsung Exynos 9810](https://en.wikichip.org/wiki/samsung/exynos/9810) has a bug that big and little cores have different ISAs. And on older Android (pre-9), the kernel incorrectly reports that features available only on some cores are available on all cores.

See https://reviews.llvm.org/D114523 for details. That patch is about FEAT_LSE, but IIUC the detection of most features that are available in armv8.2-a but may not be available in armv8.0-a should be affected by this issue.

### DIFF
--- a/crates/std_detect/src/detect/os/linux/aarch64.rs
+++ b/crates/std_detect/src/detect/os/linux/aarch64.rs
@@ -6,7 +6,7 @@ use crate::detect::{bit, cache, Feature};
 /// Try to read the features from the auxiliary vector, and if that fails, try
 /// to read them from /proc/cpuinfo.
 pub(crate) fn detect_features() -> cache::Initializer {
-    #[cfg(all(target_arch = "aarch64", target_os = "android"))]
+    #[cfg(target_os = "android")]
     let is_exynos9810 = {
         // Samsung Exynos 9810 has a bug that big and little cores have different
         // ISAs. And on older Android (pre-9), the kernel incorrectly reports
@@ -23,7 +23,7 @@ pub(crate) fn detect_features() -> cache::Initializer {
         // because Android 9+ includes the fix.
         len > 0 && arch.starts_with(b"exynos9810")
     };
-    #[cfg(not(all(target_arch = "aarch64", target_os = "android")))]
+    #[cfg(not(target_os = "android"))]
     let is_exynos9810 = false;
 
     if let Ok(auxv) = auxvec::auxv() {
--- a/crates/std_detect/src/detect/os/linux/aarch64.rs
+++ b/crates/std_detect/src/detect/os/linux/aarch64.rs
@@ -7,7 +7,7 @@ use crate::detect::{bit, cache, Feature};
 /// to read them from /proc/cpuinfo.
 pub(crate) fn detect_features() -> cache::Initializer {
     #[cfg(all(target_arch = "aarch64", target_os = "android"))]
-    {
+    let is_exynos9810 = {
         // Samsung Exynos 9810 has a bug that big and little cores have different
         // ISAs. And on older Android (pre-9), the kernel incorrectly reports
         // that features available only on some cores are available on all cores.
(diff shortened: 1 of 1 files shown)


lu-zero (approved):

taiki-e:
Hmm, I tested exynos (but 9825 since the problematic 9810 was not available) with Samsung remote test lab and it seems that ro.arch returns an empty string. So this may not actually work.

An alternative would be to query OS version (ro.build.version.release, which works in the above environment) to exclude problematic versions...

lu-zero:
If you can confirm, I'm fine with the approach.

taiki-e:
Ah, it appears that ro.arch is actually a field that was removed in Android 12. The above case used Android 12, but Android 10 and Android 11 return a valid value (with both the getprop command and the __system_property_get function).

ro arch android10
ro arch android11

It is fine that ro.arch is not available for Android 12+ because Android 9+ includes the fix.
So, I believe that this PR is fine as is.

lu-zero:
Maybe add to the comment that bit of additional information.

crates/std_detect/src/detect/os/linux/aarch64.rs
```diff
@@ -6,6 +6,23 @@ use crate::detect::{bit, cache, Feature};
...
+        // that features available only on some cores are available on all cores.
+        // https://reviews.llvm.org/D114523
+        let mut arch = [0_u8; libc::PROP_VALUE_MAX as usize];
+        let len = unsafe {
+            libc::__system_property_get(
+                b"ro.arch\0".as_ptr() as *const libc::c_char,
+                arch.as_mut_ptr() as *mut libc::c_char,
+            )
+        };
+        if len > 0 && arch.starts_with(b"exynos9810\0") {
```
Amanieu, line 22: Is this checking that the returned string starts with `exynos9810` or that it is exactly equal to `exynos9810`? In the former case you wouldn't want the `\0` in the string.

crates/std_detect/src/detect/os/linux/aarch64.rs
```diff
@@ -6,6 +6,23 @@ use crate::detect::{bit, cache, Feature};
...
+            return cache::Initializer::default();
```
Amanieu, line 23: We still want to probe for v8.0 features such as aes/sha/crc32.

crates/std_detect/src/detect/os/linux/aarch64.rs
taiki-e: I originally meant the latter, but looking at the LLVM patch they seem to be using the former.

So, I removed `\0` to match the LLVM patch's behavior.

taiki-e:
> Maybe add to the comment that bit of additional information.

Added comment about ro.arch on Android 12+ on Exynos.

crates/std_detect/src/detect/os/linux/aarch64.rs
taiki-e: IIUC this is a bit tricky because even the armv8.0 optional features may not be consistent between little and big.

So for now, all I can think of is one of the following

- Only check features that are known to be available on exynos-m3 (armv8.0 cores).
  (According to https://github.com/golang/go/issues/28431) The old kernel seems to return auxv/cpuinfo based on features available on cortex-a55 (armv8.2 cores), so it can properly handle features even if they are available on exynos-m3 but not on cortex-a55 (though such features probably do not exist).
- Continue detection on Android 9+ and return default value on pre-9.

The first one seems to make more sense, as it does not require additional __system_property_get to get the version and works on older kernels.

taiki-e: Implemented the first approach in https://github.com/rust-lang/stdarch/pull/1378/commits/9ce03f137315630630bca6acb29145d5048fb6d2.

PR #1386, opened 2023-02-28
std_detect: Respect feature dependencies on arm Linux

taiki-e:
According to LLVM's feature definitions, ARM's cryptographic extensions require NEON.
https://github.com/llvm/llvm-project/blob/main/llvm/lib/Target/ARM/ARM.td

> ```
> // True if processor supports SHA1 and SHA256.
> def FeatureSHA2           : SubtargetFeature<"sha2", "HasSHA2", "true",
>                                              "Enable SHA1 and SHA256 support", [FeatureNEON]>;
> 
> def FeatureAES            : SubtargetFeature<"aes", "HasAES", "true",
>                                              "Enable AES support", [FeatureNEON]>;
> 
> // True if processor supports Cryptography extensions.
> def FeatureCrypto         : SubtargetFeature<"crypto", "HasCrypto", "true",
>                                              "Enable support for "
>                                              "Cryptography extensions",
>                                              [FeatureNEON, FeatureSHA2, FeatureAES]>;
> ```

 (as well as ARM, AArch64's cryptographic extensions also require ASIMD (NEON); we already handle it on [linux](https://github.com/rust-lang/stdarch/blob/15c760e75c58e49f15d32e76022b04d597c308e8/crates/std_detect/src/detect/os/linux/aarch64.rs#L306) and [bsd](https://github.com/rust-lang/stdarch/blob/master/crates/std_detect/src/detect/os/aarch64.rs#L105). As for windows, ~~I'll open a PR later~~ EDIT: opened https://github.com/rust-lang/stdarch/pull/1388)

### DIFF
--- a/crates/std_detect/src/detect/os/linux/arm.rs
+++ b/crates/std_detect/src/detect/os/linux/arm.rs
@@ -16,50 +16,56 @@ pub(crate) fn detect_features() -> cache::Initializer {
     // The values are part of the platform-specific [asm/hwcap.h][hwcap]
     //
     // [hwcap]: https://github.com/torvalds/linux/blob/master/arch/arm/include/uapi/asm/hwcap.h
+    //
+    // The feature dependencies here come directly from LLVM's feature definitions:
+    // https://github.com/llvm/llvm-project/blob/main/llvm/lib/Target/ARM/ARM.td
     if let Ok(auxv) = auxvec::auxv() {
-        enable_feature(&mut value, Feature::neon, bit::test(auxv.hwcap, 12));
+        let neon = bit::test(auxv.hwcap, 12);
+        enable_feature(&mut value, Feature::neon, neon);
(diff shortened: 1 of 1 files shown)


Amanieu:
I don't think this is needed. If a CPU supports the crypto extensions then it will definitely support ASIMD.

taiki-e:
I think you are correct. And in that case, should we remove the similar code (e.g., [`&& asimd` in this line](https://github.com/rust-lang/stdarch/blob/15c760e75c58e49f15d32e76022b04d597c308e8/crates/std_detect/src/detect/os/linux/aarch64.rs#L307)) in aarch64?

Amanieu:
Yes. Those instructions use SIMD registers so it's impossible to have them without ASIMD.

taiki-e:
Thanks. I'll open a PR for removing them.

And, closing this PR as it is unneeded.

PR #1387, opened 2023-02-28
std_detect: Support run-time detection of crc/aes/sha2/crypto on arm FreeBSD

taiki-e:


crates/std_detect/src/detect/os/freebsd/arm.rs
```diff
@@ -12,9 +12,26 @@ pub(crate) fn detect_features() -> cache::Initializer {
         }
     };
 
+    // The values are defined as HWCAP32_* in machine/elf.h.
+    // https://github.com/freebsd/freebsd-src/blob/deb63adf945d446ed91a9d84124c71f15ae571d1/sys/arm64/include/elf.h
     if let Ok(auxv) = auxvec::auxv() {
-        enable_feature(&mut value, Feature::neon, auxv.hwcap & 0x00001000 != 0);
-        enable_feature(&mut value, Feature::pmull, auxv.hwcap2 & 0x00000002 != 0);
+        let neon = auxv.hwcap & 0x00001000 != 0;
```
Amanieu, line 18: It would help readability to define these as `const`s.

taiki-e: That makes sense. I changed all magic numbers in that file to constants.

Also, I removed `neon &&` conditions for cryptographic extensions based on your review in https://github.com/rust-lang/stdarch/pull/1386#issuecomment-1450594643.

PR #1388, opened 2023-02-28
std_detect: Support run-time detection of fp on aarch64 Windows

taiki-e:
According to google/cpu_features, IsProcessorFeaturePresent(PF_ARM_VFP_32_REGISTERS_AVAILABLE) returns whether fp is available.
https://github.com/google/cpu_features/blob/a6bf4f9031ec601f905660b8cef82d7478b33d64/src/impl_aarch64_windows.c#L112-L113

~~Also, update to respect feature dependencies based on LLVM's feature definitions:
https://github.com/llvm/llvm-project/blob/main/llvm/lib/Target/AArch64/AArch64.td~~

~~I don't know if there is support for [cases where fp or asimd is unavailable](https://developer.arm.com/documentation/den0024/a/AArch64-Floating-point-and-NEON) on Windows, but I think it is reasonable to unify the behavior with [linux](https://github.com/rust-lang/stdarch/blob/15c760e75c58e49f15d32e76022b04d597c308e8/crates/std_detect/src/detect/os/linux/aarch64.rs#L306) and [bsd](https://github.com/rust-lang/stdarch/blob/master/crates/std_detect/src/detect/os/aarch64.rs#L105) in any case.~~

(Tested on aarch64 Windows VM on mac m1)

### DIFF
--- a/crates/std_detect/src/detect/os/windows/aarch64.rs
+++ b/crates/std_detect/src/detect/os/windows/aarch64.rs
@@ -11,6 +11,7 @@ pub(crate) fn detect_features() -> cache::Initializer {
     // The following Microsoft documents isn't updated for aarch64.
     // https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-isprocessorfeaturepresent
     // These are defined in winnt.h of Windows SDK
+    const PF_ARM_VFP_32_REGISTERS_AVAILABLE: u32 = 18;
     const PF_ARM_NEON_INSTRUCTIONS_AVAILABLE: u32 = 19;
     const PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE: u32 = 30;
     const PF_ARM_V8_CRC32_INSTRUCTIONS_AVAILABLE: u32 = 31;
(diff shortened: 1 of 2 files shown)


crates/std_detect/src/detect/os/windows/aarch64.rs
```diff
@@ -51,28 +57,23 @@ pub(crate) fn detect_features() -> cache::Initializer {
                 Feature::dotprod,
                 IsProcessorFeaturePresent(PF_ARM_V82_DP_INSTRUCTIONS_AVAILABLE) != FALSE,
             );
+            // jsconv requires float support
             enable_feature(
                 Feature::jsconv,
-                IsProcessorFeaturePresent(PF_ARM_V83_JSCVT_INSTRUCTIONS_AVAILABLE) != FALSE,
+                fp && IsProcessorFeaturePresent(PF_ARM_V83_JSCVT_INSTRUCTIONS_AVAILABLE) != FALSE,
```
Amanieu, line 63: I don't think we need to be pedantic about things like this. It's impossible for an AArch64 cpu to support jsconv without fp.

crates/std_detect/src/detect/os/windows/aarch64.rs
```diff
@@ -51,28 +57,23 @@ pub(crate) fn detect_features() -> cache::Initializer {
...
-            enable_feature(
-                Feature::pmull,
-                IsProcessorFeaturePresent(PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE) != FALSE,
-            );
-            enable_feature(
-                Feature::sha2,
-                IsProcessorFeaturePresent(PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE) != FALSE,
-            );
+            // Cryptographic extensions require ASIMD
+            let crypto = asimd
```
Amanieu, line 72: Same here with `asimd &&`

taiki-e:
Removed all `fp &&` and `asimd &&`.

PR #1393, opened 2023-03-09
wasm32: Add relaxed simd instructions

alexcrichton:
This commit adds intrinsics to the `wasm32` to support the [relaxed SIMD proposal][proposal]. These are added with the same naming conventions of existing simd-related intrinsics for wasm which is similar to the instruction name but matches sign in a few places.

This additionally updates Wasmtime to execute tests with support for the relaxed simd proposal. No release has been made yet so this uses the `dev` release, and I can make a PR in April when the support in Wasmtime has been released to an official release. The `wasmprinter` crate is also updated to understand these instruction opcodes as well.

Documentation has been added for all intrinsics, but tests have only been added for some of them so far. I hope to follow-up later with more tests.

[proposal]: https://github.com/WebAssembly/relaxed-simd

### DIFF
--- a/crates/core_arch/src/wasm32/relaxed_simd.rs
+++ b/crates/core_arch/src/wasm32/relaxed_simd.rs
@@ -1,0 +1,449 @@
+use super::v128;
+use crate::core_arch::simd;
+
+#[cfg(test)]
+use stdarch_test::assert_instr;
+
+#[allow(improper_ctypes)]
+extern "C" {
+    #[link_name = "llvm.wasm.relaxed.swizzle"]
+    fn llvm_relaxed_swizzle(a: simd::i8x16, b: simd::i8x16) -> simd::i8x16;
+    #[link_name = "llvm.wasm.relaxed.trunc.signed"]
+    fn llvm_relaxed_trunc_signed(a: simd::f32x4) -> simd::i32x4;
+    #[link_name = "llvm.wasm.relaxed.trunc.unsigned"]
+    fn llvm_relaxed_trunc_unsigned(a: simd::f32x4) -> simd::i32x4;
+    #[link_name = "llvm.wasm.relaxed.trunc.signed.zero"]
+    fn llvm_relaxed_trunc_signed_zero(a: simd::f64x2) -> simd::i32x4;
+    #[link_name = "llvm.wasm.relaxed.trunc.unsigned.zero"]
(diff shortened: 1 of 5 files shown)


crates/core_arch/src/wasm32/relaxed_simd.rs
```diff
@@ -0,0 +1,449 @@
...
+pub fn f32x4_relaxed_nmadd(a: v128, b: v128, c: v128) -> v128 {
+    unsafe { llvm_f32x4_fms(a.as_f32x4(), b.as_f32x4(), c.as_f32x4()).v128() }
+}
+
+/// Computes `a * b + c` with either one rounding or two roundings.
+#[inline]
+#[cfg_attr(test, assert_instr(f64x2.relaxed_madd))]
+#[target_feature(enable = "relaxed-simd")]
+#[doc(alias("f64x2.relaxed_madd"))]
+pub fn f64x2_relaxed_madd(a: v128, b: v128, c: v128) -> v128 {
```
CryZe, line 150: `std` calls this [mul_add](https://doc.rust-lang.org/std/primitive.f32.html#method.mul_add), would it make sense to adopt the same naming convention here? (If not, maybe as an alias?)

alexcrichton: We already deviate from the name of the instructions themselves so I think it's reasonable to have a different name here, especially because the doc aliases cover folks searching for the instruction itself. 

Looking for prior inspiration, it appears that `clang` doesn't actually have support for relaxed simd instructions in the public `wasm_simd128.h` header. There is a `__builtin_wasm_relaxed_madd_f32x4` intrinsic but I believe that's intended to be clang-specific and not used as a public interface. In that sense no inspiration to draw from.

I'd be a bit worried about the mouthful of `f64x2_relaxed_neg_mul_add` but I don't mind changing these to that. Also happy to defer to a later date and track this as part of a tracking issue.

alexcrichton:
ping @Amanieu, would you have a chance to look at this? I'd prefer to defer the naming question to a tracking issue for stabilization myself, but I can of course rename here-and-now too.

Amanieu:
LGTM. I'm a bit uncomfortable depending on a dev version of wasmtime in CI, but it's fine for now as long as this is updated when a new version of wasmtime is released.

PR #1394, opened 2023-03-09
Allow `internal_features`

Nilstrieb:
I'm writing the lint, so it needs to be allowed here.
Needs to wait on rust-lang/compiler-team#596

### DIFF
--- a/crates/std_detect/src/lib.rs
+++ b/crates/std_detect/src/lib.rs
@@ -19,6 +19,9 @@
 #![deny(clippy::missing_inline_in_public_items)]
 #![cfg_attr(test, allow(unused_imports))]
 #![no_std]
+// FIXME(Nilstrieb): Remove this once the compiler in stdarch CI has the internal_features lint.
+#![allow(unknown_lints)]
+#![allow(internal_features)]
 
 #[cfg(test)]
 #[macro_use]


Nilstrieb:
Waiting for the rustc PR to be merged and being updated here, you might need to add the `allow()` when updating it. I'll keep this open for now to remind myself to check back later and remove the hacks.

Nilstrieb:
I think I actually need to do it this way through here. Adding an `-Ainternal_features` in bootstrap sounded nice but doesn't work with other tools that build a sysroot and fixing them is way more effort than just adding the allow here. We do sadly have to `allow(unknown_lints)`, but I'll open an issue and assign myself to remove it again as soon as possible.

@Amanieu this is ready now

CI seems to be having a weird day.....

Amanieu:
This is due to https://github.com/rust-lang/rust/pull/107224 which upgraded rustc to LLVM 16.

Nilstrieb:
So we need to wait until the revert lands?

Amanieu:
It's fine to force a merge for now, since this is a fairly straightforward change. This often happens when changes need to be done to both rustc and stdarch at the same time, but it's something that we're working towards a solution for.

PR #1399, opened 2023-03-23
Stabilize AArch64 AES/SHA1/SHA2 intrinsics

Amanieu:
These have been tested against the equivalent Clang intrinsics with random inputs.


Amanieu:
@rfcbot fcp merge

PR #1400, opened 2023-03-24
Unbreak powerpc64 tests

lu-zero:
At least locally this is needed to let the test pass.


Amanieu:
CI is failing due to style checks.

lu-zero:
I forgot to configure the editor to format on save on the ppc64 machine, sorry ...

PR #1401, opened 2023-03-24
std_detect: Unbreak auxv_from_file

lu-zero:
Discovered on powerpc64 where it crashes since it has more than 32 entries.

Thanks to @dodomorandi, @nappa85 and the rest of @rust-italia for assisting at the debugging session.

Ideally we could avoid the second allocation by using array_chunks() and from_ne_bytes but it isn't an hot path.


Amanieu:
How are you even hitting this code path? On modern Linux it should always be using the `getauxval` code path.

lu-zero:
By running the tests.

PR #1402, opened 2023-03-25
Fix the powerpc feature guards

lu-zero:
Now the documentation should be generated correctly and vsx isn't always present.


Amanieu:
This seems to have broken portable-simd in #110285. Instead of putting a `cfg` on the entire module, these functions should instead always be available but unsafe, like the other arches: the user is responsible for performing feature detection before using the corresponding intrinsics, and the types should always be available.

lu-zero:
I can prepare a PR for it.

PR #1403, opened 2023-03-26
Altivec predicates

lu-zero:
- [x] vec_all_eq
- [x] vec_all_ge
- [x] vec_all_gt
- [x] vec_all_in
- [x] vec_all_le
- [x] vec_all_lt
- [x] vec_all_nan
- [x] vec_all_ne
- [x] vec_all_nge
- [x] vec_all_ngt
- [x] vec_all_nle
- [x] vec_all_nlt
- [x] vec_all_numeric
- [x] vec_any_eq
- [x] vec_any_ge
- [x] vec_any_gt
- [x] vec_any_in
- [x] vec_any_le
- [x] vec_any_lt
- [x] vec_any_nan
- [x] vec_any_ne
- [x] vec_any_nge
- [x] vec_any_ngt
- [x] vec_any_nle
- [x] vec_any_nlt
- [x] vec_any_numeric


Amanieu:
LGTM!

2 points of concern:
- There are some remaining `TODO: vsx encoding` comments. Are you planning on addressing these before this PR is merged?
- I noticed that a lot of functions are missing `#[inline]`. Since these aren't generic functions, they are likely to never be inlined, thus hurting performance.

lu-zero:
> LGTM!
> 
> 2 points of concern:
> 
>     * There are some remaining `TODO: vsx encoding` comments. Are you planning on addressing these before this PR is merged?

It will be a new PR, I want to complete first the normal altivec and then move to vsx, power8 and power9 ISAs later.

>     * I noticed that a lot of functions are missing `#[inline]`. Since these aren't generic functions, they are likely to never be inlined, thus hurting performance.

This is a mistake, I'll address it soon :) Thank you for spotting it!

lu-zero:
I'd land it now if it is fine for you.

crates/core_arch/src/powerpc/mod.rs
```diff
@@ -1,4 +1,8 @@
 //! PowerPC intrinsics
+
+#[macro_use]
+pub(crate) mod macros;
```
Amanieu, line 4: This doesn't need to be `pub(crate)`, it's only used in the `powerpc` module.

lu-zero: Sure :)

PR #1411, opened 2023-04-12
x86: remove unnecessary parens

KisaragiEffective:
this PR is needed for https://github.com/rust-lang/rust/pull/110189, see https://github.com/rust-lang/rust/pull/110189#issuecomment-1504298677


lu-zero (approved):
Seems fine, even if that chain of `as` isn't pretty no matter what :)

PR #1416, opened 2023-04-20
Use Wasmtime 8.0.0 on CI

alexcrichton:
This moves from the "dev" release of Wasmtime, used for its relaxed-simd support, to an official release of Wasmtime just made which is the first with relaxed-simd support.


alexcrichton:
Following through on https://github.com/rust-lang/stdarch/pull/1393 to ensure that "dev" isn't used indefinitely here by accident.

PR #1421, opened 2023-04-24
std_detect: Do not use libc::getauxval on 32-bit Android

taiki-e:
This fixes the build error in https://github.com/rust-lang/rust/pull/110285#issuecomment-1519837713.

In https://github.com/rust-lang/stdarch/pull/1406 I checked only aarch64 android and missed the fact that libc crate does not provide getauxval on 32-bit android (although I believe [android provides getauxval on these targets](https://github.com/aosp-mirror/platform_bionic/commit/2c5153b043b44e9935a334ae9b2d5a4bc5258b40)). Sorry.

The second commit in this PR adds a job to CI that builds non-Linux targets that are not currently tested in CI to prevent such an oversight from occurring again in the future.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -211,3 +211,13 @@ jobs:
       if: "matrix.os == 'ubuntu-latest' && !startsWith(matrix.target, 'thumb')"
       env:
         TARGET: ${{ matrix.target }}
+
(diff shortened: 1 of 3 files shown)


lu-zero (approved):
Looks fine to me

PR #1422, opened 2023-04-25
move to array simd

Ezrashaw:
As part of [compiler-team/MCP#621](https://github.com/rust-lang/compiler-team/issues/621), this repo must be updated to use the new-style array fields instead.

Note that a very recent nightly is needed which fixed a bug where inline assembly could not be used with the array style.

Note also that this cannot be pulled to `rust-lang/rust` until beta updates with the bugfix.

cc @scottmcm

### DIFF
--- a/crates/core_arch/src/simd.rs
+++ b/crates/core_arch/src/simd.rs
@@ -3,25 +3,22 @@
 #![allow(non_camel_case_types)]
 
 macro_rules! simd_ty {
-    ($id:ident [$ety:ident]: $($elem_ty:ident),* | $($elem_name:ident),*) => {
+    ($id:ident: [$ety:ident; $ecount:literal]) => {
         #[repr(simd)]
         #[derive(Copy, Clone, Debug, PartialEq)]
-        pub(crate) struct $id($(pub $elem_ty),*);
+        pub(crate) struct $id(pub [$ety; $ecount]);
 
         #[allow(clippy::use_self)]
         impl $id {
             #[inline(always)]
-            pub(crate) const fn new($($elem_name: $elem_ty),*) -> Self {
-                $id($($elem_name),*)
+            pub(crate) const fn new(val: [$ety; $ecount]) -> Self {
+                $id(val)
             }
+
             // FIXME: Workaround rust@60637
             #[inline(always)]
             pub(crate) const fn splat(value: $ety) -> Self {
-                $id($({
-                    #[allow(non_camel_case_types, dead_code)]
-                    struct $elem_name;
-                    value
-                }),*)
+                $id([value; $ecount])
             }
 
             // FIXME: Workaround rust@60637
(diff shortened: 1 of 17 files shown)


Ezrashaw:
Gah, different architectures cause pain here.

crates/core_arch/src/simd.rs
```diff
@@ -3,25 +3,22 @@
...
-        pub(crate) struct $id($(pub $elem_ty),*);
+        pub(crate) struct $id(pub [$ety; $ecount]);
 
         #[allow(clippy::use_self)]
         impl $id {
             #[inline(always)]
-            pub(crate) const fn new($($elem_name: $elem_ty),*) -> Self {
-                $id($($elem_name),*)
+            pub(crate) const fn new(val: [$ety; $ecount]) -> Self {
+                $id(val)
```
Sp00ph, line 15: From what I've heard, the plan is to make any direct field accesses (including calling the constructor like this) illegal on `#[repr(simd)]` types and only allow creation of the type and access to the inner array through `transmute` or pointer operations, so I think it might make sense to make the struct field private and to do something similar to rust-lang/portable-simd#339 or rust-lang/portable-simd#342 in `new` and `splat`.

scottmcm: I think it'd also be fine for that to be a separate PR, if you prefer.

crates/core_arch/src/simd.rs
```diff
@@ -3,25 +3,22 @@
...
+    ($id:ident: [$ety:ident; $ecount:literal]) => {
         #[repr(simd)]
         #[derive(Copy, Clone, Debug, PartialEq)]
-        pub(crate) struct $id($(pub $elem_ty),*);
+        pub(crate) struct $id(pub [$ety; $ecount]);
 
         #[allow(clippy::use_self)]
         impl $id {
             #[inline(always)]
-            pub(crate) const fn new($($elem_name: $elem_ty),*) -> Self {
```
lu-zero, line 14: That is less ergonomic, shall we consider adding `From` implementation to support both tuple and array and keep new as it is now?

scottmcm: Note that this was always `pub(crate)`, so users couldn't call it anyway.

They can only be created using intrinsics (or memory stuff like transmutes) in stable today: <https://doc.rust-lang.org/core/arch/x86/struct.__m128i.html>.

(The portable-simd types *do* have nice safe public constructors.  But these arch ones intentionally don't.)

crates/core_arch/src/simd.rs
Ezrashaw: > direct field accesses (including calling the constructor like this) illegal

Emphasis on *plan*. This isn't happening with the current MCP, we are just removing the field-style `#[repr(simd)]`. In the longer term, I think scottmcm and other people have plans for that but not immediately.

Ezrashaw:
> @Ezrashaw With [rust-lang/rust#110672](https://github.com/rust-lang/rust/pull/110672) merged a while ago, are we good to go on this again?

Yeah, the technical side of this pull request should be good to go. Personally, I haven't been contributing to Rust recently, so if someone wants to take this up, then free totally free. However, I'll try to find some time to complete this PR in maybe a week or so.

sayantn:
This transition has already been done in a different PR, so closing this

PR #1430, opened 2023-05-15
Fix intrinsic-test rejecting the new `instructions` JSON field

adamgemmell:
Fixes CI for https://github.com/rust-lang/stdarch/pull/1427  which didn't run due to a Github outage. The second commit in the chain broke the intrinsic-test tool.


adamgemmell:
Could this be because 18.04 is now EOL?

adamgemmell:
Seems to run on my machine, though the first `apt-get update` took 500s so maybe it was timing out on GHA. That stage runs in about a minute now so maybe the mirrors were slow and another retry will sort it

PR #1431, opened 2023-05-25
Remove obscure & rarely used ARM intrinsics

Amanieu:
In almost all cases it is preferable to use the stable `asm!` instead of calling these intrinsics.

This PR removes the following unstable intrinsics:
- `__breakpoint` Clang extension, not part of ACLE.
- `brk`: undocumented
- `_rev*`, `_clz*`, `_rbit*`: use methods on integer types instead
- `__ldrex`, `__strex`: deprecated in ACLE, hard to use correctly
- Register access API: API doesn't match ACLE, better to just use asm

Also considered for deletion, but not included in this PR:
- Barriers: `__isb`, `__dsb`, `__dmb`
- Hints: `__wfi`, `__wfe`, `__sev`, `__sevl`, `__yield`, `__nop`


Amanieu:
@rfcbot fcp close

cc @JamieCunliffe @adamgemmell if you have any thoughts on these.

PR #1436, opened 2023-05-31
Stabilize `__m512i`, `__m512`, and `__m512d`

scottmcm:
While there might be some uncertainty still with all the intrinsics and other related items, could we perhaps stabilize the types?

It seems like there's not much choice to be had, since m512 is definitely f32x16 and m512d is likewise f64x8.

I didn't include `__m512bh` since `__m128bh` is also unstable.

<https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs/topic/Stabilizing.20AVX2.20.26.20AVX512.20*types*/near/362541037>


scottmcm:
FYI @dtolnay, it looks from the edit history like you attempted to check a box, but somehow rfcbot then removed your checkmark?

PR #1439, opened 2023-06-19
arm(neon): avoid snippets which triggers unused_paren lint

KisaragiEffective:
In this PR, removed parens with "simple" expression followed by cast.

The `unused_paren` lint will be [triggered](https://github.com/rust-lang/rust/pull/110189
)  at there in the future.


KisaragiEffective:
Well, I don't think I've touched around `vld2q_dup_f32` in this PR...

Amanieu:
This is caused by https://github.com/rust-lang/rust/issues/112460, but it should be fine to merge this for now.

PR #1442, opened 2023-06-23
Use const {} for the immediate args of _mm*_cmp_*

bjorn3:
Rustc_codegen_cranelift requires immediates to already be constant in MIR unlike LLVM which only needs them constant after optimizations.


bjorn3:
Required for some of the intrinsics in https://github.com/bjorn3/rustc_codegen_cranelift/pull/1380

lu-zero:
Wouldn't be better to make a shorthand macro `const_as!(var, ty)` or such?

lu-zero (approved):
Either way it looks fine to me.

bjorn3:
That would be nice if this pattern happened more. Right now I'm not sure if it is worth it.

lu-zero:
On powerpc there are few more places, but that is probably __far__ in the future :)

bjorn3:
Yeah, Cranelift doesn't support PowerPC yet: https://github.com/bytecodealliance/wasmtime/issues/1183. It does support s390x though.

lu-zero:
I might contribute support for those intrinsics as well, but again, it can wait for later :)

PR #1443, opened 2023-06-25
More altivec intrinsics

lu-zero:
`vec_xl` contains an ugly workaround since `copy_nonoverlapping` is not correctly inlined.

### DIFF
--- a/crates/core_arch/src/lib.rs
+++ b/crates/core_arch/src/lib.rs
@@ -12,6 +12,7 @@
     proc_macro_hygiene,
     stmt_expr_attributes,
     core_intrinsics,
+    intrinsics,
     no_core,
     rustc_attrs,
     stdsimd,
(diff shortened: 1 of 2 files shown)


Amanieu:
The inlining issue happens because the PowerPC LLVM backend doesn't implement `TargetTransformInfo::areInlineCompatible`, so the default implementation is used which only allows inlining if the target features for the caller and callee match exactly. The target-specific specialized implementation typically allows inlining if the callee has a subset of the target feature of the caller.

Note that `inline(always)` skips this check, so this could be one way to make `copy_nonoverlapping` work in this case (you'd have to modify the standard library though).

lu-zero:
We could have `inline(always)` for `std::intrinsics` and document the pitfall in case other architectures have the same problem.
It is probably a good idea to discuss with LLVM upstream about adding a complete `areInlineCompatible` for this arch as well...

nikic:
@Amanieu I believe https://reviews.llvm.org/D150396 will not allow inlining with incompatible target features even if alwaysinline is specified.

Amanieu:
Considering the upcoming LLVM changes, I think it's fine to merge with the current workaround. Ideally we'll want a proper implementation of `areInlineCompatible` in the future.

PR #1448, opened 2023-07-29
Remove MIPS from CI

Amanieu:
- MIPS targets have been removed from rustup, see https://github.com/rust-lang/compiler-team/issues/648.
- Removed `#[target_feature]` requirements from `const fn`, since that caused const-eval to fail.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -74,13 +74,15 @@ jobs:
         - aarch64-unknown-linux-gnu
         - riscv64gc-unknown-linux-gnu
         - powerpc64le-unknown-linux-gnu
-        - mips-unknown-linux-gnu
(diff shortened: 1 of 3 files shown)


Amanieu:
@alexcrichton It seems that there has been some breakage to WASM in stdarch. Specifically, the problem is that functions like `i8x16` are defined as `const fn` but also have a `#[target_feature(enable = "simd128")]`, which now causes them to fail in const-eval. I've removed the target feature attribute from all `const fn`, since those should never require the target feature anyways.

alexcrichton:
Apologies for the delayed response but that sounds good to me! Given that these are all constant-creating things they don't need the feature anyway to work with LLVM's codegen so all seems reasonable to me. Thanks for taking care of it!

PR #1457, opened 2023-08-09
Document movnt needs sfence

workingjubilee:
For every intrinsic that may generate any of the MOVNT family of instructions, specify it must be followed by `_mm_sfence`.

Also, ask people to not think too hard about what actually happens with write-combining memory buffers. They probably don't want to know, and in terms of the Rust abstract machine, we aren't actually entirely sure yet.

### DIFF
--- a/crates/core_arch/src/x86/avx.rs
+++ b/crates/core_arch/src/x86/avx.rs
@@ -1683,6 +1683,17 @@ pub unsafe fn _mm256_lddqu_si256(mem_addr: *const __m256i) -> __m256i {
 /// aligned memory location. To minimize caching, the data is flagged as
 /// non-temporal (unlikely to be used again soon)
 ///
+/// # Safety
+///
+/// After using this intrinsic, but before any atomic operations occur, a call
+/// to `_mm_sfence()` must be performed. A safe function that includes unsafe
+/// usage of this intrinsic must always end in `_mm_sfence()`.
+///
+/// Reading and writing to the memory stored-to by any other means, after any
+/// nontemporal store has been used to write to that memory, but before the
+/// use of `_mm_sfence()`, is discouraged. Such reads can lead to pipeline
+/// stalls and yet-unspecified program behavior.
+///
 /// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm256_stream_si256)
 #[inline]
 #[target_feature(enable = "avx")]
@@ -1695,6 +1706,17 @@ pub unsafe fn _mm256_stream_si256(mem_addr: *mut __m256i, a: __m256i) {
 /// Moves double-precision values from a 256-bit vector of `[4 x double]`
 /// to a 32-byte aligned memory location. To minimize caching, the data is
 /// flagged as non-temporal (unlikely to be used again soon).
+///
+/// # Safety
+///
+/// After using this intrinsic, but before any atomic operations occur, a call
+/// to `_mm_sfence()` must be performed. A safe function that includes unsafe
+/// usage of this intrinsic must always end in `_mm_sfence()`.
+///
+/// Reading and writing to the memory stored-to by any other means, after any
+/// nontemporal store has been used to write to that memory, but before the
+/// use of `_mm_sfence()`, is discouraged. Such reads can lead to pipeline
+/// stalls and yet-unspecified program behavior.
 ///
 /// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm256_stream_pd)
 #[inline]
@@ -1710,6 +1732,17 @@ pub unsafe fn _mm256_stream_pd(mem_addr: *mut f64, a: __m256d) {
 /// of `[8 x float]` to a 32-byte aligned memory location. To minimize
 /// caching, the data is flagged as non-temporal (unlikely to be used again
 /// soon).
+///
+/// # Safety
+///
+/// After using this intrinsic, but before any atomic operations occur, a call
+/// to `_mm_sfence()` must be performed. A safe function that includes unsafe
+/// usage of this intrinsic must always end in `_mm_sfence()`.
+///
+/// Reading and writing to the memory stored-to by any other means, after any
+/// nontemporal store has been used to write to that memory, but before the
+/// use of `_mm_sfence()`, is discouraged. Such reads can lead to pipeline
+/// stalls and yet-unspecified program behavior.
 ///
 /// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm256_stream_ps)
 #[inline]
--- a/crates/core_arch/src/x86/avx512f.rs
+++ b/crates/core_arch/src/x86/avx512f.rs
@@ -26144,6 +26144,17 @@ pub unsafe fn _mm_mask_testn_epi64_mask(k: __mmask8, a: __m128i, b: __m128i) -> 
 
 /// Store 512-bits (composed of 16 packed single-precision (32-bit) floating-point elements) from a into memory using a non-temporal memory hint. mem_addr must be aligned on a 64-byte boundary or a general-protection exception may be generated.
 ///
+/// # Safety
+///
+/// After using this intrinsic, but before any atomic operations occur, a call
+/// to `_mm_sfence()` must be performed. A safe function that includes unsafe
+/// usage of this intrinsic must always end in `_mm_sfence()`.
+///
+/// Reading and writing to the memory stored-to by any other means, after any
+/// nontemporal store has been used to write to that memory, but before the
+/// use of `_mm_sfence()`, is discouraged. Such reads can lead to pipeline
+/// stalls and yet-unspecified program behavior.
+///
 /// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm512_stream_ps&expand=5671)
 #[inline]
 #[target_feature(enable = "avx512f")]
@@ -26155,6 +26166,17 @@ pub unsafe fn _mm512_stream_ps(mem_addr: *mut f32, a: __m512) {
 
 /// Store 512-bits (composed of 8 packed double-precision (64-bit) floating-point elements) from a into memory using a non-temporal memory hint. mem_addr must be aligned on a 64-byte boundary or a general-protection exception may be generated.
 ///
+/// # Safety
+///
+/// After using this intrinsic, but before any atomic operations occur, a call
+/// to `_mm_sfence()` must be performed. A safe function that includes unsafe
+/// usage of this intrinsic must always end in `_mm_sfence()`.
+///
+/// Reading and writing to the memory stored-to by any other means, after any
+/// nontemporal store has been used to write to that memory, but before the
+/// use of `_mm_sfence()`, is discouraged. Such reads can lead to pipeline
+/// stalls and yet-unspecified program behavior.
+///
 /// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm512_stream_pd&expand=5667)
 #[inline]
 #[target_feature(enable = "avx512f")]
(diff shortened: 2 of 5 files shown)


workingjubilee:
Also see:
- https://github.com/rust-lang/rust/issues/114582
- https://github.com/llvm/llvm-project/issues/64521

crates/core_arch/src/x86/avx.rs
```diff
@@ -1683,6 +1683,17 @@ pub unsafe fn _mm256_lddqu_si256(mem_addr: *const __m256i) -> __m256i {
 /// aligned memory location. To minimize caching, the data is flagged as
 /// non-temporal (unlikely to be used again soon)
 ///
+/// # Safety
+///
+/// After using this intrinsic, but before any atomic operations occur, a call
+/// to `_mm_sfence()` must be performed. A safe function that includes unsafe
+/// usage of this intrinsic must always end in `_mm_sfence()`.
```
the8472, line 1690: I think this is a bit too strict. There are niche scenarios where one would write out data without ever reading it again or at least without reading it again on another thread.
I those cases some later release write would be incidental and is only meant to order other, regular writes.

Maybe the entire requirement could be conditional on "if the written memory is intended to be made accessible on another thread through a release operation", with the recommendation "if in doubt, add a fence".

workingjubilee: I mean, I don't think it's coherent to leave an unsatisfied obligation hanging unless it's still in `unsafe`?

the8472: I guess the conservative definition is fine for now. But it could be phrased in a way that makes it clear that it may be replaced with a more refined definition at some point.

workingjubilee: I'd prefer to tell people to always discharge the obligation (even in the case you mention, the store-store fence would serialize any remaining deferred write-buffers, which is desired if you might run out of time otherwise!) and relax things when we mechanize a better spec.

RalfJung, line 1690: We are currently moving into a different direction in https://github.com/rust-lang/rust/issues/114582: if we follow @talchas' approach, the rule will be: after using this intrinsic, but (happens-)before any other read or write of this location in any thread, a fence must occur (the fence must be in the same thread that called the intrinsic).

RalfJung: I'd prefer if we'd call that UB instead of merely "discouraged".

Ideally even further nontemporal writes (before the next sfence) would be UB... I'd find it strange to have a situation where a nontemporal store would be allowed but a regular store would not.

crates/core_arch/src/x86/avx.rs
```diff
@@ -1683,6 +1683,17 @@ pub unsafe fn _mm256_lddqu_si256(mem_addr: *const __m256i) -> __m256i {
...
+///
+/// Reading and writing to the memory stored-to by any other means, after any
+/// nontemporal store has been used to write to that memory, but before the
+/// use of `_mm_sfence()`, is discouraged. Such reads can lead to pipeline
+/// stalls and yet-unspecified program behavior.
```
workingjubilee, line 1695: Hm. Well, both x86's movntdq and Arm's stnp retain local-thread serial ordering if you use them twice on the same location, i.e. that using `_mm_stream_ps(ptr, a); _mm_stream_ps(ptr, b);` still will yield `b` when you later read it, and coalescing such writes is traditionally allowed by compilers, so I would prefer to retain the more ambiguous verbiage I used here, @RalfJung. Mostly, I said "any other means" to cast a wide net so that people avoid it rather than inviting people to try to reason about in a way that might end in them deciding it's suddenly okay to perform an atomic store or atomic load on the memory from another thread.

workingjubilee: I do agree that we should canonically allow `_mm_stream_ps(ptr, a); _mm_store_ps(ptr, b); _mm_sfence();` eventually but I would prefer it live in  an in-between state until we are clear on our mechanics.

workingjubilee:
Will address the comments after the conversation settles a bit.

crates/core_arch/src/x86/avx.rs
RalfJung: The thing is that the proposed doc just doesn't align with the proposed spec in https://github.com/rust-lang/rust/issues/114582. If you go with @talchas' original proposed spec, the doc should be something like

"After the `_mm_stream_ps` until the next time this thread calls `_mm_sfence`, any attempt to access (read or write) this memory from this thread by means other than `_mm*_stream*` operations is UB. Furthermore, any access from another threads must synchronize-with that `_mm_sfence` or else any access (read or write) from the other thread is UB, including `_mm*_stream*` operations.

Basically, the nontemporal writes performed by streaming operations should be considered not ordered even with other operations in the thread they appear in, except with other nontemporal writes, until the next sfence which establishes synchronization with all nontemporal writes of the current thread."

That would allow your example. Crucially it disallows `_mm_stream_ps(ptr, a); *ptr = a; _mm_sfence();` and `_mm_stream_ps(ptr, a); println!("{:?}", *ptr); _mm_sfence();` which is disallowed by the proposed spec. We should probably have examples in the docs.

workingjubilee: ah, `_mm_store_ps` is morally equivalent to assignment.

RalfJung: Oh sorry, I misread your example and through it said `stream` twice.

If we make `_mm_store_ps` also inline assembly we might be able to come up with a mechanism that allows it. But for regular assignment I'm quite worried that LLVM optimizations will make deductions from seeing an assignment that are incompatible with the nontemporal state of that location.

workingjubilee: Ideally our spec should closely align with the actual nontemporal store instructions, which afaict permit both of those examples, however. I'm sympathetic with feeling like we're inviting too much chaos from LLVM, but I think the people working on Alive2 are likely to be equally unexcited about LLVM doing wild optimizations that make bad assumptions.

Oddly enough I'm okay with directly saying that you shouldn't load at all before an sfence though. I'm not sure why I'm okay with that and making normal successive stores UB bothers me. I guess because it still makes sense on a naturally weak-memory system?

RalfJung: > Oddly enough I'm okay with directly saying that you shouldn't load at all before an sfence though. I'm not sure why I'm okay with that and making normal successive stores UB bothers me.

That is somewhat strange indeed. Usually "I am allowed to do a normal store" always implies "I am allowed to do a normal load", and that seems like such a fundamental reasoning principle that I'd rather not break it.

workingjubilee: I don't have any problem with "write-only" data, I guess? Isn't "write-only data is okay, actually, and it doesn't affect optimizations" why our current stance is towards endorsing https://github.com/rust-lang/unsafe-code-guidelines/issues/346 and such?

RalfJung: https://github.com/rust-lang/unsafe-code-guidelines/issues/346 is not write-only data. It is always okay to read such data at type `MaybeUninit<T>`.

workingjubilee: ...in practice, as someone actually writing code in Rust, that means that if I can take `&mut [u8]` and write to it even if it is uninitialized data, but it's UB to read from it until initialized, I, the programmer, have to reason about it as write-only data.

RalfJung: If you have an `&mut [u8]` it is always okay to read from it with something like `slice.as_ptr().cast::<MaybeUninit<u8>>().add(n).read()` (assuming you did bounds checking). So to me this does not qualify as write-only data.

workingjubilee: Again, that is functionally equivalent for the programmer.

workingjubilee: If you tell people that `slice: &mut [u8]` is okay to read from even if it is uninit, they are going to assume you mean something way more "free and clear" than you actually do, e.g. that `let byte = *slice[i];` is okay. That is, that the constraint is going to be understood as "for all safe Rust expressions that, applied to this type, perform a read operation", which is just wrong if they don't perform a type-changing cast on the slice or a reference or pointer derived from it.

The object we are talking about, at the source level, is the binding in the current case, not necessarily the resulting evaluation of any given expression involving it, even of those which do not transitively contain UB. It has a number of different expressions that can use it. And if a message has enough hidden complexities in it, like "you can read from it (but only if you do a bunch of steps that make "it" nothing like what you currently have)" then you should assume the programmer makes the demonic choice.

The entire lesson of functional programming and indeed Rust is that if you pile on enough expressions it is basically equivalent to a mutation anyways. I realize that in the Rust AM, what you described is not necessarily a state-mutating change until the final actual read, but programmers who are thinking strictly about the angelic choices are going to misinterpret statements like that.

This is also why I recommend applying a simpler overconstraint that aligns more closely with the current expected reality of these intrinsics: we can only guarantee programmers make angelic choices in the presence of a relatively simple model that aligns fairly well with existing documentation on how nontemporal stores work. Honestly, even the Intel documentation doesn't really say "oh yeah, use sfence a lot if you're also doing this with concurrency", at least, not enough in my opinion, so it all kinda sucks, but we should limit our contradictions as much as possible. And many, many people already issue the ["don't load from write-combining memory"](https://fgiesen.wordpress.com/2013/01/29/write-combining-is-not-your-friend/) advice.

RalfJung: There is a myth that "you cannot read uninitialized memory" and I think it is harmful. Instead we should teach people what actually happens: when reading memory at a certain type, that memory must be sufficiently "valid" for this type. If you read at `bool`, it must be 0 or 1. If you read at `i32`, it must be initialized. If you read at `MaybeUninit<_>`, it can be anything. I think telling people they can't read from uninit memory is a complete distraction and leads to an unnecessary complicated mental model. There's a reason that "reading from uninit memory" does *not* show up in [this list](https://doc.rust-lang.org/reference/behavior-considered-undefined.html).

But anyway, it seems unlikely we will get to an agreement on terminology here and that's all rather off-topic anyway. The on-topic question is whether we should allow people to perform regular writes in between the nontemporal write and the fence. I don't see a good motivation for doing that, and it opens some tricky questions that we'd need to carefully figure out before allowing it. For instance, if I do a nontemporal write and then a regular write, do I even still need the fence or is it now guaranteed that the next "release" operation synchronizes everything properly?

workingjubilee: Ehnn I mean yes, just...

...Anyways I think, dreadfully, to actually answer your question about the write series, that on x86 you still need the fence, from the ISA's point of view.

workingjubilee: Hmm. The local results should be consistent absolutely, but I am concerned about code that may look more like moving overlapping sizes, so
```asm
vmovntdq [somewhere], ymm09
mov [somewhere], r09
mov [flag], 1
```

The first 8 bytes are guaranteed once the flag is set due to the two regular movs participating in TSO, but the vmovntdq means 24 bytes are in an ambiguous state.

RalfJung: > do I even still need the fence or is it now guaranteed that the next "release" operation synchronizes everything properly?

If the answer to that is yes then I don't think we can allow regular writes. A regular write in Rust is guaranteed to be properly released by a release operation, after all.

We *can* have intrinsics that do "regular write to something that might be in the nontemporal state", but those need the same inline-assembly-and-Rust-replacement-code-spec treatment as streaming writes.

I'd really prefer if we didn't have to do this... the usecases we are aware of don't need this, do they?

workingjubilee: Hmm, alright then, I didn't realize that was the specific way the sequencing invariants were formed. It's *possible* the reality may be amenable to regular writes (assuming a sort of "wholesale adoption of the x86 mechanics into Rust" approach for this case) but I'd have to examine the rules... very closely.

I don't think the majority of use cases need this in practice, correct, so now that we have hashed out that issue as existing, then I am happy to let this one go. I will update the documentation changes here accordingly.

crates/core_arch/src/x86/avx.rs
RalfJung: I opened https://github.com/rust-lang/stdarch/pull/1534 with my alternative wording.

PR #1461, opened 2023-08-16
Implement RISC-V Zb intrinsics 

coastalwhite:
This is the second PR for rust-lang/rust#114544 and a follow-up to #1453. 

This PR adds all the instructions for the `zb` extension. This PR will get force-pushed after #1453 is merged.


Amanieu:
Most of these already exist as plain Rust operations or methods on the integer types. We do *not* need intrinsics for these.

I would only keep these which don't have a straightforward mapping to plain Rust code:
- `clmul*`
- `orc_b`.

coastalwhite:
This is now also done if you are okay with it :smile:

PR #1471, opened 2023-09-29
Deprecate functions that use `_mm_setcsr` or `_mm_getcsr` internally

eduardosm:
Follow-up of https://github.com/rust-lang/stdarch/pull/1454

Additionally, I removed references to `_mm_setcsr` and `_mm_getcsr` from documentation of non-deprecated functions.

cc @RalfJung

### DIFF
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -33,6 +33,11 @@ case ${TARGET} in
     i686-* | i586-*)
         export RUSTFLAGS="${RUSTFLAGS} -C relocation-model=static -Z plt=yes"
         ;;
+    # Some x86_64 targets enable by default more features beyond SSE2,
+    # which cause some instruction assertion checks to fail.
+    x86_64-*)
+        export RUSTFLAGS="${RUSTFLAGS} -C target-feature=-sse3"
+        ;;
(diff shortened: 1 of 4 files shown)


eduardosm:
Is this generating SSE4.1 instructions in SSE functions?

```
---- core_arch::x86::sse::assert__mm_move_ss_movss stdout ----
disassembly for stdarch_test_shim__mm_move_ss_movss: 
	 0: pushq %rbp
	 1: movq %rsp, %rbp
	 2: blendps $1, %xmm1, %xmm0
	 3: popq %rbp
	 4: retq
	 5: nopl (%rax)
thread 'core_arch::x86::sse::assert__mm_move_ss_movss' panicked at crates/stdarch-test/src/lib.rs:172:9:
failed to find instruction `movss` in the disassembly
```

eduardosm:
It looks like on macOS some extensions beyond SSE2 are enabled by default, so their instructions can be emitted regardless of the `#[target_feature(enable = ...)]` attribute. I added `-C target-feature=-sse3` as a workaround.

Amanieu:
Thanks!

PR #1472, opened 2023-09-29
Fix some SSE and SSE2 tests for Miri

eduardosm:
We are adding stdarch to libstd Miri testing (https://github.com/rust-lang/miri-test-libstd/pull/19).

Some tests required some fixing, while there are other that are simply ignored because it is unlikely that Miri support them in the near future.

For now, only SSE and SSE2 are being considered.

cc @RalfJung

### DIFF
--- a/crates/core_arch/src/arm_shared/neon/shift_and_insert_tests.rs
+++ b/crates/core_arch/src/arm_shared/neon/shift_and_insert_tests.rs
@@ -22,7 +22,7 @@ macro_rules! test_vsli {
             let a = [$($a as $t),*];
             let b = [$($b as $t),*];
             let n_bit_mask: $t = (1 << $n) - 1;
-            let e = [$(($a as $t & n_bit_mask) | ($b as $t << $n)),*];
+            let e = [$(($a as $t & n_bit_mask) | (($b as $t) << $n)),*];
             let r = $fn_id::<$n>(transmute(a), transmute(b));
             let mut d = e;
             d = transmute(r);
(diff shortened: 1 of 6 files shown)


Amanieu:
Can you rebase to fix CI?

eduardosm:
Why did this start failing now?

```
error: `<<` is interpreted as a start of generic arguments for `i8`, not a shift
  --> crates/core_arch/src/arm_shared/neon/shift_and_insert_tests.rs:25:60
   |
25 |             let e = [$(($a as $t & n_bit_mask) | ($b as $t << $n)),*];
   |                                                            ^^ not interpreted as shift
...
33 | test_vsli!(test_vsli_n_s8, i8 => vsli_n_s8([3, -44, 127, -56, 0, 24, -97, 10], [-128, -14, 125, -77, 27, 8, -1, 110], 5));
   | -------------------------------------------------------------------------------------------------------------------------
   | |                                                                                                                     |
   | |                                                                                                                     interpreted as generic arguments
   | in this macro invocation
   |
   = note: this error originates in the macro `test_vsli` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Amanieu:
This is probably because something changed in the latest rustc nightly. Hopefully it will get noticed and reverted soon, but otherwise this should be fairly easy to fix.

eduardosm:
It think I figured it out. It is not a change in rustc. It is the change to `simd_test` in this PR, which causes `$t` type to be reparsed, which tries to consume the `<<` token as part of the type. I will add the needed parentheses.

PR #1476, opened 2023-10-04
Stabilize Ratified RISC-V Target Features

coastalwhite:
As shortly discussed on Zulip
(https://rust-lang.zulipchat.com/#narrow/stream/250483-t-compiler.2Frisc-v/topic/Stabilization.20of.20RISC-V.20Target.20Features/near/394793704), this commit PR the ratified RISC-V instruction bases and extensions.

Specifically, this PR stabilizes the:
* Instruction-Fetch Fence (Zifencei) on v2.0
* Control and Status Register (Zcsr) on v2.0
* Pause Hint (Zihintpause) on v2.0
* Zicntr and Zihpm were ratified March 2023
* ~Total Store Ordering (Ztso) on v1.0~
* Supervisor-level Instructions (S) on v1.12
* Atomic Instructions (A) on v2.0
* Compressed Instructions (C) on v2.0
* ~Double-Precision Floating-Point (D) on v2.2~
* Embedded Base (E) (Given as `RV32E` / `RV64E`) on v2.0
* ~Single-Precision Floating-Point (F) on v2.2~
* Integer Multiplication and Division (M) on v2.0
* ~Vector Operations (V) on v1.0~
* Bit Manipulations (B) on v1.0 listed as `zba`, `zbc`, `zbs`
* Scalar Cryptography (Zk) v1.0.1 listed as `zk`, `zkn`, `zknd`, `zkne`, `zknh`, `zkr`, `zks`, `zksed`, `zksh`, `zkt`, `zbkb`, `zbkc` `zkbx`
* ~Double-Precision Floating-Point in Integer Register (Zdinx) on v1.0~
* ~Half-Precision Floating-Point (Zfh) on v1.0~
* ~Minimal Half-Precision Floating-Point (Zfhmin) on v1.0~
* ~Single-Precision Floating-Point in Integer Register (Zfinx) on v1.0~
* ~Half-Precision Floating-Point in Integer Register (Zhinx) on v1.0~
* ~Minimal Half-Precision Floating-Point in Integer Register (Zhinxmin) on v1.0~

When this is merged, a PR for the main Rust repository can be created.

### DIFF
--- a/crates/std_detect/src/detect/arch/riscv.rs
+++ b/crates/std_detect/src/detect/arch/riscv.rs
@@ -24,19 +24,39 @@ features! {
     ///
     /// The supported ratified RISC-V instruction sets are as follows:
     ///
+    /// * RV32E: `"rv32e"`
     /// * RV32I: `"rv32i"`
+    /// * RV64I: `"rv64i"`
+    /// * A: `"a"`
+    /// * B: `"b"`
+    ///   * Zba: `"zba"`
+    ///   * Zbb: `"zbb"`
+    ///   * Zbc: `"zbc"`
+    ///   * Zbs: `"zbs"`
+    /// * C: `"c"`
+    /// * D: `"d"`
+    /// * F: `"f"`
+    /// * M: `"m"`
+    /// * Q: `"q"`
+    /// * V: `"v"`
+    /// * Zicntr: `"zicntr"`
+    /// * Zicsr: `"zicsr"`
     /// * Zifencei: `"zifencei"`
     /// * Zihintpause: `"zihintpause"`
-    /// * RV64I: `"rv64i"`
-    /// * M: `"m"`
-    /// * A: `"a"`
-    /// * Zicsr: `"zicsr"`
-    /// * Zicntr: `"zicntr"`
     /// * Zihpm: `"zihpm"`
-    /// * F: `"f"`
-    /// * D: `"d"`
-    /// * Q: `"q"`
-    /// * C: `"c"`
+    /// * Zk: `"zk"`
+    ///   * Zbkb: `"zbkb"`
+    ///   * Zbkc: `"zbkc"`
+    ///   * Zbkx: `"zbkx"`
+    ///   * Zkn: `"zkn"`
+    ///     * Zknd: `"zknd"`
+    ///     * Zkne: `"zkne"`
+    ///     * Zknh: `"zknh"`
+    ///   * Zkr: `"zkr"`
+    ///   * Zks: `"zks"`
+    ///     * Zksed: `"zksed"`
+    ///     * Zksh: `"zksh"`
+    ///   * Zkt: `"zkt"`
     ///
     /// There's also bases and extensions marked as standard instruction set,
     /// but they are in frozen or draft state. These instruction sets are also
@@ -44,6 +64,8 @@ features! {
     ///
     /// Frozen RISC-V instruction sets:
     ///
+    /// * Zfh: `"zfh"`
+    /// * Zfhmin: `"zfhmin"`
     /// * Zfinx: `"zfinx"`
     /// * Zdinx: `"zdinx"`
     /// * Zhinx: `"zhinx"`
@@ -52,14 +74,9 @@ features! {
     ///
     /// Draft RISC-V instruction sets:
     ///
-    /// * RV32E: `"rv32e"`
     /// * RV128I: `"rv128i"`
-    /// * Zfh: `"zfh"`
-    /// * Zfhmin: `"zfhmin"`
-    /// * B: `"b"`
     /// * J: `"j"`
     /// * P: `"p"`
-    /// * V: `"v"`
     /// * Zam: `"zam"`
     ///
     /// Defined by Privileged Specification:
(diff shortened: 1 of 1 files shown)


Amanieu:
Could you open a PR for rust-lang/rust first which stabilizes the compiler-level feature names so that we can start a T-lang FCP on that?

coastalwhite:
Do I need to request a separate FCP for this PR or is this also okay to merge if rust-lang/rust#116485 passes the FCP?

Amanieu:
The other FCP  is sufficient.

coastalwhite:
I just rebased this onto the `master` branch and force pushed. I think this is mergable now.

Amanieu:
Can you bump the version number in the attributes to 1.76 (current nightly)?

coastalwhite:
> Can you bump the version number in the attributes to 1.76 (current nightly)?

Done

crates/std_detect/src/detect/arch/riscv.rs
```diff
@@ -146,7 +135,7 @@ features! {
     /// "Zfh" Standard Extension for 16-Bit Half-Precision Floating-Point
     @FEATURE: #[unstable(feature = "stdarch_riscv_feature_detection", issue = "111192")] zfhmin: "zfhmin";
     /// "Zfhmin" Standard Extension for Minimal Half-Precision Floating-Point Support
-    @FEATURE: #[unstable(feature = "stdarch_riscv_feature_detection", issue = "111192")] b: "b";
+    @FEATURE: #[stable(feature = "riscv_ratified", since = "1.76.0")] b: "b";
```
Amanieu, line 138: "b" as an extension doesn't actually exist. Instead we should only expose the underlying zba/zbb/zbs extension.

crates/std_detect/src/detect/arch/riscv.rs
```diff
@@ -157,7 +146,7 @@ features! {
     @FEATURE: #[unstable(feature = "stdarch_riscv_feature_detection", issue = "111192")] zam: "zam";
     /// "Zam" Standard Extension for Misaligned Atomics
 
-    @FEATURE: #[unstable(feature = "stdarch_riscv_feature_detection", issue = "111192")] s: "s";
+    @FEATURE: #[stable(feature = "riscv_ratified", since = "1.76.0")] s: "s";
```
Amanieu, line 149: None of our feature detection (via hwcaps/cpuinfo) actually enables this feature, so it doesn't make sense to stabilize it.

crates/std_detect/src/detect/arch/riscv.rs
```diff
@@ -70,61 +87,33 @@ features! {
...
+    @FEATURE: #[stable(feature = "riscv_ratified", since = "1.76.0")] a: "a";
     /// "A" Standard Extension for Atomic Instructions
-    @FEATURE: #[unstable(feature = "stdarch_riscv_feature_detection", issue = "111192")] zicsr: "zicsr";
+    @FEATURE: #[stable(feature = "riscv_ratified", since = "1.76.0")] zicsr: "zicsr";
     /// "Zicsr", Control and Status Register (CSR) Instructions
-    @FEATURE: #[unstable(feature = "stdarch_riscv_feature_detection", issue = "111192")] zicntr: "zicntr";
+    @FEATURE: #[stable(feature = "riscv_ratified", since = "1.76.0")] zicntr: "zicntr";
     /// "Zicntr", Standard Extension for Base Counters and Timers
-    @FEATURE: #[unstable(feature = "stdarch_riscv_feature_detection", issue = "111192")] zihpm: "zihpm";
+    @FEATURE: #[stable(feature = "riscv_ratified", since = "1.76.0")] zihpm: "zihpm";
```
Amanieu, line 108: Same here: zicsr, zicntr, zihpm, rv32i, rv64i, zifencei, zihintpause will all never be enabled, because it's not exposed in hwcaps. These should not be stabilized as is.

It's not an issue for feature that are already exposed on the compiler side (e.g. zba, zbb) since in that case we just return the static configuration from `cfg(target_feature)`.

crates/std_detect/src/detect/arch/riscv.rs
coastalwhite: The "zb* does exist in the target features. Should we expose that?

crates/std_detect/src/detect/arch/riscv.rs
coastalwhite: Since the linux kernel does now detect them, should be add them to the linux hwcaps?

<https://github.com/torvalds/linux/blob/master/arch/riscv/include/asm/hwcap.h>

Amanieu: For now I would prefer sticking to the same set of features that was stabilized in the compiler itself.

crates/std_detect/src/detect/arch/riscv.rs
Amanieu: I don't see "zb" in `llc -march=riscv64 -mattr=help` or in https://github.com/rust-lang/rust/pull/116485.

coastalwhite: Okay, I removed the `zb` and `b` features.

crates/std_detect/src/detect/arch/riscv.rs
coastalwhite: These are also removed.

crates/std_detect/src/detect/arch/riscv.rs
```diff
@@ -168,39 +155,41 @@ features! {
     @FEATURE: #[unstable(feature = "stdarch_riscv_feature_detection", issue = "111192")] h: "h";
     /// Hypervisor Extension
 
-    @FEATURE: #[unstable(feature = "stdarch_riscv_feature_detection", issue = "111192")] zba: "zba";
+    @FEATURE: #[stable(feature = "riscv_ratified", since = "1.76.0")] zb: "zb";
+    /// "Zb" Standard Extension for Bit Manipulation
```
Amanieu, line 159: This should be removed.

Amanieu:
CI should be fixed, can you rebase?

coastalwhite:
Done :smile:

PR #1482, opened 2023-10-11
Fix UB in _mm_movemask_ps, _mm_movemask_pd, _mm256_movemask_ps and _mm256_movemask_pd

eduardosm:
The `simd_bitmask` intrinsic requires each element to be all-1 or all-0, while the x86 functions only check for the highest bit.


eduardosm:
Hmmm

```
LLVM ERROR: IO failure on output stream: No space left on device
error: could not compile `core_arch` (lib)
Error: Process completed with exit code 101.
```

Amanieu:
Can you do a quick test on godbolt to ensure this still compiles down to 1 instruction?

eduardosm:
Of course: https://godbolt.org/z/n6EMz7dbv

Issue #1484, opened 2023-10-19
NEON intrinsics are broken on big-endian


Amanieu:
These are currently broken because the order of elements inside vectors is reversed on big-endian systems: the ARM ABI requires that element 0 is located at the highest address of the vector type. However LLVM intrinsics expect element 0 to be located at the lowest address.

See https://llvm.org/docs/BigEndianNEON.html and `arm_neon.h` in Clang for more details.

RalfJung:
> the ARM ABI requires that element 0 is located at the highest address of the vector type. However LLVM intrinsics expect element 0 to be located at the lowest address.

What exactly does this mean? Is there a bug in LLVM? If so, where is it tracked?

Or is the problem that Rust stdarch wants to expose the intrinsics the way they work on hardware, but LLVM doesn't provide those semantics? If so, could that be fixed by doing appropriate translation of indices before calling the intrinsics?

Amanieu:
The short answer is that, on big-endian, LLVM portable vectors have a different element ordering than the one in the vector types used by the NEON intrinsics.

The C intrinsics work around this by reversing the element ordering in vectors before & after each intrinsic. We need to do the same in stdarch.

RalfJung:
Oh I see, so this is a mismatch about the `simd_x` intrinsics vs vendor-specific intrinsics? Okay makes sense.

OTOH this is good news for portable-simd, seems like there we'll be getting consistent behavior across platforms without extra work then.

he32:
As stated elsewhere, I reported the issue https://github.com/rust-lang/rust/issues/129819 earlier, and now have workarounds in place so that I'm able to produce a working rust compiler on big-endian NetBSD/aarch64 by avoiding attempts to use the NEON extensions in that mode.

However, since those extensions are available in the CPU, a better solution would be to fix this issue and then probably to revert the workarounds.

[...]

PR #1485, opened 2023-10-19
Disable NEON intrinsics on big-endian ARM

Amanieu:
These are currently broken because the order of elements inside vectors is reversed on big-endian systems: the ARM ABI requires that element 0 is located at the highest address of the vector type. However LLVM intrinsics expect element 0 to be located at the lowest address.

See https://llvm.org/docs/BigEndianNEON.html and `arm_neon.h` in Clang for more details.

This is tracked in #1484.

Although this is a breaking change, this is acceptable for 2 reasons:
- big endian ARM targets are only tier 3.
- it is preferable to stop existing code from compiling than to let it run and produce incorrect results.

### DIFF
--- a/crates/core_arch/src/aarch64/mod.rs
+++ b/crates/core_arch/src/aarch64/mod.rs
@@ -6,7 +6,10 @@
 //! [arm_ref]: http://infocenter.arm.com/help/topic/com.arm.doc.ihi0073a/IHI0073A_arm_neon_intrinsics_ref.pdf
 //! [arm_dat]: https://developer.arm.com/technologies/neon/intrinsics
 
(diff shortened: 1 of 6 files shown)


Amanieu:
@rust-lang/libs-api @rust-lang/libs

This is a breaking change, but a necessary one in my opinion until the intrinsics are fixed. They currently silently produce the wrong results, which has led to a bug in hashbrown on big-endian ARM: https://github.com/rust-lang/rust/issues/116880

Amanieu:
cc @adamgemmell @JamieCunliffe @jacobbramley

jacobbramley:
Fixing this is not likely to be quick or easy, so I agree that disabling them for BE targets is probably the best approach for now.

PR #1490, opened 2023-10-30
Fix some minor intrinsic-test issues.

jacobbramley:
Notably, fix some warnings that flood the console, making it awkward to debug Arm tests.

### DIFF
--- a/crates/intrinsic-test/src/types.rs
+++ b/crates/intrinsic-test/src/types.rs
@@ -1,6 +1,8 @@
 use std::fmt;
 use std::str::FromStr;
 
+use itertools::Itertools as _;
+
 use crate::values::value_for_array;
 use crate::Language;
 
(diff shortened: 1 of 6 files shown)


crates/intrinsic-test/src/types.rs
```diff
@@ -288,82 +290,79 @@ impl IntrinsicType {
...
+    ///
     /// This is determistic based on the pass number.
     ///
     /// * `loads`: The number of values that need to be loaded from the argument array
     /// * e.g for argument type uint32x2, loads=2 results in a string representing 4 32-bit values
     ///
     /// Returns a string such as
-    /// * `0x1, 0x7F, 0xFF` if `language` is `Language::C`
-    /// * `0x1 as _, 0x7F as _, 0xFF as _` if `language` is `Language::Rust`
+    /// * `[0x1, 0x7F, 0xFF]` if `language` is `Language::C`
```
Amanieu, line 302: Should this be `{}` instead?

Amanieu (commented):
LGTM with the minor typo in the comment fixed.

crates/intrinsic-test/src/types.rs
jacobbramley: Fixed in (force-pushed) https://github.com/rust-lang/stdarch/commit/326fe6f04c8393cbdfe66cdbf8a673dde14600b3.

PR #1491, opened 2023-10-31
std_detect: Add support for LoongArch

heiher:
Add run-time feature detection for LoongArch.


lu-zero (commented):
It looks good, is there a way to test it? (e.g. qemu?)

heiher:
> It looks good, is there a way to test it? (e.g. qemu?)

Thank you for the comment. Currently no. The qemu-loongarch64 is already in ubuntu-23.04, but the loongarch64 cross toolchain is still missing.  I will add CI when these are ready.

PR #1499, opened 2023-11-18
Re-implement some AVX functions without LLVM intrinsics

eduardosm:


bjorn3:
Do these have the same behavior for NaN (including payload preservation if applicable), Infinity, -0.0 and so on?

eduardosm:
They produce the same assembly as the previous implementation.

RalfJung:
> They produce the same assembly as the previous implementation.

Producing the same assembly does not mean that LLVM guarantees the same behavior for NaNs. They might get optimized in different ways.

eduardosm:
If you are asking if there is written documentation that allows to formally prove that both versions are functionally identical, there isn't (or at least I couldn't find it), so let's speculate.

For the shuffle part, it would surprise me if it wasn't lossless (i.e., copy a bunch of bits from one place to another without caring if they represent a NaN or some weird value).

For the FP-to-FP cast, I would consider it a LLVM bug if it didn't follow IEEE semantics (unless [fast-math flags](https://llvm.org/docs/LangRef.html#fast-math-flags) were used). Since SSE/AVX is available, it would be pointless that it falled back to X87.

Issue #1217, opened 2021-09-11
arm neon vld1*x tests fail on arm hw due to misaligned pointers


hkratz:
vld1*x* intrinsics are compiled to have alignment requirements, e.g. the assembly emitted for:
```rust
pub unsafe fn vld1_f32_x2(a: *const f32) -> float32x2x2_t { ... }
```
is
```
<stdarch_test_shim_vld1_f32_x2_vld1>:
      f4200a9f        vld1.32 {d0-d1}, [r0 :64]
      e12fff1e        bx      lr
```
and requires `a`to be 64-bit aligned (denoted by the `:64` in the assembly). Clang does the same AFAICS.

Two problems:
1) The alignment requirements are neither documented in Rust nor obvious and aarch64 has no such requirements.
2) Our tests do not pass properly aligned pointers. Execution of the test suite on real arm hardware causes bus errors.

Possible solutions:
1) Relax the alignment requirements at a slight general performance loss. Haven't checked if this can be done easily.
2) Document and maybe assert the requirements and fix the tests.

cc @SparrowLii

Amanieu:
The alignment should be that of `f32`. This is what clang does for the same intrinsic.

hkratz:
> The alignment should be that of `f32`. This is what clang does for the same intrinsic.

In my test clang compiles the load with 64-bit alignment as well (`vld1.32 {d16, d17}, [r1:64]`): [Godbolt](https://godbolt.org/z/fvs1M9GGY)

hkratz:
As we are directly calling the LLVM intrinsics here without any possibility to specify alignment solution 2) seems to be the only way.

SparrowLii:
> As we are directly calling the LLVM intrinsics here without any possibility to specify alignment solution 2) seems to be the only way.

I agree. We can describe 1) in the document and try to find a better way in the future

hkratz:
I don't know if the LLVM behavior is intended. For `vld1_f32_x4` they require 256-bit alignment ([clang Godbolt](https://godbolt.org/z/q98rvM8cT)).
 
GCC does not even have the `vld1_*_x*` intrinsics for armv7 ([issue](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=71233)), so we can't compare.

[...]

PR #1505, opened 2023-11-29
Work around CI failures for the ARM target

Amanieu:
These seem to have been introduced by recent LLVM changes.

* The instruction limit for vld*/vst* has been raised. This is not a significant issue, it is only used for testing.
* vld*/vst* instructions are generated with overly strict alignments: https://github.com/rust-lang/stdarch/issues/1217
* vtbl/vtbx instrinsics are failing intrinsic-test for unknown reasons.


Amanieu:
@jacobbramley Do you think the vtbl/vtbx issue might be related to https://github.com/rust-lang/rust/issues/112460?

PR #1508, opened 2023-12-12
std_detect: add avx512fp16

usamoi:
get cpuid number from https://cdrdv2-public.intel.com/678970/intel-avx512-fp16.pdf

### DIFF
--- a/crates/std_detect/src/detect/arch/x86.rs
+++ b/crates/std_detect/src/detect/arch/x86.rs
@@ -75,6 +75,7 @@ features! {
     /// * `"avx512bitalg"`
     /// * `"avx512bf16"`
     /// * `"avx512vp2intersect"`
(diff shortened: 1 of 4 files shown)


Amanieu:

Amanieu:
I noticed that `avx512fp16` is missing from the list of target features in rustc. Additionally, adding a new feature will require an FCP. I think the best way forward is for you to submit a PR to add this target feature to rustc, at which point we can start an FCP on that PR which will cover both the target feature and the feature detection in stdarch.

usamoi:
`avx512fp16` is in the list of target features in latest nightly rustc now. Is an FCP still required? @Amanieu

Amanieu:
This is fine to merge, but we're waiting on https://github.com/rust-lang/rust/pull/121088 to fix CI-related issues.

PR #1516, opened 2024-01-10
Add CPU feature detection for macOS/aarch64.

makotokato:


lu-zero (approved):
Looks fine and seems correct on m2

crates/std_detect/src/detect/os/macos/aarch64.rs
```diff
@@ -0,0 +1,98 @@
+//! Run-time feature detection for aarch64 on macOS.
+
+use crate::detect::{cache, Feature};
+
+#[inline]
+fn _sysctlbyname(name: &str) -> bool {
+    use libc;
+
+    let mut enabled: i32 = 0;
+    let mut enabled_len: usize = 4;
```
Amanieu, line 10: Are you sure this is the correct size? I couldn't find anything in Apple's documentation that indicates what size to use.

crates/std_detect/src/detect/os/macos/aarch64.rs
```diff
@@ -0,0 +1,98 @@
...
+            name.as_ptr() as *const i8,
+            enabled_ptr,
+            &mut enabled_len,
+            core::ptr::null_mut(),
+            0,
+        )
+    };
+
+    match ret {
+        0 => enabled != 0,
```
Amanieu, line 24: Should the returned size be checked here as well?

crates/std_detect/src/detect/os/macos/aarch64.rs
makotokato: Other proeject such as [Google cpufeature](https://github.com/google/cpu_features/blob/main/src/impl_aarch64_macos_or_iphone.c) and [Apple WebKit](https://searchfox.org/wubkat/source/Source/JavaScriptCore/assembler/CPU.cpp#127) uses `int`.

crates/std_detect/src/detect/os/macos/aarch64.rs
makotokato: It is unnecessary to check size as long as I look Apple's WebKit code.

PR #1519, opened 2024-01-24
CI: add a success conclusion job

Kobzol:
This PR adds a new CI job that only succeeds if all other CI jobs have succeeded. This should make it easier to check if CI was green. After this PR, we can easily gate CI behind the "success" job in `team`, to avoid having to enumerate all the individual jobs, and to make sure that if new jobs are added in the future, they will be gated without having to modify `team`.


Kobzol:
Ok, looks like webhooks and communication with bors works.

Issue #1432, opened 2023-05-25
AArch64: Mismatch between compiler features and runtime detection


Amanieu:
While reviewing the list of target features, I noticed 2 inconsistencies between the [list used by the compiler](https://github.com/rust-lang/rust/blob/a2b1646c597329d0a25efa3889b66650f65de1de/compiler/rustc_codegen_ssa/src/target_features.rs#L59) and the [one used by stdarch for runtime feature detection](https://github.com/rust-lang/stdarch/blob/master/crates/std_detect/src/detect/arch/aarch64.rs).

Specifically these features are present as a separate feature in the runtime detection list but are folded into another feature in the compiler:
- `fp` can be detected at runtime but can't be enabled. It is instead implied by the `neon` feature.
- `pmull` can be detected at runtime but can't be enabled. It is instead implied by the `aes` feature, and all the related intrinsics use the `aes` target feature.

cc @JamieCunliffe @adamgemmell @workingjubilee

sayantn:
This was resolved in #1527

PR #1527, opened 2024-02-15
Improve feature detect for combined aarch64 features

adamgemmell:
LLVM's `ssbs` and `mte` target_features represent two Arm features. Linux's HWCAP also represents the same two features, so this is just a documentation update.

LLVM's `ras` target_feature represents two Arm features - FEAT_RAS and FEAT_RASv1p1. There is no runtime detection for this, so this is a no-op in stdarch.

LLVM's `aes` feature covers both `FEAT_AES` and `FEAT_PMULL`, but Linux exposes seperate feature bits. This patch makes the `aes` target_feature correctly shortcut runtime `pmull` detection and also makes the `aes` feature check for `pmull` at runtime to bring it in line with the target_feature behaviour. In practice I think this makes the two runtime features identical since the ID_AA64ISAR0_EL1 register does not allow for PMULL without AES. I haven't added a `pmull` target_feature because it would behave identically to the `aes` one, but I could do if you see fit.

https://github.com/rust-lang/stdarch/issues/1432

For reference:
Linux HWCAPs are documented here: https://www.kernel.org/doc/html/latest/arch/arm64/elf_hwcaps.html
They reference the ID registers documented here: https://developer.arm.com/documentation/ddi0595/2021-12/AArch64-Registers
LLVM target features are defined here: https://github.com/llvm/llvm-project/blob/main/llvm/lib/Target/AArch64/AArch64.td

### DIFF
--- a/crates/std_detect/src/detect/arch/aarch64.rs
+++ b/crates/std_detect/src/detect/arch/aarch64.rs
@@ -30,7 +30,7 @@ features! {
     /// * `"fhm"` - FEAT_FHM
     /// * `"dit"` - FEAT_DIT
     /// * `"flagm"` - FEAT_FLAGM
-    /// * `"ssbs"` - FEAT_SSBS
+    /// * `"ssbs"` - FEAT_SSBS & FEAT_SSBS2
     /// * `"sb"` - FEAT_SB
     /// * `"paca"` - FEAT_PAuth (address authentication)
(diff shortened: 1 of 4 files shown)


adamgemmell:
I've bisected the CI failure down to the LLVM 18 upgrade - I'll be able to look into it next week if no-one gets there first.

Amanieu (approved):
LGTM once CI is fixed (when the nightly updates)

PR #1528, opened 2024-02-16
Update test expectations for aarch64

nikic:
vcgez/vclez now generate different (and more obvious) instructions.

fmadd/fmsub are now preferred over the slower fmla/fmls in more cases. I believe this is specific to the LANE=0 case, but I couldn't figure out how to switch testing to use LANE=1 without also breaking the case where only LANE=0 is valid.


nikic:
I believe the fma change is https://github.com/llvm/llvm-project/commit/38c92c1ee2f07e3260c94d51834a97e84f93c708. I didn't try to track down the other one.

PR #1542, opened 2024-02-28
[wasm32] Add an intrinsic for the throw instruction

coolreader18:
For use in rust-lang/rust#121438. I was conservative in the wording of the docs cause I feel like this is the right place to put the intrinsic, but it likely won't be stabilized (at least in its current state).

### DIFF
--- a/crates/core_arch/src/wasm32/mod.rs
+++ b/crates/core_arch/src/wasm32/mod.rs
@@ -31,3 +31,25 @@ pub use self::memory::*;
 pub fn unreachable() -> ! {
     crate::intrinsics::abort()
 }
+
+extern "C" {
+    #[link_name = "llvm.wasm.throw"]
+    fn wasm_throw(tag: i32, ptr: *mut u8) -> !;
+}
+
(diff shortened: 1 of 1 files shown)


coolreader18:
```
   Compiling stdarch-test v0.1.0 (/checkout/crates/stdarch-test)
LLVM ERROR: undefined tag symbol cannot be weak
error: could not compile `core_arch` (lib test)
```

Ohh. Fair. I guess the other one has to come first.

coolreader18:
Ah. The test is failing because wasmtime doesn't have exception support yet :/

coolreader18:
@Amanieu what are your thoughts of just not having a `assert_instr` test for this?

Amanieu:
That's probably fine for now, but should be addressed as soon as it can be tested in CI.

coolreader18:
I was thinking of having it just be perma-unstable for the foreseeable future - the llvm intrinsic is likely to change, and I don't see it being super useful at least until panic=unwind via exception-handling is stable. It's more of just an implementation detail for libpanic_unwind to use.

coolreader18:
Eh whatever, I'll make an issue.

PR #1556, opened 2024-04-07
Make std_detect::detect::check_for unstably public

GnomedDev:
This is a requirement for my idea of how to make feature detection usable in core when std is present. See more information at [my Zulip post proposing the idea.](https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs/topic/Allowing.20runtime.20CPU.20feature.20detection.20in.20core/near/431776274)

I do not see any downside from this being merged, as if the idea turns out to not be wanted it doesn't affect any public API outside of libstd's usage due to the unstable attribute.

### DIFF
--- a/crates/std_detect/src/detect/arch/mod.rs
+++ b/crates/std_detect/src/detect/arch/mod.rs
@@ -53,7 +53,7 @@ cfg_if! {
     } else {
         // Unimplemented architecture:
         #[doc(hidden)]
-        pub(crate) enum Feature {
+        pub enum Feature {
             Null
         }
         #[doc(hidden)]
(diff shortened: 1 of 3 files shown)


Amanieu:
OK let's defer this for now until the LLVM issue is sorted.

Amanieu:
I would prefer seeing this move in a different direction. Essentially all of the `Feature` enums should be moved to core, and the feature detection logic should be moved to std. Essentially I would like std_detect to be moved out of this repo into the main rust-lang/rust repo, leaving this repo only for the arch-specific intrinsics and related tests.

GnomedDev:
Yeah, I was working on this for a bit and couldn't get the build system working but I was doing absolute hackiness (`include`ing std_detect into core with only the feature enum, then properly depending on std_detect trying to use the feature enum re-exported from core).

PR #1552, opened 2024-03-13
Stabilize AArch64 SHA3 intrinsics

tarcieri:
See also: rust-lang/rust#117225

PR #1558, opened 2024-04-09
typos: fix bad precision comments

messense:


lu-zero:
`intrinsic-test` fails with an unused variable, but the commit does not touch it...

messense:
> `intrinsic-test` fails with an unused variable, but the commit does not touch it...

See https://github.com/rust-lang/stdarch/pull/1552#issuecomment-2028455560

PR #1566, opened 2024-04-16
Enable testing for AArch64 Windows

dpaoliello:
Enables testing for AArch64 Windows as "no run", as there are no AArch64 GitHub runners available.

* Fixes setting up the `GITHUB_ENV` for Windows (use Bash instead of PowerShell).
* Instruct dumpbin not to print the raw assembly bytes.
* Fix rewriting the `shll` instructions to handle dumpbin's format.
* Skip tests where dumpbin does not know how to handle the instruction.
* Rewrite `ins` as `mov` for MSVC.
* In one case, expect `dup` instead of `mov` for MSVC.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -98,10 +98,10 @@ jobs:
         # Windows targets
         - x86_64-pc-windows-msvc
         - i686-pc-windows-msvc
+        - aarch64-pc-windows-msvc
         # FIXME: Disassembly not implemented for the # following targets:
         # - x86_64-pc-windows-gnu:
         # - i686-pc-windows-gnu:
-        # - aarch64-pc-windows-msvc:
 
         include:
         - target: i686-unknown-linux-gnu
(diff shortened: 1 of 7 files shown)


Amanieu:
`mov` is an alias for `ins` when moving vector elements, so it's just the disassembler choosing to disassemble things differently. The actual generated code is the same. `uxtll2` is not a valid instruction, the correct name is `uxtl2`, so that's just the disassembler being wrong.

It may be worth using `llvm-objdump` for disassembly on all targets for consistency.

dpaoliello:
> `mov` is an alias for `ins` when moving vector elements, so it's just the disassembler choosing to disassemble things differently. The actual generated code is the same. `uxtll2` is not a valid instruction, the correct name is `uxtl2`, so that's just the disassembler being wrong.

Ah, thanks! This reduced the failures to 1.

> It may be worth using `llvm-objdump` for disassembly on all targets for consistency.

I tried this, but it didn't work since `llvm-objdump` doesn't seem to read in the PDB files, and so the disassembly didn't contain any function names. Let me see if I can disable split-debuginfo for tests on Windows and try again...

PR #1571, opened 2024-05-01
Fix `std_detect_env_override` feature on Windows by avoiding `libc`

### LINKED ISSUE #1563: std_detect: Don't unnecessarily depend on libc for windows-msvc
I'm trying to reduce std's dependency on libc when it isn't actually needed.

Currently std_detect depends on libc for these features: https://github.com/rust-lang/stdarch/blob/9814fb0ec86e12ca68a3999660cee9617e232651/crates/std_detect/Cargo.toml#L37-L39

However, for windows-msvc, the only feature actually using libc in std_detect is `std_detect_env_override` https://github.com/rust-lang/stdarch/blob/9814fb0ec86e12ca68a3999660cee9617e232651/crates/std_detect/src/detect/cache.rs#L120-L137

The easier way to address this issue is to make it so the on windows-msvc targets, libc is only used for this feature. The harder way is to use the Windows native equivalent of `libc::getenv`.

dpaoliello:
As noted in #1563, the only dependency on `libc` when building `stdarch` for Windows was a call to `getenv` when the `std_detect_env_override` feature was enabled. Since the `libc` dependency is only enabled for non-Windows platforms, enabling the `std_detect_env_override` feature on Windows would cause a build break.

This change replaces the use of `getenv` with a call to `GetEnvironmentVariableA` on Windows and enables testing the `std_detect_env_override` feature.

Fixes #1563

r? @ChrisDenton

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: Install Rust
-      run: rustup update nightly && rustup default nightly
+      run: rustup update nightly --no-self-update && rustup default nightly
     - run: ci/style.sh
 
   docs:
@@ -25,7 +25,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: Install Rust
-      run: rustup update nightly && rustup default nightly
+      run: rustup update nightly --no-self-update && rustup default nightly
     - run: ci/dox.sh
       env:
         CI: 1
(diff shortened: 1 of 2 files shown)


ChrisDenton (approved):
lgtm

Though it looks like CI is stuck? cc @Amanieu

Amanieu:
I think your changes to the Github Actions configuration broke something.

dpaoliello:
> I think your changes to the Github Actions configuration broke something.

Ah, yep, I messed up the `matrix` syntax - sorry!

dpaoliello:
@Amanieu CI has been fixed up - I also added `--no-self-update` to each of the rustup invocations since the self-update was causing issues with the Windows builds.

PR #1573, opened 2024-05-06
feat: stabilization for stdarch_arm_crc32

ssukanmi:
Stabilization for CRC32 intrinsics on ARM, for which the FCP has been completed
https://github.com/rust-lang/rust/issues/117215

### DIFF
--- a/crates/core_arch/src/arm_shared/crc.rs
+++ b/crates/core_arch/src/arm_shared/crc.rs
@@ -58,7 +58,14 @@ use stdarch_test::assert_instr;
 #[target_feature(enable = "crc")]
 #[cfg_attr(target_arch = "arm", target_feature(enable = "v8"))]
 #[cfg_attr(test, assert_instr(crc32b))]
-#[unstable(feature = "stdarch_arm_crc32", issue = "117215")]
+#[cfg_attr(
+    target_arch = "arm",
+    unstable(feature = "stdarch_aarch32_crc32", issue = "125085")
+)]
+#[cfg_attr(
+    not(target_arch = "arm"),
+    stable(feature = "stdarch_aarch64_crc32", since = "1.80.0")
+)]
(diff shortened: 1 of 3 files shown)


crates/core_arch/src/arm_shared/crc.rs
```diff
@@ -58,7 +58,7 @@ use stdarch_test::assert_instr;
 #[target_feature(enable = "crc")]
 #[cfg_attr(target_arch = "arm", target_feature(enable = "v8"))]
 #[cfg_attr(test, assert_instr(crc32b))]
-#[unstable(feature = "stdarch_arm_crc32", issue = "117215")]
+#[stable(feature = "stdarch_arm_crc32", since = "1.80.0")]
```
Amanieu, line 61: Since target features for 32-bit ARM are not stabilized yet, we only want to stabilize these on AArch64 for now. You can do this by using `cfg_attr` like this:
```rust
#[cfg_attr(not(target_arch = "arm"), stable(feature = "stdarch_arm_crc32", since = "1.80.0"))]
```

Amanieu: Right, you actually do need to include both, like this:
```rust
#[cfg_attr(target_arch = "arm", unstable(feature = "stdarch_aarch32_crc32", issue = "XXXX"))]
#[cfg_attr(not(target_arch = "arm"), stable(feature = "stdarch_aarch64_crc32", since = "1.80.0"))]
```

You will need to create a new tracking issue for the stabilization of the intrinsics on 32-bit ARM, blocked on the target features being stabilized.

ssukanmi:
Added new PR with suggested change and created a new issue for intrinsics on 32 bit ARM https://github.com/rust-lang/rust/issues/125085

Amanieu:
Thanks! Now you just need to send a PR to rust-lang/rust to update the library/stdarch submodule.

PR #1575, opened 2024-05-30
use `simd_saturating_{add, sub}` on neon

folkertdev:
The idea behind this is that e.g. miri and cranelift don't know all of the llvm intrinsics (and adding them is possible but a lot of extra work). Using the generic simd function means logic only needs to be implemented once for all targets.

The `simd_saturating_*` functions generate the same instructions for the saturating add/sub operations https://godbolt.org/z/zc35doTc6

### DIFF
--- a/crates/core_arch/src/arm_shared/neon/generated.rs
+++ b/crates/core_arch/src/arm_shared/neon/generated.rs
@@ -36978,15 +36978,7 @@ pub fn vqabsq_s32(a: int32x4_t) -> int32x4_t {
     unstable(feature = "stdarch_arm_neon_intrinsics", issue = "111800")
 )]
 pub fn vqadd_s8(a: int8x8_t, b: int8x8_t) -> int8x8_t {
-    unsafe extern "unadjusted" {
-        #[cfg_attr(
-            any(target_arch = "aarch64", target_arch = "arm64ec"),
-            link_name = "llvm.aarch64.neon.sqadd.v8i8"
-        )]
-        #[cfg_attr(target_arch = "arm", link_name = "llvm.sadd.sat.v8i8")]
-        fn _vqadd_s8(a: int8x8_t, b: int8x8_t) -> int8x8_t;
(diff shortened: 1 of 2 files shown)


folkertdev:
I don't think this idea works because of this LLVM issue I opened https://github.com/llvm/llvm-project/issues/94463. So unless we can get that fixed (or somehow accept much worse codegen) this PR does not work

Amanieu:
OK let's defer this for now until the LLVM issue is sorted.

folkertdev:
@Jamesbarford given that you work at ARM, is there any chance you could pass this on to the right person (specifically the LLVM issue https://github.com/llvm/llvm-project/issues/94463).  Not only does it clean up stdarch a bit, but it's especially valuable for `std::simd`.

sayantn:
We can still use `simd_saturating_add` for arm, right? There are actually quite a few such occurrences in ARM (e.g. using `llvm.fma` instead of just using `simd_fma`). Also there are a few which use wrong LLVM intrinsics, I will send a PR to correct those soon

folkertdev:
I actually fixed https://github.com/llvm/llvm-project/issues/94463 in https://github.com/llvm/llvm-project/pull/140454. I just re-ran CI here to check if there are any other issues, and there were, resulting in https://github.com/llvm/llvm-project/issues/142323 and a fix in https://github.com/llvm/llvm-project/pull/142342. 

So, things are looking good for being able to use these operations for `aarch64` with the next llvm release. 

For `arm`: we can use the platform-independent versions if they don't cause regressions. If that's the case (looks like it for these operations at least) then yes, we can (and should!) use the  `simd_` intrinsics.

sayantn:
The ARM intrinsics *are* actually using the same LLVM intrinsics that `simd_saturating_{add,sub}` use, so it shouldn't cause any regressions

folkertdev:
@Amanieu this works now and is ready for review

PR #1576, opened 2024-06-03
Fix duplicated argument name in extern block

bjorn3:
While rustc accepts it just fine due to what is arguably a bug, this duplication does become an issue when generating functions from the llvm intrinsic declarations while reusing the argument names.


bjorn3:
CI failure is unrelated (the test thinks the int3 padding should be counted towards the instruction count and rejects it due to having too many instructions)

PR #1582, opened 2024-06-11
Use longer associated types in the Altivec traits

lu-zero:
Address https://github.com/rust-lang/rust/issues/126260

### DIFF
--- a/crates/core_arch/src/powerpc/altivec.rs
+++ b/crates/core_arch/src/powerpc/altivec.rs
@@ -410,8 +410,8 @@ mod sealed {
 
     #[unstable(feature = "stdarch_powerpc", issue = "111145")]
     pub trait VectorInsert {
-        type S;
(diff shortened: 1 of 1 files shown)


crates/core_arch/src/powerpc/altivec.rs
```diff
@@ -3233,18 +3233,18 @@ mod sealed {
...
+        type Shift;
+        unsafe fn vec_rl(self, b: Self::Shift) -> Self;
     }
 
     macro_rules! impl_vec_rl {
         ($fun:ident ($a:ident)) => {
             #[unstable(feature = "stdarch_powerpc", issue = "111145")]
             impl VectorRl for $a {
-                type B = t_u!($a);
+                type Shift = t_u!($a);
```
workingjubilee, line 3244: Should this be `Bits` for `B`? that would preserve the initial-letter relationship.

lu-zero: It would had been a better name indeed :)

PR #1584, opened 2024-06-16
Use stdsimd instead of LLVM imports in more instances.

TDecking:
Continuation of #790. Back then, there were instances where it was not possible to use the generic SIMD intrinsics due to LLVM bugs. These bugs have since then been fixed.


RalfJung:
@bjorn3 @antoyo the cranelift and GCC backends also seem to have special handling for some of these intrinsics that are now dead code and could possibly be removed.

bjorn3:
Removed in https://github.com/rust-lang/rustc_codegen_cranelift/commit/14d020f88d8840987d161d2425c30778499a6ac3, thanks!

Issue #1533, opened 2024-02-17
_mm512_reduce_add_ps and friends are setting fast-math flags they should not set


RalfJung:
Today I learned about the existence of the `simd_reduce_add_unordered` intrinsic. When called on a float, this compiles to LLVM's `vector.reduce.fadd` with the "fast" flag set, which means that passing in NAN or INF is UB and optimizations are allowed "to treat the sign of a zero argument or zero result as insignificant" (which I think means the sign of input zeros is non-deterministically swapped and returned zeros have non-deterministic sign).

This intrinsic is not used a lot in stdarch, but it has a total of 8 uses (all in `avx512f.rs`). 4 of these are integer intrinsics, where this should be entirely equivalent to `simd_reduce_add`; not sure why the "unordered" version is used. The other 4 are float intrinsics, `_mm512_reduce_add_ps` being the first:

https://github.com/rust-lang/stdarch/blob/4d9c0bb591336792c4c4baf293d0acc944e57e28/crates/core_arch/src/x86/avx512f.rs#L31262-L31270
```rust
/// Reduce the packed single-precision (32-bit) floating-point elements in a by addition. Returns the sum of all elements in a.
///
/// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm512_reduce_add_ps&expand=4562)
#[inline]
#[target_feature(enable = "avx512f")]
#[unstable(feature = "stdarch_x86_avx512", issue = "111137")]
pub unsafe fn _mm512_reduce_add_ps(a: __m512) -> f32 {
    simd_reduce_add_unordered(a.as_f32x16())
}
```

Neither the docs here nor Intel's docs mention that this is UB on NAN or INF, and the concerns around signed zeros and doing the addition in an unspecified order. Given that the Intel docs should be the authoritative docs (since this is a vendor intrinsic), why is it even correct to use fast-math flags here? Either the docs need to be updated to state the fast-math preconditions, or the implementation needs to be updated to avoid the fast-math flag. Maybe it should only use "reassoc", not the full but unsafe "fast" flag? But even that should probably be mentioned in the docs.

Amanieu:
It is definitely incorrect for these intrinsics to be using the fast-math flag.

Here is the LLVM IR that clang generates for this intrinsic:
```llvm
%0 = tail call reassoc noundef float @llvm.vector.reduce.fadd.v16f32(float -0.000000e+00, <16 x float> %x)
```

[...]

Issue #1178, opened 2021-06-07
Missing x86 vendor intrinsics (SSE2, SSE 4.1, AVX2)


newpavlov:
Previous issue: #40

# AVX2
 * [ ]  [`_mm256_stream_load_si256`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm256_stream_load_si256&expand=5236)
* [ ]  [`_mm_broadcastsi128_si256`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_broadcastsi128_si256&expand=5236)

# MMX 
EDIT(@workingjubilee): Direct MMX support is no longer in scope for `std::arch`, see:
- https://github.com/rust-lang/stdarch/pull/890

# SSE
* [ ]  [`_mm_free`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_free&expand=5236)
* [ ]  [`_mm_storeu_si16`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_storeu_si16&expand=5236)
* [ ]  [`_mm_loadu_si16`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_loadu_si16&expand=5236)
* [ ]  [`_mm_malloc`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_malloc&expand=5236)
* [ ]  [`_mm_storeu_si64`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_storeu_si64&expand=5236)

# SSE2
* [ ]  [`_mm_loadu_si32`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_loadu_si32&expand=5236)
* [ ]  [`_mm_storeu_si32`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_storeu_si32&expand=5236)

# SSE4.1

* [ ]  [`_mm_stream_load_si128`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_stream_load_si128&expand=5236)

[...]

PR #1594, opened 2024-06-24
Various Fixes and enhancements in x86 intrinsics

sayantn:
- Updated the x86-intel.xml to be a more recent update of Intel Intrinsics Guide (v3.6.8)
 - Added functionality to auto-generate a `missing-x86.md` for ease of implementation
 - Updated disassembly to allow windows-gnu targets (they have the same implementation as linux targets, as that uses `objdump` from binutils, and binutils is a dependency of GCC). Add the `x86_64-pc-windows-gnu` target in CI
 - Fixed some of the stream intrinsics.
 - modified floating-point reduce-add and reduce-mul intrinsics to NOT use `simd_reduce_add_unordered` and `simd_reduce_mul_unordered` as Intel specifies a strict associativity. Follow GCC and hand-implement the associativity ourselves (#1533)
 - Fixed `_load_mask32` etc in AVX512BW (they should have taken a `__mmask32`/`__mmask64` pointer, but took `u32`/`u64` pointer)
 - As moves never modify any flags, add `preserves_flags` to the `asm!` blocks for moves
 - Fix `_mm_loadu_si64` (it had target-feature sse, but needs sse2), `_mm256_extract_epi64`, `_mm256_extract_epi32`, `_mm256_cvtsi256_si32` (these had target-feature avx2, but need avx).
 -  Fixed `_mm_cvtt` intrinsics (they were actually calling vcvtss2si, when they should call vcvttss2si)
 - Removed all MMX support from stdarch-verify, and made the target-feature verification stricter
 - Implemented the missing intrinsics mentioned in #1178 with feature-gate `simd-x86-updates` (rust-lang/rust#126936)
 - Fixed `_mm512_kunpackb`
 - Modified the reduce-max and reduce-min intrinsics to preserve associativity specified by Intel and to use the comparison function they described (which is NOT `maxnum` from LLVM)
 - Bumped the OS in CI Docker containers to Ubunto 24.04 (except for in `armv7-unknown-linux-gnueabihf` and `x86_64-unknown-linux-gnu-emulated`)

Modifying fma has been moved to #1597 
Masked load/stores are on standby due to rust-lang/rust#126919


crates/core_arch/src/x86/sse41.rs
```diff
@@ -1148,6 +1142,35 @@ pub unsafe fn _mm_test_mix_ones_zeros(a: __m128i, mask: __m128i) -> i32 {
...
+/// mutates, a call to [`_mm_sfence`] must be performed by the thread that used the intrinsic. In
+/// particular, functions that call this intrinsic should generally call `_mm_sfence` before they
+/// return.
+///
+/// See [`_mm_sfence`] for details.
+#[inline]
+#[target_feature(enable = "sse4.1")]
+#[cfg_attr(test, assert_instr(movntdqa))]
+#[stable(feature = "simd_x86", since = "1.27.0")]
+pub unsafe fn _mm_stream_load_si128(mem_addr: *const __m128i) -> __m128i {
```
workingjubilee, line 1163: Could you split out the addition of new stable intrinsics under an existing feature flag into a new PR, so that they can be FCPed if necessary?

sayantn: There are 5-6 missed functions in sse2, sse3, ssse3 and avx2. What feature flag should I use?

workingjubilee: Oh, I have no objection to you proposing they be insta-stabilized under that feature flag, it's just better to have insta-stable additions be in their own PR.

sayantn: Oh, ok. So they should be insta-stabilized under `simd_x86`? and what should be the since argument be? (I don't know enough about stabilization)

workingjubilee: The `since` argument should be either the same version as the rest, or the current nightly, I think? I'm not entirely sure either. This is an edge case of a "should've been in the original intrinsic set" thing.

Yes, T-libs-api will look at it and discuss whether or not to simply merge it or whether to FCP it as a formality or what. In general, adding or stabilizing new public API should be split from changing implementation of existing public API, unless someone says something like "let's change this one thing and then stabilize it".

sayantn: Ok, I will send a PR for these missing intrinsics. We cannot use simd_x86 feature with any version other than 1.27.0 (it doesn't compile)

sayantn: Also can we do anything about `_mm_malloc` and `_mm_free`?

workingjubilee: I don't think we should try that here.

crates/core_arch/src/x86/avx2.rs
```diff
@@ -3124,6 +3137,35 @@ pub unsafe fn _mm256_srlv_epi64(a: __m256i, count: __m256i) -> __m256i {
     transmute(psrlvq256(a.as_i64x4(), count.as_i64x4()))
 }
 
+/// Load 256-bits of integer data from memory into dst using a non-temporal memory hint. mem_addr
+/// must be aligned on a 32-byte boundary or a general-protection exception may be generated. To
+/// minimize caching, the data is flagged as non-temporal (unlikely to be used again soon)
+///
+/// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm256_stream_load_si256)
+///
+/// # Safety of non-temporal stores
```
Amanieu, line 3146: This text only applies to non-temporal stores, but this is a non-temporal load which doesn't have weak memory ordering semantics. This whole section can be removed.

crates/core_arch/src/x86/sse41.rs
```diff
@@ -1142,6 +1142,35 @@ pub unsafe fn _mm_test_mix_ones_zeros(a: __m128i, mask: __m128i) -> i32 {
     _mm_testnzc_si128(a, mask)
 }
 
+/// Load 128-bits of integer data from memory into dstt. mem_addr must be aligned on a 16-byte
```
Amanieu, line 1145: dstt -> dst

crates/core_arch/src/x86/sse41.rs
```diff
@@ -1142,6 +1142,35 @@ pub unsafe fn _mm_test_mix_ones_zeros(a: __m128i, mask: __m128i) -> i32 {
...
+/// boundary or a general-protection exception may be generated. To minimize caching, the data
+/// is flagged as non-temporal (unlikely to be used again soon)
+///
+/// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_stream_load_si128)
+///
+/// # Safety of non-temporal stores
```
Amanieu, line 1151: Again, doesn't apply to load.

crates/stdarch-verify/tests/x86-intel.rs
```diff
@@ -284,6 +273,7 @@ fn verify_all_signatures() {
                 "_mm_cvtsi64x_sd",
                 "_bextr2_u64",
                 "_mm_tzcnt_64",
+                "_mm_broadcastsi128_si256",
```
Amanieu, line 276: Can you write a quick test for this to avoid needing to add yet another exception to this list?

sayantn: I think it should be on that list - as that is just a list of all aliases. But ok, I will write a test

crates/core_arch/src/x86/sse41.rs
```diff
@@ -1936,4 +1965,14 @@ mod tests {
         let r = _mm_test_mix_ones_zeros(a, mask);
         assert_eq!(r, 0);
     }
+
+    #[simd_test(enable = "sse4.1")]
+    // Miri cannot support this until it is clear how it fits in the Rust memory model
+    // (non-temporal store)
+    #[cfg_attr(miri, ignore)]
```
Amanieu, line 1963: This is no longer needed.

crates/core_arch/src/x86/avx2.rs
```diff
@@ -5153,6 +5195,16 @@ mod tests {
         assert_eq_m256i(r, e);
     }
 
+    #[simd_test(enable = "avx2")]
+    // Miri cannot support this until it is clear how it fits in the Rust memory model
+    // (non-temporal store)
+    #[cfg_attr(miri, ignore)]
```
Amanieu, line 5205: Same here.

crates/core_arch/src/x86/avx2.rs
```diff
@@ -3124,7 +3137,25 @@ pub unsafe fn _mm256_srlv_epi64(a: __m256i, count: __m256i) -> __m256i {
     transmute(psrlvq256(a.as_i64x4(), count.as_i64x4()))
 }
 
-// TODO _mm256_stream_load_si256 (__m256i const* mem_addr)
+/// Load 256-bits of integer data from memory into dst using a non-temporal memory hint. mem_addr
+/// must be aligned on a 32-byte boundary or a general-protection exception may be generated. To
+/// minimize caching, the data is flagged as non-temporal (unlikely to be used again soon)
+///
+/// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm256_stream_load_si256)
```
RalfJung, line 3144: This (and all other AVX2 non-temporal operations) should get the same safety comment that the older non-temporal stores have. See e.g. [here](https://doc.rust-lang.org/nightly/core/arch/x86/fn._mm_stream_si128.html).

Amanieu: I checked and only non-temporal *stores* have special memory orderings on x86. x86 non-temporal loads work just like normal loads.

sayantn: @Amanieu told that that doesn't apply to streaming loads, only streaming stores.

RalfJung: Oh, I didn't realize non-temporal loads even are a thing. More nightmare waiting to happen, I guess...

ruriww:
> I checked and only non-temporal stores have special memory orderings on x86. x86 non-temporal loads work just like normal loads.

Is there a reason the non-temporal loads are written with inline assembly then? It generates extra instructions when offsetting. https://godbolt.org/z/os8bG3Efr

PR #1595, opened 2024-06-25
Remove `has_cpuid`

workingjubilee:
Implements the resolution of https://github.com/rust-lang/rust/issues/60123


workingjubilee:
The FCP isn't over yet but I figured I might as well while I was looking at stdarch for other reasons.

workingjubilee:
The FCP *is* over.

Amanieu:
We still want to avoid executing the CPUID instruction on SGX targets.

workingjubilee:
so like that?

PR #1597, opened 2024-06-25
Refactor avx512f

TDecking:
* Fused multiply-add functions have been reworked and can now be used by miri.
* Square root functions have been reworked and can now be used by miri.
* The definitions of some functions with explicit rounding have been simplified.
* Some functions now correctly use `_MM_FROUND_CUR_DIRECTION`.
* Some integer functions have been reworked and can now be used by miri.
* Some missing intrinsics were added.
* Masked integer comparisons now properly use the mask registers.
* Some documentation issues were fixed.


TDecking:
@sayantn Do you mind if I open this? There is a bit of an overlap between this PR and yours, but some things in this PR is not present in your work.

sayantn:
So the only overlap is fma intrinsics and masked loads? I have no problem with you implementing the fma (honestly I didn't know about the `simd_fma` intrinsic. But I think right now you should not do the masked load/stores. `simd_masked_load` aligns with only the _element_'s alignment, so it will never generate the aligned load instructions. See rust-lang/rust#126919. Also, typically in stdarch it is preferred to link with llvm and use the simd intrinsics using the `core::simd` types instead of the `__m128i` etc. I will remove the fma enhancements from my PR, it will remain draft for some time.

TDecking:
@sayantn I've removed the masked load changes on my end. Our PRs should now be orthogonal to each other.

sayantn:
Yes I will modify my PR in a while.

sayantn:
can you also please do the floating-point abs using `simd_fabs`

TDecking:
@sayantn done.

sayantn:
Thanks.

sayantn:
I have already done the remaining gather-scatter in avx512f and reduce ops in avx512bw

PR #1601, opened 2024-06-30
Update Intel SDE

sayantn:
Update Intel SDE to v9.33.0 (The latest v9.38.0 is buggy in Linux)

Disabled `assert_instr` in the emulated run. Some interaction between `pinkit` and `objdump` results in an `SIGSEGV`. We still have `assert_instr` in the other runs - which is more than enough verification.

Added support for `SSE4a` and `TBM` in the emulated test, for future also enabled `XOP`. Now the emulated one supports all x86 targets features supported by rust.


crates/core_arch/src/x86_64/xsave.rs
```diff
@@ -201,6 +202,7 @@ mod tests {
         xsave::_xsaveopt64(b.ptr(), m);
         assert_eq!(a, b);
     }
+    */
```
Amanieu, line 205: What is the problem with these? I don't want to leave commented-out code in the repo, it's a maintenance nightmare. If it needs to be removed then it should be completely removed.

sayantn: All 4 of these were commented out because as #209 points out, these are privileged instructions, so we cannot test these without access to the kernel. I found out that one of these is actually not a privileged instruction, so I uncommented it.

sayantn: I have now conditionally ignored them only for SDE, as `xsaveopt` and `xsave` is not well-supported by SDE, but are not privileged, so can be run in the other workflows. the `xsaves` etc intrinsics are privileged, and cannot be tested

PR #1602, opened 2024-06-30
Refactor avx512bw

TDecking:
* The following operations have been rewritten and can now be consumed by miri:
  * Saturating arithmetic
  * Comparison operations
  * Maximum/Minimum
  * Byte average
  * High multiplication
  * Integer abs
* Masked comparison operations now properly use the mask registers
* All missing intrinsics are added


crates/core_arch/src/x86/avx512bw.rs
```diff
@@ -9263,104 +10113,310 @@ pub unsafe fn _kand_mask64(a: __mmask64, b: __mmask64) -> __mmask64 {
...
 
-/// Compute the bitwise NOT of 32-bit mask a, and store the result in k.
+/// Compute the bitwise NOT of 32-bit mask a, and store the result in k.
+///
+/// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_knot_mask32&expand=3234)
+#[inline]
+#[target_feature(enable = "avx512bw")]
+#[unstable(feature = "stdarch_x86_avx512", issue = "111137")]
+pub unsafe fn _knot_mask32(a: __mmask32) -> __mmask32 {
+    a ^ 0b11111111_11111111_11111111_11111111
```
Amanieu, line 10123: This would be much more readable as `!a`.

crates/core_arch/src/x86/avx512bw.rs
```diff
@@ -9263,104 +10113,310 @@ pub unsafe fn _kand_mask64(a: __mmask64, b: __mmask64) -> __mmask64 {
...
+}
+
+/// Compute the bitwise NOT of 64-bit mask a, and store the result in k.
+///
+/// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_knot_mask64&expand=3235)
+#[inline]
+#[target_feature(enable = "avx512bw")]
+#[unstable(feature = "stdarch_x86_avx512", issue = "111137")]
+pub unsafe fn _knot_mask64(a: __mmask64) -> __mmask64 {
+    a ^ 0b11111111_11111111_11111111_11111111_11111111_11111111_11111111_11111111
```
Amanieu, line 10133: Same here.

Amanieu (commented):
LGTM, I will merge once conflicts are resolved.

crates/core_arch/src/x86/avx512bw.rs
TDecking: Fixed. It might be worth pointing out that `_knot_mask16` in `avx512f.rs` and `_knot_mask8` in `avx512dq.rs` are written in the same way and should be changed in the future.

crates/core_arch/src/x86/avx512bw.rs
TDecking: Fixed.

PR #1603, opened 2024-07-01
Fix incorrect reduction operations in avx512f

TDecking:
Some of the masked reduction operations had bogus values for the masked out elements.


crates/core_arch/src/x86/avx512f.rs
```diff
@@ -31718,7 +31714,7 @@ pub unsafe fn _mm512_mask_reduce_min_epi32(k: __mmask16, a: __m512i) -> i32 {
     simd_reduce_min(simd_select_bitmask(
         k,
         a.as_i32x16(),
-        _mm512_undefined_epi32().as_i32x16(),
+        i32x16::splat(0x7fffffff),
```
Amanieu, line 31717: Use `i32::MAX` here, it's clearer. Same for the others.

TDecking: Done.

PR #1605, opened 2024-07-02
Implement AVX512_FP16

sayantn:
This PR adds the AVX512_FP16 intrinsics in Rust. These intrinsics will be behind the feature gate `#[feature(stdarch_x86_avx512_f16)]` (rust-lang/rust#127213).

### Progress:
 - [x] Set intrinsics
 - [x] Arithmetic operations (Add-Sub-Mul-Div on Real numbers)
 - [x] FMA
 - [x] Cast
 - [x] Compare
 - [x] Convert
 - [x] Math (Reciprocal, Sqrt, RSqrt, Max, Min, Reduce)
 - [x] FP-Support (FPClass, `getexp`, `getmant`, `scale`)
 - [x] Load-Store
 - [x] Complex number operations (mul-cmul-conj on Complex numbers)
 - [x] AVX-NE-CONVERT

This also adds some missing inlining in `avx512ifma` and updates the `x86-intel.xml` file to v3.6.9

The `set1_pch` intrinsics were not implemented due to a lack of complex number type.
`cmpph` and `fpclassph` intrinsics use inline asm because of no `i1` support yet.


tgross35:
Have you run into any weird behavior with these, or do things seem to be working smoothly? (ignoring the ABI issue for system function calls, that is)

sayantn:
No problems yet, just that `simd_fabs` doesn't accept a `f16` argument, so i will just use an and operation. I am actively avoiding doing f16 operations in rust, but that's not a blocker for sure.

sayantn:
cc @tgross35 @beetrees

crates/core_arch/src/simd.rs
```diff
@@ -186,9 +187,20 @@ simd_ty!(
 simd_ty!(i32x4[i32]: x0, x1, x2, x3);
 simd_ty!(i64x2[i64]: x0, x1);
 
+simd_ty!(
+    #[unstable(feature = "f16", issue = "116909")]
```
Amanieu, line 191: We don't need stability attributes for these since they are private types and are only used internally.

crates/core_arch/src/x86/test.rs
```diff
@@ -36,6 +36,16 @@ pub unsafe fn get_m128(a: __m128, idx: usize) -> f32 {
     transmute::<_, [f32; 4]>(a)[idx]
 }
 
+#[track_caller]
+#[target_feature(enable = "avx512fp16")]
+#[unstable(feature = "stdarch_x86_avx512_f16", issue = "127213")]
```
Amanieu, line 41: Stability attributes are not needed for internal functions, only public API.

PR #1606, opened 2024-07-07
Use generic simd in wasm intrinsics

TDecking:


CryZe:
Are there any tests to ensure this doesn't degrade the quality?

TDecking:
`assert_instr` checks for the presence of a given instruction.

PR #1607, opened 2024-07-07
Implement missing in `sse4a` and `tbm` intrinsics

sayantn:
The `extracti`, `inserti` and `bextri` intrinsics were implemented. They were left out because of #248. This also tidies up the `tbm` module and divides it into x86 and x86_64 specific modules. This completes all the `SSE4a` and `TBM` intrinsics. These are under rust-lang/rust#126936 `#[feature(simd_x86_updates)]`

### DIFF
--- a/crates/core_arch/src/x86/sse4a.rs
+++ b/crates/core_arch/src/x86/sse4a.rs
@@ -9,16 +9,17 @@ use stdarch_test::assert_instr;
 extern "C" {
     #[link_name = "llvm.x86.sse4a.extrq"]
     fn extrq(x: i64x2, y: i8x16) -> i64x2;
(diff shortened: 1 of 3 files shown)


crates/core_arch/src/x86/sse4a.rs
```diff
@@ -39,6 +40,27 @@ pub unsafe fn _mm_extract_si64(x: __m128i, y: __m128i) -> __m128i {
     transmute(extrq(x.as_i64x2(), y.as_i8x16()))
 }
 
+/// Extracts the specified bits from the lower 64 bits of the 128-bit integer vector operad at the
```
Amanieu, line 43:
```suggestion
/// Extracts the specified bits from the lower 64 bits of the 128-bit integer vector operand at the
```

PR #1609, opened 2024-07-11
Add missing `#[target_feature(enable = "simd128")]`

daxpedda:
I believe that adding these was missed in #874 and adding this now should not be a breaking change.

Cc @alexcrichton.

### DIFF
--- a/crates/core_arch/src/wasm32/simd128.rs
+++ b/crates/core_arch/src/wasm32/simd128.rs
@@ -660,6 +660,7 @@ pub unsafe fn v128_store64_lane<const L: usize>(v: v128, m: *mut u64) {
         a15 = 15,
     )
 )]
+#[target_feature(enable = "simd128")]
 #[doc(alias("v128.const"))]
 #[stable(feature = "wasm_simd", since = "1.54.0")]
 #[rustc_const_stable(feature = "wasm_simd", since = "1.54.0")]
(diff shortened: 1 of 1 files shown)


crates/core_arch/src/wasm32/simd128.rs
```diff
@@ -33,6 +33,7 @@ types! {
     /// type in WebAssembly. Operations on `v128` can only be performed with the
     /// functions in this module.
     // N.B., internals here are arbitrary.
+    #[target_feature(enable = "simd128")]
```
daxpedda, line 36: Right, this doesn't work on types.
I guess we also don't want to add `cfg(target_feature)` here because otherwise using it would require `build-std`.

Amanieu:
I believe these were intentionally omitted to allow using SIMD type constructors in a const context.

Amanieu:
And in the end, just using the SIMD types doesn't require the feature to be available, the compiler will just fall back to treating it as an array of scalars.

alexcrichton:
Ah yes at the time this wasn't possible due to the usage of `const`, but if the compiler now allows `#[target_feature]` on `const` functions I think this is ok to land. I also agree with @Amanieu though that these should be optional and not required.

daxpedda:
> I also agree with @Amanieu though that these should be optional and not required.

Does `#[target_feature(enable = "simd128")]` cause LLVM to generate SIMD128 instructions here even without using `-Ctarget-feature=simd128`?

alexcrichton:
Right yeah, the `-C` flag applies to all functions within a crate and `#[target_feature]` only applies to a single function, and LLVM only needs it enabled for the function-at-hand.

daxpedda:
I see, thanks!
Will go ahead and close this then.

PR #1610, opened 2024-07-11
wasm32: Add `simd128` to enabled features for relaxed intrinsics

alexcrichton:
It looks like LLVM requires that `simd128` is active to use these intrinsics and `relaxed-simd` isn't implicitly enabling them. This is probably something to fix at the LLVM layer as well but for now enable both the `simd128` feature as well as the `relaxed-simd` feature to fix things on our side.

This is done to help fix the test failure in #117468


alexcrichton:
I filed https://github.com/llvm/llvm-project/issues/98502 for a possible upstream LLVM issue

PR #1612, opened 2024-07-14
Revert "wasm32: Add `simd128` to enabled features for relaxed intrinsics"

daxpedda:
This reverts https://github.com/rust-lang/stdarch/pull/1610.
The idea was to use this as a workaround until LLVM addresses it (see https://github.com/llvm/llvm-project/issues/98502).
It turned out that LLVM is likely to see this as something that should be fixed in the frontend, which has now been done in https://github.com/rust-lang/rust/pull/117468.

Cc @alexcrichton.


aheejin:
FYI, I fixed it in the LLVM backend: https://github.com/llvm/llvm-project/pull/99803

PR #1618, opened 2024-07-23
initial commit to enable amx

ziyizhang-1:
As mentioned in [x86_amx_intrinsics](https://github.com/rust-lang/rust/issues/126622)

Completed AMX Intrinsics:
`amx-tile`
- _tile_loadconfig
- _tile_storeconfig
- _tile_loadd
- _tile_release
- _tile_stored
- _tile_stream_loadd
- _tile_zero
`amx-int8`
- _tile_dpbssd
- _tile_dpbsud
- _tile_dpbusd
- _tile_dpbuud
`amx-bf16`
- _tile_dpbf16ps
`amx-fp16`
- _tile_dpfp16ps
`amx-complex`
- _tile_cmmimfp16ps
- _tile_cmmrlfp16ps

### DIFF
--- a/crates/core_arch/Cargo.toml
+++ b/crates/core_arch/Cargo.toml
@@ -23,3 +23,6 @@ maintenance = { status = "experimental" }
 [dev-dependencies]
 stdarch-test = { version = "0.*", path = "../stdarch-test" }
 std_detect = { version = "0.*", path = "../std_detect" }
+
+[target.'cfg(all(target_arch = "x86_64", target_os = "linux"))'.dev-dependencies]
+syscalls = { version = "0.6.18", default-features = false }
--- a/crates/core_arch/src/x86_64/amx.rs
+++ b/crates/core_arch/src/x86_64/amx.rs
@@ -1,0 +1,604 @@
+/// Load tile configuration from a 64-byte memory location specified by mem_addr.
+/// The tile configuration format is specified below, and includes the tile type pallette,
+/// the number of bytes per row, and the number of rows. If the specified pallette_id is zero,
+/// that signifies the init state for both the tile config and the tile data, and the tiles are zeroed.
+/// Any invalid configurations will result in #GP fault.
+///
+/// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_tile_loadconfig&ig_expand=6875)
+#[inline]
+#[target_feature(enable = "amx-tile")]
+#[unstable(feature = "x86_amx_intrinsics", issue = "126622")]
+pub unsafe fn _tile_loadconfig(mem_addr: *const u8) {
+    ldtilecfg(mem_addr);
+}
+
+/// Stores the current tile configuration to a 64-byte memory location specified by mem_addr.
+/// The tile configuration format is specified below, and includes the tile type pallette,
+/// the number of bytes per row, and the number of rows. If tiles are not configured, all zeroes will be stored to memory.
+///
+/// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_tile_storeconfig&ig_expand=6879)
+#[inline]
+#[target_feature(enable = "amx-tile")]
+#[unstable(feature = "x86_amx_intrinsics", issue = "126622")]
+pub unsafe fn _tile_storeconfig(mem_addr: *mut u8) {
+    sttilecfg(mem_addr);
+}
+
+/// Load tile rows from memory specifieid by base address and stride into destination tile dst using the tile configuration previously configured via _tile_loadconfig.
+///
+/// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_tile_loadd&ig_expand=6877)
+#[inline]
+#[rustc_legacy_const_generics(0)]
+#[target_feature(enable = "amx-tile")]
+#[unstable(feature = "x86_amx_intrinsics", issue = "126622")]
+pub unsafe fn _tile_loadd<const DST: i32>(base: *const u8, stride: usize) {
+    static_assert_uimm_bits!(DST, 3);
+    tileloadd64(DST as i8, base, stride);
+}
+
+/// Release the tile configuration to return to the init state, which releases all storage it currently holds.
+///
+/// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_tile_release&ig_expand=6878)
+#[inline]
+#[target_feature(enable = "amx-tile")]
+#[unstable(feature = "x86_amx_intrinsics", issue = "126622")]
+pub unsafe fn _tile_release() {
+    tilerelease();
+}
+
+/// Store the tile specified by src to memory specifieid by base address and stride using the tile configuration previously configured via _tile_loadconfig.
+///
+/// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_tile_stored&ig_expand=6881)
+#[inline]
+#[rustc_legacy_const_generics(0)]
+#[target_feature(enable = "amx-tile")]
+#[unstable(feature = "x86_amx_intrinsics", issue = "126622")]
+pub unsafe fn _tile_stored<const DST: i32>(base: *mut u8, stride: usize) {
+    static_assert_uimm_bits!(DST, 3);
+    tilestored64(DST as i8, base, stride);
+}
+
+/// Load tile rows from memory specifieid by base address and stride into destination tile dst using the tile configuration
+/// previously configured via _tile_loadconfig. This intrinsic provides a hint to the implementation that the data will
+/// likely not be reused in the near future and the data caching can be optimized accordingly.
+///
+/// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_tile_stream_loadd&ig_expand=6883)
+#[inline]
+#[rustc_legacy_const_generics(0)]
+#[target_feature(enable = "amx-tile")]
+#[unstable(feature = "x86_amx_intrinsics", issue = "126622")]
+pub unsafe fn _tile_stream_loadd<const DST: i32>(base: *const u8, stride: usize) {
+    static_assert_uimm_bits!(DST, 3);
+    tileloaddt164(DST as i8, base, stride);
+}
+
+/// Zero the tile specified by tdest.
+///
+/// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_tile_zero&ig_expand=6885)
+#[inline]
+#[rustc_legacy_const_generics(0)]
+#[target_feature(enable = "amx-tile")]
+#[unstable(feature = "x86_amx_intrinsics", issue = "126622")]
+pub unsafe fn _tile_zero<const DST: i32>() {
+    static_assert_uimm_bits!(DST, 3);
+    tilezero(DST as i8);
+}
+
+/// Compute dot-product of BF16 (16-bit) floating-point pairs in tiles a and b,
+/// accumulating the intermediate single-precision (32-bit) floating-point elements
+/// with elements in dst, and store the 32-bit result back to tile dst.
+///
+/// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_tile_dpbf16ps&ig_expand=6864)
+#[inline]
+#[rustc_legacy_const_generics(0, 1, 2)]
+#[target_feature(enable = "amx-bf16")]
+#[unstable(feature = "x86_amx_intrinsics", issue = "126622")]
+pub unsafe fn _tile_dpbf16ps<const DST: i32, const A: i32, const B: i32>() {
+    static_assert_uimm_bits!(DST, 3);
+    static_assert_uimm_bits!(A, 3);
+    static_assert_uimm_bits!(B, 3);
+    tdpbf16ps(DST as i8, A as i8, B as i8);
+}
+
+/// Compute dot-product of bytes in tiles with a source/destination accumulator.
+/// Multiply groups of 4 adjacent pairs of signed 8-bit integers in a with corresponding
+/// signed 8-bit integers in b, producing 4 intermediate 32-bit results.
+/// Sum these 4 results with the corresponding 32-bit integer in dst, and store the 32-bit result back to tile dst.
+///
+/// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_tile_dpbssd&ig_expand=6866)
+#[inline]
+#[rustc_legacy_const_generics(0, 1, 2)]
+#[target_feature(enable = "amx-int8")]
+#[unstable(feature = "x86_amx_intrinsics", issue = "126622")]
+pub unsafe fn _tile_dpbssd<const DST: i32, const A: i32, const B: i32>() {
+    static_assert_uimm_bits!(DST, 3);
+    static_assert_uimm_bits!(A, 3);
+    static_assert_uimm_bits!(B, 3);
+    tdpbssd(DST as i8, A as i8, B as i8);
+}
+
+/// Compute dot-product of bytes in tiles with a source/destination accumulator.
+/// Multiply groups of 4 adjacent pairs of signed 8-bit integers in a with corresponding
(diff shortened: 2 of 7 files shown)


crates/core_arch/src/x86_64/mod.rs
```diff
@@ -3,6 +3,40 @@
...
+            palette: 0,
+            start_row: 0,
+            reserved_a0: [0; 14],
+            colsb: [0; 8],
+            reserved_b0: [0; 8],
+            rows: [0; 8],
+            reserved_c0: [0; 8],
+        }
+    }
+}
```
sayantn, line 38: Can't this be `Derive`d?

crates/stdarch-test/src/lib.rs
```diff
@@ -201,3 +201,42 @@ pub fn assert_skip_test_ok(name: &str, missing_features: &[&str]) {
 
 // See comment in `assert-instr-macro` crate for why this exists
 pub static mut _DONT_DEDUP: *const u8 = std::ptr::null();
+
+#[cfg(target_arch = "x86_64")]
+#[inline]
+pub unsafe fn _init_amx() {
```
sayantn, line 207: These syscalls are required only on the linux kernel, and you can use the `syscall` crate as a `dev-dependency`. Windows by default enables amx support. Also, this should be in `core_arch/src/x86/test.rs`

crates/core_arch/src/x86_64/amxtile.rs
```diff
@@ -0,0 +1,209 @@
...
+/// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_tile_storeconfig&ig_expand=6879)
+#[inline]
+#[target_feature(enable = "amx-tile")]
+#[unstable(feature = "x86_amx_intrinsics", issue = "126622")]
+pub unsafe fn _tile_storeconfig(mem_addr: *mut __tilecfg) {
+    asm!(
+        "sttilecfg [{p}]",
+        p = in(reg) mem_addr,
+        options(nostack, preserves_flags),
+    );
```
sayantn, line 34: You can use `llvm.x86.sttilecfg` here

crates/core_arch/src/x86_64/amxtile.rs
```diff
@@ -0,0 +1,209 @@
...
+/// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_tile_loadconfig&ig_expand=6875)
+#[inline]
+#[target_feature(enable = "amx-tile")]
+#[unstable(feature = "x86_amx_intrinsics", issue = "126622")]
+pub unsafe fn _tile_loadconfig(mem_addr: *const __tilecfg) {
+    asm!(
+        "ldtilecfg [{p}]",
+        p = in(reg) mem_addr,
+        options(nostack, preserves_flags),
+    );
```
sayantn, line 18: `llvm.x86.ldtilecfg` here

sayantn:
Can you please run `cargo test -p stdarch-verify` with the `PRINT_MISSING_LISTS_MARKDOWN` constant set to `true`? This would update the `missing-x86.md` file to reflect the new added intrinsics. Also, please do not edit the `x86-intel.xml` file manually. Download the latest from Intel's website (I think v3.6.9 currently) and replace this

sayantn (changes requested):

crates/core_arch/src/x86_64/mod.rs
ziyizhang-1: Thx for the advice, implement several basic traits for __tilecfg via `derive`

crates/stdarch-test/src/lib.rs
ziyizhang-1: Since only a few testcases of amx intrinsics require this function and probably not expose to users, so maybe it is too expensive to introduce the syscall crate

crates/core_arch/src/x86_64/amxtile.rs
ziyizhang-1: replaced~

sayantn:
Even though windows doesn't need the syscall, you still need to have the function to compile. Maybe add an empty `init_amx` function with `#[cfg(all(target_arch = "x86_64", windows))]`

ziyizhang-1:
> Even though windows doesn't need the syscall, you still need to have the function to compile. Maybe add an empty `init_amx` function with `#[cfg(all(target_arch = "x86_64", windows))]`

yep, that's a good workaround to prevent repeated condition check

ziyizhang-1:
> Can you please run `cargo test -p stdarch-verify` with the `PRINT_MISSING_LISTS_MARKDOWN` constant set to `true`? This would update the `missing-x86.md` file to reflect the new added intrinsics. Also, please do not edit the `x86-intel.xml` file manually. Download the latest from Intel's website (I think v3.6.9 currently) and replace this.
> 
> Also, there are very few amx intrinsics, so they can be implemented in a single `amx.rs` file instead of 5 files

updated `x86-intel.xml` to v3.6.9, please help to review the change

crates/core_arch/src/x86_64/mod.rs
```diff
@@ -3,6 +3,26 @@
...
+    /// number of bytes of one row in each tile
+    pub colsb: [u16; 8],
+    /// reserved, must be zero
+    reserved_b0: [u16; 8],
+    /// number of rows in each tile
+    pub rows: [u8; 8],
+    /// reserved, must be zero
+    reserved_c0: [u8; 8],
+}
+
```
sayantn, line 25: Can you add a method in `__tilecfg` that takes the `pub` fields as input and returns the correctly configured cfg (as the `reserved` fields are private, this struct can't be constructed outside `core_arch`)?

crates/stdarch-verify/tests/x86-intel.rs
```diff
@@ -711,10 +713,10 @@ fn equate(
     match (t, &intel[..]) {
         (&Type::PrimFloat(32), "float") => {}
         (&Type::PrimFloat(64), "double") => {}
-        (&Type::PrimSigned(8), "__int8" | "char") => {}
+        (&Type::PrimSigned(8), "__int8" | "constexpr int" | "char") => {}
         (&Type::PrimSigned(16), "__int16" | "short") => {}
-        (&Type::PrimSigned(32), "__int32" | "const int" | "int") => {}
-        (&Type::PrimSigned(64), "__int64" | "long long") => {}
+        (&Type::PrimSigned(32), "__int32" | "const int" | "int" | "size_t") => {}
```
sayantn, line 718: This is not needed, you are using `i64` as `size_t`

crates/core_arch/src/x86/test.rs
```diff
@@ -139,3 +139,47 @@ pub unsafe fn assert_eq_m512d(a: __m512d, b: __m512d) {
...
+                );
+                if ret != 0 {
+                    panic!("Failed to enable AMX");
+                }
+            }
+            3 => {}
+            _ => unreachable!(),
+        }
+    }
+}
```
sayantn, line 185: See [Intel's sample code](https://github.com/intel/AMX-TMUL-Code-Samples/blob/main/src%2Ftest-amxtile.c#L85-L99). They use only one syscall to directly set xtiledata without checking for xtilecfg (as xtiledata requires xtilecfg, that call would simply fail if xtilecfg is not enabled). Also, as you are only using it in one file (`x86_64/amx.rs`), it makes sense to move these functions to the `tests` module of amx.rs

sayantn (changes requested):

sayantn:
The `x86_64-unknown-linux-gnu-emulated` CI runs all the tests (using Intel SDE), so your implementation amx-complex and amx-fp16 has been tested

crates/core_arch/src/x86_64/mod.rs
Amanieu: The GCC and Clang intrinsics don't define a `__tilecfg` type and just use `void *`. We shouldn't be inventing our own type here.

crates/core_arch/src/x86/test.rs
```diff
@@ -139,3 +139,47 @@ pub unsafe fn assert_eq_m512d(a: __m512d, b: __m512d) {
         panic!("{:?} != {:?}", a, b);
     }
 }
+
+#[cfg(all(target_arch = "x86_64", windows))]
+#[target_feature(enable = "amx-tile")]
+pub(crate) fn _init_amx() {}
+
+#[cfg(all(target_arch = "x86_64", not(windows)))]
```
Amanieu, line 147: This should specifically be `cfg(target_os = "linux")`. The syscalls are specific to Linux and won't work on other OSes like BSD.

crates/stdarch-test/src/lib.rs
Amanieu: It's fine to add dev-dependencies. They are only used when testing this crate in CI.

crates/core_arch/src/x86_64/mod.rs
ziyizhang-1: Oh, thx for the notification, I did not notice the function signature of the ldtilecfg and sttilecfg in llvm which can be found [here](https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/IR/IntrinsicsX86.td). So I will make the `ldtilecfg` to accept `*const i8` instead of `*const __tilecfg` and similar to `sttilecfg`

crates/stdarch-verify/tests/x86-intel.rs
ziyizhang-1: I will remove that from 32-bit as I do not think amx can be supported on those processors

sayantn: yes, amx is x64-only

crates/core_arch/src/x86/test.rs
ziyizhang-1: The first syscall is arch_prctl(ARCH_GET_XCOM_PERM) which is a read-only operation and has 2~3x less overhead compared with the latter syscall arch_prctl(ARCH_REQ_XCOMP_PERM), which might execute a write operation. The `xtiledata` flag only needs to be set once for a process during its lifetime. So we can apply the first "read" syscall to check if there is a need to follow up another "request" syscall in case the xtiledata has already been turned on.

crates/core_arch/src/x86_64/mod.rs
ziyizhang-1: I think the __tilecfg is necessary to call amx intrinsics, so if we do not define it here, users might be forced to do so. How about we provide a usable definition here? Since the interface accepts `*const i8`, it is also okay for users to pass their self-defined struct to the function.

crates/stdarch-verify/tests/x86-intel.rs
```diff
@@ -711,10 +713,10 @@ fn equate(
     match (t, &intel[..]) {
         (&Type::PrimFloat(32), "float") => {}
         (&Type::PrimFloat(64), "double") => {}
-        (&Type::PrimSigned(8), "__int8" | "char") => {}
+        (&Type::PrimSigned(8), "__int8" | "constexpr int" | "char") => {}
         (&Type::PrimSigned(16), "__int16" | "short") => {}
         (&Type::PrimSigned(32), "__int32" | "const int" | "int") => {}
-        (&Type::PrimSigned(64), "__int64" | "long long") => {}
+        (&Type::PrimSigned(64), "__int64" | "long long" | "size_t") => {}
```
sayantn, line 719: `size_t` should be an unsigned type, preferably `usize` (you can define the LLVM intrinsic signature with `stride` as `usize`, LLVM doesn't care about the signedness or type of integer arguments, just their bit width)

crates/core_arch/Cargo.toml
```diff
@@ -23,3 +23,6 @@ maintenance = { status = "experimental" }
 [dev-dependencies]
 stdarch-test = { version = "0.*", path = "../stdarch-test" }
 std_detect = { version = "0.*", path = "../std_detect" }
+
+[target.'cfg(target_arch = "x86_64")'.dev-dependencies]
+syscalls = { version = "0.*", default-features = false }
```
sayantn, line 28:
```suggestion
[target.'cfg(all(target_arch = "x86_64", target_os = "linux"))'.dev-dependencies]
syscalls = { version = "0.6.18", default-features = false }
```

crates/stdarch-verify/tests/x86-intel.rs
ziyizhang-1: Added `usize` as a primitive in x86-intel, please check if that what you expect

crates/core_arch/src/x86_64/amx.rs
```diff
@@ -0,0 +1,576 @@
...
+
+/// Store the tile specified by src to memory specifieid by base address and stride using the tile configuration previously configured via _tile_loadconfig.
+///
+/// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_tile_stored&ig_expand=6881)
+#[inline]
+#[allow(non_upper_case_globals)]
+#[rustc_legacy_const_generics(0)]
+#[target_feature(enable = "amx-tile")]
+#[unstable(feature = "x86_amx_intrinsics", issue = "126622")]
+pub unsafe fn _tile_stored<const dst: i8>(base: *mut i8, stride: usize) {
```
sayantn, line 58: Normally in Rust, const generics are named in all-caps

crates/core_arch/src/x86_64/amx.rs
```diff
@@ -0,0 +1,576 @@
...
+#[unstable(feature = "x86_amx_intrinsics", issue = "126622")]
+pub unsafe fn _tile_cmmrlfp16ps<const dst: i8, const A: i8, const B: i8>() {
+    static_assert_uimm_bits!(dst, 3);
+    static_assert_uimm_bits!(A, 3);
+    static_assert_uimm_bits!(B, 3);
+    tcmmrlfp16ps(dst, A, B);
+}
+
+#[allow(improper_ctypes)]
+#[allow(clashing_extern_declarations)]
```
sayantn, line 240: Why is this lint turned off? Seems like stagnant code

crates/core_arch/src/x86_64/amx.rs
ziyizhang-1: renamed

crates/core_arch/src/x86_64/amx.rs
ziyizhang-1: Good question, those are redundant, removed

sayantn (approved):

sayantn:
@Amanieu should we keep the `__tilecfg` type? It indeed makes for a nice abstraction (all C programs that use amx have to define a `struct` or use an array because C lacks this type)

Amanieu:
I don't think we should be inventing our own API in this crate, we should strictly follow the vendor's API. What if Intel decides to add its own `__tilecfg` type in the future that is incompatible with ours? I think it is best to leave this to an external crate. That crate can also deal with other AMX-related issues such as the syscalls needed for initialization.

crates/core_arch/src/x86_64/amx.rs
```diff
@@ -0,0 +1,563 @@
...
+/// The tile configuration format is specified below, and includes the tile type pallette,
+/// the number of bytes per row, and the number of rows. If the specified pallette_id is zero,
+/// that signifies the init state for both the tile config and the tile data, and the tiles are zeroed.
+/// Any invalid configurations will result in #GP fault.
+///
+/// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_tile_loadconfig&ig_expand=6875)
+#[inline]
+#[target_feature(enable = "amx-tile")]
+#[unstable(feature = "x86_amx_intrinsics", issue = "126622")]
+pub unsafe fn _tile_loadconfig(mem_addr: *const i8) {
```
Amanieu, line 11: For void pointers in Rust we generally use `*mut u8`. This is what the allocator API uses for example. At the LLVM level there is no distinction between `i8` and `u8` which is why LLVM intrinsics are always show taking `i8`.

ziyizhang-1: make sense, let's make it more idiomatic

ziyizhang-1:
I understood the concern raised by @Amanieu; there is indeed some probability that Intel will change the definition of the tile configuration in the future, for example, increasing/decreasing the number of tiles, which can lead to a change of the config layout. Anyway, I just moved `__tilecfg` to the tests mod, and considering the current feature `x86_amx_intrinsics` is still unstable, any different points of view can be welcomed and discussed.

Amanieu:
Everything else looks good, so this can be merged once the `i32` issue is fixed.

PR #1622, opened 2024-08-02
Add TBI helpers for AArch64

ghost:
Adds a trait to `core_arch` for AArch64 to set a value in the top byte of a pointer (which should logically become the canonical address of the allocation), and to retrieve the value stored in the top byte of a pointer, if any.

The intent is for a Rust-provided mechanism for such adjustments to pointers, so that any internal pointer tracking could be updated as necessary, such as to enable this pattern within Miri.

### DIFF
--- a/crates/core_arch/src/aarch64/tbi.rs
+++ b/crates/core_arch/src/aarch64/tbi.rs
@@ -1,0 +1,229 @@
+extern crate alloc;
+use crate::arch::asm;
+use alloc::boxed::Box;
+use core::convert::{AsMut, AsRef};
+use core::ops::{Deref, DerefMut, Drop};
+
+/// A TBI-enabled [Box].
+///
+/// This allows setting the top byte to arbitrary values via top-byte ignore (TBI).
+#[derive(Debug)]
+#[unstable(feature = "stdarch_aarch64_tbi", issue = "none")]
+pub struct TBIBox<T>(
+    Option<Box<T>>,
+    /* original pointer for deallocation: */ *mut T,
+    // Without keeping a copy of the top byte around, it is difficult to cleanly extract the top byte from the
+    // underlying pointer on-demand since `from_raw` requires consuming `self.0`.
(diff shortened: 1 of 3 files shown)


crates/core_arch/src/aarch64/tbi.rs
```diff
@@ -0,0 +1,229 @@
...
+            original_ptr.map_addr(|addr| (addr & 0x00ffffffffffffff) | ((top_byte as usize) << 56));
+        unsafe {
+            asm!(
+                "/* pretend that we're memcpy'ing from {original_ptr} to {ptr}... */",
+                original_ptr = in(reg) original_ptr,
+                ptr = in(reg) ptr
+            )
+        };
+        // Reconstruct the `Box` using the address with the new top byte and return that, wrapped as a TBIBox
+        Self(Some(unsafe { Box::from_raw(ptr) }), original_ptr, top_byte)
```
RalfJung, line 36: This doesn't hide things sufficiently well from the compiler. The compiler can still see that `ptr` is just the same as `original_ptr` with an offset applied -- making this still UB.

You need the computation of `ptr` itself to occur in a "black box".

mrkajetanp: Ah yes I see, maybe I could move the step that actually sets the top byte i.e. `addr | top_byte_shifted` into the inline asm block and do it there with an explicit `ORR`, do you think that would work?

RalfJung: Yeah that would be better.

mrkajetanp:
Because the TBIBox interacts with Box and does allocations, it does not actually fit into stdarch - this will be moved into `alloc`, though there'll be an RFC coming first.

adamgemmell:
Though the [work on MTE has stalled](https://github.com/rust-lang/rfcs/pull/3700) recently, I don't think this PR fits into our future approach.

PR #1626, opened 2024-08-07
NVPTX: Add f16 SIMD intrinsics

kjetilkjeka:
This implements parts of the `f16x2` intrinsics as discussed in https://github.com/rust-lang/rust/issues/125440#issuecomment-2128911257

It's unfortunately not possible to test that these intrinsics produce the correct instructions with the current test infrastructure. This was discussed on [zulip](https://rust-lang.zulipchat.com/#narrow/stream/208962-t-libs.2Fstdarch/topic/Running.20.23.5Bassert_instr.5D.20on.20a.20no_std.20target.20.28nvptx.29)

I have created assembly tests in my own branch of `rust-lang` repo that verifies the correct instructions are used. I have added `#[assert_instr]` attributes nonetheless. I'm not sure if it's desirable to follow up on `rust-lang` with a PR including the tests or not?

### DIFF
--- a/crates/core_arch/src/nvptx/packed.rs
+++ b/crates/core_arch/src/nvptx/packed.rs
@@ -1,0 +1,143 @@
+//! NVPTX Packed data types (SIMD)
+//!
+//! Packed Data Types is what PTX calls SIMD types. See [PTX ISA (Packed Data Types)](https://docs.nvidia.com/cuda/parallel-thread-execution/#packed-data-types) for a full reference.
+
+// Note: #[assert_instr] tests are not actually being run on nvptx due to being a `no_std` target incapable of running tests. Something like FileCheck would be appropriate for verifying the correct instruction is used.
+
+use crate::intrinsics::simd::*;
+
+#[allow(improper_ctypes)]
+extern "C" {
+    #[link_name = "llvm.minnum.v2f16"]
+    fn llvm_f16x2_minnum(a: f16x2, b: f16x2) -> f16x2;
+    #[link_name = "llvm.minimum.v2f16"]
+    fn llvm_f16x2_minimum(a: f16x2, b: f16x2) -> f16x2;
+    #[link_name = "llvm.maxnum.v2f16"]
+    fn llvm_f16x2_maxnum(a: f16x2, b: f16x2) -> f16x2;
+    #[link_name = "llvm.maximum.v2f16"]
+    fn llvm_f16x2_maximum(a: f16x2, b: f16x2) -> f16x2;
+}
+
+types! {
+    #![unstable(feature = "stdarch_nvptx", issue = "111199")]
+
+    /// PTX-specific 32-bit wide floating point (f16 x 2) vector type
+    pub struct f16x2(2 x f16);
+
+}
+
+/// Add two values, round to nearest even
+///
+/// <https://docs.nvidia.com/cuda/parallel-thread-execution/#half-precision-floating-point-instructions-add>
+///
+/// Corresponds to the CUDA C intrinsics:
+///  - [`__hadd2`](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH____HALF2__ARITHMETIC.html#group__CUDA__MATH____HALF2__ARITHMETIC_1g921c795176eaa31265bd80ef4fe4b8e6)
+///  - [`__hadd2_rn`](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH____HALF2__ARITHMETIC.html#group__CUDA__MATH____HALF2__ARITHMETIC_1g6cd8ddb2c3d670e1a10c3eb2e7644f82)
+#[inline]
+#[cfg_attr(test, assert_instr(add.rn.f16x22))]
+#[unstable(feature = "stdarch_nvptx", issue = "111199")]
+pub unsafe fn f16x2_add(a: f16x2, b: f16x2) -> f16x2 {
+    simd_add(a, b)
+}
+
+/// Subtract two values, round to nearest even
+///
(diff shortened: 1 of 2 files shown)


crates/core_arch/src/nvptx/packed.rs
```diff
@@ -0,0 +1,107 @@
+//! NVPTX Packed data types (SIMD)
+//!
+//! Packed Data Types is what PTX calls SIMD types. See [PTX ISA (Packed Data Types)](https://docs.nvidia.com/cuda/parallel-thread-execution/#packed-data-types) for a full reference.
+//!
+//! Note: #[assert_instr] tests are not actually being run on nvptx due to being a `no_std` target incapable of running tests. Something like FileCheck would be appropriate for verifying the correct instruction is used.
```
Amanieu, line 5: This should be in normal comments, not doc comments. It is not helpful to users.

crates/core_arch/src/nvptx/packed.rs
```diff
@@ -0,0 +1,107 @@
...
+        #[repr(simd)]
+        #[allow(clippy::missing_inline_in_public_items)]
+        pub struct $name($($fields)*);
+    )*)
+}
+
+simd_types! {
+    /// PTX-specific 32-bit wide floating point (f16 x 2) vector type
+    #[unstable(feature = "stdarch_nvptx", issue = "111199")]
+    pub struct f16x2(f16, f16);
```
Amanieu, line 35: You should use the `types!` macro to define SIMD types. Make sure to pull, the syntax for this macro just got changed.

Amanieu: The syntax for the macro has changed, see the other architectures.

crates/core_arch/src/nvptx/packed.rs
```diff
@@ -0,0 +1,93 @@
...
+    simd_neg(a)
+}
+
+/// Find the minimum of two values
+///
+/// <https://docs.nvidia.com/cuda/parallel-thread-execution/#half-precision-floating-point-instructions-min>
+#[inline]
+#[cfg_attr(test, assert_instr(min.NaN.f16x2))]
+#[unstable(feature = "stdarch_nvptx", issue = "111199")]
+pub unsafe fn f16x2_min(a: f16x2, b: f16x2) -> f16x2 {
```
CryZe, line 81: Are these intrinsics names based on C or something? If not, this one probably should be called `f16x2_minimum` then, as it follows the IEEE-754 `minimum` rather than the `minNum` algorithm, which according to the Nvidia docs seems to be configurable.

kjetilkjeka: Thanks for commenting. I'm not familiar with the intrinsic naming conventions so I appreciate the help.

I named after the PTX `min` instruction which it is supposed to lower to. Judging from your comment you have already seen it, but for completeness this is the instruction in the ISA reference https://docs.nvidia.com/cuda/parallel-thread-execution/#half-precision-floating-point-instructions-min

I guess that it's the `NaN` part that makes it conforming to IEE-754 `minimum` and naming it only `f16x2_min` is wrong in either case. I guess `f16x2_minimum` (after the standard) or  `f16x2_min_nan` (after the ISA) is more descriptive. I like `_minimum` but I also see that most other targets have simd intrinsics named after the instruction. 

Please let me know if I should change it to `f16x2_minimum` or `f16x2_min_nan` (or keep it as is).

kjetilkjeka: I circumvented the problem in this MR by instead implementing the "minnum" intrinsic. 

I'll just create a follow up when the correct naming is established

kjetilkjeka:
Thanks for reviewing! 

I have fixed the two comments from @Amanieu 

I have circumvented the naming issue around min/minimum brought up by @CryZe by instead implementing the intrinsics that makes sense to have named as `min`/`max`. I can add the `minimum`/`min_nan` and `maximum`/`max_nan` when the correct naming is confirmed (either here or as follow up).

I see that there might be yet another slight naming issue. I have named the functions as `f16x2_<intrinsic>` while the intrinsic (in PTX) is named `<intrinsic>.f16x2`. What is the desired naming convention here?

Amanieu:
CUDA seems to have an [API](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH____HALF2__ARITHMETIC.html) for `f16` intrinsics, should we be mirroring that API instead?

kjetilkjeka:
> CUDA seems to have an [API](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH____HALF2__ARITHMETIC.html) for f16 intrinsics, should we be mirroring that API instead?

I see that the names are consistent with the ptx ISA for functions. I will change all the function names to be consistent with this naming (both CUDA and PTX ISA). I will refrain from adopting the leading double underscore `__` as this is a C detail related to not having namespaces. I'm very happy with this solution.  

The part I'm a bit more unsure about is naming the type `half2`. The CUDA C type is, at least in one aspect, closer to something found in `core::simd` as a big part of the problem it solves is being able to use the same vector instructions both on x86_64 and nvptx64 targets. These, `core::arch::nvptx` types are only usable on `ptx` and if it is acceptable to keep `f16x2` I think that is the better name (being consistent with the ptx ISA and as a bonus also familiar for Rust users)

crates/core_arch/src/nvptx/packed.rs
kjetilkjeka: Re-added with consistent naming to CUDA C API. `min_nan` and `max_nan`

kjetilkjeka:
I updated the PR a little while ago. I also re-run the rustc assembly tests I have in my branch to verify the correct instructions are still being produced. I also ported a much used function in our (my workplace) internal code and verified that tests were passing. (And there was also a notable performance boost on the relevant benchmarks).

And with that, I think this one should be ready for merging.

PR #1635, opened 2024-09-06
std_detect: Add sme-b16b16 as an explicit aarch64 target feature

mrkajetanp:
LLVM 20 split out what used to be called b16b16 and correspond to aarch64 FEAT_SVE_B16B16 into sve-b16b16 and sme-b16b16.
Add sme-b16b16 as an explicit feature and update the detection accordingly.

### DIFF
--- a/crates/std_detect/src/detect/os/linux/aarch64.rs
+++ b/crates/std_detect/src/detect/os/linux/aarch64.rs
@@ -469,11 +469,7 @@ impl AtHwcap {
                 self.svesha3 && sve2 && self.sha512 && self.sha3 && self.sha1 && self.sha2,
             );
             enable_feature(Feature::sve2_bitperm, self.svebitperm && self.sve2);
-            // SVE_B16B16 can be implemented either for SVE or SME
-            enable_feature(
-                Feature::sve_b16b16,
-                self.bf16 && (self.sveb16b16 || self.smeb16b16),
-            );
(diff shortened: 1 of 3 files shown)


crates/std_detect/src/detect/os/linux/aarch64.rs
```diff
@@ -500,6 +496,10 @@ impl AtHwcap {
             let sme2 = self.sme2 && sme;
             enable_feature(Feature::sme2, sme2);
             enable_feature(Feature::sme2p1, self.sme2p1 && sme2);
+            enable_feature(
+                Feature::sme_b16b16,
+                sme2 && self.bf16 && self.sveb16b16 && self.smeb16b16,
```
Amanieu, line 501: Is it necessary to check for `sveb16b16` here as well? I would expect this to be available for chips that only support SME but not SVE like the latest apple ones.

mrkajetanp: Yeah we should check because there's a dependency in the architecture and if you try enabling this in LLVM it'll also enable sve-b16b16.

The check is just being overly cautious because having this one implies having the other one on the architecture level anyway so that part of the check should never fail.

Slightly counterintuitive due to the naming but this is still fine for SME-only devices because sve-b16b16 does not actually depend on sve, in that case it would just only give you the SME instructions.

PR #1647, opened 2024-09-22
Add space in order to fix rendering of Rounding options

marxin:
Before this change, one of the options is wrongly rendered (apparently anything single word after `:`):
Screenshot from 2024-09-22 19-35-33


marxin:
Btw. @Amanieu, when can I expect the changes being propagated to https://doc.rust-lang.org/core/index.html ?

PR #1649, opened 2024-09-28
Clean up examples, make it more "Rusty"

nyurik:


examples/connect5.rs
```diff
@@ -651,10 +636,8 @@ fn eval(pos: &Pos, _ply: i32) -> i32 {
                     return 2560;
                 }
             }
-        } else {
-            if check_patterndead4(pos, atk) > 0 {
```
Amanieu, line 655: I'm not a fan of this clippy lint, I think in many cases it actually makes the code *less* readable.

nyurik: @Amanieu which clippy lint are you not a fan of?  Most of these are simple stylistic changes that are pretty standard in the Rust code... to my knowledge

Amanieu: Specifically the one that turns `else { if X {} }` into `else if X {}`. I often find that the additional nesting improves readability, especially if the `X` condition is completely unrelated to the condition in the outer `if`.

Amanieu: Could you undo the change to `else` here and above?

nyurik:
thx, fixed

PR #1652, opened 2024-09-28
Add all crates to workspace

nyurik:
I am not certain why some crates are missing - it might be by accident or on purpose, so feel free to reject.  This makes sure no crate is missed by accident, and also removed the non-existent `wasm-assert-instr-tests` crate.

P.S. Also, added some crate-level lints, but perhaps these should be added to all crates in the workspace?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,16 +1,8 @@
 [workspace]
 resolver = "1"
 members = [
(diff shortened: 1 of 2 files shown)


Amanieu:
It's fine to keep these lint allows crate-specific for now.

PR #1656, opened 2024-10-08
docs: do not inline x86 instrinsics into x86-64

notriddle:
This reduces the size of the `core/arch` HTML from about 187M to 132M.


Amanieu:
While this does reduce the size, it actually makes the documentation incorrect: `std::arch::x86` can't be accessed from x86-64: you need to use `std::arch::x86_64` instead.

notriddle:
I can remove the incorrect `cfg(any())` part, but is the change to an explicitly-visible glob import acceptable?

PR #1662, opened 2024-10-31
core_arch: powerpc: Fix documentation for vec_ctu

Gelbpunkt:
Like the name suggests, this converts to unsigned integers rather than signed. I assume this was copy pasted from `vec_cts`.


lu-zero (approved):
Thank you

PR #1667, opened 2024-11-05
Avoid using unexpected cfgs in user code

Urgau:
This PR adds:
 - the ability to declare a feature without cfg checking
 - and mark every feature with missing corresponding target feature cfgs as such

This is a prerequisites to https://github.com/rust-lang/rust/pull/132577 which wants to enable reporting of the `unexpected_cfgs` lint in *external macros* and is unfortunately triggering on the macros call generated by stdarch, like `is_x86_feature_detected!`, ...

I have computed the list of missing target feature cfgs by comparing:
 - `rg "[ ]+@feature: .*: \"(.*)\";" -r '$1' --no-filename crates/std_detect/src/detect/ | sort | uniq`
 - with `tests/ui/check-cfg/well-known-values.stderr`[^1] (from the main Rust repo)

([actual.txt](https://github.com/user-attachments/files/17637673/actual.txt), [referenced.txt](https://github.com/user-attachments/files/17637674/referenced.txt), [diff-actual-referenced.txt](https://github.com/user-attachments/files/17637678/diff-actual-referenced.txt))

[^1]: https://github.com/rust-lang/rust/blob/e8c698bb3bdc121ac7f65919bd16d22f6567a3f1/tests/ui/check-cfg/well-known-values.stderr#L177

### DIFF
--- a/crates/std_detect/src/detect/arch/aarch64.rs
+++ b/crates/std_detect/src/detect/arch/aarch64.rs
@@ -160,6 +160,7 @@ features! {
     @FEATURE: #[unstable(feature = "stdarch_aarch64_feature_detection", issue = "127764")] fp8fma: "fp8fma";
     /// FEAT_FP8FMA (F8FMA Instructions)
     @FEATURE: #[unstable(feature = "stdarch_aarch64_feature_detection", issue = "127764")] fpmr: "fpmr";
+    without cfg check: true;
     /// FEAT_FPMR (Special-purpose AArch64-FPMR register)
     @FEATURE: #[stable(feature = "simd_aarch64", since = "1.60.0")] frintts: "frintts";
     /// FEAT_FRINTTS (float to integer rounding instructions)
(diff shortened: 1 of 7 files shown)


crates/std_detect/src/detect/arch/aarch64.rs
```diff
@@ -244,6 +245,7 @@ features! {
     @FEATURE: #[stable(feature = "simd_aarch64", since = "1.60.0")] sve2_aes: "sve2-aes";
     /// FEAT_SVE_AES & FEAT_SVE_PMULL128 (SVE2 AES crypto)
     @FEATURE: #[unstable(feature = "stdarch_aarch64_feature_detection", issue = "127764")] sve_b16b16: "sve-b16b16";
```
Amanieu, line 247: This is actually a valid target feature, but it is only exposed on LLVM 20 (we are currently on LLVM 19). See compiler/rustc_codegen_llvm/src/llvm_util.rs

Amanieu:
Could we have some way of checking, in stdarch CI, that when rustc adds support for a feature we don't forget to re-enable the flag here?

Urgau:
> Could we have some way of checking, in stdarch CI, that when rustc adds support for a feature we don't forget to re-enable the flag here?

Sure, I just pushed (with the small fix) a commit that add a test function as part of the `feature!` macro. That automatically generated test function denies both the `unexpected_cfgs`, and `unfulfilled_lint_expectations` lint such as the for the known target feature cfgs it will trigger the `unexpected_cfgs` lint if it isn't marked with `without cfg check: true` and if it's an unexpected cfg it will trigger the lint which is expected with an `#[expect]` attribute.

I have successfully tested it locally with the modification from upstream.

PR #1674, opened 2024-11-16
Fix RISC-V Packed SIMD intrinsics compilation

SpriteOvO:
Fixes rust-lang/rust#129593.

Currently the P extension is still unratified, so there is no support for it in the upstream LLVM for now, and thus no LLVM built-in functions or serialization of instructions are provided.

This PR implements quick fixes to some of the instructions to fix them causing compilation failures (see issue rust-lang/rust#129593). Calling it a "quick fix" because ideally we could replace all `.insn` with LLVM built-in functions in the future when LLVM supports the P extension.

This PR also adds `assert_instr(unknown)` to each function so that we can at least make sure they compile. Since there is no serialization yet, the stdout of `assert_instr` is like

```
---- core_arch::riscv_shared::p::assert_kabs16_unkno1wn stdout ----
disassembly for stdarch_test_shim_kabs16_unkno1wn: 
         0: <unknown>
         1: ret
thread 'core_arch::riscv_shared::p::assert_kabs16_unkno1wn' panicked at crates/stdarch-test/src/lib.rs:172:9:
failed to find instruction `unkno1wn` in the disassembly
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    core_arch::riscv_shared::p::assert_kabs16_unkno1wn
```

So I left "unknown" here, also with the idea that if LLVM upstream provides support for the P extension at some point in the future, we can know in time and then update our implementation.

r? @workingjubilee 
CC @Amanieu

### DIFF
--- a/crates/core_arch/src/riscv_shared/p.rs
+++ b/crates/core_arch/src/riscv_shared/p.rs
@@ -2,9 +2,22 @@
 //!
 //! RV64 only part is placed in riscv64 folder.
 use crate::arch::asm;
+#[cfg(test)]
+use stdarch_test::assert_instr;
+
+// FIXME: Currently the P extension is still unratified, so there is no support
+// for it in the upstream LLVM for now, and thus no LLVM built-in functions or
+// serialization of instructions are provided.
+//
+// We add `assert_instr(unknown)` to each function so that we can at least make
+// sure they compile. Since there is no serialization yet, we can only write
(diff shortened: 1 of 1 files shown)


Amanieu:
This shouldn't be checking for `unknown`: that will just cause the tests to break when the disassembler is updated in the future. It's fine to leave them without instruction tests for now.

PR #1675, opened 2024-11-16
Fix incorrect documentation for _mm_loadu_si64 function

BLANKatGITHUB:
The documentation incorrectly stated that the function loads 16 bits of integer data.  This commit corrects the documentation to accurately reflect that the function loads  64 bits (8 bytes) of integer data from memory into a new vector.

closes rust-lang/rust#133062


Amanieu:

PR #1677, opened 2024-11-17
Add numeric instructions for Wasm not available in `core`

daxpedda:
Tracking issue: https://github.com/rust-lang/rust/issues/133908

The recently added [`wasm32v1-none` target](https://doc.rust-lang.org/nightly/rustc/platform-support/wasm32v1-none.html) comes without `std`. This has caused some `f32` and `f64` methods to not be available for this target:
- [`f32/f64::ceil()`](https://doc.rust-lang.org/1.82.0/std/primitive.f32.html#method.ceil)
- [`f32/f64::floor()`](https://doc.rust-lang.org/1.82.0/std/primitive.f32.html#method.floor)
- [`f32/f64::trunc()`](https://doc.rust-lang.org/1.82.0/std/primitive.f32.html#method.trunc) and [`f32/f64::fract()`](https://doc.rust-lang.org/1.82.0/std/primitive.f32.html#method.fract)
- [`f32/f64::round_ties_even()`](https://doc.rust-lang.org/1.82.0/std/primitive.f32.html#method.round_ties_even)
- [`f32/f64::sqrt()`](https://doc.rust-lang.org/1.82.0/std/primitive.f32.html#method.sqrt)

However, all these methods are available in Wasm via instructions that are part of Wasm v1. This PR exposes these methods as platform specific intrinsics.

It should be noted that there is a tracking issue (https://github.com/rust-lang/rust/issues/50145) for bringing [`f32/f64::sqrt()`](https://doc.rust-lang.org/1.82.0/std/primitive.f32.html#method.sqrt) to `core`, which would make adding this instruction obsolete.

I also made these functions insta-stable, is that alright? Should I make an ACP first? I'm also targetting v1.84, the release of the new [`wasm32v1-none` target](https://doc.rust-lang.org/nightly/rustc/platform-support/wasm32v1-none.html). Considering that the fork date is the 22nd, its unlikely this will make it in time.

Cc @alexcrichton.

### DIFF
--- a/crates/core_arch/src/wasm32/mod.rs
+++ b/crates/core_arch/src/wasm32/mod.rs
@@ -32,6 +32,148 @@ pub fn unreachable() -> ! {
     crate::intrinsics::abort()
 }
 
+/// Generates the [`f32.ceil`] instruction, returning the smallest integer greater than or equal to `a`.
+///
+/// This method is useful when targeting `no_std` and is equivalent to [`std::f32::ceil()`].
+///
+/// [`std::f32::ceil()`]: https://doc.rust-lang.org/std/primitive.f32.html#method.ceil
+/// [`f32.ceil`]: https://webassembly.github.io/spec/core/syntax/instructions.html#syntax-instr-numeric
+#[cfg_attr(test, assert_instr(f32.ceil))]
+#[inline]
+#[must_use = "method returns a new number and does not mutate the original value"]
+#[unstable(feature = "wasm_numeric_instr", issue = "133908")]
+pub fn f32_ceil(a: f32) -> f32 {
+    unsafe { crate::intrinsics::ceilf32(a) }
+}
+
+/// Generates the [`f32.floor`] instruction, returning the largest integer less than or equal to `a`.
+///
+/// This method is useful when targeting `no_std` and is equivalent to [`std::f32::floor()`].
+///
+/// [`std::f32::floor()`]: https://doc.rust-lang.org/std/primitive.f32.html#method.floor
+/// [`f32.floor`]: https://webassembly.github.io/spec/core/syntax/instructions.html#syntax-instr-numeric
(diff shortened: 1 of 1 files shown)


alexcrichton:
Oh nice! Seems reasonable to me :)

I thought I had also seen some chatter on the Rust Zulip about making more floating-point methods available in libcore, so it might be good to sync up on that perhaps? I think it might also be best to start these out as unstable and then perhaps quickly follow up with a stabilization request with the rationale outlined?

For code changes, perhaps the documentation of one of these functions could explain why it's available as an intrinsic and why you might not want to use it most of the time (e.g. the float methods all do the same thing). Other methods could then all link to that one instance of the documentation to avoid duplicating everything here.

alexcrichton:
Took some digging as I also forgot, but I found:

* https://github.com/rust-lang/rust/pull/131304
* https://github.com/rust-lang/rust/issues/50145

I don't think that should hold this up personally though, but probably is worth considering for stabilization

daxpedda:
Ah, thanks. I was already aware of those and left a note in the OP.
AFAIU, taking into account our list of intrinsics, only [`f32/f64::sqrt()`](https://doc.rust-lang.org/1.82.0/std/primitive.f32.html#method.sqrt) is still being considered.

alexcrichton (approved):

daxpedda:
CI seems to be unblocked again.
@Amanieu this should be ready to go now.

Amanieu:
This seems fine to add on nightly, but I wouldn't expect these to be actually stabilized: I would much prefer if we could have this functionality work directly on floats, even with `#[no_std]`.

If you still think this is worth it, please create a tracking issue and add the number the unstable attributes.

daxpedda:
Tracking issue created: https://github.com/rust-lang/rust/issues/133908.

daxpedda:
@Amanieu friendly ping.

PR #1679, opened 2024-11-19
Fix types for ARM SIMD32 intrinsics

Amanieu:
These were previously defined using vector types which is incorrect. Instead, `{u}int{8x4,16x2}_t` are aliases for `i32` and `u32`.

This also fixes CI since these types don't need to be passed in NEON registers and this was triggering a newly added compiler warning.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -113,10 +113,11 @@ jobs:
           os: ubuntu-latest
         - tuple: thumbv7em-none-eabihf
           os: ubuntu-latest
-        - tuple: loongarch64-unknown-linux-gnu
-          os: ubuntu-latest
-        - tuple: wasm32-wasip1
-          os: ubuntu-latest
+        # Temporarily disabled due to fallout from SIMD ABI checking.
+        #- tuple: loongarch64-unknown-linux-gnu
+        #  os: ubuntu-latest
+        #- tuple: wasm32-wasip1
+        #  os: ubuntu-latest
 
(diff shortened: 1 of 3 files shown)


Amanieu:
@hev LoongArch seems to have been broken by https://github.com/rust-lang/rust/issues/116558 as well, but I think the lint may be triggering incorrectly in that case? Perhaps this needs to be fixed on the compiler side?

Amanieu:
Sorry that should be @heiher.

heiher:
> @heiher LoongArch seems to have been broken by [rust-lang/rust#116558](https://github.com/rust-lang/rust/issues/116558) as well, but I think the lint may be triggering incorrectly in that case? Perhaps this needs to be fixed on the compiler side?

Thank you for pointing it out. I have created a PR for it: https://github.com/rust-lang/rust/pull/133249

Amanieu:
cc @alexcrichton I've disabled WASM in CI because it was failing for the same reason as loongarch: the compiler is emitting an error because it thinks that SIMD types can't be passed as argument/return values when `simd128` is not enabled and the vector constructor functions don't have this attribute. However in practice it's fine since when that target feature is absent LLVM will just lower to scalar operations.

alexcrichton:
Thanks for the heads up! I'll take a look early next week into what the best fix would be

alexcrichton:
Followed up with https://github.com/rust-lang/stdarch/pull/1686

PR #1682, opened 2024-11-25
Rework SIMD zeroing

TDecking:
This PR adds a dedicated `ZERO` constant for the internal SIMD types and uses it intead of current idioms such as `splat(0)` or `setzero_pd().as_<type>()`. The result is improved code generation in non-optimized builds as well as a small line count reduction.


Amanieu:

PR #1688, opened 2024-12-10
core_arch: Add LoongArch basic intrinsics

heiher:


Amanieu:
Is the intent here to match the C intrinsics exactly? In which case I would expect this to use `legacy_const_generics` and `struct rdtime` like the C header does.

Alternatively it's fine to leave this as it is if the intent is for the Rust intrinsics to have different signatures from the C ones.

heiher:
> Is the intent here to match the C intrinsics exactly?

Thank you for your reply. These are not exact bindings for C intrinsics; they are just helper functions. Additional functions may be added as needed, even if they do not exist in the C intrinsics.

Amanieu:
This seems fine for now, but we will want to review before stabilization whether it is better to directly mirror the C API instead.

Issue #1672, opened 2024-11-12
`core_arch::x86::xsave::tests::test_xsavec{,64}` fails on Windows


marxin:
It happens the test fails on Windows on my Zen PC: `AMD Ryzen 9 7900X 12-Core Processor`:

```
---- core_arch::x86::xsave::tests::test_xsavec stdout ----
thread 'core_arch::x86::xsave::tests::test_xsavec' panicked at crates\core_arch\src\x86\xsave.rs:254:9:
assertion `left == right` failed
[...]

PR #1690, opened 2024-12-13
Remove buffers in tests for `{f,}xsave{c,}{64,}`

marxin:
Fixes: #1672


Amanieu:
We should maybe also consider deprecating these intrinsics. I don't think there is a way to use them correctly without inline assembly.

sayantn:
Also, IG we can now re-enable the x86_64-pc-windows-gnu-dev test

PR #1692, opened 2024-12-15
Fix Wasm `f*_sqrt()` doc

daxpedda:
Caught in https://github.com/rust-lang/rust/pull/134333.

While trying to figure out why this what not caught in the CI, I was unable to reproduce the error locally, `cargo doc --target wasm32-unknown-unknown` doesn't seem to do the trick. Maybe something to do with how `stdarch` interacts with `core` I don't understand.

### DIFF
--- a/crates/core_arch/src/wasm32/mod.rs
+++ b/crates/core_arch/src/wasm32/mod.rs
@@ -94,7 +94,7 @@ pub fn f32_nearest(a: f32) -> f32 {
 /// This method is useful when targeting `no_std` and is equivalent to [`std::f32::sqrt()`].
 ///
 /// [`std::f32::sqrt()`]: https://doc.rust-lang.org/std/primitive.f32.html#method.sqrt
-/// [`f32.nearest`]: https://webassembly.github.io/spec/core/syntax/instructions.html#syntax-instr-numeric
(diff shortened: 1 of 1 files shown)


Amanieu:
You need to enable the `rustdoc::broken-intra-doc-links` lint which I believe is not enabled by default. See https://stackoverflow.com/questions/70511431/how-can-i-make-the-rustdocbroken-intra-doc-links-lint-an-error-from-the-comm.

daxpedda:
> You need to enable the `rustdoc::broken-intra-doc-links` lint which I believe is not enabled by default.

Apologies, I should have mentioned, I already that by adding `#![deny(rustdoc::broken_intra_doc_links)]` and `RUSTDOCFLAGS=-Drustdoc::broken_inta_doc_links`. Also tried with `rustdoc::all` (which showed a bunch of other issues, but not that one, which I also made sure of by temporarily fixing all the other issues).

I also made sure that the functions were compiled in the first place by adding a `compile_error!` inside a function, which correctly triggered. Also added `#![feature(wasm_numeric_instr)]` just to stave off any lingering paranoia.

Amanieu:
Fair enough, I don't know why it isn't working then.

PR #1693, opened 2024-12-16
New ARM intrinsic generator

Jamesbarford:
# New Arm intrinsic generator

A new generator for creating ARM intrinsics using YAML. The input which superseeds the old generators bespoke file format that was proving cumbersome to maintain and tricky to expand to support other architectures than neon. The YAML provides a more expressive and maintainable way to declare intrinsics.

## Summary of changes
- Remove the old generator
- Add new generator
- Add YAML files for generating rust instrinsics
- `aarch64/neon/generated.rs` and `arm_shared/neon/generated.rs` are both newly created from the new input as such has a slightly different output and code style.


Amanieu:
Some initial notes (I'm travelling and will do a full review later):
- I'm very glad to see this! It indeed seems much more maintainable than the old generator.
- This should be called stdarch-gen-arm rather than stdarch-gen2.
- We have existing manual implementations of intrinsics in crates/core_arch/src/arm_shared/neon/mod.rs, for example `vbsl_p16`. Ideally we should use automatic generation for all intrinsics and remove the manual implementations.

Jamesbarford:
Thanks, hope your travels are going smoothly! It can be a pretty hectic time of year.

I'll hold off changes until you've had a more thorough perusal.

With respect to the intrinsics floating about in the example you gave of `crates/core_arch/src/arm_shared/neon/mod.rs`. I was thinking of doing a two stage approach; replace what we are currently generating with the new generator. Then port over the other intrinsics when the opportune moment arises to do so.

This is based off it being time consuming porting the old intrinsics over coupled with the desire to unblock other units of work that cannot be achieved without the new generator.

I do, however, undoubtedly agree that the goal should be to have all of these intrinsics generated with the new generator.

Jamesbarford:
Thanks for your review, I believe the latest commit should have taken care of your comments.

Jamesbarford:
Are there any other changes that are required? I'd be of the opinion the manually created intrinsics can be done as a follow-up because this PR is arguably quite large

PR #1699, opened 2025-01-12
add `is_s390x_feature_detected`

folkertdev:
tracking issue: https://github.com/rust-lang/rust/issues/135413
relevant issue: https://github.com/rust-lang/rust/issues/130869

this PR adds the `is_s390x_feature_detected` macro.

Is an ACP needed here? 

Below is a bunch of context.

## Feature detection

This PR uses only the `getauxval` function. That is sufficient for the current feature flags.

Based on the official documents, the `stfl` instruction can/should be used to test for "[Facility Indications]". Unfortunately, it appears that this instruction can only be executed by the kernel, and in any case stores its result at address `200`, that users cannot read. Bummer.

There is also the `stfle` instruction, that provides more facility indications. We can actually use that from user space, e.g.

```rust
struct ExtendedFacilityList([u64; 4]);

impl ExtendedFacilityList {
    unsafe fn new() -> Self {
        let mut result: [u64; 4] = [0; 4];
        unsafe {
            core::arch::asm!(
                "lgr %r0, {0}",
                // equivalently ".insn s, 0xb2b00000, 0({1})",
                "stfle 0({1})",
                in(reg) result.len() as u64 - 1,
                in(reg_addr) result.as_mut_ptr() ,
                options(nostack, preserves_flags )
            );
        }
        Self(result)
    }

    const fn get(&self, n: usize) -> bool {
        // of course they number bits from the left...
        self.0[n / 64] & (1 << (63 - (n % 64))) != 0
    }
}
```

But, for now, this is not needed and `getauxval` is sufficient.

## Feature names

On nightly, `s390x` currently has two target features, introduced in https://github.com/rust-lang/rust/pull/127506:

- `vector` for SIMD capabilities
- `backchain` related to stack unwinding

The "vector" name appears to be an LLVM concept: in `qemu`, [`clang`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-mvx) and the [linux kernel](https://github.com/torvalds/linux/blob/b62cef9a5c673f1b8083159f5dc03c1c5daced2f/arch/s390/include/asm/elf.h#L141) the "vector" extension is called `vx`. There is nothing called "backchain" in the [Facility Indications], so it looks like that is not a feature that can be queried for at runtime.

Anyway, for now, only `"vector"` is supported by the `is_s390x_feature_detected` macro, and it maps internally to the `vx` facility indication. For some of the vector instructions, the extensions like `vector-enhancements-1` will likely also be needed, but this PR does not include them.

This [LLVM file](https://github.com/llvm/llvm-project/blob/d047dbd95ed3ef4ace54eaa7c32fe3954317f926/llvm/lib/Target/SystemZ/SystemZFeatures.td) lists many more features that are not currently supported.

## Testing

I could not find a qemu CPU that works with rust programs and also does not support `vx`. I did test the logic using the `vxrs_ext2` field, and a custom CPU:

```toml
runner = "qemu-s390x -cpu qemu,vx=on,vxeh=on,vxeh2=off -L /usr/s390x-linux-gnu"
```

I verified that toggling the `vxeh2` flag is observed by the implementation.

Obviously that is not an automated test, so for now the details of the implementation go untested.

[Facility Indications]: https://www.ibm.com/support/pages/sites/default/files/2021-05/SA22-7871-10.pdf#page=63

cc @taiki-e, hope I'm not stepping on your toes here, but this nerdsniped me.

### DIFF
--- a/crates/std_detect/README.md
+++ b/crates/std_detect/README.md
@@ -5,7 +5,7 @@ The private `std::detect` module implements run-time feature detection in Rust's
 standard library. This allows detecting whether the CPU the binary runs on
 supports certain features, like SIMD instructions.
 
-# Usage 
+# Usage
 
 `std::detect` APIs are available as part of `libstd`. Prefer using it via the
 standard library than through this crate. Unstable features of `std::detect` are
@@ -19,7 +19,7 @@ from the platform.
 You can then manually include `std_detect` as a dependency to get similar
 run-time feature detection support than the one offered by Rust's standard
 library. We intend to make `std_detect` more flexible and configurable in this
-regard to better serve the needs of `#[no_std]` targets. 
+regard to better serve the needs of `#[no_std]` targets.
 
 # Features
 
@@ -53,8 +53,8 @@ crate from working on applications in which `std` is not available.
   [`cupid`](https://crates.io/crates/cupid) crate.
 
 * Linux/Android:
-  * `arm{32, 64}`, `mips{32,64}{,el}`, `powerpc{32,64}{,le}`, `riscv{32,64}`, `loongarch64`: `std_detect`
-    supports these on Linux by querying ELF auxiliary vectors (using `getauxval`
+  * `arm{32, 64}`, `mips{32,64}{,el}`, `powerpc{32,64}{,le}`, `riscv{32,64}`, `loongarch64`, `s390x`:
+    `std_detect` supports these on Linux by querying ELF auxiliary vectors (using `getauxval`
     when available), and if that fails, by querying `/proc/cpuinfo`.
   * `arm64`: partial support for doing run-time feature detection by directly
     querying `mrs` is implemented for Linux >= 4.11, but not enabled by default.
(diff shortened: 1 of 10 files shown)


taiki-e:
> ```rust
>              core::arch::asm!(
>                 "lgr %r0, {0}",
>                 // equivalently ".insn s, 0xb2b00000, 0({1})",
>                 "stfle 0({1})",
>                 in(reg) result.len() as u64 - 1,
>                 in(reg_addr) result.as_mut_ptr() ,
>                 options(nostack, preserves_flags )
>             );
> ```

Btw, this inline assembly has UBs. (I missed it because it was not included in the change by PR...)

- r0 is clobbered (by both `lgr ..` and `stfile`), but not marked clobbered in assembly. (Need to add `out("r0") _,`, or use `inout("r0") result.len() as u64 - 1 => _,` and remove `lgr ...`)

  Section for stfile in PoP says:

  > bits 56-63 of general register 0 are updated to contain one less than the number of doublewords needed to contain all of the facility bits assigned for the configuration

- condition code is modified (by `stfile`), but preserves_flags is used. (Need to remove `preserves_flags`)

  Section for stfile in PoP says:

  > Resulting Condition Code:
  > 0 Complete facility list stored
  > 1 --
  > 2 --
  > 3 Incomplete facility list stored

PR #1700, opened 2025-01-14
Stabilize `wasm_numeric_instr`

daxpedda:
Awaiting approval in #133908.


daxpedda:
It was decided not to stabilize this after all.
See https://github.com/rust-lang/rust/issues/133908#issuecomment-2590508742.

PR #1701, opened 2025-01-16
Fix: Update to Support Rust's Unstable Features and Correct Module Imports

krishpranav:
This PR addresses issues with unstable features like f128 and incorrect module imports in the core_arch crate. It updates the code to work with the nightly version of Rust, enabling the necessary unstable features and fixing issues related to the use of std in a no_std environment.


tgross35:
This patch does not address the linked issue, does not follow the style of this crate, has a completely unrelated commit message, and the code is visibly incorrect. Help bringing up SIMD for the new float types is very welcome, but please make an effort to understand the code before submitting it.

PR #1702, opened 2025-01-17
Stdarch support for `wasm32-wali-linux-musl` Tier-3 Linux target

arjunr2:
To support the new Tier-3 target `wasm32-wali-linux-musl` within the rust compiler : rust-lang/rust#135651
 
Minimally, this just disables dynamic-linking related Linux features which is not supported yet in Wasm.

### DIFF
--- a/crates/std_detect/src/detect/os/linux/auxvec.rs
+++ b/crates/std_detect/src/detect/os/linux/auxvec.rs
@@ -74,6 +74,7 @@ pub(crate) fn auxv() -> Result<AuxVec, ()> {
         not(all(target_os = "linux", target_env = "gnu")),
         // TODO: libc crate currently doesn't provide getauxval on 32-bit Android.
         not(all(target_os = "android", target_pointer_width = "64")),
+        not(target_arch = "wasm32")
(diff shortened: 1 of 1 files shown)


sayantn:
Please rebase

bjorn3:
Is there a reason wali doesn't provide getauxvval?

arjunr2:
> Is there a reason wali doesn't provide getauxvval?

Wasm doesn't have a dynamic linking specification yet. It can be added in the future if dynamic linking is added to Wasm

bjorn3:
Directly linking to `getauxval` rather than using `dlsym` should work if libc for wali is guaranteed to support it, right? The only reason std_detect is using `dlsym` is because older libc versions for some targets don't support `getauxval`.

arjunr2:
That's true, we could probably just disable `std_detect_dlsym_getauxval` when building `stdarch` for `rustc`. That should work right?

WALI does expose `getauxval` in the binary but support for it is not fully specified/handled yet (particularly architecture-specific types/values)  As such, preventing it from being called for the time being is preferable.

PR #1705, opened 2025-01-30
Fix build and CLI behaviour for stdarch-gen-arm.

jacobbramley:
Fix the `stdarch-gen-arm` build by correcting misuse of `proc_macro2::Punct`. Our usage was always incorrect, but recent versions of `Punct::new` include assertions that show it.

Also fix the `generated.rs` path inference.


crates/core_arch/src/aarch64/neon/generated.rs
sayantn, line 1: Seems like you forgot to re-generate this file

jacobbramley:
Someone commented that I'd forgotten to regenerate the files. For some reason, I can't see the comment on GitHub, but they were right; the instructions at the top of each `generated.rs` still point to the old generator. I'll fix that shortly.

PR #1706, opened 2025-02-05
Add keylocker (`kl` and `widekl`) intrinsics and runtime feature detection

sayantn:
This is an effort towards rust-lang/rust#134813

`_mm_loadiwkey` is only allowed to be called from CPL0, so we are skipping testing it.

### DIFF
--- a/crates/stdarch-test/src/lib.rs
+++ b/crates/stdarch-test/src/lib.rs
@@ -119,6 +119,9 @@ pub fn assert(shim_addr: usize, fnname: &str, expected: &str) {
                 // it from the slightly more restrictive 22 instructions below.
                 "cpuid" => 30,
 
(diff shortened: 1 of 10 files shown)


crates/stdarch-test/src/lib.rs
```diff
@@ -119,6 +119,9 @@ pub fn assert(shim_addr: usize, fnname: &str, expected: &str) {
                 // it from the slightly more restrictive 22 instructions below.
                 "cpuid" => 30,
 
+                // It takes at least 23 because of the many moves, also there are padding NOPs in x86
+                "aesencwide128kl" | "aesdecwide128kl" | "aesencwide256kl" | "aesdecwide256kl" => 29,
```
Amanieu, line 123: Rather than special casing padding nops, we could change the test suite to ignore any instructions after `ret`.

PR #1712, opened 2025-02-13
S390x float rounding

folkertdev:
tracking issue: https://github.com/rust-lang/rust/issues/135681

cc @uweigand


folkertdev:
for bookkeeping, https://github.com/llvm/llvm-project/commit/adacbf68ebeaeaf9d07a5beaff163da9ca0534a9 implements `llvm.roundeven` using the `vfisb`/`vfidb` instruction. We'll have to wait for that to be available in nightly before we can test it.

Right now `roundeven` uses a fallback implementation that works, but is not as efficient as a single instruction.

PR #1713, opened 2025-02-13
powerpc: use more target-independent llvm intrinsics (`min`, `max`, `round`, `countlz`)

folkertdev:
These are available now, and much easier to support for miri and the gcc/cranelift backends then target-specific intrinsics.

### DIFF
--- a/crates/core_arch/src/powerpc/altivec.rs
+++ b/crates/core_arch/src/powerpc/altivec.rs
@@ -1455,8 +1455,7 @@ mod sealed {
     #[cfg_attr(test, assert_instr(vspltb, IMM4 = 15))]
     unsafe fn vspltb<const IMM4: u32>(a: vector_signed_char) -> vector_signed_char {
         static_assert_uimm_bits!(IMM4, 4);
-        let b = u8x16::splat(IMM4 as u8);
-        vec_perm(a, a, transmute(b))
+        simd_shuffle(a, a, const { u32x16::from_array([IMM4; 16]) })
     }
 
     #[inline]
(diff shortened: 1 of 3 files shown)


crates/core_arch/src/powerpc/altivec.rs
```diff
@@ -368,21 +368,14 @@ unsafe extern "C" {
...
     #[link_name = "llvm.ppc.altivec.vrlb"]
     fn vrlb(a: vector_signed_char, b: vector_unsigned_char) -> vector_signed_char;
     #[link_name = "llvm.ppc.altivec.vrlh"]
     fn vrlh(a: vector_signed_short, b: vector_unsigned_short) -> vector_signed_short;
     #[link_name = "llvm.ppc.altivec.vrlw"]
     fn vrlw(a: vector_signed_int, c: vector_unsigned_int) -> vector_signed_int;
 
-    #[link_name = "llvm.ppc.altivec.vrfin"]
+    #[link_name = "llvm.nearbyint.v4f32"]
     fn vrfin(a: vector_float) -> vector_float;
```
folkertdev, line 379: I've verified that these in fact produce identical assembly https://godbolt.org/z/Wx1KWezbe

lu-zero (approved):
Looks fine, thank you.

folkertdev:
I've added 2 more commits

- use `simd_shuffle` in the implementation of `vec_splat`. This removes a bunch of lines of mask construction.
- use `simd_neg` in `impl Neg`, because the old implementation did not always combine with other instructions well (at least on s390x), and using the specific intrinsic just seems right and gives the backend the most information. In isolation the old and new version are identical https://godbolt.org/z/98r73bW5b

crates/core_arch/src/powerpc/altivec.rs
```diff
@@ -158,32 +158,32 @@ unsafe extern "C" {
     #[link_name = "llvm.ppc.altivec.vmulosh"]
     fn vmulosh(a: vector_signed_short, b: vector_signed_short) -> vector_signed_int;
 
-    #[link_name = "llvm.ppc.altivec.vmaxsb"]
+    #[link_name = "llvm.smax.v16i8"]
```
Amanieu, line 161: Since these are generic LLVM vector intrinsics, it would make more sense to provide these in `core::intrinsics::simd` and then use them from stdarch, similar to how `simd_neg` is used.

Amanieu: I see this is already a pattern in quite a few places in stdarch. This doesn't need to block this PR, it can be resolved separately.

folkertdev: yeah, makes sense. I'll add that then.

PR #1716, opened 2025-02-15
mark arm intrinsics as safe

usamoi:
This patch marks all neon intrinsics that do not operate on memory, crc and crypto intrinsics as safe.

ACP: https://github.com/rust-lang/libs-team/issues/494


Amanieu:
ARM maintainers should have a second pass at reviewing this.

cc @JamieCunliffe @adamgemmell @Jamesbarford

Jamesbarford:
It might be beneficial to have https://github.com/rust-lang/stdarch/pull/1708 merged prior to these changes as https://github.com/rust-lang/stdarch/pull/1708 converts _most_ of the manually defined intrinsics to the YAML file.

Amanieu:
This needs another rebase since the f16 intrinsics were added.

crates/stdarch-gen-arm/spec/neon/arm_shared.spec.yml
```diff
@@ -14184,8 +13816,7 @@ intrinsics:
     visibility: private
     doc: "Store multiple single-element structures from one, two, three, or four registers."
     arguments: ["addr: {type[1]}", "val: {neon_type[2]}", "align: {type[3]}"]
-    safety:
-      unsafe: [neon]
+    safety: safe
```
Jamesbarford, line 13819: This should be marked as unsafe as it operates on memory; for example [vst1_s8](https://developer.arm.com/architectures/instruction-sets/intrinsics/vst1_s8):

> Store multiple single-element structures from one, two, three, or four registers. This instruction stores elements to memory from one, two, three, or four SIMD&FP registers, without interleaving. Every element of each register is stored.

Jamesbarford (changes requested):

Jamesbarford:
LGTM

PR #1722, opened 2025-02-23
Changed altivec.rs to new intrinsic declaration

BLANKatGITHUB:


BLANKatGITHUB:
Well in rust-lang/rust#132735 the pr said to migrate it to the same style like in compiler and standard library , I will change  it if its not ok

eduardosm:
The goal is to avoid having an `extern "rust-intrinsic"` block, which can also be achieved by not needing to declare the intrinsic.

folkertdev:
The code does note that the intrinsic declaration is there to 

> // Workaround ptr::copy_nonoverlapping not being inlined

there are some issues around target features preventing inlining, so it may be related to that. That problem also may have been fixed in the meantime (by more aggressively inlining on the rust side), but it's something to check for.

folkertdev:
Looks like that is no longer a problem (in this case) https://godbolt.org/z/vYzxKMdeo, and both versions generate the same assembly. So just using `core::ptr::copy_nonoverlapping` should be fine. (I don't even see a reason to prefer the intrinsic from `core::intrinsics`, but maybe there is one?)

crates/core_arch/src/powerpc/altivec.rs
```diff
@@ -688,10 +688,13 @@ mod sealed {
                 let addr = (b as *const u8).offset(a);
 
                 // Workaround ptr::copy_nonoverlapping not being inlined
-                unsafe extern "rust-intrinsic" {
+                    #[rustc_intrinsic]
+                    #[rustc_intrinsic_must_be_overridden]
```
RalfJung, line 692: Note that the `rustc_intrinsic_must_be_overridden` attribute is being removed, please do not use it any more. The new new way to declare an intrinsic is:
```rust
#[rustc_intrinsic]
unsafe fn unreachable() -> !;
```

RalfJung:
We used to do something similar in `core::ptr` where we sometimes called the intrinsic directly to avoid the extra debug checks that exist in `core::intrinsics::copy_nonoverlapping`. But we no longer seem to do that so I agree we shouldn't have to do it here either -- just call `core::intrinsics::copy_nonoverlapping`.

RalfJung:
You can then also remove this line; stdarch no longer should declare any intrinsics itself:

https://github.com/rust-lang/stdarch//blob/f32eeffb47d783a703017851b1cf6bdecf1601c0/crates/core_arch/src/lib.rs#L16

BLANKatGITHUB:
> You can then also remove this line; stdarch no longer should declare any intrinsics itself:
> 
> https://github.com/rust-lang/stdarch//blob/f32eeffb47d783a703017851b1cf6bdecf1601c0/crates/core_arch/src/lib.rs#L16

wont it potentially break some of the functionality?

folkertdev:
No, you've removed all usages of `rust-intrinsic`, so this feature is no longer needed (and in fact we'd want adding a new `rustc-intrinsic` to fail). Also CI would catch it if something did break, so there is no risk.

crates/core_arch/src/powerpc/altivec.rs
```diff
@@ -687,15 +687,9 @@ mod sealed {
...
-                // Workaround ptr::copy_nonoverlapping not being inlined
-                unsafe extern "rust-intrinsic" {
-                    #[rustc_nounwind]
-                    pub fn copy_nonoverlapping<T>(src: *const T, dst: *mut T, count: usize);
-                }
-
                 let mut r = mem::MaybeUninit::uninit();
 
-                copy_nonoverlapping(
+                core::ptr::copy_nonoverlapping(
```
RalfJung, line 692:
```suggestion
                crate::ptr::copy_nonoverlapping(
```

crates/core_arch/src/powerpc/altivec.rs
```diff
@@ -742,13 +736,7 @@ mod sealed {
...
                 let addr = (b as *mut u8).offset(a);
 
-                // Workaround ptr::copy_nonoverlapping not being inlined
-                unsafe extern "rust-intrinsic" {
-                    #[rustc_nounwind]
-                    pub fn copy_nonoverlapping<T>(src: *const T, dst: *mut T, count: usize);
-                }
-
-                copy_nonoverlapping(
+                core::ptr::copy_nonoverlapping(
```
RalfJung, line 739:
```suggestion
                crate::ptr::copy_nonoverlapping(
```

RalfJung (commented):
This crate has a sumewhat funny import structure due to how it is integrated with libcore. Everything else uses `crate::ptr`, not `core::ptr`, so just to be sure let's also use that here.

PR #1731, opened 2025-02-27
Fix doctests failing due to unused_unsafe

ehuss:
The function calls in these doctests were switched to safe in https://github.com/rust-lang/stdarch/pull/1714 which broke these tests. This unblocks updating upstream.


ehuss:
@Amanieu It looks like something went wrong with GitHub ("The hosted runner encountered an error while running your job. (Error Type: Disconnect)."). It looks like maybe you tried to trigger a re-run? But that doesn't put it back in the queue. Can you please try merging it again?

PR #1732, opened 2025-02-28
Feat - Aarch64 FEAT_FAMINMAX

Jamesbarford:
These intrinsics have been excluded from the tests because they require a Clang version with LLVM 20, and QEMU does not yet support the famax or famin instructions.

### DIFF
--- a/crates/stdarch-gen-arm/spec/neon/aarch64.spec.yml
+++ b/crates/stdarch-gen-arm/spec/neon/aarch64.spec.yml
@@ -13973,3 +13973,45 @@ intrinsics:
           - - r
             - a
             - FnCall: ["vdup{neon_type[1].N}", [{FnCall: [simd_extract!, [b, 'LANE as u32']]}]]
+
+  - name: "vamax{neon_type.no}"
+    doc: "Multi-vector floating-point absolute maximum"
+    arguments: ["a: {neon_type}", "b: {neon_type}"]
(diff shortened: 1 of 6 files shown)


crates/stdarch-gen-arm/spec/neon/aarch64.spec.yml
```diff
@@ -13973,3 +13973,43 @@ intrinsics:
...
             - FnCall: ["vdup{neon_type[1].N}", [{FnCall: [simd_extract!, [b, 'LANE as u32']]}]]
+
+  - name: "vamax{neon_type.no}"
+    doc: "Multi-vector floating-point absolute maximum"
+    arguments: ["a: {neon_type}", "b: {neon_type}"]
+    return_type: "{neon_type}"
+    attr:
+      - FnCall: [target_feature, ['enable = "neon,faminmax"']]
+      - FnCall: [cfg_attr, [test, {FnCall: [assert_instr, [nop]]}]]
+      - FnCall: [stable, ['feature = "neon_intrinsics"', 'since = "1.59.0"']]
```
Amanieu, line 13984: New intrinsics should always be added as unstable and have a tracking issue.

Jamesbarford: I've added the issue here; https://github.com/rust-lang/rust/issues/137933

PR #1734, opened 2025-03-01
powerpc: use the `simd_fma` intrinsic for `vec_madd`

folkertdev:
I had hoped to also replace the implementation of

```rust
    #[inline]
    #[target_feature(enable = "altivec")]
    #[cfg_attr(test, assert_instr(vnmsubfp))]
    pub unsafe fn vec_vnmsubfp(a: vector_float, b: vector_float, c: vector_float) -> vector_float {
        simd_neg(simd_fma(a, b, simd_neg(c)))
    }
```

but that does not optimize for earlier versions of powerpc apparently https://godbolt.org/z/zvWfra8PT

### DIFF
--- a/crates/core_arch/src/powerpc/altivec.rs
+++ b/crates/core_arch/src/powerpc/altivec.rs
@@ -129,8 +129,6 @@ unsafe extern "C" {
         b: vector_signed_short,
         c: vector_signed_int,
     ) -> vector_signed_int;
-    #[link_name = "llvm.ppc.altivec.vmaddfp"]
-    fn vmaddfp(a: vector_float, b: vector_float, c: vector_float) -> vector_float;
     #[link_name = "llvm.ppc.altivec.vnmsubfp"]
     fn vnmsubfp(a: vector_float, b: vector_float, c: vector_float) -> vector_float;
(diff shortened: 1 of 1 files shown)


gilamn5tr:
FWIW, We talked a little about this internally, pwr5 doesn't have a vector unit. Although ppc970 and pwr6 have vector units this will not transform as the negate operation has no hardware support for altivec so will get broken up.  The PowerPC vector units prior to Power7 were poor performing so you were often better off to use scalar code. 

Not sure if that helps.

folkertdev:
huh, so on those CPUs it is faster to run

    vspltisb 5, -1
    vslw 5, 5, 5
    vsubfp 4, 5, 4
    vmaddfp 2, 2, 3, 4
    vsubfp 2, 5, 2

instead of a single

    vnmsubfp 2, 2, 3, 4

or is it rather that some of the intermediate steps aren't profitable for those CPUs, and so the final optimization is never found?

gilamn5tr:
Its not really either.  The vnmsubfp instruction produced incorrect results so was replaced with the functionally equivalent set of instructions. I did find the change set https://reviews.llvm.org/D80617 The issue is specific to VMX instruction, the VSX instructions are ok.

PR #1740, opened 2025-03-05
Don't field-project (`.0`) into SIMD types

scottmcm:
Part of https://github.com/rust-lang/compiler-team/issues/838, as people have already mentioned in comments like

https://github.com/rust-lang/stdarch/blob/f3be311e4d0405912d8d282ecd6206aeddb3ef1d/crates/core_arch/src/macros.rs#L112-L113

### DIFF
--- a/crates/core_arch/src/macros.rs
+++ b/crates/core_arch/src/macros.rs
@@ -132,7 +132,7 @@ macro_rules! types {
         impl crate::fmt::Debug for $name {
             #[inline]
             fn fmt(&self, f: &mut crate::fmt::Formatter<'_>) -> crate::fmt::Result {
-                crate::core_arch::simd::debug_simd_finish(f, stringify!($name), self.0)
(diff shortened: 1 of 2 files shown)


folkertdev:
it's because `f16` is just not really supported on s390x currently, e.g. even a simple `f16` comparison fails to compile with a very similar error:

https://godbolt.org/z/xGTEevEjb

tgross35:
In std/build.rs there is a list of platforms that have problems with f16 and f128. Most of them are pretty quiet  and only show up when you try to use specific ops, but s390x stands out since it’s basically an instant crash if `half` shows up in the IR.

The workaround was using `#[inline]` on any functions containing these types so we don’t codegen, since https://github.com/rust-lang/rust/pull/133050 that’s automatic but limited. Maybe this PR is hitting a case that isn’t auto-inlined?

folkertdev:
yeah the `#[inline]` on the manual `PartialEq` implementation seems to have done the trick

PR #1744, opened 2025-03-16
s390x: documentation fixes

folkertdev:
adds a link to the module docs, and uses the official intrinsic names as the title for the docs. I'm not sure what to do with the additional documentation, I've left it where i added it before, but really it's not our job to document these functions so we can also remove that text?


Amanieu:
I'm not too picky about the documentation on individual functions, you really should be referring to the vendor's documentation if you're using the intrinsics.

This seems to have conflicts, but otherwise LGTM.

PR #1746, opened 2025-03-18
std_detect: Always avoid dlsym on *-linux-{musl,ohos}* targets

taiki-e:
Similar to https://github.com/rust-lang/stdarch/pull/1375 (for `*-linux-gnu*`) and https://github.com/rust-lang/stdarch/pull/1406 (for `*-android*`), but for `-linux-{musl,ohos}*`:

* `*-linux-musl*` targets ([at least since Rust 1.15](https://github.com/rust-lang/rust/blob/1.15.0/src/ci/docker/x86_64-musl/build-musl.sh#L15)) use musl newer than [musl 1.1.0 that added `getauxval`](https://git.musl-libc.org/cgit/musl/tree/WHATSNEW?h=v1.1.0#n1197)
* `*-linux-ohos*` targets use a [fork of musl 1.2](https://gitee.com/openharmony/docs/blob/master/en/application-dev/reference/native-lib/musl.md)

On musl with static linking, there is a known problem that `getauxval` is not always available, independent of version requirements: https://github.com/rust-lang/rust/issues/89626
However, this seems to be due to the fact that `compiler-builtins` is built before `libc` (which is a dependency of `std`) links musl. So IIUC the `std` and its dependent can use `getauxval` without this problem (`std_detect` is a part of `std`). Actually `std` uses it since Rust 1.78: https://github.com/rust-lang/rust/commit/9da004ea1997ac0eedbb882fa52960417725bd48

---

After this PR, remaining linux/android targets that use dlsym are:

- 32-bit `*-android*`: This is actually a issue in `libc` crate https://github.com/rust-lang/stdarch/pull/1421
  - Opened https://github.com/rust-lang/libc/pull/4338 to solve that issue.
- `*-linux-uclibc*`: uclibc-ng [supports `getauxval` since 1.0.43](https://github.com/wbx-github/uclibc-ng/commit/d869bb1600942c01a77539128f9ba5b5b55ad647)
  - This is not available in some years old toolchains mentioned in the examples in the platform support documentation.
    - By the way, since Rust 1.78, the `std` cannot be compiled with these toolchains without a [patch](https://github.com/taiki-e/rust-cross-toolchain/blob/87b464d4fff151835570243b023911539536973c/docker/test-base/patches/std%2Barmv7-unknown-linux-uclibceabihf.diff) because the commit mentioned above that started using `getauxval` is included. (Either the `std` or the platform support documentation should be fixed.)
- `*-linux-none`: This target does not use any libc APIs. 
  - However, at this time, it does [not support `std` in the first place and only for x86_64 which doesn't use `getauxval`](https://github.com/rust-lang/rust/blob/75530e9f72a1990ed2305e16fd51d02f47048f12/src/doc/rustc/src/platform-support/x86_64-unknown-linux-none.md#L4).
  - By the way, [mustang](https://github.com/sunfishcode/mustang) provides [`getauxval`](https://github.com/sunfishcode/origin/blob/06cba9ba73a84f929ed3710f14bbe6db0ad797c8/src/getauxval.rs).

### DIFF
--- a/crates/std_detect/src/detect/os/linux/auxvec.rs
+++ b/crates/std_detect/src/detect/os/linux/auxvec.rs
@@ -58,10 +58,13 @@ pub(crate) struct AuxVec {
 /// feature detection on some platforms.
 ///
 ///  Note: The `std_detect_dlsym_getauxval` cargo feature is ignored on
-/// `*-linux-gnu*` and `*-android*` targets because we can safely assume `getauxval`
+/// `*-linux-{gnu,musl,ohos}*` and `*-android*` targets because we can safely assume `getauxval`
 /// is linked to the binary.
 /// - `*-linux-gnu*` targets ([since Rust 1.64](https://blog.rust-lang.org/2022/08/01/Increasing-glibc-kernel-requirements.html))
 ///   have glibc requirements higher than [glibc 2.16 that added `getauxval`](https://sourceware.org/legacy-ml/libc-announce/2012/msg00000.html).
+/// - `*-linux-musl*` targets ([at least since Rust 1.15](https://github.com/rust-lang/rust/blob/1.15.0/src/ci/docker/x86_64-musl/build-musl.sh#L15))
+///   use musl newer than [musl 1.1.0 that added `getauxval`](https://git.musl-libc.org/cgit/musl/tree/WHATSNEW?h=v1.1.0#n1197)
+/// - `*-linux-ohos*` targets use a [fork of musl 1.2](https://gitee.com/openharmony/docs/blob/master/en/application-dev/reference/native-lib/musl.md)
 /// - `*-android*` targets ([since Rust 1.68](https://blog.rust-lang.org/2023/01/09/android-ndk-update-r25.html))
 ///   have the minimum supported API level higher than [Android 4.3 (API level 18) that added `getauxval`](https://github.com/aosp-mirror/platform_bionic/blob/d3ebc2f7c49a9893b114124d4a6b315f3a328764/libc/include/sys/auxv.h#L49).
 ///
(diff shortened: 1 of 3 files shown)


crates/std_detect/src/detect/os/linux/auxvec.rs
```diff
@@ -73,7 +76,10 @@ pub(crate) struct AuxVec {
...
     #[cfg(all(
         feature = "std_detect_dlsym_getauxval",
-        not(all(target_os = "linux", target_env = "gnu")),
+        not(all(
+            target_os = "linux",
+            any(target_env = "gnu", target_env = "musl", target_env = "ohos"),
+        )),
         // TODO: libc crate currently doesn't provide getauxval on 32-bit Android.
         not(all(target_os = "android", target_pointer_width = "64")),
     ))]
```
Amanieu, line 85: There's a lot of duplication in this function which makes it hard to review. I think this is a good opportunity to move the choice of whether to use `libc::getauxval` or look it up dynamically into the local `getauxval` function.

taiki-e: Done in https://github.com/rust-lang/stdarch/pull/1746/commits/2960b074140080208da46870a96ead4b7d03ae35.

I actually found a difference between the two blocks...

https://github.com/rust-lang/stdarch/blob/5ba56c7924058623d230f39f9073c1e42c775775/crates/std_detect/src/detect/os/linux/auxvec.rs#L83-L89

https://github.com/rust-lang/stdarch/blob/5ba56c7924058623d230f39f9073c1e42c775775/crates/std_detect/src/detect/os/linux/auxvec.rs#L130-L137

Issue #231, opened 2017-12-10
add more /proc/cpuinfo tests


gnzlbg:
@randombit has a [huge corpus of cpuinfo files](https://github.com/randombit/cpuinfo) for all the architectures we currently care about. We should figure out a way to test the run-time feature detection system against that.

marshallpierce:
I did some work on ring's runtime detection with cpuinfo, so we could lift from that: https://github.com/briansmith/ring/pull/392/files#diff-43684ec91e5967978239dce34e7beeb7

That ended up not being used in ring (decided to not bother with cpuinfo because the platforms in question could have capabilities queried in other ways) but it might save a few minutes of implementation. :)

PR #1747, opened 2025-03-18
std_detect: Remove /proc/cpuinfo-based detection

taiki-e:
Status: All blockers have been resolved. ~~Blocked on https://github.com/rust-lang/stdarch/pull/1746, https://github.com/rust-lang/libc/pull/4338, and stdarch PR that uses the fix in libc (https://github.com/rust-lang/stdarch/pull/1778).~~

---

Assuming https://github.com/rust-lang/stdarch/pull/1746, https://github.com/rust-lang/libc/pull/4338, and stdarch PR that uses the fix in libc (https://github.com/rust-lang/stdarch/pull/1778) will be merged, the only targets that might use this implementation are:

> - `*-linux-uclibc*`: uclibc-ng [supports `getauxval` since 1.0.43](https://github.com/wbx-github/uclibc-ng/commit/d869bb1600942c01a77539128f9ba5b5b55ad647)
>   - This is not available in some years old toolchains mentioned in the examples in the platform support documentation.
>     - By the way, since Rust 1.78, the `std` cannot be compiled with these toolchains without a [patch](https://github.com/taiki-e/rust-cross-toolchain/blob/87b464d4fff151835570243b023911539536973c/docker/test-base/patches/std%2Barmv7-unknown-linux-uclibceabihf.diff) because the [commit](https://github.com/rust-lang/rust/commit/9da004ea1997ac0eedbb882fa52960417725bd48) that started using `getauxval` is included. (Either the `std` or the platform support documentation should be fixed.)
> - `*-linux-none`: This target does not use any libc APIs. 
>   - However, at this time, it does [not support `std` in the first place and only for x86_64 which doesn't use `getauxval`](https://github.com/rust-lang/rust/blob/75530e9f72a1990ed2305e16fd51d02f47048f12/src/doc/rustc/src/platform-support/x86_64-unknown-linux-none.md#L4).
>   - By the way, [mustang](https://github.com/sunfishcode/mustang) provides [`getauxval`](https://github.com/sunfishcode/origin/blob/06cba9ba73a84f929ed3710f14bbe6db0ad797c8/src/getauxval.rs).

Looking at only the builtin target, included there are:

```console
$ rustc --print target-list | grep -E 'linux-uclibc|linux-none'
armv5te-unknown-linux-uclibceabi
armv7-unknown-linux-uclibceabi
armv7-unknown-linux-uclibceabihf
mips-unknown-linux-uclibc
mipsel-unknown-linux-uclibc
x86_64-unknown-linux-none
```

These are all tier 3 targets, and only arm actually uses cpuinfo-based detection. x86_64 does not use auxv/cpuinfo, and mips does not implement cpuinfo-based detection (https://github.com/rust-lang/stdarch/issues/355). And, arm's feature detection API is currently unstable.

Also even if we consider future possible targets, file-related libc APIs should not be available for `*-linux-none`, so cpuinfo-based detection is effectively only for `*-linux-uclibc*` targets (except when writing syscalls in inline assembly). And the version of `*-linux-uclibc*` targets where `getauxval` is not available has been broken since 1.78 (and probably no one other than me was aware of it)...

So I think it is okay to remove this implementation, which will effectively no longer be used once the above PRs are merged.

(The other parts of std_detect_file_io (`/proc/self/auxv`) will be similarly effectively unused, but they are [used in the tests for the auxv-based detection](https://github.com/rust-lang/stdarch/blob/57a69c5c3418e7444a6452e2e9de14c9adcf27de/crates/std_detect/src/detect/os/linux/aarch64.rs#L524), so I'm leaving it for now.)

Another advantage of doing this is that Linux (getauxval/prctl(PR_GET_AUXV)), FreeBSD(elf_aux_info), and OpenBSD(elf_aux_info) all support auxv-based detection and HWCAP* values are the same (although there is the case that it exists on only one side), so we can merge code for them like [this](https://github.com/taiki-e/portable-atomic/blob/88a278fdfb424a751fb8e21e414fcba86b71a71d/src/imp/detect/auxv.rs).

Closes https://github.com/rust-lang/stdarch/issues/231
Closes https://github.com/rust-lang/stdarch/issues/355


taiki-e:
All blockers have been resolved.

PR #1509, opened 2023-12-13
SVE types and intrinsics

JamieCunliffe:
All the SVE intrinsics along with updates to the intrinsic test tool to allow for SVE to be tested.

CI will fail until the rustc changes are merged.

PR #1758, opened 2025-03-24
Re-organize `intrinsic-test` to enable seamless addition of behaviour testing for more architectures

madhav-madhusoodanan:
## Updates so far
1. Moved the ARM-specific testing to the `arm` module, exposing only a `test` function at `src/main.rs`
2. Created a `SupportedArchitectureTest` trait that has been implemented for `arm` and should be implemented for other architectures too. Allows us to expose functionality like building C/Rust test files and comparing outputs.
3. Moved common functionality (CLI parsing, create files, compilations, compare outputs) into the `common` module for reuse by other architectures too.
4. Implemented a `match` block for selection of architectures using the `target` CLI variable.

## Reasoning
1. The `Intrinsic` type may be specific to architectures, hence the implementation of a trait helps us to abstract the usage of such data types within the  architecture-specific module

## New architecture of `intrinsic-test`
IMG_9057


Amanieu:
The module should be called `arm` rather than `aarch`.

Amanieu:
@JamieCunliffe feel free to have a look at this since you are the original author of intrinsic-test. This is part of a GSoC project to extend intrinsic-test to other architectures.

Amanieu:
`--a32` is not currently tested in CI, but it is used for 32-bit ARM targets (`arm*` and `thumb*`). Some ARM intrinsics are only available on 64-bit ARM targets and this flag skips those.

crates/intrinsic-test/src/common/mod.rs
```diff
@@ -0,0 +1,16 @@
+use crate::common::types::ProcessedCli;
+
+pub mod compare;
+pub mod gen_c;
+pub mod gen_rust;
+pub mod types;
+pub mod values;
+
+/// Architectures must support this trait
```
madhav-madhusoodanan, line 9: This is the core addition around which this entire restructure is based upon.
Please let me know if the restructure looks good.

Looking forward to learning with this PR and the next logical ones (addition of more architectures for behavioral testing).

crates/intrinsic-test/src/arm/functions.rs
```diff
@@ -0,0 +1,270 @@
...
+
+        let compiler_command = if target == "aarch64_be-unknown-linux-gnu" {
+            let Some(cxx_toolchain_dir) = cxx_toolchain_dir else {
+                panic!(
+                    "When setting `--target aarch64_be-unknown-linux-gnu` the C++ compilers toolchain directory must be set with `--cxx-toolchain-dir <dest>`"
+                );
+            };
+
+            /* clang++ cannot link an aarch64_be object file, so we invoke
+             * aarch64_be-unknown-linux-gnu's C++ linker. This ensures that we
```
madhav-madhusoodanan, line 174: May I ask why clang++ cannot link an aarch64_be object file?
I'd be grateful for any resources about this.

Amanieu: It's not that clang itself isn't capable of it, but rather that there is no linker toolchain and sysroot (containing aarch64_be libc and other support libraries) installed for clang to use. Hence the need to use a custom toolchain instead. This code currently runs on an aarch64 system so it's able to use the system toolchain for little-endian aarch64.

madhav-madhusoodanan: Got it.

madhav-madhusoodanan:
I have attached a [diagram](https://github.com/rust-lang/stdarch/pull/1758#issue-2944197780) of the new structure of `intrinsic-test` so that it may make the review process easier.

Please let me know how best I can make the review process easier, or if there are any concerns that need to be addressed.

Amanieu:
The basic idea with the traits is good, but I think this makes too much code arm-specific. For example, I would like to see `types.rs` and `intrinsics.rs` stay in the common mode and instead have `Type` be extended to support non-arm types (either as additional enum variants or by keeping the same types but displaying them differently depending on the architecture).

@adamgemmell Woud you like to have a pass at reviewing this since you worked on this?

madhav-madhusoodanan:
Thank you for telling me.

I think making `intrinsic.rs` common seems like a lower hanging fruit than `types.rs`. I'll first work towards making `intrinsic.rs` common.

I might need to spend some more time familiarizing myself with the x86 side of things to understand how `types.rs` could be made common.

crates/intrinsic-test/src/arm/functions.rs
```diff
@@ -0,0 +1,277 @@
...
+             * are testing the intrinsics against LLVM.
+             *
+             * Note: setting `--sysroot=<...>` which is the obvious thing to do
+             * does not work as it gets caught up with `#include_next <stdlib.h>`
+             * not existing... */
+            format!(
```
madhav-madhusoodanan, line 182: Would it be a good idea to make this part common too?

Parts like:
1. Include-paths, target
2. Output binary
3. Linking steps
4. Optional cleanup step, etc

could be organized within a struct, post which we could add in functionality (in `common` module) to convert the struct into the commands.

Amanieu: Yes it's a good idea. In general there will be 2 cases to consider:
- When running natively on target hardware, we can just use the host toolchain. This is what currently happens for the aarch64 tests which run on an aarch64 Github runner.
- When running emulated using qemu, we need to use a custom toolchain. We still need to use clang, but we need to point it to the C headers and libraries of the cross-compilation toolchain.

All architectures can be generalized to one of these 2 cases.

madhav-madhusoodanan: I see
Awesome, I'll do that too.

crates/intrinsic-test/src/common/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+/// to be successfully tested.
+pub trait SupportedArchitectureTest {
```
Amanieu, line 13: You may actually want to consider an `Architecture` enum instead of a trait and just special casing the spots in the code that depend on the architecture. This somewhat inverts the whole design, but I personally find it much easier to work with in practice.

madhav-madhusoodanan: Yes, I am considering the shift to enums for the purpose.

adamgemmell:
> Are we using the `--a32` flag in the CLI of `intrinsic-test` anywhere? I was unable to find its usage. Would it be alright if I remove it, or is there some reasoning that I may be missing?

I believe it's inferred now since [here](https://github.com/rust-lang/stdarch/commit/20e6936644703e46b468fe176e805b8e7c855cf2#diff-7c34e890416eb66d4462c7f31278409c662bb808a57c3e0a4d13c232c64533a9L416) where that PR added a third target. I think it can be removed.

crates/intrinsic-test/src/arm/functions.rs
adamgemmell: > When running natively on target hardware, we can just use the host toolchain. This is what currently happens for the aarch64 tests which run on an aarch64 Github runner.

I don't believe this is true - stdarch still uses an x86 runner and [installs](https://github.com/rust-lang/stdarch/blob/master/ci/docker/aarch64-unknown-linux-gnu/Dockerfile) the cross-compile toolchain. I think this is generally good as it means people can always run `run-docker.sh` on any host which is the source of truth for CI (given some tests are sensitive to clang/objdump versions and the like)

madhav-madhusoodanan: I see

adamgemmell (commented):
Looks good! I think types.rs would map fine to x86 with a little modification.

I'd just like to add that Scalable Vectors will be added to this tool at some point - the Arm code is [here](https://github.com/rust-lang/stdarch/pull/1509/files#diff-a795e2ce4f5eb0ee8b426778d8216e8910efd05f6740bac2ebc7392b9e4b0094R115) and it would probably be useful for RISC-V in the future too.

Don't worry about conflicts with that PR (it's very old and needs a rework) and I don't think there's anything particularly different about how intrinsic-test needs to work but scalable vectors do add an extra dimension of complexity to the tool.

madhav-madhusoodanan:
Over the last [4 commits](https://github.com/rust-lang/stdarch/pull/1758/files/90249a3d1a13c380c1f1706701d453d75bd73109..cc615b6056070d94886fccfe15d9669833f95185) I changed from the `Argument` type to an `Argument<T, M>` generic type.

This allowed me to refactor a lot more code into the `common` module, since `Argument` has a lot of code that could be reused across architectures. Where it differed was in the usage of the `IntrinsicType` enum (now the `IntrinsicType` struct after removing the recursion-based `Ptr` and `Type` variants) and the `ArgPrep` struct

At the moment, each architecture can define:
1. Their own `Metadata` type (which `Constraint` used to be for `arm` target intrinsics), represented by the generic type `M`.
2. Their own version of `<arch>IntrinsicType` struct, which wraps over the core `IntrinsicType` struct and can handle architecture-specific needs (represented by the generic type `T`).
    - The common parts of the definition is now handled by a macro called `base_intrinsictype_trait_def_macro`.
    - Reason I didn't make this into an `enum` is I think it might be easier to extend architecture-specific versions of `IntrinsicType` by adding a new module, please let me know if this is a good idea.

I've made some changes in the C testfile-generation part, will pinpoint them in the next message.

After this part, I intend to work on:
1. Wrapping the functionality that handles the compilation commands into a struct
5. Moving the architecture-specific `<arch>ArchitectureTest` structs into an enum

crates/intrinsic-test/src/common/mod.rs
madhav-madhusoodanan: @Amanieu I was looking into this and was wondering whether dynamic dispatch would be more ergonomic.

The concern being that there is a lot of common code and enums might force us to use `match`-blocks for similar functionality.

Amanieu: In this end this was only a suggestion. If you feel that dynamic dispatch is a better fit then feel free to go ahead.

madhav-madhusoodanan:
Updating what I did over the last 4 commits:

1. Converted the way the C compilation commands were being constructed from raw `format!` calls into a structured manner.
2. Updated the execution command for Rust test files, removing the `aarch64_be`-specific commands (as suggested by @adamgemmell in this [comment](https://github.com/rust-lang/stdarch/pull/1758#discussion_r2044942632)).
3. Implemented dynamic dispatch for the construction of `<arch>ArchitectureTest` structs. 
    - This was done since the structs are intended to hold similar functionality (viz. building C/Rust files and comparing outputs)
    - The arch-specific modules accommodates any arch-specific differences
    - Dynamic dispatch seems like a great way to call these functions without needing to implement `match` blocks.
 4. Cleaned up any unnecessary function arguments
 
What I'll do next:
1. It seems like `arm/function.rs` has some code that could be made common to other arch too. Will look into it.
2. Files need to be renamed for clarity.

crates/intrinsic-test/src/arm/functions.rs
```diff
@@ -0,0 +1,265 @@
...
+// The number of times each intrinsic will be called.
+const PASSES: u32 = 20;
+
+fn gen_code_c(
+    indentation: Indentation,
+    intrinsic: &Intrinsic<ArmIntrinsicType, Constraint>,
+    constraints: &[&Argument<ArmIntrinsicType, Constraint>],
+    name: String,
+    target: &str,
+) -> String {
```
madhav-madhusoodanan, line 24: I think it might be best for the `gen_code_c` and `gen_code_rust` functions to stay module-specific. What do you think @Amanieu @adamgemmell ?

Constraints seem like a very ARM thing.

crates/intrinsic-test/src/arm/functions.rs
```diff
@@ -0,0 +1,265 @@
...
+
+    if !target.contains("v7") {
+        command = command.add_arch_flags(vec!["fanimalux", "lut", "sha3"]);
+    }
+
+    command = if target == "aarch64_be-unknown-linux-gnu" {
+        command
+            .set_linker(
+                cxx_toolchain_dir.unwrap_or("").to_string() + "/bin/aarch64_be-none-linux-gnu-g++",
+            )
```
madhav-madhusoodanan, line 185: I also wanted to ask if the `if` blocks are being too unclean here. I adopted a factory pattern to address this.
Does this look appropriate?

crates/intrinsic-test/src/arm/functions.rs
adamgemmell: I think overall you should try to keep as much generic as possible until you find an example of an intrinsic that wouldn't work otherwise - no point trying to guess what the thousands of intrinsics will want at this point.

If an architecture doesn't have any Constraints you could pass it an empty slice and that would be fine. Constraints are the "top-level" of generated code per intrinsic so it would be a bit awkward for the generator to go to platform-specific code and then jump back to generic code that might need to be aware of what constraints were set up and how they were named.

Constraints really are for anything const (i.e. needs to end up as an immediate at the assembly level), including enums. Scanning x86_64 intrinsics I can found a few examples that would require a constraint (if you were to keep this model around):

- https://doc.rust-lang.org/core/arch/x86_64/fn._mm512_getmant_round_ps.html takes some enums - the test tool doesn't currently need to handle any enums but I [chose a new Constraint variant](https://github.com/rust-lang/stdarch/pull/1509/files#diff-70215bec33928357e4667880d262ce905ab7acaa123e5465bbcc447c05bb2fb8) to represent each enum in SVE. This might not be best overall given there's loads of enums x86.
- https://doc.rust-lang.org/core/arch/x86_64/fn._mm512_fpclass_pd_mask.html takes a const IMM8 from 0 to 0xFF, not using the full i32 range of the function parameter.

I think what you end up doing depends on how your intrinsic reference is written too.

crates/intrinsic-test/src/arm/functions.rs
adamgemmell: If statements for these targets are unavoidable, don't worry about them.

Why do you have a `set_target` call and then manually below set the target again if it's clang?

madhav-madhusoodanan: Ohh thanks for pointing that out.

Initially I thought `-target` and `—target` had different meanings. I intended to remove them, I forgot about that.

crates/intrinsic-test/src/arm/functions.rs
madhav-madhusoodanan: I see

madhav-madhusoodanan: Okay now I understand constraints. I was under the impression that they described something about the register lanes or the like.

Really, they just mention the possible min-max values of the argument.

adamgemmell: Yes - but also a Constraint is required for every const value otherwise the generated intrinsic call won't pass a const. Perhaps a rename is appropriate.

They may end up needing a slight rework. Currently every possible combination of constraints is tested with the core loop, which works out fine for Neon but there was [an SVE intrinsic ](https://developer.arm.com/architectures/instruction-sets/intrinsics/svdupq[_n]_b8)that resulted in a huge test file. I bet you'll find one in the x86 intrinsics too.

adamgemmell:
Just in case you're surprised down the line - aarch64 big endian currently isn't tested in CI. https://github.com/rust-lang/stdarch/pull/1785 or https://github.com/rust-lang/stdarch/pull/1786 will fix that. It might be worth running that locally if you haven't already.

madhav-madhusoodanan:
Just noticed that the `i686-pc-windows-msvc` target CI run is failing for some CI flows for PRs.

madhav-madhusoodanan:
> Just in case you're surprised down the line - aarch64 big endian currently isn't tested in CI. #1785 or #1786 will fix that. It might be worth running that locally if you haven't already.

Got it, let me try doing that.

crates/intrinsic-test/src/arm/config.rs
```diff
@@ -0,0 +1,34 @@
+pub fn build_notices(line_prefix: &str) -> String {
+    format!(
+        "\
+{line_prefix}This is a transient test file, not intended for distribution. Some aspects of the
+{line_prefix}test are derived from a JSON specification, published under the same license as the
+{line_prefix}`intrinsic-test` crate.\n
+"
+    )
+}
```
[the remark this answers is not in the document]
adamgemmell: We would like to keep this notice for Arm intrinsics if possible. For us it's not technically mandatory to include, but it reduces risk for us.

I'm not in a position to speak for what's best for other architectures unfortunately.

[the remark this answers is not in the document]
madhav-madhusoodanan: I see

madhav-madhusoodanan:
In the last commit, I migrated from Argument<T, M> to an Argument<T> configuration, since I realized that the `Constraint` type would be common across architectures.

Presently, I think I’m at that stage where I can optimise further when I add in more architectures to `intrinsic-test`.

Would it be a good strategy on making this PR merge-worthy and then continue making such re-organization efforts in future PRs (as and when I add more architectures)?

There was one [suggestion](https://github.com/rust-lang/stdarch/pull/1758#pullrequestreview-2768927071) made by @adamgemmell  about updating the types.rs (now [common/intrinsic_helpers.rs](https://github.com/rust-lang/stdarch/blob/e72b4c18946d370e7bf93bf7d3f0f9babd7c55ec/crates/intrinsic-test/src/common/intrinsic_helpers.rs)) to make it supportive for x86 too.

Would it be a good idea if I can work on that in the next PR (in which I’d bring behavioural testing coverage for x86)?

madhav-madhusoodanan:
Okay, I'm more or less done with resolving minor issues now.
Is this PR good enough to be merged? @adamgemmell @Amanieu ?

crates/intrinsic-test/src/common/write_file.rs
```diff
@@ -0,0 +1,53 @@
...
+    intrinsics_name_list
+}
+
+pub fn write_rust_testfiles<T: IntrinsicTypeDefinition>(
+    intrinsics: Vec<&dyn IntrinsicDefinition<T>>,
+    rust_target: &str,
+    notice: &str,
+    definitions: &str,
+    cfg: &str,
+) -> Vec<String> {
```
madhav-madhusoodanan, line 37: One change I made in the last commit was adding the "definitions" argument so that we could accommodate the f16 formatting related changes made in this [commit](https://github.com/rust-lang/stdarch/commit/ffa1235a80a6062bd1ba5bdfdf17b1b6f05cf72a#diff-7c34e890416eb66d4462c7f31278409c662bb808a57c3e0a4d13c232c64533a9)

crates/intrinsic-test/src/common/intrinsic.rs
```diff
@@ -0,0 +1,231 @@
...
+        passes: u32,
+    ) -> String {
+        let constraints = self.arguments().as_constraint_parameters_rust();
+        let constraints = if !constraints.is_empty() {
+            format!("::<{constraints}>")
+        } else {
+            constraints
+        };
+
+        let indentation2 = indentation.nested();
```
Amanieu, line 80: The f16 handling in the ARM-specific code should actually be moved into the common code: f16 support is also going to be need for some x86 intrinsics.

crates/intrinsic-test/src/common/intrinsic.rs
```diff
@@ -0,0 +1,231 @@
...
+            "{indentation}for (int i=0; i<{passes}; i++) {{\n\
+                {loaded_args}\
+                {body_indentation}auto __return_value = {intrinsic_call}({args});\n\
+                {print_result}\n\
+            {indentation}}}",
+            loaded_args = self.arguments().load_values_c(body_indentation),
+            intrinsic_call = self.name(),
+            args = self.arguments().as_call_param_c(),
+            print_result = self.print_result_c(body_indentation, additional)
+        )
```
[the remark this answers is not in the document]
Amanieu: In fact I think all the methods below and including this one should just be removed from the trait: there isn't really any reason for target-specific customization of these.

crates/intrinsic-test/src/common/intrinsic.rs
```diff
@@ -0,0 +1,231 @@
...
+        let constraints = arguments
+            .iter()
+            .filter(|&i| i.has_constraint())
+            .collect_vec();
+
+        let indentation = Indentation::default();
+        generate_c_program(
+            notices,
+            header_files,
+            "aarch64",
```
Amanieu, line 153: `"aarch64"` in the common code.

Amanieu (changes requested):

crates/intrinsic-test/src/common/intrinsic.rs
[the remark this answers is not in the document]
madhav-madhusoodanan: I thought having the common functionality in the trait would make it more ergonomic.
I overlooked the part where the specific functionality won't need to be adjusted for different architectures.

I'll update this, thank you for pointing out.

crates/intrinsic-test/src/common/intrinsic.rs
madhav-madhusoodanan: Got it

crates/intrinsic-test/src/common/intrinsic_helpers.rs
```diff
@@ -0,0 +1,408 @@
...
+            name.clone()
+        }
+    }
+}
+
+pub trait IntrinsicTypeDefinition: BaseIntrinsicTypeDefinition {
+    /// Determines the load function for this type.
+    /// can be implemented in an `impl` block
+    fn get_load_function(&self, _language: Language) -> String {
+        unimplemented!("Different architectures must implement get_load_function!")
```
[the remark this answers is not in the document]
madhav-madhusoodanan: Here too, I thought it might help with debugging in case the function isn't implemented downstream. 
I'll remove this.

adamgemmell (approved):
Looks like a nice first step to me, thanks!

Amanieu (approved):

madhav-madhusoodanan:
Thank you so much @adamgemmell @Amanieu !

There was quite a lot of learning for me on the architecture design front.

crates/intrinsic-test/src/common/compare.rs
```diff
@@ -0,0 +1,90 @@
...
+            let c = Command::new("sh")
+                .arg("-c")
+                .arg(format!("{runner} ./c_programs/{intrinsic_name}"))
+                .output();
+
+            let rust = Command::new("sh")
+                .current_dir("rust_programs")
+                .arg("-c")
+                .arg(format!(
+                    "cargo {toolchain} run --target {target} --bin {intrinsic_name} --release",
```
adamgemmell, line 23: This should be run with the runner too - it's masked on CI due to the CARGO_TARGET_*_RUNNER env variable that's set in the dockerfiles

madhav-madhusoodanan: Ohh I see

PR #1762, opened 2025-03-30
std_detect: Use riscv_hwprobe on RISC-V Linux/Android

taiki-e:
On RISC-V, detection using auxv only supports single-letter extensions.
 So, this PR uses riscv_hwprobe that supports multi-letter extensions if available to support more target features. <https://www.kernel.org/doc/html/latest/arch/riscv/hwprobe.html>

I had originally planned to open a PR after the nightly release that included https://github.com/rust-lang/rust/pull/138742 merged today, but there was a question about this (https://github.com/rust-lang/rust/issues/139139), so I opened it earlier.

Closes https://github.com/rust-lang/rust/issues/139139

Tested with QEMU 9.2.1 user mode:

<details><summary>riscv64gc-unknown-linux-gnu</summary>

```
rv32i: false
rv32e: false
rv64i: true
rv128i: false
zicsr: true
zicntr: false
zihpm: false
zifencei: false
zihintpause: false
m: true
a: true
zacas: true
zawrs: false
zam: false
ztso: true
f: true
d: true
q: false
zfh: true
zfhmin: true
zfinx: false
zdinx: false
zhinx: false
zhinxmin: false
c: true
zba: true
zbb: true
zbc: true
zbs: true
zbkb: true
zbkc: true
zbkx: true
zknd: true
zkne: true
zknh: true
zksed: true
zksh: true
zkr: false
zksed: true
zksh: true
zkr: false
zkn: true
zks: true
zkt: true
v: true
zvfh: true
zvfhmin: true
zve32x: true
zve32f: true
zve64x: true
zve64f: true
zve64d: true
zvkb: true
zvbb: true
zvbc: true
zvkg: true
zvkned: true
zvknha: true
zvknhb: true
zvksed: true
zvksh: true
zvkn: true
zvknc: true
zvkng: true
zvks: true
zvksc: true
zvksg: true
zvkt: true
unaligned-scalar-mem: true
unaligned-vector-mem: false
j: false
p: false
```

</details>

<details><summary>riscv32gc-unknown-linux-gnu</summary>

```
rv32i: true
rv32e: false
rv64i: false
rv128i: false
zicsr: true
zicntr: false
zihpm: false
zifencei: false
zihintpause: false
m: true
a: true
zacas: true
zawrs: false
zam: false
ztso: true
f: true
d: true
q: false
zfh: true
zfhmin: true
zfinx: false
zdinx: false
zhinx: false
zhinxmin: false
c: true
zba: true
zbb: true
zbc: true
zbs: true
zbkb: true
zbkc: true
zbkx: true
zknd: true
zkne: true
zknh: true
zksed: true
zksh: true
zkr: false
zksed: true
zksh: true
zkr: false
zkn: true
zks: true
zkt: true
v: true
zvfh: true
zvfhmin: true
zve32x: true
zve32f: true
zve64x: true
zve64f: true
zve64d: true
zvkb: true
zvbb: true
zvbc: true
zvkg: true
zvkned: true
zvknha: true
zvknhb: true
zvksed: true
zvksh: true
zvkn: true
zvknc: true
zvkng: true
zvks: true
zvksc: true
zvksg: true
zvkt: true
unaligned-scalar-mem: true
unaligned-vector-mem: false
j: false
p: false
```

</details>

### DIFF
--- a/crates/std_detect/src/detect/arch/riscv.rs
+++ b/crates/std_detect/src/detect/arch/riscv.rs
@@ -125,6 +125,10 @@ features! {
 
     @FEATURE: #[stable(feature = "riscv_ratified", since = "1.78.0")] a: "a";
     /// "A" Extension for Atomic Instructions
+    @FEATURE: #[unstable(feature = "stdarch_riscv_feature_detection", issue = "111192")] zacas: "zacas";
+    /// Atomic Compare-And-Swap Instructions
+    @FEATURE: #[unstable(feature = "stdarch_riscv_feature_detection", issue = "111192")] zawrs: "zawrs";
+    /// Wait on Reservation Set
     @FEATURE: #[unstable(feature = "stdarch_riscv_feature_detection", issue = "111192")] zam: "zam";
     without cfg check: true;
     /// "Zam" Extension for Misaligned Atomics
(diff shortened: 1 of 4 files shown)


crates/std_detect/src/detect/os/linux/riscv.rs
```diff
@@ -1,11 +1,269 @@
...
+        }
+        // FIXME: should be enough with hwprobe only, but our code below checks h and e
+        // unavailable in neither uapi/asm/hwprobe.h nor uapi/asm/hwcap.h.
+        // https://github.com/torvalds/linux/blob/master/arch/riscv/include/uapi/asm/hwcap.h
+        // return value;
+    }
+
+    // FIXME: As said in the above FIXME, we currently alway checks auxv too.
+    // // riscv_hwprobe requires Linux 6.4, so we fallback to auxv-based detection on
+    // // old Linux kernel.
```
taiki-e, line 265: And if so, we can resolve there FIXMEs.

crates/std_detect/src/detect/os/linux/riscv.rs
```diff
@@ -1,11 +1,269 @@
...
+fn _riscv_hwprobe(out: &mut [riscv_hwprobe]) -> bool {
+    unsafe fn __riscv_hwprobe(
+        pairs: *mut riscv_hwprobe,
+        pair_count: libc::size_t,
+        cpu_set_size: libc::size_t,
+        cpus: *mut libc::c_ulong,
+        flags: libc::c_uint,
+    ) -> libc::c_long {
+        unsafe {
+            libc::syscall(
```
taiki-e, line 97: We could use inline assembly for Linux here (like [portable-atomic does](https://github.com/taiki-e/portable-atomic/blob/a1e1319fb7dddb6ee095f3c5c70072125d9fa8fc/src/imp/detect/riscv_linux.rs#L49-L85)), but since this module is currently behind the libc feature, and since std currently always depends on libc on all Linux targets, there is no need to complicate the code here.

a4lg:
That's pretty much the same as what I've testing and thanks for making progress earlier than me.
So, I'll review some of the details as a RISC-V toolchain maintainer (mainly on GNU).

crates/std_detect/src/detect/os/linux/riscv.rs
```diff
@@ -46,16 +309,20 @@ pub(crate) fn detect_features() -> cache::Initializer {
...
+    #[cfg(target_arch = "riscv32")]
     enable_feature(
         &mut value,
         Feature::rv32e,
         bit::test(auxv.hwcap, (b'e' - b'a').into()),
     );
+    // FIXME: h is not exposed in uapi/asm/hwcap.h and uapi/asm/hwprobe.h
     enable_feature(
         &mut value,
         Feature::h,
```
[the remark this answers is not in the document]
Amanieu: I would argue that all the supervisor features should probably just be removed from feature detection and from the compiler. They are not relevant for the language/compiler.

a4lg:
@taiki-e @Amanieu

We need to discuss about one topic before merging this: the fact that Linux supports disabling vector support *per thread* (current one / next context created by following `execve` call) by a `prctl` call: `PR_RISCV_V_SET_CONTROL`.

The point is, support for vector extensions retrieved by a `riscv_hwprobe` call is not affected whether the vector extension *for the current thread* (or the next context created by following `execve` call) is enabled.

IMHO, the solution here would be, checking (on the first feature test call) whether vector extensions for the current thread are enabled by `prctl` with `PR_RISCV_V_GET_CONTROL` and disabling vector-related extensions if not. This method cannot track per-thread status but should be fine on most cases. Force-enabling vector extensions is also a solution but can be a problem on a library.

taiki-e:
Given the structure of Rust/LLVM's target feature, I wonder if it is UB to call code that is not written in (inline or external) assembly in the time between disallowing vector instructions and re-allowing it.

For example, if I compile code with `-C target-feature=+v`, the [compiler will use vector instructions for code that copies a size larger than a register from memory](https://godbolt.org/z/3Msdbfv1b), but if I disallowing vector instructions and then reach that code, I will get a trap, right?

Amanieu:
My reading of https://docs.kernel.org/arch/riscv/vector.html is that it's mainly intended for system-level management of vector availability. They explicitly say that:
> To get the availability of V in an ELF program, please read COMPAT_HWCAP_ISA_V bit of ELF_HWCAP in the auxiliary vector.

So we should just do that: hwcaps will reflect whether vector support is enabled for the current process.

Note that hwprobe does *not* disable V if vector support is disabled for the process, so we need to use hwcaps in this case (or manually check with `prctl(PR_RISCV_V_GET_CONTROL)`).

a4lg:
@Amanieu I confirmed that `ELF_HWCAP` is in fact affected by availability of the `V` extension when the program starts (unlike `riscv_hwprobe`; I remembered `ELF_HWCAP` incorrectly and I worried a breaking change if `ELF_HWCAP` is not affected by `prctl` (in fact, that was not the case)).
Still, I think we'd better use `prctl` not to accidentally enable vector extensions through `riscv_hwprobe`.

taiki-e:
Updated PR (mostly based on feedback).

- Changed V extension detection to use hwcap.
- Removed RISC-V privileged features (supervisor/hypervisor).
  - Because these target features are not in rustc's target_feature list and also detecting them from std/std_detect running in user mode is useless.
- Enabled cfg check for RISC-V vector-related target features.
  - Because https://github.com/rust-lang/rust/pull/138742 is now available on the nightly toolchain.

---

> Note that hwprobe does _not_ disable V if vector support is disabled for the process, so we need to use hwcaps in this 
case (or manually check with `prctl(PR_RISCV_V_GET_CONTROL)`).

> Still, I think we'd better use `prctl` not to accidentally enable vector extensions through `riscv_hwprobe`.

I *thought* that the use of hwprobe and prctl is probably more preferable because hwcap only supports detection of V extension, but hwprobe also supports detection of cases where only subset of V extension are available.

However, qemu-user 9.2.1 does not seem to support that, so for now I implemented it using hwcap.

```
[crates/std_detect/src/detect/os/linux/riscv.rs:211:25] libc::prctl(70) = -1
```

a4lg:
Let me give a time to test that (I'm struggling to prepare RISC-V 32-bit Linux environment with QEMU system emulation while preparing 64-bit counterpart worked well).

taiki-e:
(I tested this with qemu-riscv32 (user mode) using toolchain used in setup-cross-toolchain-action, and updated PR description to include the result.)

a4lg:
Okay, finally I've built full Linux 6.14 + Buildroot + glibc RISC-V environments (32/64) + Rust toolchain and tested your code is working as expected (I could not accept your changes while no real Linux environment is tested). I will likely make some tidying but after you make most of the changes (because I found your `prctl`-based version branch).

Recommendation: approval

crates/std_detect/src/detect/os/linux/riscv.rs
```diff
@@ -1,11 +1,118 @@
...
+// const RISCV_HWPROBE_EXT_ZVE64D: u64 = 1 << 41;
+// const RISCV_HWPROBE_EXT_ZIMOP: u64 = 1 << 42;
+// const RISCV_HWPROBE_EXT_ZCA: u64 = 1 << 43;
+// const RISCV_HWPROBE_EXT_ZCB: u64 = 1 << 44;
+// const RISCV_HWPROBE_EXT_ZCD: u64 = 1 << 45;
+// const RISCV_HWPROBE_EXT_ZCF: u64 = 1 << 46;
+// const RISCV_HWPROBE_EXT_ZCMOP: u64 = 1 << 47;
+const RISCV_HWPROBE_EXT_ZAWRS: u64 = 1 << 48;
+// const RISCV_HWPROBE_EXT_SUPM: u64 = 1 << 49;
+
```
taiki-e, line 78: By the way, Linux 6.15 will also add the following.
(Although we cannot include them in this PR because 6.15 is not yet released and could be changed.)

```rs
const RISCV_HWPROBE_EXT_ZICNTR: u64 = 1 << 50;
const RISCV_HWPROBE_EXT_ZIHPM: u64 = 1 << 51;
const RISCV_HWPROBE_EXT_ZFBFMIN: u64 = 1 << 52;
const RISCV_HWPROBE_EXT_ZVFBFMIN: u64 = 1 << 53;
const RISCV_HWPROBE_EXT_ZVFBFWMA: u64 = 1 << 54;
const RISCV_HWPROBE_EXT_ZICBOM: u64 = 1 << 55;
const RISCV_HWPROBE_EXT_ZAAMO: u64 = 1 << 56;
const RISCV_HWPROBE_EXT_ZALRSC: u64 = 1 << 57;
```

a4lg: Wow, I'm writing full `riscv_hwprobe` feature probing (including OS-independent logic for `stdarch`) and related feature sets (for Rust) and... I think that will exceed 93, the current limit of the 12-byte cache.

crates/std_detect/src/detect/os/linux/riscv.rs
```diff
@@ -1,11 +1,268 @@
...
+            enable_feature(Feature::zkt, zkt);
+            let zkn = zbkb & zbkc & zbkx & zkne & zknd & zknh;
+            enable_feature(Feature::zkn, zkn);
+            // enable_feature(Feature::zk, zkn & zkr & zkt);
+            enable_feature(Feature::zks, zbkb & zbkc & zbkx & zksed & zksh);
+            // Standard Vector Extensions
+            enable_feature(Feature::v, ima_ext_0 & RISCV_HWPROBE_IMA_V != 0);
+            enable_feature(Feature::zvfh, ima_ext_0 & RISCV_HWPROBE_EXT_ZVFH != 0);
+            enable_feature(Feature::zvfhmin, ima_ext_0 & RISCV_HWPROBE_EXT_ZVFHMIN != 0);
+            enable_feature(Feature::zve32x, ima_ext_0 & RISCV_HWPROBE_EXT_ZVE32X != 0);
```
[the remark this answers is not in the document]
a4lg: Sorry, that was my outdated review.

a4lg:
I noticed that I haven't pressed the "Submit review" button.

crates/std_detect/src/detect/arch/riscv.rs
```diff
@@ -194,23 +190,62 @@ features! {
...
-    without cfg check: true;
-    /// "Svpbmt" Extension for Page-Based Memory Types
-    @FEATURE: #[unstable(feature = "stdarch_riscv_feature_detection", issue = "111192")] svinval: "svinval";
-    without cfg check: true;
-    /// "Svinval" Extension for Fine-Grained Address-Translation Cache Invalidation
-    @FEATURE: #[unstable(feature = "stdarch_riscv_feature_detection", issue = "111192")] h: "h";
-    without cfg check: true;
-    /// "H" Extension for Hypervisor Support
+    // FIXME: we can implement this by getting the current vlen
+    // zvl*b: Minimum Vector Length Standard Extensions
```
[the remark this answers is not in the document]
taiki-e: It indeed makes more sense to have it as a core::arch intrinsic, since nothing OS-specific is needed to get the VLEN, and the actual length information can be obtained in a single call.

Removed that code comment.

crates/std_detect/src/detect/os/linux/riscv.rs
```diff
@@ -46,29 +153,172 @@ pub(crate) fn detect_features() -> cache::Initializer {
...
+        },
+    ];
+    if _riscv_hwprobe(&mut out) {
+        let mut enable_feature = |feature, enable| {
+            if enable {
+                value.set(feature as u32);
+            }
+        };
+        if out[0].key != -1 {
+            let ima_ext_0 = out[0].value;
```
a4lg, line 209: It seems using this key alone is not robust enough (checking the `IMA` base is necessary).

I opened #1770 for my similar proposal with OS-independent extension implication logic and with fixes including this part.

a4lg (changes requested):

a4lg:
I opened #1770 to shape the best feature detection logic in my mind.

a4lg:
Interestingly, in the process reviewing your PR, I found that your direct `syscall` approach (not calling glibc's `__riscv_hwprobe` and userland vDSO) succeeds to workaround a weird behavior I found when we call glibc's `__riscv_hwprobe` which calls vDSO prepared by the Linux kernel (likely a Linux-side bug possibly related to vDSO).

crates/std_detect/src/detect/os/linux/riscv.rs
taiki-e: Good catch. Earlier versions of this PR actually included that check, but it was accidentally removed when I changed to use hwcap.
https://github.com/rust-lang/stdarch/commit/3203fd550d499bde03f74f02a810b9d1b3ac2a02#diff-dfab5b675012c4be84a409e8c4f92c8e172e797caf07a23516d1fda390fb9818R152

taiki-e:
Closing in favor of https://github.com/rust-lang/stdarch/pull/1770.

PR #1767, opened 2025-04-10
fix broken intra doc links

bend-n:
there appears to have been some broken doc links here, which are inhibiting my stdarch update
this pr fixes those broken doc links.

### DIFF
--- a/crates/core_arch/src/x86/avx512fp16.rs
+++ b/crates/core_arch/src/x86/avx512fp16.rs
@@ -264,8 +264,8 @@ pub fn _mm512_setzero_ph() -> __m512h {
 
 /// Return vector of type `__m128h` with indetermination elements.
 /// Despite using the word "undefined" (following Intel's naming scheme), this non-deterministically
-/// picks some valid value and is not equivalent to [`mem::MaybeUninit`].
-/// In practice, this is typically equivalent to [`mem::zeroed`].
+/// picks some valid value and is not equivalent to [`mem::MaybeUninit`](crate::mem::MaybeUninit).
+/// In practice, this is typically equivalent to [`mem::zeroed`](crate::mem::zeroed).
 ///
 /// [Intel's documentation](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_undefined_ph)
 #[inline]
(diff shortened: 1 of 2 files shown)


crates/core_arch/src/x86/avx512fp16.rs
```diff
@@ -264,8 +264,8 @@ pub fn _mm512_setzero_ph() -> __m512h {
 
 /// Return vector of type `__m128h` with indetermination elements.
 /// Despite using the word "undefined" (following Intel's naming scheme), this non-deterministically
-/// picks some valid value and is not equivalent to [`mem::MaybeUninit`].
-/// In practice, this is typically equivalent to [`mem::zeroed`].
+/// picks some valid value and is not equivalent to [`mem::MaybeUninit`](core::mem::MaybeUninit).
```
Amanieu, line 267: Since `core_arch` is directly included in `core`, I don't think this will work. Wouldn't the correct link be `crate::mem::MaybeUninit`?

bend-n: im not sure that works... i guess the safest way to do it is to directly link to https://doc.rust-lang.org/std/mem/union.MaybeUninit.html ?

Amanieu: Have you checked that using `core::mem::MaybeUninit` actually builds correctly in rust-lang/rust and passes tests? If so then it's probably fine.

bend-n: i dont actually know as i cant figure out how to reproduce the CI failure locally.

sayantn: You can probably repro it locally by running `./x doc`. This will build the docs, and should repro the failure (if you have x86 system, then it should definitely work, otherwise you might need to add and specify a x86 target, as these doc problems are in x86)

sayantn: weird! it repros in my x86 machine. But anyway, I tried with `crate::mem`, and it worked. The rustc repo always uses `crate` for `core`, because that enables use of the latest features (while compiling rustc, `core` is the previous stage compiler)

bend-n: ill take your word for it

sayantn:
I think instead of adding specific lints to `lib.rs`, we should add `-D warnings` to `ci/dox.sh`, otherwise things like this will continue to happen as rustdoc gets more and more lints

crates/core_arch/src/x86/avx512fp16.rs
```diff
@@ -277,8 +277,8 @@ pub fn _mm_undefined_ph() -> __m128h {
 
 /// Return vector of type `__m256h` with indetermination elements.
 /// Despite using the word "undefined" (following Intel's naming scheme), this non-deterministically
-/// picks some valid value and is not equivalent to [`mem::MaybeUninit`].
-/// In practice, this is typically equivalent to [`mem::zeroed`].
+/// picks some valid value and is not equivalent to [`mem::MaybeUninit`](crate::mem::::MaybeUninit).
```
Amanieu, line 280: You seem to have added too many `::` here.

sayantn:
You have to add `-- -D warnings` to `ci/dox.sh` (note the initial `--`)

bend-n:
i tried that with just `cargo doc` and it didnt seem to work locally?

sayantn:
Ok finally figured it out, `RUSTDOCFLAGS=-D warnings` environment variable is needed to pass arguments to `rustdoc` through `cargo`

PR #1768, opened 2025-04-10
Disable cfg check for the recently-merged target features to allow stdarch update

sayantn:
The stdarch update is currently giving doc errors, which #1767 fixes. But then it will give unknown cfg errors due to the recently added amx and riscv target features, so this pr temporarily disables those checks. Revert after the bump is successful

### DIFF
--- a/crates/std_detect/src/detect/arch/riscv.rs
+++ b/crates/std_detect/src/detect/arch/riscv.rs
@@ -104,15 +104,20 @@ features! {
     /// RV128I Base Integer Instruction Set
 
     @FEATURE: #[unstable(feature = "stdarch_riscv_feature_detection", issue = "111192")] zicsr: "zicsr";
+    without cfg check: true;
     /// "Zicsr" Extension for Control and Status Register (CSR) Instructions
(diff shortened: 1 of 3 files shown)


a4lg:
@sayantn
I'd like to ask how do we know that unknown cfg errors are fixed (and/or which CI failures to track) to know the timing to revert this PR.
Just making documentation seems working on my local environment so I think there is a failing configuration.

sayantn:
These unknown cfg errors are obtained when stdarch is built as a part of libstd, by itself stdarch should compile fine - the bootstrapping architecture of rustc creates the problem because the stage0 compiler doesn't yet know about those new target features.

It should be good to revert once a stdarch update gets merged, or the bootstrap version gets bumped

sayantn:
afaik this repo doesn't use `cfg(bootstrap)` because the `update cfgs` commit (which removes all `cfg(bootstrap)` items after the bootstrap bump) wouldn't work here because this is still a submodule, not a subtree. Using `cfg(bootstrap)` here will require a PR here after bootstrap bump, and a stdarch submodule update in rustc, which might create a lot of problems, because stdarch updates are typically problematic as stdarch CI and rustc CI don't match perfectly

PR #1769, opened 2025-04-11
RISC-V: Tidying (batch 2025-04-11)

a4lg:
This PR updates the RISC-V definition file (`arch/riscv.rs`; mainly its documentation) for various parts (better links, indentation, more consistency on the extension naming and its capitalization).

This is a subset of upcoming PR but separate for being:

1. independent from the rest and
2. probably unarguable.

This PR consists of four commits:

1. Improve extension consistency (capitalization)
2. Move ratified extensions and make some clarification changes (by sorting or raising / lowering indentation)
3. Changing the ISA manual link ([before](https://github.com/riscv/riscv-isa-manual/) / [after](https://riscv.org/specifications/ratified/))  
   while GitHub is a good place to browse the latest draft, it's hard to determine which is the ratified specification.  So, I think a page in RISC-V International would be better.
4. Fix minor section separation issue

### DIFF
--- a/crates/std_detect/src/detect/arch/riscv.rs
+++ b/crates/std_detect/src/detect/arch/riscv.rs
@@ -20,6 +20,8 @@ features! {
     /// ISA prefix X. These sets are highly platform specific and should be
     /// detected with their own platform support crates.
     ///
+    /// [ISA manual]: https://riscv.org/specifications/ratified/
+    ///
     /// # Unprivileged Specification
     ///
     /// The supported ratified RISC-V instruction sets are as follows:
@@ -83,8 +85,6 @@ features! {
     /// * Svnapot: `"svnapot"`
     /// * Svpbmt: `"svpbmt"`
     /// * Svinval: `"svinval"`
-    ///
-    /// [ISA manual]: https://github.com/riscv/riscv-isa-manual/
     #[stable(feature = "riscv_ratified", since = "1.78.0")]
 
     @FEATURE: #[unstable(feature = "stdarch_riscv_feature_detection", issue = "111192")] rv32i: "rv32i";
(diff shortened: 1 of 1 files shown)


crates/std_detect/src/detect/arch/riscv.rs
```diff
@@ -44,34 +46,31 @@ features! {
...
+    ///   * Zfhmin: `"zfhmin"`
     /// * Zfinx: `"zfinx"`
     /// * Zdinx: `"zdinx"`
     /// * Zhinx: `"zhinx"`
-    /// * Zhinxmin: `"zhinxmin"`
+    ///   * Zhinxmin: `"zhinxmin"`
+    /// * Zbkb: `"zbkb"`
+    /// * Zbkc: `"zbkc"`
+    /// * Zbkx: `"zbkx"`
+    /// * Zk: `"zk"`
```
Amanieu, line 58: Is it intentional that all the `Zk*` extensions aren't indented under `Zk`?

crates/std_detect/src/detect/arch/riscv.rs
```diff
@@ -81,13 +80,11 @@ features! {
     ///
     /// Defined by Privileged Specification:
     ///
-    /// * Supervisor: `"s"`
+    /// * *Supervisor-Level ISA* (not "S" extension): `"s"`
+    /// * H (hypervisor): `"h"`
     /// * Svnapot: `"svnapot"`
     /// * Svpbmt: `"svpbmt"`
     /// * Svinval: `"svinval"`
```
Amanieu, line 87: I still think we should just remove privileged features entirely. They're unlikely to pass stabilization.

crates/std_detect/src/detect/arch/riscv.rs
a4lg: Probably (for someone made this).
But I would argue that this will lead to a misunderstanding so that `Zk` contains `Zks` (that is incorrect).

Rather, I'd prefer lower indentation with fewer grouping (leaving groupings with (1) no superset-subset misconception will not occur or (2) there is a true superset-subset relation).

crates/std_detect/src/detect/arch/riscv.rs
a4lg: In fact, #1770 actually removes this.

My intentions of this change are (1) like putting makeup on a dead (before its funeral) and (2) ensure that the documentation gets better even in case if PR #1770 (which actually removes this line) stalls for some reason.

PR #1503, opened 2023-11-22
Generate constant Neon vectors within a const block

GeorgeWort:
This allows the constant vector to be passed directly to LLVM which fixes a few issues where the LLVM intrinsic expects a constant vector.

This relies on https://github.com/rust-lang/rust/pull/118171 to be useful

PR #1771, opened 2025-04-11
Fix intrinsic-test for debug test suites

adamgemmell:
The intrinsic-test tool itself was being built in dev/release profiles but the test programs it generates were all being tested in release mode. This PR pipes the GitHub Actions profile into the tool through a command line argument.

### DIFF
--- a/ci/run-docker.sh
+++ b/ci/run-docker.sh
@@ -30,6 +30,7 @@ run() {
       --env NORUN \
       --env RUSTFLAGS \
       --env STDARCH_TEST_NORUN \
+      --env PROFILE \
       --volume "${HOME}/.cargo":/cargo \
       --volume "$(rustc --print sysroot)":/rust:ro \
       --volume "$(pwd)":/checkout:ro \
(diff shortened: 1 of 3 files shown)


Amanieu:
This seems to have revealed compiler crashes.

ci/run-docker.sh
```diff
@@ -30,6 +30,7 @@ run() {
       --env NORUN \
       --env RUSTFLAGS \
       --env STDARCH_TEST_NORUN \
+      --env PROFILE \
```
adamgemmell, line 33: Without this line the debug CI wasn't actually doing anything different to the release CI and it turns out quite a few platforms are broken in debug mode. Will keep digging.

adamgemmell: There's some handwritten arm intrinsics that have not been fixed by https://github.com/rust-lang/stdarch/pull/1503, e.g. https://github.com/rust-lang/stdarch/blob/90914540cb3d5f48abecd1619dfc5425b76f1c8d/crates/core_arch/src/arm/neon.rs#L29

The wasm failure is due a hardcoded path to `target/release`. Not sure about freebsd, haven't looked into it yet.

adamgemmell:
Superceded by https://github.com/rust-lang/stdarch/pull/1949/

PR #1773, opened 2025-04-13
Fix pointer type for `_mm512_loadu_si512`

Amanieu:
The C intrinsic uses a void pointer but we must specify a type in Rust. This PR makes it consistent with the equivalent SSE and AVX intrinsics, as well as `_mm512_store_si512`.


Amanieu:
Closing in favor of #1775

PR #1774, opened 2025-04-13
Remove the workarounds for some resolved issues

sayantn:
- Re-enable `dox` CI for `armv7-unknown-linux-gnueabihf`
 - Remove workarounds for llvm/llvm-project#98306

### DIFF
--- a/ci/dox.sh
+++ b/ci/dox.sh
@@ -25,9 +25,7 @@ dox() {
 if [ -z "$1" ]; then
   dox i686-unknown-linux-gnu
   dox x86_64-unknown-linux-gnu
-  # Disabled temporarily,
-  # See https://github.com/rust-lang/rust/issues/134511
-  #dox armv7-unknown-linux-gnueabihf
+  dox armv7-unknown-linux-gnueabihf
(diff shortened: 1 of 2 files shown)


sayantn:
There seems to be some instruction check failures (specifically `vdup_n` and `vmov_n` intrinsics both are generating `vmov`, and before this they both generated `vdup`). @Amanieu do I need to fix that or is that ok to have?

Amanieu:
> There seems to be some instruction check failures (specifically `vdup_n` and `vmov_n` intrinsics both are generating `vmov`, and before this they both generated `vdup`). @Amanieu do I need to fix that or is that ok to have?

I'd just remove the tests on those (set the assert_instr to `nop` or just remove it). They are data movement operations and we don't guarantee a specific instruction sequence for them.

Amanieu:
That looks like a regression on the LLVM side, it should be emitting the `2` variant of the instruction.

For now this can be accepted with a FIXME, but an issue should be opened in LLVM about it.

sayantn:
Seems like rust-lang/rust#60637 is still causing some problems in LLVM optimizing for darwin and ios, so I am delegating that

PR #1776, opened 2025-04-14
Various CI and Testing enhancements

sayantn:
- Upgrades all dockers to `Ubuntu-25.04`, all x86 macos runners to `macos-15-large` (which is a new runner image with x86 macos), arm macos runners to `macos-15` and windows runners to `windows-2025` 
 - Re-enable all conditionally disabled tests in x86 (todo: try other archs too)
 - Add PPC64 to CI
 - Removes the `cupid` dependency of `std_detect`  and the `env-override-no-avx` CI run
 - Change `simd_test_macro` test skipping logic a little, separate `SKIP_TESTS` into feature-based and function-name-based parts
 - Replace `cfg(stdarch_intel_sde)` with `STDARCH_TEST_SKIP_FUNCTION` as it is powerful enough.
 - Add `test-everything` annotations to more archs to ensure test coverage
 - Add x86 run with `-Ctarget-feature=+avx512f`
 - Remove `STDARCH_DISABLE_DEDUP_GUARD` because it was unused
 - [TRY] Enable `assert_instr` for PPC64LE

PR #1779, opened 2025-04-17
[RFC] std_detect: RISC-V platform guide documentation

a4lg:
To help users make a decision for feature checking on a RISC-V system:

1.  Whether to use this runtime detection macro (or prefer to use static feature enablement with `-Ctarget-feature` compiler option) and
2.  Determine the platform and versions to support,

This commit adds a platform guide with minimum supported platform versions.

# Screenshot

Because I could not make a single HTML version of the resulting documentation, I will paste a screenshot (slightly modified for readability) here.

Image: [screenshot-is_riscv_feature_detected.png](https://github.com/user-attachments/assets/7f021c9d-3d56-4ae1-938a-08e2aac27311)

# Background / Problem

## Absence of Architectural Runtime Feature Detection on RISC-V

Unlike x86 architectures with an unprivileged `CPUID` instruction and Arm/AArch64 architectures [with a series of system registers](https://github.com/rust-lang/stdarch/blob/d8785a32e63090e98243c0b4b5d6d2f5f8518b5a/crates/std_detect/src/detect/os/aarch64.rs) (available to the operating system level (EL1) but some operating systems chose to expose them to the user mode (EL0) through instruction emulation), RISC-V does not provide any fine-grained feature detection facility itself.

The only standard RISC-V architectural thing provided is the `misa` CSR but it is only available to the firmware level (M-mode, even higher than the S-mode which the operating system normally runs on, and the firmware normally does not provide the contents of the `misa` CSR to lower modes). Even if this CSR is accessible, it's very coarse (only provides existence of single-letter extensions).

As a result, runtime feature detection on RISC-V entirely depends on the platform-specific facility which heavily depends on the platform and its version *per feature* (e.g. the main source of the runtime feature detection on Linux is the Devicetree (optionally, command line arguments and boot time benchmark) and the kernel exports a subset through the auxiliary vector and the `riscv_hwprobe` system call but note that this filtering is performed on the kernel which means only features *supported* by the kernel are exposed to the user mode; the word *support* means just the given extension is on the allow list in the kernel and even some *unsupported* extensions work without explicit kernel support).

This is a mess and the author considers just providing the list of features supported by the macro is insufficient for users who *actually* wants to use runtime feature detection.

## Existence of a Supported but Not Useful Feature: `"zkr"` (and `"zk"`)

The problem is not just the platform and the version.  For instance, there is an example of `"zkr"` (the Zkr extension) which is supported by this macro but unlikely to return `true` by a result of true runtime feature detection because the `seed` CSR provided by this extension is normally inaccessible from the user-mode.
For instance, OpenSBI ― a reference RISC-V firmware implementation ― changes the `mseccfg` CSR so that the `seed` CSR is accessible from the S-mode (OS) but not from the U-mode (user) by default. Worse than that, this behavior is not even configurable from the OS.

It also affects the Zk extension `"zk"` (fully-featured scalar cryptography) which contains the Zkr extension.

# Proposal

This commit makes the large portion of the supported feature list table-based to show platform support (with minimum supported versions) *per feature* and a few footnotes.

Tables are basically sorted (purely) alphabetically to make finding an extension easier but this change makes relations between extensions less obvious.  The author still prefers this format because the user will find an extension to use with the ISA manual  (which describes relations between extensions anyway).

It also adds some notes as a RISC-V platform guide.

## Note: About Reverse Implication (Extension Groups)

It intentionally omits the description of the reverse implication related to *extension groups* (such like implication of `B` *from* its members: `Zba`, `Zbb` and `Zbs` extensions) because it currently does not synchronize well with the `-Ctarget-feature` compiler option (due to missing reverse implication checks using `cfg` and due to constraints of the current Rust's feature handling).

Instead, it only describes forward implications (like `D` implying `F`) due to the fact that it relatively synchronizes well between Rust and `stdarch` for this kind of feature handling (not fully synchronized though).

# Request for Comments

Although this PR *can* be a regular PR right now, I first request comments because the change is very significant.

## Is this Guide to be Here?

Because this kind of change is large and the resulting documentation is far more verbose than the rest (runtime feature detection macro on other architectures), I'm not sure whether this kind of the guide is suitable here.

Even if this kind of the guide is helpful for users, it may not be the right place.

What do you think?

## Are Version Numbers Required in the Extension Table?

The author prefers to include version numbers (also due to the fact that there are large gaps of runtime feature detection on the Linux kernel version 6.4 and later) but if that's too much, the author considers just showing Yes/No (with footnotes) will be still helpful for users (far better than nothing).

## Formatting: Trailing Spaces

Changes in the "Performance Hints" section are the first examples in `stdarch` that utilize *a new line by two trailing spaces*.
Though it is not prohibited per coding rules, this is unusual as a Rust source code (and may need to ask whether this is okay).

### DIFF
--- a/crates/std_detect/src/detect/arch/riscv.rs
+++ b/crates/std_detect/src/detect/arch/riscv.rs
@@ -22,87 +22,121 @@ features! {
     ///
     /// [ISA manual]: https://riscv.org/specifications/ratified/
     ///
+    /// # Platform-specific behavior and Availability
+    ///
+    /// Runtime detection depends on the platform-specific feature detection
+    /// facility and its availability per feature is
+    /// highly platform/version-specific.
+    ///
+    /// Still, a best-effort attempt is performed to enable subset/dependent
+    /// features if a superset feature is enabled regardless of the platform.
+    /// For instance, if the A extension (`"a"`) is enabled, its subsets (the
+    /// Zalrsc and Zaamo extensions; `"zalrsc"` and `"zaamo"`) are also enabled.
+    /// Likewise, if the F extension (`"f"`) is enabled, one of its dependencies
+    /// (the Zicsr extension `"zicsr"`) is also enabled.
+    ///
     /// # Unprivileged Specification
     ///
-    /// The supported ratified RISC-V instruction sets are as follows:
+    /// The supported ratified RISC-V instruction sets are as follows (OS
+    /// columns denote runtime feature detection support with or without minimum
+    /// kernel version; "No" by default):
     ///
-    /// * RV32E: `"rv32e"`
-    /// * RV32I: `"rv32i"`
-    /// * RV64I: `"rv64i"`
-    /// * A: `"a"`
-    ///   * Zaamo: `"zaamo"`
-    ///   * Zalrsc: `"zalrsc"`
-    /// * B: `"b"`
-    ///   * Zba: `"zba"`
-    ///   * Zbb: `"zbb"`
-    ///   * Zbs: `"zbs"`
-    /// * C: `"c"`
-    ///   * Zca: `"zca"`
-    ///   * Zcd: `"zcd"` (if D is enabled)
-    ///   * Zcf: `"zcf"` (if F is enabled on RV32)
-    /// * D: `"d"`
-    /// * F: `"f"`
-    /// * M: `"m"`
-    /// * Q: `"q"`
-    /// * V: `"v"`
-    ///   * Zve32x: `"zve32x"`
-    ///   * Zve32f: `"zve32f"`
-    ///   * Zve64x: `"zve64x"`
-    ///   * Zve64f: `"zve64f"`
-    ///   * Zve64d: `"zve64d"`
-    /// * Zicboz: `"zicboz"`
-    /// * Zicntr: `"zicntr"`
-    /// * Zicond: `"zicond"`
-    /// * Zicsr: `"zicsr"`
-    /// * Zifencei: `"zifencei"`
-    /// * Zihintntl: `"zihintntl"`
-    /// * Zihintpause: `"zihintpause"`
-    /// * Zihpm: `"zihpm"`
-    /// * Zimop: `"zimop"`
-    /// * Zacas: `"zacas"`
-    /// * Zawrs: `"zawrs"`
-    /// * Zfa: `"zfa"`
-    /// * Zfh: `"zfh"`
-    ///   * Zfhmin: `"zfhmin"`
-    /// * Zfinx: `"zfinx"`
-    /// * Zdinx: `"zdinx"`
-    /// * Zhinx: `"zhinx"`
-    ///   * Zhinxmin: `"zhinxmin"`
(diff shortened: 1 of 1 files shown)


sayantn:
Maybe a bad question, but what do the gaps in the `Linux` column mean? Has Linux not added support for them yet?

a4lg:
@sayantn

I meant:
*   A system call `riscv_hwprobe` which is new in the kernel version 6.4 (for Linux kernel between 4.16 and 6.3, this crate entirely depends on the auxiliary vector).
*   Multiple batches of runtime feature addition in the `riscv_hwprobe` system call (e.g. version 6.8 adds support for 37 extensions).

sayantn:
No, I meant like

```
    /// | `"zfinx"`       | Zfinx       |                    |
    /// | `"zhinx"`       | Zhinx       |                    |
    /// | `"zhinxmin"`    | Zhinxmin    |                    |
```

In the doc comment for `is_riscv_feature_detected`, what do the empty values for the third column mean?

a4lg:
@sayantn Ah, sorry.  *Just unsupported* extensions are filled with empty value and I meant "No" as the default value (without anything to note about) but... seems misleading.

sayantn:
Just to be clear, the kernel doesn't support it yet, right? Are there any extensions which the kernel supports, but `std_detect` doesn't implement it?

a4lg:
> Just to be clear, the kernel doesn't support it yet, right?

Yes.

> Are there any extensions which the kernel supports, but `std_detect` doesn't implement it?

As of the Linux kernel 6.14 (the latest release), the only extension not supported by `std_detect` but supported by the `riscv_hwprobe` system call is the `Supm` extension.
But the position of this extension is very special.

This extension (a part of the pointer masking specification) denotes that existence of *some* facility to change the behavior of pointer masking (enabling it makes some upper pointer bits reserved for custom purposes and ignored on hardware access). However, it provides existence of *some* facility and nothing architectural (no instructions, no CSRs...).

For instance, it just denotes the existence of `prctl`-based pointer masking control in Linux and that's too OS-dependent. That's why I am requested by a reviewer *not* to support it in Rust (as in rust-lang/rust#139440).

sayantn:
Ok thanks for the info. I was trying to see which target features are available in our CI runs in #1776

Amanieu:
> ## Is this Guide to be Here?
> 
> Because this kind of change is large and the resulting documentation is far more verbose than the rest (runtime feature detection macro on other architectures), I'm not sure whether this kind of the guide is suitable here.
> 
> Even if this kind of the guide is helpful for users, it may not be the right place.
> 
> What do you think?

I think it makes sense for RISC-V specifically because unlike other architectures it doesn't have a standard mechanism for describing target features.

> ## Are Version Numbers Required in the Extension Table?
> 
> The author prefers to include version numbers (also due to the fact that there are large gaps of runtime feature detection on the Linux kernel version 6.4 and later) but if that's too much, the author considers just showing Yes/No (with footnotes) will be still helpful for users (far better than nothing).

Version numbers are fine, but I don't feel too strongly about it either way. I'm happy with this PR as it is.

PR #1780, opened 2025-04-17
attempt to `#[allow(unnecessary_transmutes)]` in neon

bend-n:
im not sure why the `#![allow]` in `lib.rs` doesnt take effect, but for some reason it doesnt: https://github.com/rust-lang/rust/pull/136083#issuecomment-2804984503

this might fix that? im not sure


bend-n:
oh so i should put the #![allow] in core_arch/src/mod.rs then, right?

Amanieu:
The old allow in lib.rs should be removed since it is now redundant.

PR #1783, opened 2025-04-20
Add power9 and power8 target-features

lu-zero:


sayantn:
you need to add them to `detect/arch/powerpc.rs` too

PR #1788, opened 2025-04-28
Fix errors in decoupling `avx512vl` and `avx512dq` from `avx512fp16`

sayantn:
This does the required fixes to stdarch for rust-lang/rust#140389.

Also changes the tests for the same reason.

Related: llvm/llvm-project#136209, #1781, https://github.com/rust-lang/rust/pull/138940#issuecomment-2826131769

### DIFF
--- a/crates/core_arch/src/x86/avx.rs
+++ b/crates/core_arch/src/x86/avx.rs
@@ -1425,7 +1425,10 @@ pub fn _mm256_insert_epi32<const INDEX: i32>(a: __m256i, i: i32) -> __m256i {
 /// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm256_load_pd)
 #[inline]
 #[target_feature(enable = "avx")]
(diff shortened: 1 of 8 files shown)


crates/core_arch/src/x86/avx.rs
```diff
@@ -1425,7 +1425,7 @@ pub fn _mm256_insert_epi32<const INDEX: i32>(a: __m256i, i: i32) -> __m256i {
 /// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm256_load_pd)
 #[inline]
 #[target_feature(enable = "avx")]
-#[cfg_attr(test, assert_instr(vmovap))]
+#[cfg_attr(all(test, not(target_env = "msvc")), assert_instr(vmovap))]
```
Amanieu, line 1428: Can you change these to be `all(target_arch = "x86", target_env = "msvc")` and add a comment pointing to https://github.com/rust-lang/rust/pull/139261 as the cause.

PR #1790, opened 2025-04-29
Implement SHA512, SM3 and SM4 intrinsics

sayantn:
This PR implements all the SHA512, SM3 and SM4 intrinsics (rust-lang/rust#126624).

This also fixes a issue in stdarch-verify where it was silently not verifying most intrinsics as they were marked safe as part of `target_feature_11`.

### DIFF
--- a/crates/core_arch/missing-x86.md
+++ b/crates/core_arch/missing-x86.md
@@ -213,31 +213,6 @@
 </p></details>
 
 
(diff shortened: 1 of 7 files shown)


sayantn:
```
failed to verify `vdupq_n_p8`
  * mismatched arguments: PrimUnsigned(8) != PrimPoly(8)
```

@Amanieu this intrinsic should have `p8` parameter right? Can we change this?

PR #1795, opened 2025-05-07
Stabilize keylocker intrinsics and runtime detection

sayantn:
parent rust-lang/rust#134813

related: rust-lang/rust#140766

Currently draft because this requires the rustc pr to merge


sayantn:
Libs-api FCP is now complete, so this should be fine to merge now

PR #1796, opened 2025-05-07
Stabilize `sha512`, `sm3` and `sm4` intrinsics and runtime detection

sayantn:
parent: rust-lang/rust#126624

related: rust-lang/rust#140767

currently draft because this requires the rustc PR to merge


sayantn:
Libs-api FCP is now complete, so this should be fine to merge now

PR #1802, opened 2025-05-19
`avx512_target_feature` is now stable on nightly

folkertdev:
and enabling it causes errors on CI

### DIFF
--- a/crates/core_arch/src/lib.rs
+++ b/crates/core_arch/src/lib.rs
@@ -22,7 +22,6 @@
     sse4a_target_feature,
     riscv_target_feature,
     arm_target_feature,
-    avx512_target_feature,
     mips_target_feature,
     powerpc_target_feature,
     s390x_target_feature,
--- a/crates/core_arch/src/powerpc/altivec.rs
+++ b/crates/core_arch/src/powerpc/altivec.rs
@@ -547,7 +547,7 @@ mod sealed {
             #[target_feature(enable = "altivec")]
             #[cfg_attr(test, assert_instr($instr))]
             pub unsafe fn $fun(a: isize, b: *const $ty) -> t_t_l!($ty) {
-                let addr = (b as *const i8).offset(a);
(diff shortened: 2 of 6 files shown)


folkertdev:
Just writing up what we found here (yet more reasons to make this a submodule): https://github.com/rust-lang/rust/pull/127013 changed the formatting of `f16` values (to something much more helpful) in rust. In C, and `f16` is apparently formatted as a hexadecimal value. So we see results like this in the intrinsics tests:

```
C: result -1: float16x4x2_t(float16x4_t(0x0000, 0x0000, 0x37ff, 0x37ff), float16x4_t(0x0400, 0x0400, 0x3800, 0x3800))
Rust: result -1: float16x4x2_t(float16x4_t(0.0, 0.0, 0.4998, 0.4998), float16x4_t(6.104e-5, 6.104e-5, 0.5, 0.5))
```

I think the quick fix is to make stdarch have the old formatting behavior, so I implemented some hacks to make that happen. Long-term, hopefully there is a better way to compare the C and rust output?

crates/core_arch/src/macros.rs
```diff
@@ -132,12 +132,37 @@ macro_rules! types {
...
             }
         }
     )*);
 }
 
+// the `intrinsic-test` crate compares the output of C and Rust intrinsics. Currently, It uses
+// a string representation of the output value to compare. In C, f16 values are currently printed
+// as hexadecimal integers. Since https://github.com/rust-lang/rust/pull/127013, rust does print
+// them as decimal floating point values. To keep the intrinsics tests working, for now, format
+// vectors containing f16 values like C prints them.
```
Amanieu, line 146: f16 vector should use the same default formatting as f16. Could you change this to only apply to the Rust output of intrinsic-test instead of all users of the Debug impl?

folkertdev: Right, it takes a bunch more lines but it's probably better this way.

PR #1804, opened 2025-05-22
Add back `std_detect_env_override`

sayantn:
#1776 removed the machinery to manually enable/disable target features at runtime. It was overlooked that this is actually useful for testing, so this pr adds it back (the `env_override_no_avx` CI run is not added back)

cc @lu-zero


sayantn:
Oh yes, my mistake

PR #1807, opened 2025-05-27
Fix uses of invalid LLVM intrinsics

sayantn:
see https://github.com/rust-lang/rust/pull/140763#issuecomment-2912252256 for context

This PR also starts checking for IR validity in tests, it is almost free, but adds a lot of confidence as we are dealing with LLVM intrinsics

### DIFF
--- a/crates/stdarch-gen-arm/spec/neon/aarch64.spec.yml
+++ b/crates/stdarch-gen-arm/spec/neon/aarch64.spec.yml
@@ -2943,7 +2943,7 @@ intrinsics:
       - LLVMLink:
           name: "rbit.{neon_type}"
           links:
-            - link: "llvm.aarch64.neon.rbit.{neon_type}"
+            - link: "llvm.bitreverse.{neon_type}"
               arch: aarch64,arm64ec
 
   - name: "vrbit{neon_type[0].no}"
@@ -3096,7 +3096,7 @@ intrinsics:
       - LLVMLink:
           name: "frintn.{neon_type}"
           links:
-            - link: "llvm.aarch64.neon.frintn.{neon_type}"
+            - link: "llvm.roundeven.{neon_type}"
               arch: aarch64,arm64ec
 
   - name: "vrndns_{type}"
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -10,7 +10,7 @@ set -ex
 #export RUST_TEST_NOCAPTURE=1
 #export RUST_TEST_THREADS=1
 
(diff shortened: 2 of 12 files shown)


folkertdev:
Can you reproduce that failure in isolation? this seems to work fine https://godbolt.org/z/Mozez3P8f so I must be missing something

folkertdev:
an https://godbolt.org/z/9azEhx4vz does it, because it uses the 32-bit version. I was confused because the example from that llvm issue does work when I put it into godbolt

crates/stdarch-gen-arm/spec/neon/aarch64.spec.yml
```diff
@@ -2943,7 +2943,7 @@ intrinsics:
       - LLVMLink:
           name: "rbit.{neon_type}"
           links:
-            - link: "llvm.aarch64.neon.rbit.{neon_type}"
+            - link: "llvm.bitreverse.{neon_type}"
               arch: aarch64,arm64ec
```
folkertdev, line 2947: Can't we use [`simd_bitreverse`](https://doc.rust-lang.org/std/intrinsics/simd/fn.simd_bitreverse.html) directly instead? That does seem to happen for e.g. `simd_and`

sayantn: I don't know too much about how `stdarch-gen-arm` operates, particularly I don't know how I can change the impl. Still, I would try

folkertdev: I mean model it after this:

```yml
  - name: "vand{neon_type.no}"
    doc: Vector bitwise and
    arguments: ["a: {neon_type}", "b: {neon_type}"]
    return_type: "{neon_type}"
    attr:
      - *neon-v7
      - FnCall: [cfg_attr, [*test-is-arm, {FnCall: [assert_instr, [vand]]}]]
      - FnCall: [cfg_attr, [*neon-target-aarch64-arm64ec, {FnCall: [assert_instr, [and]]}]]
      - *neon-not-arm-stable
      - *neon-cfg-arm-unstable
    safety: safe
    types:
      - int8x8_t
      - int8x16_t
      - int16x4_t
      - int16x8_t
      - int32x2_t
      - int32x4_t
      - uint8x8_t
      - uint8x16_t
      - uint16x4_t
      - uint16x8_t
      - uint32x2_t
      - uint32x4_t
      - int64x1_t
      - int64x2_t
      - uint64x1_t
      - uint64x2_t
    compose:
      - FnCall:
          - simd_and
          - - a
            - b
```

PR #1812, opened 2025-05-30
Coding style modernization to deal with Clippy warnings

a4lg:
In the development of `stdarch`, we encounter many warnings (mainly from Clippy; currently 133 as counted by Rust Analyzer and 132 as counted by `cargo clippy`) and hid real warnings we need to take care of.

However, the author found that source of such changes are mainly from old Rust code (which does not know newer coding styles) and decided to deal with it by this PR.

# Policy

I allowed `clippy::collapsible_if` for now (to be reviewed) because it's just nested `if`s.
I allowed `clippy::obfuscated_if_else` because a candidate to fix this issue looked more complex than before.
I allowed `clippy::manual_range_contains` (as suggested by @Amanieu) because *fixing* the warning will complicate the code.

Other than that, I mostly used suggestions generated by Clippy.

# Further Simplification

In `crates/stdarch-gen-arm/src/fn_suffix.rs`, I also simplified the code (use a new variable to deal with Clippy warnings and use `format!()` macro) where no warnings are generated.  It makes the code more simple and easier to maintain.

### DIFF
--- a/crates/intrinsic-test/src/common/argument.rs
+++ b/crates/intrinsic-test/src/common/argument.rs
@@ -33,7 +33,7 @@ where
     }
 
     pub fn has_constraint(&self) -> bool {
-        !self.constraint.is_some()
+        self.constraint.is_none()
     }
 
     pub fn type_and_name_from_c(arg: &str) -> (&str, &str) {
@@ -65,7 +65,7 @@ where
     pub fn from_c(
         pos: usize,
         arg: &str,
-        target: &String,
+        target: &str,
         constraint: Option<Constraint>,
     ) -> Argument<T> {
         let (ty, var_name) = Self::type_and_name_from_c(arg);
@@ -127,15 +127,14 @@ where
     /// e.g `const int32x2_t a_vals = {0x3effffff, 0x3effffff, 0x3f7fffff}`, if loads=2.
     pub fn gen_arglists_c(&self, indentation: Indentation, loads: u32) -> String {
         self.iter()
-            .filter_map(|arg| {
-                (!arg.has_constraint()).then(|| {
-                    format!(
-                        "{indentation}const {ty} {name}_vals[] = {values};",
-                        ty = arg.ty.c_scalar_type(),
-                        name = arg.name,
-                        values = arg.ty.populate_random(indentation, loads, &Language::C)
-                    )
-                })
+            .filter(|&arg| !arg.has_constraint())
+            .map(|arg| {
+                format!(
+                    "{indentation}const {ty} {name}_vals[] = {values};",
+                    ty = arg.ty.c_scalar_type(),
+                    name = arg.name,
+                    values = arg.ty.populate_random(indentation, loads, &Language::C)
+                )
             })
             .collect::<Vec<_>>()
             .join("\n")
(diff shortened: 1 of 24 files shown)


examples/connect5.rs
```diff
@@ -556,7 +555,7 @@ fn search(pos: &Pos, alpha: i32, beta: i32, depth: i32, _ply: i32) -> i32 {
     }
 
     assert_ne!(bm, MOVE_NONE);
-    assert!(bs >= -EVAL_INF && bs <= EVAL_INF);
+    assert!((-EVAL_INF..=EVAL_INF).contains(&bs));
```
Amanieu, line 558: I think this is actually less clear than the previous code. I'd like this specific line to be reverted.

crates/intrinsic-test/src/common/argument.rs
```diff
@@ -145,17 +144,16 @@ where
...
-                    format!(
-                        "{indentation}{bind} {name}: [{ty}; {load_size}] = {values};",
-                        bind = arg.rust_vals_array_binding(),
-                        name = arg.rust_vals_array_name(),
-                        ty = arg.ty.rust_scalar_type(),
-                        load_size = arg.ty.num_lanes() * arg.ty.num_vectors() + loads - 1,
-                        values = arg.ty.populate_random(indentation, loads, &Language::Rust)
-                    )
-                })
+            .filter(|&arg| (!arg.has_constraint()))
```
Amanieu, line 147:
```suggestion
            .filter(|&arg| !arg.has_constraint())
```

crates/intrinsic-test/src/common/argument.rs
```diff
@@ -168,22 +166,18 @@ where
...
-                        ty = arg.to_c_type(),
-                        name = arg.name,
-                        load = if arg.is_simd() {
-                            arg.ty.get_load_function(Language::C)
-                        } else {
-                            "*".to_string()
-                        }
-                    )
-                })
+            .filter(|&arg| (!arg.has_constraint()))
```
Amanieu, line 169:
```suggestion
            .filter(|&arg| !arg.has_constraint())
```

crates/intrinsic-test/src/common/gen_rust.rs
```diff
@@ -130,7 +130,7 @@ pub fn setup_rust_file_paths(identifiers: &Vec<String>) -> BTreeMap<&String, Str
     identifiers
         .par_iter()
         .map(|identifier| {
-            let rust_dir = format!(r#"rust_programs/{}"#, identifier);
+            let rust_dir = format!(r#"rust_programs/{identifier}"#);
```
Amanieu, line 133:
```suggestion
            let rust_dir = format!("rust_programs/{identifier}");
```

This doesn't need to be a raw string.

crates/intrinsic-test/src/common/gen_rust.rs
```diff
@@ -130,7 +130,7 @@ pub fn setup_rust_file_paths(identifiers: &Vec<String>) -> BTreeMap<&String, Str
...
             let _ = std::fs::create_dir_all(&rust_dir);
             let rust_filename = format!(r#"{rust_dir}/main.rs"#);
```
Amanieu, line 135:
```suggestion
            let rust_filename = format!("{rust_dir}/main.rs");
```

crates/intrinsic-test/src/common/argument.rs
a4lg: I'll fix that, including two more occurrences of `.filter(|&arg| (!arg.has_constraint()))`.

a4lg:
Updated the PR with all your suggestions plus two similar changes.

PR #1813, opened 2025-05-30
intrinsic-test: Minor fixes (2025-05)

a4lg:
This is built on #1812.

While I'm making #1812, I found two likely bugs in the crate `intrinsic-test`.

1.  Some type names for Arm were not created correctly (because it used `rust_prefix` where `c_prefix` is necessary).
2.  `has_constraints()` worked opposite to the behavior as implied by the name (i.e. returns `true` if there's *no* constraints).

What's strange is, those changes did not affect CI and I doubt that CI does not use all tests generated by `intrinsic-test`.

### DIFF
--- a/crates/intrinsic-test/src/common/argument.rs
+++ b/crates/intrinsic-test/src/common/argument.rs
@@ -33,7 +33,7 @@ where
     }
 
     pub fn has_constraint(&self) -> bool {
-        self.constraint.is_none()
+        self.constraint.is_some()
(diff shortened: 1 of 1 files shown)


a4lg:
After #1812 is merged, it's rebased on the latest master.

adamgemmell:
@madhav-madhusoodanan I would have expected this to have broken CI too - something's masking failures here

PR #1814, opened 2025-05-30
Adding the x86 part of behavioural testing for std::arch intrinsics

madhav-madhusoodanan:


crates/intrinsic-test/src/common/intrinsic_helpers.rs
```diff
@@ -42,10 +49,15 @@ impl fmt::Display for TypeKind {
...
-                Self::UInt => "uint",
+                Self::Double => "double",
+                Self::Int(true) => "int",
+                Self::Int(false) => "uint",
                 Self::Poly => "poly",
                 Self::Void => "void",
+                Self::Char(true) => "char",
+                Self::Char(false) => "unsigned char",
+                Self::Short(true) => "short",
+                Self::Short(false) => "unsigned short",
```
madhav-madhusoodanan: @Amanieu @adamgemmell Are we using this trait (`fmt::Display for TypeKind`) anywhere in the crate? I was searching for its usage and I was unable to find any.

If we are not using this trait, may I delete the definition?

madhav-madhusoodanan:
Pardon the git history, I will be sorting that out shortly

crates/intrinsic-test/src/common/intrinsic_helpers.rs
Amanieu: I suspect this may be useful for debug output during development. However since it isn't used for functionality then it's fine to make any changes you want to it.

madhav-madhusoodanan: Got it.

crates/intrinsic-test/src/common/intrinsic_helpers.rs
```diff
@@ -12,23 +13,28 @@ use super::values::value_for_array;
 pub enum TypeKind {
     BFloat,
     Float,
-    Int,
-    UInt,
+
+    // if signed, then the inner value is true
+    Int(bool),
```
Amanieu: Rather than using a `bool`, considering instead something like:
```rust
enum Signed {
    Signed,
    Unsigned,
}
```

This ends up making the code much more readable.

crates/intrinsic-test/src/x86/config.rs
```diff
@@ -0,0 +1,9 @@
+pub fn build_notices(line_prefix: &str) -> String {
+    format!(
+        "\
+{line_prefix}This is a transient test file, not intended for distribution. Some aspects of the
+{line_prefix}test are derived from a JSON specification, published under the same license as the
```
Amanieu: x86 uses an XML specification, not JSON.

madhav-madhusoodanan:
The CI is giving me an odd error, even after I rebased from master:

```bash
  Updating crates.io index
     Locking 5 packages to latest compatible versions
      Adding quick-xml v0.37.5
      Adding serde-xml-rs v0.8.1
      Adding thiserror v1.0.69
      Adding thiserror-impl v1.0.69
      Adding xml-rs v0.8.26
   Compiling core_arch v0.1.5 (/Users/runner/work/stdarch/stdarch/crates/core_arch)
error: the feature `generic_arg_infer` has been stable since 1.89.0-nightly and no longer requires an attribute to enable
  --> crates/core_arch/src/lib.rs:34:5
   |
34 |     generic_arg_infer,
   |     ^^^^^^^^^^^^^^^^^
   |
   = help: to override `-D warnings` add `#[allow(stable_features)]`
```

Amanieu:
> The CI is giving me an odd error, even after I rebased from master:
> 
> ```shell
>   Updating crates.io index
>      Locking 5 packages to latest compatible versions
>       Adding quick-xml v0.37.5
>       Adding serde-xml-rs v0.8.1
>       Adding thiserror v1.0.69
>       Adding thiserror-impl v1.0.69
>       Adding xml-rs v0.8.26
>    Compiling core_arch v0.1.5 (/Users/runner/work/stdarch/stdarch/crates/core_arch)
> error: the feature `generic_arg_infer` has been stable since 1.89.0-nightly and no longer requires an attribute to enable
>   --> crates/core_arch/src/lib.rs:34:5
>    |
> 34 |     generic_arg_infer,
>    |     ^^^^^^^^^^^^^^^^^
>    |
>    = note: `-D stable-features` implied by `-D warnings`
>    = help: to override `-D warnings` add `#[allow(stable_features)]`
> ```

Yes, this regularly happens due to changes in nightly Rust. In this case you can just remove the feature from lib.rs.

folkertdev:
Hi, cool project!

I'm currently investigating how we can speed up the arm/aarch64 version of `intrinsic-test`. The CI job with it currently takes 28 minutes, and that is just not acceptible; it slows everything/everyone down.

On the other hand, I don't want to create horrible merge conflicts for you. The two main changes that I want to make are

- combine tests into larger files (one per core), this doesn't really change anything relevant to you
- write into a `&mut impl std::io::Write` instead of concatenating strings

The latter change mostly just helps with making the former possible. In the experimental https://github.com/rust-lang/stdarch/pull/1856 I've now gotten CI time down to 8 minutes.

So, to prevent merge conflicts (especially now that we have additional reviewing capacity) I'd encourage you to split out and merge changes early if they mostly touch the existing code, instead of accumulating them in one big PR. Some changes that look like they can be split out are: 

- removing some of the `Box` wrapping (e.g.  some of the changes in `crates/intrinsic-test/src/arm/types.rs`) 
- the `pub enum Sign` and its use in `TypeKind::Int`
- the cleanup in crates/intrinsic-test/src/common/gen_c.rs

There might be others that I missed.

Feel free to `r? @folkertdev` me on any such cleanup PRs.

madhav-madhusoodanan:
Hi @folkertdev, thank you so much for the suggestion and the concern about merge conflicts (and about looking into how we can optimize compile times!)
I was checking out #1856 and I agree with you on the need for the same.

I shall create small PRs with the changes that needs to be made on the ARM and the common modules, once I reconcile the changes with this PR itself.

crates/intrinsic-test/src/common/gen_rust.rs
```diff
@@ -88,11 +88,15 @@ pub fn compile_rust_programs(
...
 
-    if let Some(toolchain) = toolchain {
-        if !toolchain.is_empty() {
-            cargo_command.arg(toolchain);
+    match toolchain {
+        None => return true,
+        Some(toolchain) => {
+            if !toolchain.is_empty() {
+                cargo_command.arg(toolchain);
+            }
```
folkertdev: I'll refactor so that this is not needed. it's really confusing

crates/intrinsic-test/src/arm/mod.rs
```diff
@@ -69,28 +69,35 @@ impl SupportedArchitectureTest for ArmArchitectureTest {
...
         let (chunk_size, chunk_count) = chunk_info(self.intrinsics.len());
 
-        let cpp_compiler = compile::build_cpp_compilation(&self.cli_options).unwrap();
+        let cpp_compiler = compile::build_cpp_compilation(&self.cli_options);
+
+        match fs::exists("c_programs") {
+            Ok(false) => fs::create_dir("c_programs").unwrap(),
+            Ok(true) => {}
+            _ => return false,
+        }
```
folkertdev: I believe `create_dir_all` has the semantics that you want here (just return succes if the path already exists)

madhav-madhusoodanan: Ohh, I noticed your comments only after I pushed some updates.
I'll fix this, thank you for pointing out.

folkertdev:
closing this now that it's been subsumed by https://github.com/rust-lang/stdarch/pull/1894

PR #1815, opened 2025-05-30
use `simd_bitreverse` on `aarch64`

folkertdev:


crates/stdarch-gen-arm/spec/neon/aarch64.spec.yml
```diff
@@ -2940,11 +2940,9 @@ intrinsics:
...
-      - LLVMLink:
-          name: "rbit.{neon_type}"
-          links:
-            - link: "llvm.bitreverse.{neon_type}"
-              arch: aarch64,arm64ec
+      - FnCall:
+          - simd_bitreverse
+          - - a
 
   - name: "vrbit{neon_type[0].no}"
```
folkertdev, line 2947: it seems possible to use `simd_bitreverse` here too, but I can't get it to generate the right thing.

sayantn:
Thanks, I was procrastinating on it 😅

PR #1818, opened 2025-06-02
Fix Neon f16 stability

adamgemmell:
In https://rust-lang.zulipchat.com/#narrow/channel/219381-t-libs/topic/is.20.60arch.3A.3Aaarch64.3A.3Afloat16x8_t.60.20supposed.20to.20be.20stable.3F/with/520762685 it was noted that some f16 Neon types were incorrectly stable when introduced and that some f16 Neon intrinsics were incorrectly changed from unstable to stable recently.

The latter is a regression from stable to beta and could be nominated for a beta backport, but I couldn't find any process or precedent for doing this on stdarch.


Amanieu:
As much as this is correct, it's still going to be a pain to backport...

Amanieu:
If the FCP on https://github.com/rust-lang/rust/issues/136306 passes then we could avoid backporting entirely.

PR #1819, opened 2025-06-02
Stabilize AVX512 intrinsics

Amanieu:
FCP completed in tracking issue rust-lang/rust#111137


sayantn:
I think this should have a place in the relnotes, this is a quite significant step for SIMD support in Rust ❤️

Amanieu:
Tracking issue for the relnotes: https://github.com/rust-lang/rust/issues/141629

sayantn:
we can probably stabilize avx512fp16 too by the next release (26th June). I am adding the `release-blog-post` label to the AVX512 relnotes issue now

PR #1820, opened 2025-06-02
Fix some more uses of LLVM intrinsics

sayantn:
- Use rust intrinsics for more ARM intrinsics
 - Fix incorrect intrinsic names (these were autoupgraded by LLVM, so didn't result in invalid IR)
 - Remove uses of deprecated type-specific pointers, LLVM only supports opaque pointers now.

@folkertdev is it okay to replace the `llvm.{u,s}{max,min}` calls with something like `simd_select(simd_gt(a, b), a, b))`? I am just worried if it will optimize well

edit: seems like it is not a problem actually https://godbolt.org/z/6r1javWc6

### DIFF
--- a/crates/core_arch/src/aarch64/neon/generated.rs
+++ b/crates/core_arch/src/aarch64/neon/generated.rs
@@ -10072,14 +10072,7 @@ pub fn vextq_p64<const N: i32>(a: poly64x2_t, b: poly64x2_t) -> poly64x2_t {
 #[stable(feature = "neon_intrinsics", since = "1.59.0")]
 #[cfg_attr(test, assert_instr(fmadd))]
 pub fn vfma_f64(a: float64x1_t, b: float64x1_t, c: float64x1_t) -> float64x1_t {
-    unsafe extern "unadjusted" {
-        #[cfg_attr(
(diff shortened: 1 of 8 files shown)


folkertdev:
Just try it, you can check whether that works by running the docker container. We should be testing for the specific instruction (adjust for your target):

```
TARGET=aarch64-unknown-linux-gnu sh ci/run-docker.sh aarch64-unknown-linux-gnu
```

If it doesn't work, let me know because we should get that fixed in LLVM.

Amanieu:
> @folkertdev is it okay to replace the `llvm.{u,s}{max,min}` calls with something like `simd_select(simd_gt(a, b), a, b))`? I am just worried if it will optimize well

No, those 2 are not equivalent, specifically with regards to negative zero and NaN.

folkertdev:
`umax` and `smax` are only for integers right? For floats the expressions are indeed not equivalent in some nasty ways.

Amanieu:
~~The fact that LLVM generates the same code for both suggests a bug in the LLVM backend.~~

Nevermind, it's fine if this is only for integers.

sayantn:
These intrinsics deal with integers, so there should be no difference

Issue #1822, opened 2025-06-03
Bug in implementation of `_mm256_bsrli_epi128`


karthikbhargavan:
The `_mm256_bsrli_epi128` intrinsic in `core::arch::x86::avx2` can be used to shift right each 128-bit lane in a 256-bit vector by an `IMM8` number of bytes.

The implementation of this intrinsic is buggy when the shift argument `IMM8` is greater than 15.
In particular, it behaves differently from the Intel documentation and from the corresponding C intrinsic in clang.

The relevant Intel documentation for this intrinsic is here: https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm256_bsrli_epi128&ig_expand=628

When the shift argument `imm8` is greater than 15, the resulting vector should contain all zeroes.
Indeed, this is what clang does: https://godbolt.org/z/6o3W96qhP

However, the Rust implementation shifts right by `IMM8 % 16' and so produces a different result.
See: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=500d1a08d16b30a6cc8547b607918854

(This example results in particularly weird behavior since shifting right by 1..15 bytes yields all 0s, but shifting right by > 16 yields a non-zero result.)

The bug is in this line of this commit: https://github.com/rust-lang/stdarch/blob/35595690d7881b20bd419875c2b64f284534d237/crates/core_arch/src/x86/avx2.rs#L2782.

Removing the `% 16` would make this implementation consistent with Intel's documentation and with clang.

Note: This issue was found by Aniket Mishra, an intern at Cryspen working on verifying parts of the Rust core library, specifically this challenge: https://github.com/model-checking/verify-rust-std/issues/173

karthikbhargavan:
There appears to be a similar bug in `_mm512_bsrli_epi128`

https://github.com/rust-lang/stdarch/blob/35595690d7881b20bd419875c2b64f284534d237/crates/core_arch/src/x86/avx512bw.rs#L8962

sayantn:
Fixed in #1823

PR #1823, opened 2025-06-03
Fix in erroneous implementation of _mm256_bsrli_epi128

satiscugcat:
This fixes the error mentioned in issue #1822


crates/core_arch/src/x86/avx2.rs
```diff
@@ -191,7 +191,7 @@ pub fn _mm256_alignr_epi8<const IMM8: i32>(a: __m256i, b: __m256i) -> __m256i {
             return transmute(a);
         }
 
-        let r: i8x32 = match IMM8 % 16 {
+        let r: i8x32 = match IMM8 {
```
bjorn3: You have to also add a value to the default arm. It is currently `unreachable_unchecked()`, which makes `IMM8` values > 15 UB.
Edit: Maybe you should change the `IMM8 == 16` if above to be `IMM8 >= 16`? Also there is handling for `IMM8 > 16` above already: `(_mm256_setzero_si256(), a)` which would become unreachable in that case.

bjorn3:
Did you try reverifying _mm256_alignr_epi8 with whatever tool you used to find the _mm256_bsrli_epi128 bug? And if so how did the `unreachable_unchecked()` bug slip past it?

crates/core_arch/src/x86/avx2.rs
satiscugcat: Oh right, I was mistaken. I think specifically values between 17-31 (inclusive) are UB without the %, which makes it not redundant. My mistake!

satiscugcat:
> Did you try reverifying _mm256_alignr_epi8 with whatever tool you used to find the _mm256_bsrli_epi128 bug? And if so how did the `unreachable_unchecked()` bug slip past it?

The _mm256_bsrli_epi128 bug was found manually upon noticing a discrepancy in the intel specifications and the rust implementations in that case, and then testing it as it is tested in issue #1822 . Thus the other bug slipped past me too. My bad!

sayantn:
Could you check pls if this bug is also there for `_mm256_bslli_epi128` and the `_mm512` variants too. They might be affected as the impl is very similar

satiscugcat:
> Could you check pls if this bug is also there for `_mm256_bslli_epi128` and the `_mm512` variants too. They might be affected as the impl is very similar

Sure! We'll work on testing the variants from tomorrow.

sayantn:
just checked, seems like the same bug is there for `_mm512_bsrli_epi128`, but not for any of the `bslli` variants (it is actually amusing how different the implementations are for `bslli` and `bsrli`, but anyway it seems like the `bslli` implementation is much more robust). 

Actually, would you mind changing the impl of the `bsrli` variants to be like as the `bslli` variants? It would also fix this bug, and make it more consistent

satiscugcat:
> just checked, seems like the same bug is there for `_mm512_bsrli_epi128`, but not for any of the `bslli` variants (it is actually amusing how different the implementations are for `bslli` and `bsrli`, but anyway it seems like the `bslli` implementation is much more robust).
> 
> Actually, would you mind changing the impl of the `bsrli` variants to be like as the `bslli` variants? It would also fix this bug, and make it more consistent

I've made the appropriate changes.

sayantn (commented):
lgtm. Could you also see if this refactoring is also possible for `_mm{256,512}_alignr_epi8`. You could see the impl for `_mm_alignr_epi8` for some reference (sorry for this many requests 😅)

satiscugcat:
I don't understnd why the CI failed, running ci/run-docker.sh went through fine on my end.

satiscugcat:
Ok, everything should be fine now. Let me know if there's anything else you think might need attention.

PR #1824, opened 2025-06-04
Remove `compiler-builtins` from `rustc-dep-of-std` dependencies

tgross35:


tgross35:
I think this should still work since alloc has the dep on builtins and it isn't used directly, but am not sure. Testing in https://github.com/rust-lang/rust/pull/141993.

PR #1829, opened 2025-06-11
Add lockfile

Kobzol:
Needed to unblock https://github.com/rust-lang/rust/pull/141899.

The problem is that on Rust's CI, we want to test at least compiling stdarch's `core_arch` and possibly run some tests (which are normally not part of stdlib). But on CI we run everything with `--locked`, so we need a lockfile.

An alternative is to include stdarch in the stdlib's workspace. However, if I understand it correctly, stdlib only uses `std_detect`, but we want to build also other crates (such as `core_arch`). So we would have to include all packages from stdarch that we want to build/test on rust's CI into the stdlib workspace, which I'm not sure is a good idea (?). 

Happy to hear feedback!


Cargo.lock
```diff
@@ -0,0 +1,965 @@
...
+
+[[package]]
+name = "lazy_static"
+version = "1.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e2abad23fbc42b3700f2f279844dc832adb2b2eb069b2df918f455c4e18cc646"
+
+[[package]]
+name = "libc"
+version = "0.2.152"
```
sayantn: why is this `0.2.152`? The latest crates.io version is `0.2.172`

Kobzol: :shrug: Ask Cargo :D I just did `cargo +nightly build` to generate it.

Amanieu: This would have used whatever existing Cargo.lock you had as a base. Could you run `cargo update` or delete your existing Cargo.lock before re-building it?

Amanieu:
This needs a rebase now that #1830 has landed.

Cargo.lock
Kobzol: I thought I recreated it, but indeed it seems that a new lockfile has libc `0.2.172`. Thanks!

Kobzol:
Rebased and recreated the lockfile.

PR #1832, opened 2025-06-12
s390x: z17 target feature detection

folkertdev:
tracking issue: https://github.com/rust-lang/rust/issues/130869

I didn't add feature detection yet in https://github.com/rust-lang/stdarch/pull/1826. 

The implementation follows the information in https://github.com/rust-lang/rust/issues/135413#issuecomment-2886439455, in particular that `HWCAP` is used to check for the `vector` feature because it needs kernel support, and `stfle` is used for the other features because they only need hardware support.

cc @uweigand @taiki-e

### DIFF
--- a/crates/std_detect/src/detect/os/linux/s390x.rs
+++ b/crates/std_detect/src/detect/os/linux/s390x.rs
@@ -5,19 +5,11 @@ use crate::detect::{Feature, bit, cache};
 
 /// Try to read the features from the auxiliary vector
 pub(crate) fn detect_features() -> cache::Initializer {
-    if let Ok(auxv) = auxvec::auxv() {
-        let hwcap: AtHwcap = auxv.into();
-        return hwcap.cache();
-    }
-
-    cache::Initializer::default()
+    let opt_hwcap: Option<AtHwcap> = auxvec::auxv().ok().map(Into::into);
+    let facilities = ExtendedFacilityList::new();
+    cache(opt_hwcap, facilities)
 }
 
-/// These values are part of the platform-specific [asm/elf.h][kernel], and are a selection of the
-/// fields found in the [Facility Indications].
-///
-/// [Facility Indications]: https://www.ibm.com/support/pages/sites/default/files/2021-05/SA22-7871-10.pdf#page=63
-/// [kernel]: https://github.com/torvalds/linux/blob/b62cef9a5c673f1b8083159f5dc03c1c5daced2f/arch/s390/include/asm/elf.h#L129
 #[derive(Debug, Default, PartialEq)]
 struct AtHwcap {
     esan3: bool,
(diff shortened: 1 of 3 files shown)


crates/std_detect/src/detect/os/linux/s390x.rs
```diff
@@ -76,57 +68,80 @@ impl From<auxvec::AuxVec> for AtHwcap {
...
     }
+
+    {
+        let mut enable_if_set = |bit_index, f| {
+            if facilities.get_bit(bit_index) {
+                value.set(f as u32);
+            }
+        };
+
+        // vector and related
```
taiki-e: These facilities need vector register support, so I think they should be gated with `if hwcap.vxrs`.

crates/std_detect/src/detect/os/linux/s390x.rs
```diff
@@ -76,57 +68,79 @@ impl From<auxvec::AuxVec> for AtHwcap {
...
+}
 
-            // bit 151 of the extended facility list
-            enable_feature(Feature::deflate_conversion, self.dflt);
+/// Initializes the cache from the feature bits.
+///
+/// These values are part of the platform-specific [asm/elf.h][kernel], and are a selection of the
+/// fields found in the [Facility Indications].
+///
+/// [Facility Indications]: https://www.ibm.com/support/pages/sites/default/files/2021-05/SA22-7871-10.pdf#page=63
```
taiki-e, line 100: Where did you get the information about the facility bits that will be added in z17? (It is not included in the documentation at this link.)

uweigand: I can confirm that the z17 facility bit values are correct.  The new version of the Principles of Operation including z17 features will be released shortly after the general availability of the new machine (which is next week).

crates/std_detect/src/detect/os/linux/s390x.rs
```diff
@@ -76,57 +68,77 @@ impl From<auxvec::AuxVec> for AtHwcap {
...
-        value
+
+        // others
+
+        enable_if_set(45, Feature::high_word);
+        enable_if_set(73, Feature::transactional_execution);
+        enable_if_set(133, Feature::guarded_storage);
+        enable_if_set(150, Feature::enhanced_sort);
+        enable_if_set(151, Feature::deflate_conversion);
+        // added in z17
```
uweigand: Do we need this comment?  If so, should we add this information also for the vector-related facilities, and also specific machine information for pre-z17 machines?

crates/std_detect/src/detect/os/linux/s390x.rs
```diff
@@ -76,57 +68,77 @@ impl From<auxvec::AuxVec> for AtHwcap {
...
+        enable_if_set(84, Feature::miscellaneous_extensions_4);
+        enable_if_set(86, Feature::message_security_assist_extension12);
```
uweigand: We should really include *all* currently defined features here.  What I see missing is the older variants of miscellaneous_extensions and messact_security_assist_extension.  These all have defined facility bits.

folkertdev: We can only add features here when they are also recognized by the compiler. I'll add them there, but that does not need to block this PR I think

crates/std_detect/src/detect/os/linux/s390x.rs
folkertdev: nah, this list doesn't really have structure to it, but it should be OK given the official docs.

crates/std_detect/src/detect/os/linux/s390x.rs
folkertdev: ah, I forgot that we did already add these to the compiler. Now they're in this PR too.

uweigand: Thanks!  Looks good to me now.

folkertdev:
the build broke because a feature got stabilized:

```
    Updating crates.io index
   Compiling core_arch v0.1.5 (/Users/runner/work/stdarch/stdarch/crates/core_arch)
error: the feature `generic_arg_infer` has been stable since 1.89.0-nightly and no longer requires an attribute to enable
  --> crates/core_arch/src/lib.rs:34:5
   |
34 |     generic_arg_infer,
   |     ^^^^^^^^^^^^^^^^^
   |
   = help: to override `-D warnings` add `#[allow(stable_features)]`

error: could not compile `core_arch` (lib) due to 1 previous error
```

I'll add a commit to remove that line

crates/std_detect/src/detect/os/linux/s390x.rs
uweigand: The new PoP is now available here: https://www.ibm.com/docs/en/module_1678991624569/pdf/SA22-7832-14.pdf

PR #1834, opened 2025-06-23
Mark `avx` broadcast intrinsics as safe

okaneco:
The scalar functions are wrappers around `_mm_set1_*` functions which are already safe.

`_mm256_broadcast_ps` and `_mm256_broadcast_pd` take their input argument by reference instead of raw pointer.


sayantn:
The CI failure seems strange, could you force push it once to retry the CI? You can probably just do `git commit --amend --allow-empty --no-edit && git push --force` if the worktree is clean

okaneco:
CI failed twice on `Run ./ci/run-docker.sh armv7-unknown-linux-gnueabihf`.

When I first submitted the PR, the failing `x86_64-unknown-freebsd` didn't seem to even start so it timed out.

sayantn:
The ARM failures bisect to rust-lang/rust#142768
I am not sure why the freebsd run didn't start, probably spurious since it didn't happen this time

PR #1835, opened 2025-06-26
Add Josh sync scripts

Kobzol:
This pretty much 1:1 mirrors the sync scripts of rustc-dev-guide. If you want, I can also setup a CI bot (similar to rdg or miri) that will attempt to create PRs with pull synces once every few days, and post to Zulip if there is a merge conflict.


tgross35:
I have a couple changes from the RDG version in https://github.com/rust-lang/compiler-builtins/tree/674910e0fa6f0fb2cc055f4f7051ff0eb53c7735/crates/josh-sync that might be worth picking up here. The biggest thing is dropping `clap` since that adds 30 extra deps to the lockfile, bit of overkill for parsing two possible commands. (anyhow and xshell are dropped too , but those are lighter weight to start).

Also just updating the commit messages to crosslink the original ref (e.g. https://github.com/rust-lang/compiler-builtins/commit/6e475af52394f18ccbd57efabf586cee742b2d5a). You can figure this out from the `rust-version` file but it's nice to have it show up in the log of anything that got touched in the sync. Similar for the PR message (will probably add an option to just open a PR via `gh` at some point to avoid percent encoding something more detailed).

Kobzol:
My original plan was to create a separate repo that would contain the josh sync scripts, and they could be reused for all subtrees, either as a library or as a binary that could be `cargo install`-ed. Now that we have probably all of the submodules that we wanted to have as subtrees (maybe with the exception of Cargo), I wonder if it would maybe be better to create this repo, to avoid having N separate copies of these scripts scattered across various repositories :thinking:

RalfJung:
Yeah, that sounds like a good idea.

I'd prefer not to have to install `gh` on my machine if that's an option. :)

Kobzol:
This script doesn't use `gh` (haven't looked at Trevor's script yet). Ideally it should install everything on its own, i.e. download and run josh and then just do operations using the `git` binary.

tgross35:
Oh, yeah, mine only needs josh+git as well (only the deps and commit messages are different). I was just thinking a "would you like to open a PR via `gh`? Y/N" option alternative to the link.

RalfJung:
If it generates the link it might as well `open` it then rather than relying on `gh`. ;)

Amanieu:
> My original plan was to create a separate repo that would contain the josh sync scripts, and they could be reused for all subtrees, either as a library or as a binary that could be `cargo install`-ed. Now that we have probably all of the submodules that we wanted to have as subtrees (maybe with the exception of Cargo), I wonder if it would maybe be better to create this repo, to avoid having N separate copies of these scripts scattered across various repositories 🤔

Would it make sense to centralize all the sync scripts in rust-lang/rust since this is where all the repos are synced as sub-trees?

Kobzol:
> Would it make sense to centralize all the sync scripts in rust-lang/rust since this is where all the repos are synced as sub-trees?

I was thinking about that, but I'm not sure it's that useful, because both pulls and pushes are invoked from the subtree repo, not from the rust-lang/rust repo.

tgross35:
Oh we don't actually need this anymore do we? Since https://github.com/rust-lang/josh-sync exists.

Kobzol:
Oh, I forgot to close the PR :) Anyway, I'll send another one where I document josh-sync and remove the scripts.

PR #1837, opened 2025-06-28
loongarch: Add basic support for LoongArch32

heiher:


crates/core_arch/src/loongarch_shared/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+    __frsqrte_s(a)
+}
+
+/// Calculate the approximate double-precision result of dividing 1.0 by the square root
+#[inline]
+#[target_feature(enable = "frecipe")]
+#[unstable(feature = "stdarch_loongarch", issue = "117427")]
+pub unsafe fn frsqrte_d(a: f64) -> f64 {
+    __frsqrte_d(a)
+}
```
folkertdev, line 244: looks like llvm will just optimize this (no target feature is even needed): https://godbolt.org/z/5s6hvKE51.

Overall we prefer to have fewer (manual) `extern "unadjusted¨` items. Also you should probably add some tests that these instructions actually produce the instructions you expect, like so

```rust
#[cfg_attr(all(test, target_feature = "frecipe"), assert_instr(frsqrt.d))]
```

assuming that `frecipe` is enabled on CI.

heiher: > looks like llvm will just optimize this (no target feature is even needed): https://godbolt.org/z/5s6hvKE51.

They're different -- `frsqrte_d_manual` generates `frsqrt.d`, not `frsqrt**e**.d`.

> 
> Overall we prefer to have fewer (manual) `extern "unadjusted¨` items. Also you should probably add some tests that these instructions actually produce the instructions you expect, like so
> 
> ```rust
> #[cfg_attr(all(test, target_feature = "frecipe"), assert_instr(frsqrt.d))]
> ```
> 
> assuming that `frecipe` is enabled on CI.

Sure, I'll add more test cases. Perhaps in a separate PR to keep things focused. Thanks!

folkertdev: ah, my bad. so the "e" is for "estimate" apparently. Maybe use that word? (most functions on floats approximate the answer, but apparently this intrinsic deliberately uses less precision to achieve higher speed).

heiher: Yep, exactly. 😄

sayantn:
We will also have to add a LA32 CI run. It doesn't seem like `qemu` supports LA32 yet (can we run it on `qemu-loongarch64` work?), and `rustc` has only bare-metal targets with this arch (we can probably work around that by having a `loongarch32-unknown-linux-gnu.json` target spec file). Also it doesn't seem like the Ubuntu repo has a `loongarch32-gcc` package (is it available from somewhere else, like `aarch64_be-gcc`?)

tgross35:
Usually we don't run CI for T3 targets, e.g. Mips is skipped. They need to move to T2 to get to blocking CI.

riscv64gc is technically T3, but that one is probably fine since it's a superset of the T2 `riscv32` targets.

crates/core_arch/src/loongarch32/mod.rs
```diff
@@ -0,0 +1,47 @@
+//! `LoongArch32` intrinsics
+
+use crate::arch::asm;
+
+#[allow(improper_ctypes)]
+unsafe extern "unadjusted" {
+    #[link_name = "llvm.loongarch.cacop.w"]
+    fn __cacop(a: i32, b: i32, c: i32);
+    #[link_name = "llvm.loongarch.csrrd.w"]
+    fn __csrrd(a: i32) -> i32;
```
Amanieu, line 10: Would it make sense to have these in the common code returning `isize`? It is technically a breaking change but this is still an unstable feature.

heiher: Thanks for your comments. I don't quite understand why this is considered a breaking change. Previously, LoongArch64 exported `csrrd(i32) -> i64`, and it sill does. Now, LoongArch32 is introducing `csrrd(i32) -> i32`, which doesn't change the existing behavior.

Amanieu: I'm suggesting having a single shared intrinsic `csrrd(i32) -> isize` instead of separate ones for loongarch32 and loongarch64.

heiher: Ah, I see -- I misunderstood your point. On 32-bit target, the link name for `__csrrd` is `llvm.*.csrrd.w`, while on 64-bit it's `csrrd.d`. That means we'd need to use `#[cfg()]` in the shared code? Would it be better to just keep things as they are?

PR #1838, opened 2025-06-29
std_detect: Tidying of slice length

a4lg:
It's nothing more than a small tidying.
But since `stdarch` moved on to a Josh subtree, we (`stdarch` contributors) can test whether purely `stdarch` changes are still accepted through this repo.

We don't need to put the length of the `riscv_hwprobe` array into a variable.
This commit removes that variable and gives the length of the output slice to the `__riscv_hwprobe` directly.

### DIFF
--- a/crates/std_detect/src/detect/os/linux/riscv.rs
+++ b/crates/std_detect/src/detect/os/linux/riscv.rs
@@ -124,8 +124,7 @@ fn _riscv_hwprobe(out: &mut [riscv_hwprobe]) -> bool {
         }
     }
 
-    let len = out.len();
-    unsafe { __riscv_hwprobe(out.as_mut_ptr(), len, 0, ptr::null_mut(), 0) == 0 }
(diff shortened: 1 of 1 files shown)


a4lg:
Hmm, CI is failing with a reason unrelated to this change itself.

folkertdev:
Generally, changes should still be made here if possible. Certainly right now (CI doesn't run in the `rust-lang/rust` repo), but this has been the policy for e.g. miri which has been a subtree for a while.

I have a possible fix for the issue here https://github.com/rust-lang/rust/pull/143194, but that'll take at least a day to propagate to nightly and fix CI here. (that is of course one of the reasons to make stdarch a subtree: no more such breakage).

a4lg:
@folkertdev Thanks for letting me know (both about future contribution and *current* CI failures).
Now, it seems stdarch CI is working as nightly-2025-07-03 contains the fix you mentioned.

PR #1841, opened 2025-06-30
Use `simd_extract!` macro instead of calling `simd_extract` directly

clarfonthey:
Was poking at this due to the desire to potentially change this intrinsic to use a const generic. The macro has been presumably updated to allow an explicit type parameter after this change was made, so, this uses that instead to allow potentially updating the calling convention later.

Side note: while looking into this, it feels unlikely we'll be able to do that *anyway* due to the fact that casting types is not allowed in generic constants, and some intrinsics have already stabilised `i32` const generics over `u32` ones. But this change felt okay anyway, so, I wanted to offer it as a PR.


clarfonthey:
Side note: ARM test failures despite not modifying any ARM code indicate something spurious. I don't think my change is the issue; it should *literally* not affect anything besides making some not-fully-qualified paths fully-qualified.

sayantn:
yeah, it is a rustc bug. The fix is in the process of being merged

PR #1843, opened 2025-07-04
Remove `std_detect` dev dependency in `core_arch`

Kobzol:
Instead of `core_arch` loading the `is_XYZ_feature_enabled` macros from the local `std_detect` crate, they are loaded from `::std::arch` instead. This means that nothing in this repository will depend on `std_detect` anymore (except for tests, these will be ported later). The downside is that `core_arch` will need to wait for nightly std bumps to see changes in the detection macros, but I assume that should be rare.

This is one of the steps required to move `std_detect` fully into `rust-lang/rust`.

### DIFF
--- a/crates/core_arch/Cargo.toml
+++ b/crates/core_arch/Cargo.toml
@@ -22,7 +22,6 @@ maintenance = { status = "experimental" }
 
 [dev-dependencies]
 stdarch-test = { version = "0.*", path = "../stdarch-test" }
-std_detect = { version = "0.*", path = "../std_detect" }
 
 [target.'cfg(all(target_arch = "x86_64", target_os = "linux"))'.dev-dependencies]
 syscalls = { version = "0.6.18", default-features = false }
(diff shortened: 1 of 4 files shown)


crates/core_arch/Cargo.toml
```diff
@@ -22,7 +22,7 @@ maintenance = { status = "experimental" }
 
 [dev-dependencies]
 stdarch-test = { version = "0.*", path = "../stdarch-test" }
-std_detect = { version = "0.*", path = "../std_detect" }
+#std_detect = { version = "0.*", path = "../std_detect" }
```
folkertdev: probably just remove this fully?

```suggestion
```

folkertdev (commented):
changes to the detection macros should just happen in r-l/r then, right? At least when adding new target features that nicer anyway (in that case today we must wait for a nightly bump). It's only internal logic changes (which, as you say, are rare) that would get bitten by this, but then it should be easy enough to move such a PR over to r-l/r

crates/core_arch/Cargo.toml
Kobzol: Right, I was just testing this at first. Removed, thanks!

Kobzol:
> changes to the detection macros should just happen in r-l/r then, right?

Yes. Hopefully the whole of `std_detect` will more into r-l/r soon-ish (https://github.com/rust-lang/rust/pull/143412).

PR #1844, opened 2025-07-04
Perform first rustc pull

Kobzol:
Performs the first rustc pull using https://github.com/rust-lang/josh-sync. Also adds the required `josh-sync.toml` config file and the `rust-version` file.

If everything goes through smootly, I will also do a push, and then document how to perform synces on this repo, and update rustc-dev-guide with the instructions.

CC @RalfJung


rust-version
```diff
@@ -0,0 +1 @@
+1b61d43bdbf875183b1f436302d62ff93f9a6bba
```
RalfJung, line 1: ideally CI would use this to determine the Rust version to test, that'd made it stable under rustc repo changes... but that's a future TODO I think :)

Kobzol: That's an interesting idea. It would mean that if there is a breaking change in rustc, we would only recognize it during the next pull (instead of recognizing it the next day when nightly is bumped). So we should combine this technique with having CI that will do an automatic pull every night to give us faster feedback when r-l/r changes break something in stdarch. I assume that's what miri does, right?

RalfJung: Specifically we have a nightly CI job that [starts with](https://github.com/rust-lang/miri/blob/645b974b22d9a22dbe393af1729b51526cdab7a1/.github/workflows/setup/action.yml#L48):
```
git ls-remote https://github.com/rust-lang/rust/ HEAD | cut -f 1 > rust-version
```
So we test against the latest rustc every night, and we get a Zulip message if that job fails. But even if rustc breaks, that doesn't affect unrelated PRs. If CI fails in a PR, that's *always* the fault of the PR because everything is pinned. IMO that is a quite important property to have.

This is orthogonal to the automatic pull, though we do both in the same CI cronjob.

Kobzol: That is nice, yeah. I'll take a look at it once we have rustc-josh-sync for stdarch up & running.

RalfJung: Happy to explain any questions about our CI setup that might come up. :)  We spent quite a bit of effort on it over the years and I think it has served us well.

PR #1845, opened 2025-07-04
get rid of all clippy warnings

hkBst:
I'm trying to remove clippy warnings from rustc, which is how I found myself here.

This does get rid of all clippy warnings, but some of the generated code expanded, which seems bad. Please let me know if there is a better way of doing this.


crates/core_arch/src/aarch64/neon/generated.rs
```diff
@@ -19491,49 +19489,115 @@ pub fn vqtbl1q_s8(a: int8x16_t, b: uint8x16_t) -> int8x16_t {
...
+    let a: uint8x16_t =
+        unsafe { simd_shuffle!(a, a, [15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0]) };
+    let b: uint8x8_t = unsafe { simd_shuffle!(b, b, [7, 6, 5, 4, 3, 2, 1, 0]) };
     unsafe {
-        let x = transmute(vqtbl1(transmute(a), b));
-        x
+        let ret_val: uint8x8_t = transmute(vqtbl1(transmute(a), b));
+        simd_shuffle!(ret_val, ret_val, [7, 6, 5, 4, 3, 2, 1, 0])
     }
 }
```
folkertdev: the duplication might be due to a fix in transmuting on big-endian systems that landed in a recent nightly. i.e. it might not actually be an effect of your changes, but rather of stdarch being out of date.

apparently aarch64 on BE is a thing?

crates/stdarch-gen-arm/spec/neon/aarch64.spec.yml
```diff
@@ -12289,16 +12286,13 @@ intrinsics:
...
-                    - FnCall: [transmute, [b]]
-                    - c
-      - Identifier: [x, Symbol]
+      - FnCall:
+          - transmute
+          - - FnCall:
+                - "{type[3]}"
+                - - FnCall: [transmute, [a]]
+                  - FnCall: [transmute, [b]]
+                  - c
```
folkertdev: hmm no it really is this change (and the similar one below). Somehow removing the intermediate variable triggers some sort of pattern that adds all the shuffles. I'm not really familiar with the generating code, so for now I'd suggest removing these changes, then the rest can be merged without issue.

folkertdev: then maybe open a PR with this change and the one below, so that we can ping someone who is more familiar with `stdarch-gen-arm`.

hkBst: Thanks @folkertdev that seems a good plan.

hkBst:
Now without the changes that are causing test failures and are otherwise suspect.

PR #1846, opened 2025-07-04
Add intrinsics for the new AMX target features

sayantn:
WIP

 - [ ] Tests
 - [ ] Links to Intrinsics Guide (not updated yet, probably will be after DiamondRapids is released)


sayantn:
@Amanieu should I go ahead and open this PR (with some modifications in the `stdarch_verify` logic to get around the fact that these aren't in the Intrinsics Guide yet)?

Amanieu:
It's fine to add these, but please add a FIXME: since these aren't in the XML they are not being tested.

sayantn:
Just to be clear, I will be adding the normal tests (I'm just being lazy), only thing that won't be tested is the presence of links to the Intel docs

sayantn:
Finally got around to finishing this lol

PR #1849, opened 2025-07-08
Fix remaining aarch64 clippy issues

folkertdev:
cc @hkBst 

Turns out we needed `big_endian_inverse: false` to tell the generator to not perform the shuffles. I also cleaned up some stray whitespace and compacted some array literals.


hkBst:
Thanks @folkertdev! Excellent work!

hkBst (approved):

PR #1852, opened 2025-07-08
`intrinsic-test`: use runner also for rust

folkertdev:
Let's see what this does on CI. It uses `runner` also for running the rust programs, so that they can be run outside of the docker container (assuming qemu and clang etc.)

### DIFF
--- a/crates/intrinsic-test/src/arm/config.rs
+++ b/crates/intrinsic-test/src/arm/config.rs
@@ -114,7 +114,6 @@ pub const AARCH_CONFIGURATIONS: &str = r#"
 #![cfg_attr(any(target_arch = "aarch64", target_arch = "arm64ec"), feature(stdarch_neon_fcma))]
 #![cfg_attr(any(target_arch = "aarch64", target_arch = "arm64ec"), feature(stdarch_neon_dotprod))]
 #![cfg_attr(any(target_arch = "aarch64", target_arch = "arm64ec"), feature(stdarch_neon_i8mm))]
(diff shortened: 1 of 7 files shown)


crates/intrinsic-test/src/arm/config.rs
```diff
@@ -114,7 +114,6 @@ pub const AARCH_CONFIGURATIONS: &str = r#"
 #![cfg_attr(any(target_arch = "aarch64", target_arch = "arm64ec"), feature(stdarch_neon_fcma))]
 #![cfg_attr(any(target_arch = "aarch64", target_arch = "arm64ec"), feature(stdarch_neon_dotprod))]
 #![cfg_attr(any(target_arch = "aarch64", target_arch = "arm64ec"), feature(stdarch_neon_i8mm))]
-#![cfg_attr(any(target_arch = "aarch64", target_arch = "arm64ec"), feature(stdarch_neon_sha3))]
```
folkertdev, line 117: this target feature has been stable for some time, and was generating warning spam.

crates/intrinsic-test/src/common/cli.rs
```diff
@@ -102,7 +102,7 @@ impl ProcessedCli {
         Self {
             toolchain,
             cpp_compiler,
-            c_runner,
+            runner: c_runner,
```
Amanieu: Rename `c_runner` to `runner` here.

folkertdev: fixed

PR #1853, opened 2025-07-08
Perform the first rustc pull.. for the second time

Kobzol:
https://github.com/rust-lang/stdarch/pull/1844 was created correctly, but sadly I didn't realize that this repo used squash merges instead of merge commits (sorry..).

Now the repo has been reconfigured to use merge commits. I think that if we repeat the pull, we can get away from this situation without any force push.

If it doesn't work for some reason, then we just force push this away anyway, so there's nothing to lose :)


Kobzol:
Ok, mergd with a master commit, good.

PR #1855, opened 2025-07-09
`intrinsic-test`: combine rust intrinsics in one program

folkertdev:
Basically just a big switch on the name, executing the (otherwise unchanged) logic for the intrinsic. Locally, this gives a 5 second speedup when processing the first 500 aarch64 intrinsics.

Let's see what CI says here.


folkertdev:
CI said meh, and this code is kind of messy. I'll redo it on top of one of the other branches.

PR #1856, opened 2025-07-09
`intrinsic-test`: combine files for better performance

folkertdev:
Combines the C and rust files into a file per thread. Especially for C this appears to skip a lot of work. Let's see how that works out in practice.


folkertdev:
So we're down from 28 minutes to just 8 here for `aarch64-unknown-linux-gnu`. It's not even the longest-running job at that point. 

Now to get this broken up into something that isn't torture to review...

folkertdev:
These ideas are now implemented in https://github.com/rust-lang/stdarch/pull/1863 and its parents.

PR #1859, opened 2025-07-10
`intrinsic-test`: allow names that don't match rust's naming conventions

folkertdev:
Locally, it looked like lots of time is spent by cargo printing warnings for incorrect names (lowercase constants, etc). Let's see if this matters in practice.


folkertdev:
apparently it's OK if it just gets `/dev/null`'d.

PR #1860, opened 2025-07-10
`stdarch-test`: various cleanups

folkertdev:
My goal here was to make the instruction matching stricter, to prevent accepting e.g. `fminnm` when the test specifies the `fmin` instruction must be used. 

But especially on x86 there are lots of cases where a prefix is (deliberately?) used.

Anyway, I also made some other cleanups. Can be reviewed commit-by-commit.

### DIFF
--- a/crates/stdarch-test/src/lib.rs
+++ b/crates/stdarch-test/src/lib.rs
@@ -92,7 +92,13 @@ pub fn assert(shim_addr: usize, fnname: &str, expected: &str) {
     };
 
     // Check whether the given instruction is part of the disassemblied body.
-    let found = expected == "nop" || instrs.iter().any(|s| s.starts_with(expected));
+    let found = expected == "nop"
+        || instrs.iter().any(|instruction| {
+            // Check that the next character is non-alphabetic. This prevents false negatives
+            // when e.g. `fminnm` was used but `fmin` was expected.
+            instruction.starts_with(expected)
+                && !instruction[expected.len()..].starts_with(|c: char| c.is_ascii_alphabetic())
+        });
 
     // Look for subroutine call instructions in the disassembly to detect whether
     // inlining failed: all intrinsics are `#[inline(always)]`, so calling one
--- a/crates/stdarch-test/src/lib.rs
+++ b/crates/stdarch-test/src/lib.rs
@@ -61,11 +61,12 @@ pub fn assert(shim_addr: usize, fnname: &str, expected: &str) {
     black_box(shim_addr);
 
     //eprintln!("shim name: {fnname}");
-    let function = &DISASSEMBLY
-        .get(&Function::new(fnname))
-        .unwrap_or_else(|| panic!("function \"{fnname}\" not found in the disassembly"));
+    let Some(function) = &DISASSEMBLY.get(&Function::new(fnname)) else {
+        panic!("function `{fnname}` not found in the disassembly")
+    };
     //eprintln!("  function: {:?}", function);
 
+    // Trim any filler instructions.
     let mut instrs = &function.instrs[..];
     while instrs.last().is_some_and(|s| s == "nop" || s == "int3") {
         instrs = &instrs[..instrs.len() - 1];
@@ -80,11 +81,17 @@ pub fn assert(shim_addr: usize, fnname: &str, expected: &str) {
     // 2. It is a mark, indicating that the instruction will be
     // compiled into other instructions - mainly because of llvm
     // optimization.
-    let expected = if expected == "unknown" {
-        "<unknown>" // Workaround for rust-lang/stdarch#1674, todo: remove when the issue is fixed
-    } else {
-        expected
+    let expected = match expected {
+        // `<unknown>` is what LLVM will generate for unknown instructions. We use this to fail
+        // loudly when LLVM does start supporting these instructions.
+        //
+        // This was introduced in https://github.com/rust-lang/stdarch/pull/1674 to work around the
+        // RISC-V P extension not yet being supported.
+        "unknown" => "<unknown>",
+        _ => expected,
     };
+
+    // Check whether the given instruction is part of the disassemblied body.
     let found = expected == "nop" || instrs.iter().any(|s| s.starts_with(expected));
 
     // Look for subroutine call instructions in the disassembly to detect whether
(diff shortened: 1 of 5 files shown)


crates/stdarch-test/src/lib.rs
```diff
@@ -84,12 +81,26 @@ pub fn assert(shim_addr: usize, fnname: &str, expected: &str) {
...
+        "unknown" => "<unknown>",
+        _ => expected,
     };
-    let found = expected == "nop" || instrs.iter().any(|s| s.starts_with(expected));
+
+    // Check whether the given instruction is part of the disassemblied body.
+    let found = expected == "nop"
+        || instrs.iter().any(|instruction| {
+            instruction.starts_with(expected)
+            // Check that the next character is non-alphabetic. This prevents false negatives
```
Amanieu: I would expect digits (`is_ascii_digit`) to also be part of the instruction name.

crates/stdarch-test/src/lib.rs
```diff
@@ -84,12 +81,26 @@ pub fn assert(shim_addr: usize, fnname: &str, expected: &str) {
...
+            // when e.g. `fminnm` was used but `fmin` was expected.
+            //
+            // TODO: resolve the conflicts (x86_64 has a bunch, probably others)
```
Amanieu: Are you intending to have this merged as it is, or wait for x86 conflicts to be resolved first? I'm happy to merge either way.

crates/stdarch-test/src/lib.rs
folkertdev: That is what I had earlier, but it gave lots of hits for `ld1` where the test just asserts on `ld`

crates/stdarch-test/src/lib.rs
folkertdev: This is good to go, we can work through the false positives over time (it's not the most fun task).

crates/stdarch-test/src/lib.rs
Amanieu: I don't see any intrinsics that specifically match on "ld". All the load instrinsics explicitly specify the number since this is an important part of the instruction.

folkertdev: hmm maybe it was a different instruction? In any case it gave a lot of false positives somewhere.

Amanieu: I would prefer if this included digits and had a way to explicitly opt-in to wildcard matching.

folkertdev:
I've updated the logic to use `is_ascii_alphanumeric`, though it's still commented out.

A selection of the false positives:

for x86 

```
---- core_arch::x86::avx::assert__mm256_store_pd_vmovap stdout ----
disassembly for stdarch_test_shim__mm256_store_pd_vmovap: 
	 0: vmovaps %ymm0,(%rdi)
	 1: vzeroupper
	 2: ret

thread 'core_arch::x86::avx::assert__mm256_store_pd_vmovap' panicked at crates/stdarch-test/src/lib.rs:204:9:
failed to find instruction `vmovap` in the disassembly

---- core_arch::x86::avx::assert__mm256_storeu_pd_vmovup stdout ----
disassembly for stdarch_test_shim__mm256_storeu_pd_vmovup: 
	 0: vmovups %ymm0,(%rdi)
	 1: vzeroupper
	 2: ret

thread 'core_arch::x86::avx::assert__mm256_storeu_pd_vmovup' panicked at crates/stdarch-test/src/lib.rs:204:9:
failed to find instruction `vmovup` in the disassembly

---- core_arch::x86::fma::assert__mm_fnmsub_ss_vfnmsub stdout ----
disassembly for stdarch_test_shim__mm_fnmsub_ss_vfnmsub: 
	 0: vfnmsub213ss %xmm2,%xmm1,%xmm0
	 1: ret

thread 'core_arch::x86::fma::assert__mm_fnmsub_ss_vfnmsub' panicked at crates/stdarch-test/src/lib.rs:204:9:
failed to find instruction `vfnmsub` in the disassembly

---- core_arch::x86::pclmulqdq::assert__mm_clmulepi64_si128_pclmul stdout ----
disassembly for stdarch_test_shim__mm_clmulepi64_si128_pclmul: 
	 0: pclmullqlqdq %xmm1,%xmm0
	 1: ret

thread 'core_arch::x86::pclmulqdq::assert__mm_clmulepi64_si128_pclmul' panicked at crates/stdarch-test/src/lib.rs:204:9:
failed to find instruction `pclmul` in the disassembly
```

for aarch64 I was thinking of `zip1`/`zip2`. I've now fixed that but there are a bunch that remain

```
---- core_arch::aarch64::neon::generated::assert_vmull_high_p64_pmull stdout ----
disassembly for stdarch_test_shim_vmull_high_p64_pmull: 
	 0: pmull2 v0.1q, v0.2d, v1.2d
	 1: mov x1, v0.d[1]
	 2: fmov x0, d0
	 3: ret

thread 'core_arch::aarch64::neon::generated::assert_vmull_high_p64_pmull' panicked at crates/stdarch-test/src/lib.rs:204:9:
failed to find instruction `pmull` in the disassembly

---- core_arch::aarch64::neon::generated::assert_vmull_high_p8_pmull stdout ----
disassembly for stdarch_test_shim_vmull_high_p8_pmull: 
	 0: pmull2 v0.8h, v0.16b, v1.16b
	 1: ret

thread 'core_arch::aarch64::neon::generated::assert_vmull_high_p8_pmull' panicked at crates/stdarch-test/src/lib.rs:204:9:
failed to find instruction `pmull` in the disassembly
```

it looks like for some instructions, the `2` version is used for some types/widths. So checking for the exact instruction would mean a bunch of duplication in the spec document.

Perhaps we can extend the macro to accept an explicit `prefix = "foo"` or similar, so that the default is to perform a strict compare, but we can opt-in into a less strict match.

Amanieu:
> it looks like for some instructions, the `2` version is used for some types/widths. So checking for the exact instruction would mean a bunch of duplication in the spec document.

The `2` version is used for the _high versions of the intrinsics which operate on the high half instead of the low half of a SIMD register for narrowing/widening operations. So it does make sense to check for it.

> Perhaps we can extend the macro to accept an explicit `prefix = "foo"` or similar, so that the default is to perform a strict compare, but we can opt-in into a less strict match.

That would be a good idea.

> for x86

The store operations only involve data movement and could probably have the instruction check removed.

folkertdev:
Yeah, well I fixed all aarch64 issues, but for x86_64 it's a lot

```
test result: FAILED. 11021 passed; 898 failed;
```

so eh I'd propose we merge this now, and then we'll see about fixing those

PR #1861, opened 2025-07-10
`intrinsic-test`: streamline c compilation

folkertdev:
Preparation for combining the C files. That is the biggest speedup, so I want to do that first to make later refactorings less painful.

### DIFF
--- a/crates/intrinsic-test/src/arm/compile.rs
+++ b/crates/intrinsic-test/src/arm/compile.rs
@@ -1,23 +1,23 @@
+use crate::common::cli::ProcessedCli;
 use crate::common::compile_c::CompilationCommandBuilder;
 use crate::common::gen_c::compile_c_programs;
 
-pub fn compile_c_arm(
-    intrinsics_name_list: &[String],
-    compiler: &str,
-    target: &str,
-    cxx_toolchain_dir: Option<&str>,
-) -> bool {
+pub fn compile_c_arm(config: &ProcessedCli, intrinsics_name_list: &[String]) -> bool {
+    let Some(ref cpp_compiler) = config.cpp_compiler else {
+        return true;
+    };
+
     // -ffp-contract=off emulates Rust's approach of not fusing separate mul-add operations
     let mut command = CompilationCommandBuilder::new()
         .add_arch_flags(vec!["armv8.6-a", "crypto", "crc", "dotprod", "fp16"])
-        .set_compiler(compiler)
-        .set_target(target)
+        .set_compiler(cpp_compiler)
+        .set_target(&config.target)
         .set_opt_level("2")
-        .set_cxx_toolchain_dir(cxx_toolchain_dir)
+        .set_cxx_toolchain_dir(config.cxx_toolchain_dir.as_deref())
         .set_project_root("c_programs")
         .add_extra_flags(vec!["-ffp-contract=off", "-Wno-narrowing"]);
 
-    if !target.contains("v7") {
+    if !config.target.contains("v7") {
         command = command.add_arch_flags(vec!["faminmax", "lut", "sha3"]);
     }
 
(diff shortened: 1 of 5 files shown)


ci/docker/aarch64_be-unknown-linux-gnu/Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN mkdir /toolchains && mv "./${TOOLCHAIN}" /toolchains
 ENV AARCH64_BE_TOOLCHAIN="/toolchains/${TOOLCHAIN}"
 ENV AARCH64_BE_LIBC="${AARCH64_BE_TOOLCHAIN}/aarch64_be-none-linux-gnu/libc"
 
-ENV CARGO_TARGET_AARCH64_BE_UNKNOWN_LINUX_GNU_LINKER="${AARCH64_BE_TOOLCHAIN}/bin/aarch64_be-none-linux-gnu-gcc"
+ENV CARGO_TARGET_AARCH64_BE_UNKNOWN_LINUX_GNU_LINKER="${AARCH64_BE_TOOLCHAIN}/bin/aarch64_be-none-linux-gnu-g++"
```
tgross35: gcc can still be used to link c++ object files right?

folkertdev: Well, I ran into some linker issues with it earlier, and this is the linker that is picked here (and fixed those linker issues, though maybe other fixes were possible):

https://github.com/rust-lang/stdarch/blob/af2aa767b55625fc33c13f2f9102222fe428b38c/crates/intrinsic-test/src/arm/compile.rs#L34-L37

So this solved a problem and seemed more consistent. Are there downsides to using `g++`?

tgross35: None that I know of, I was just thinking it would be a change without any effect. Interesting that it makes a difference

edit: Oh, if it needs to link the c++ libraries then that makes more sense.

folkertdev: True, but on the other hand it did work up to now (and, it might still, I just had trouble with it earlier). But I figured we may as well be consistent here.

folkertdev: digging in a little further, this is the linker command that we execute. Apparently `clang19++` cannot itself construct the final binary, so `intrinsic-test-programs.cpp` is first compiled to an object file by `clang19++`, then linked using this custom linker. With `gcc` this fails:

```
[crates/intrinsic-test/src/common/compile_c.rs:209:17] &cmd = Command {
    program: "/toolchains/arm-gnu-toolchain-14.2.rel1-x86_64-aarch64_be-none-linux-gnu/bin/aarch64_be-none-linux-gnu-gcc",
    args: [
        "/toolchains/arm-gnu-toolchain-14.2.rel1-x86_64-aarch64_be-none-linux-gnu/bin/aarch64_be-none-linux-gnu-gcc",
        "mod_0.o",
        "mod_1.o",
        "mod_2.o",
        "mod_3.o",
        "mod_4.o",
        "mod_5.o",
        "mod_6.o",
        "mod_7.o",
        "mod_8.o",
        "mod_9.o",
        "mod_10.o",
        "mod_11.o",
        "mod_12.o",
        "mod_13.o",
        "mod_14.o",
        "mod_15.o",
        "intrinsic-test-programs.o",
        "-o",
        "intrinsic-test-programs",
    ],
    cwd: Some(
        "c_programs",
    ),
    create_pidfd: false,
}
```

This gives errors like this

```
aarch64_be-none-linux-gnu/bin/ld: mod_0.o: in function run___crc32b()':
mod_0.cpp:(.text+0x1c): undefined reference to std::cout'
aarch64_be-none-linux-gnu/bin/ld: mod_0.cpp:(.text+0x2c): undefined reference to std::cout'
aarch64_be-none-linux-gnu/bin/ld: mod_0.cpp:(.text+0x5c): undefined reference to std::ctype<char>::_M_widen_init() const'
aarch64_be-none-linux-gnu/bin/ld: mod_0.cpp:(.text+0x7c): undefined reference to std::ostream::put(char)'
aarch64_be-none-linux-gnu/bin/ld: mod_0.cpp:(.text+0x80): undefined reference to std::ostream::flush()'
aarch64_be-none-linux-gnu/bin/ld: mod_0.cpp:(.text+0xa4): undefined reference to std::basic_ostream<char, std::char_traits<char> >& std::__ostream_insert<char, std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*, long)'
```

So yeah, we're linking C++ now, and then using `g++` seems like the best option.

sayantn: > `clang19++` cannot itself construct the final binary

that seems weird!

crates/intrinsic-test/src/arm/compile.rs
```diff
@@ -1,64 +1,51 @@
...
+            &format!("{cxx_toolchain_dir}/aarch64_be-none-linux-gnu/include/c++/14.2.1"),
+            "--include-directory",
+            &format!("{cxx_toolchain_dir}/aarch64_be-none-linux-gnu/include/c++/14.2.1/aarch64_be-none-linux-gnu"),
+            "-L",
+            &format!("{cxx_toolchain_dir}/lib/gcc/aarch64_be-none-linux-gnu/14.2.1"),
+            "-L",
+            &format!("{cxx_toolchain_dir}/aarch64_be-none-linux-gnu/libc/usr/lib"),
+            "-B",
+            &format!("{cxx_toolchain_dir}/lib/gcc/aarch64_be-none-linux-gnu/14.2.1"),
+        ]);
```
folkertdev: So, I got this to work, meaning we can drop the custom linker stuff. I've also reverted the linker change from `gcc` to `g++` for now, it might be needed later but we can change it then.

folkertdev:
Lmk if this is still too tough to review btw, I could attempt to break it up further.

sayantn:
LGTM, just resolve those merge conflicts (just version bump for gcc). Also, shouldn't there be a better way to add the libraries than just hardcoding the librarypath, like at least making the version a parameter?

PR #1862, opened 2025-07-10
`intrinsic-test`: combine C files for faster compilation

folkertdev:
There are 3 main changes

- emit the arguments (the arrays of inputs) inside of the test function, and rename the test function from `main` to `run_{intrinsic_name}`.
- generate number-of-cores files, and evenly divide all intrinsics over these files. Each file now has the functions for running many different intrinsic tests
- generate a `main` that is just a gigantic switch to select the right test to run

Parts of this are still inefficient, and for the full effect the rust side also needs work, but I suspect this is the biggest win (if not, it's actually cargo, we'll see I guess), and sort of a manageable change to review (hopefully).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,8 @@ members = [
   "examples",
 ]
 exclude = [
-  "crates/wasm-assert-instr-tests"
+  "crates/wasm-assert-instr-tests",
+  "rust_programs",
 ]
 
 [profile.release]
--- a/crates/intrinsic-test/src/arm/mod.rs
+++ b/crates/intrinsic-test/src/arm/mod.rs
@@ -4,21 +4,33 @@ mod intrinsic;
 mod json_parser;
 mod types;
 
+use std::fs::File;
+
+use rayon::prelude::*;
+
+use crate::arm::config::POLY128_OSTREAM_DEF;
 use crate::common::SupportedArchitectureTest;
 use crate::common::cli::ProcessedCli;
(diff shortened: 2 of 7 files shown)


folkertdev:
This brings down the `aarch64_be` job to 10 minutes (down from 40m), so it is no longer the longest-running job (that is now the `x86_64-unknown-linux-gnu`). So in terms of CI throughput, this gives us all of the benefits we're going to get.

Combining the rust files is still useful for getting faster feedback for failures though.

cc @madhav-madhusoodanan some of the C compilation code is a bit too arm-specific now, but I didn't want to generalize in the abstract. So we can see later how to make it work for both arm and x86_64. Shouldn't be hard at all, just needs some minor refactoring I think.

madhav-madhusoodanan:
> This brings down the `aarch64_be` job to 10 minutes (down from 40m), so it is no longer the longest-running job (that is now the `x86_64-unknown-linux-gnu`). So in terms of CI throughput, this gives us all of the benefits we're going to get.
> 
> Combining the rust files is still useful for getting faster feedback for failures though.
> 
> cc @madhav-madhusoodanan some of the C compilation code is a bit too arm-specific now, but I didn't want to generalize in the abstract. So we can see later how to make it work for both arm and x86_64. Shouldn't be hard at all, just needs some minor refactoring I think.

On it!
I just merged the master branch into my feature branch. There is some work on the way the C code is constructed and compiled, but it's totally doable!

PR #1868, opened 2025-07-11
`stdarch-gen-arm`: remove `lazy_static`, use `LazyLock` instead

folkertdev:


crates/stdarch-gen-arm/src/load_store_tests.rs
```diff
@@ -801,13 +801,11 @@ fn assert_vector_matches_u64(vector: svuint64_t, expected: svuint64_t) {{
...
-    );
-}
+    )
+});
 
-lazy_static! {
-    static ref MANUAL_TESTS: String = format!(
-        "#[simd_test(enable = \"sve\")]
-unsafe fn test_ffr() {{
+static MANUAL_TESTS: &str = "#[simd_test(enable = \"sve\")]
```
tgross35: Could this just be a const?

tgross35 (approved):

crates/stdarch-gen-arm/src/load_store_tests.rs
folkertdev: yeah actually. I'm normally a bit worried about const's getting inlined but this is a string and it's used just once so it should be OK.

Issue #1659, opened 2024-10-24
aarch64: use portable SIMD intrinsics where possible


RalfJung:
We have a number of portable SIMD intrinsics in [`core::intrinsics::simd`](https://doc.rust-lang.org/nightly/std/intrinsics/simd/index.html). The advantage of using those is that they have to only be implemented once by non-LLVM backends (Miri, cranelift, gcc) and then they work consistently across all architectures. The x86 parts of stdarch thankfully make good use of them, which heavily reduces the number of intrinsics Miri and the other backends have to special-case by hand.

It would be good to do the same for aarch64. AFAIK aarch64 has completely standard float semantics including the behavior on subnormals, so there shouldn't be a reason to use architecture-specific intrinsics. (arm-32 is a different case since the neon intrinsics there flush subnormals to 0, which makes it technically not correct to use the portable intrinsics. But also at least LLVM doesn't implement the portable intrinsics properly on arm-32 so they actually do flush subnormals, except if optimizations kick in in which case we can get unsoundness.)

workingjubilee:
> It would be good to do the same for aarch64. AFAIK aarch64 has completely standard float semantics including the behavior on subnormals, so there shouldn't be a reason to use architecture-specific intrinsics

correct.

PR #1871, opened 2025-07-12
`aarch64`: use more of `intrinsics::simd` for min/max

folkertdev:
cc https://github.com/rust-lang/stdarch/pull/1858 https://github.com/rust-lang/stdarch/issues/1659

### DIFF
--- a/crates/core_arch/src/aarch64/neon/generated.rs
+++ b/crates/core_arch/src/aarch64/neon/generated.rs
@@ -13381,14 +13381,7 @@ pub fn vmaxvq_f64(a: float64x2_t) -> f64 {
 #[stable(feature = "neon_intrinsics", since = "1.59.0")]
 #[cfg_attr(test, assert_instr(smaxv))]
 pub fn vmaxv_s8(a: int8x8_t) -> i8 {
-    unsafe extern "unadjusted" {
-        #[cfg_attr(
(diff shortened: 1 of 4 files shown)


crates/core_arch/src/aarch64/neon/generated.rs
```diff
@@ -13229,14 +13229,7 @@ pub fn vmaxh_f16(a: f16, b: f16) -> f16 {
...
 pub fn vmaxnm_f64(a: float64x1_t, b: float64x1_t) -> float64x1_t {
-    unsafe extern "unadjusted" {
-        #[cfg_attr(
-            any(target_arch = "aarch64", target_arch = "arm64ec"),
-            link_name = "llvm.aarch64.neon.fmaxnm.v1f64"
-        )]
-        fn _vmaxnm_f64(a: float64x1_t, b: float64x1_t) -> float64x1_t;
-    }
-    unsafe { _vmaxnm_f64(a, b) }
+    unsafe { simd_fmax(a, b) }
```
RalfJung, line 13232: I don't think these have the same semantics. `simd_fmax` is documented to follow IEEE 754-2008 semantics, which means the behavior of signed zeros is non-deterministic. Also, due to the usual LLVM SNaN shenanigans, `minnum(SNaN, x)` can either return a NaN or x. IMO this commit should be reverted. Also see https://github.com/rust-lang/rust/issues/153395.

PR #1872, opened 2025-07-12
`aarch64`: use `intrinsics::simd` for horizontal add and `abs`

folkertdev:


crates/core_arch/src/aarch64/neon/generated.rs
```diff
@@ -298,46 +298,40 @@ pub fn vabsq_f64(a: float64x2_t) -> float64x2_t {
...
+    unsafe {
+        let neg: int64x1_t = simd_neg(a);
+        let mask: int64x1_t = simd_ge(a, neg);
+        simd_select(mask, a, neg)
     }
-    unsafe { _vabs_s64(a) }
 }
 #[doc = "Absolute Value (wrapping)."]
-#[doc = "[Arm's documentation](https://developer.arm.com/architectures/instruction-sets/intrinsics/vabsd_s64)"]
+#[doc = "[Arm's documentation](https://developer.arm.com/architectures/instruction-sets/intrinsics/vabsq_s64)"]
```
folkertdev, line 308: the diff here is wonky because the order of the definitions changed.

crates/stdarch-gen-arm/spec/neon/aarch64.spec.yml
```diff
@@ -13023,6 +13023,25 @@ intrinsics:
...
+    types:
+      - [i64, i64]
+    compose:
+      # FIXME(llvm): this should be `i64::wrapping_abs` but it optimizes differently.
+      # See https://github.com/llvm/llvm-project/issues/148388.
+      - LLVMLink:
+          name: "vabsd_s64"
+          links:
+             - link: "llvm.aarch64.neon.abs.i64"
+               arch: aarch64,arm64ec
```
folkertdev: cc https://github.com/llvm/llvm-project/issues/148388

Just kept this as-is for now, but from what I can tell `i64::wrapping_abs` would be equivalent.

Amanieu: The LLVM issue is closed, is this comment still relevant?

folkertdev: I updated the comment. it's kind of subtle, but the intrinsic keeps the value in a simd register, while the scalar version uses GPRs.

PR #1873, opened 2025-07-13
[do not merge] Remove std_detect from CI

Kobzol:
Testing the stdarch side of https://github.com/rust-lang/rust/pull/143412.


Kobzol:
Good, CI is green (https://github.com/rust-lang/stdarch/actions/runs/16248563487), so removal of `std_detect` from `stdarch`'s CI should be good.

PR #1874, opened 2025-07-15
loongarch: Mark partial intrinsics as safe

heiher:


crates/core_arch/src/loongarch_shared/mod.rs
```diff
@@ -75,59 +79,63 @@ unsafe extern "unadjusted" {
...
 /// Generates the memory barrier instruction
 #[inline]
 #[unstable(feature = "stdarch_loongarch", issue = "117427")]
-pub unsafe fn dbar<const IMM15: i32>() {
+pub fn dbar<const IMM15: i32>() {
     static_assert_uimm_bits!(IMM15, 15);
-    __dbar(IMM15);
+    unsafe {
+        __dbar(IMM15);
+    }
```
folkertdev: nit: this would save 2 lines

```suggestion
    unsafe { __dbar(IMM15) };
```

crates/core_arch/src/loongarch_shared/mod.rs
```diff
@@ -5,20 +5,24 @@ use crate::arch::asm;
...
     (val, tid)
 }
 
 /// Reads the upper 32-bit stable counter value and the counter ID
 #[inline]
 #[unstable(feature = "stdarch_loongarch", issue = "117427")]
-pub unsafe fn rdtimeh_w() -> (i32, isize) {
+pub fn rdtimeh_w() -> (i32, isize) {
     let val: i32;
     let tid: isize;
```
folkertdev: nit: this would also work

```suggestion
    let (val, tid): (i32, isize);
```

crates/stdarch-gen-loongarch/src/main.rs
```diff
@@ -456,7 +457,10 @@ fn gen_bind_body(
             };
             rustc_legacy_const_generics = "rustc_legacy_const_generics(2, 3)";
         }
-        format!("pub unsafe fn {current_name}{fn_inputs} {fn_output}")
+        format!(
+            "pub {}fn {current_name}{fn_inputs} {fn_output}",
+            if is_mem { "unsafe " } else { "" }
+        )
```
folkertdev, line 463: I understand this as a heuristic, but is it truly the case that only pointer reads/writes cause unsafety for these intrinsics?

crates/core_arch/src/loongarch64/lasx/generated.rs
```diff
@@ -1495,3501 +1495,3725 @@ unsafe extern "unadjusted" {
...
 #[inline]
 #[target_feature(enable = "lasx")]
 #[rustc_legacy_const_generics(1)]
 #[unstable(feature = "stdarch_loongarch", issue = "117427")]
-pub unsafe fn lasx_xvslei_b<const IMM_S5: i32>(a: v32i8) -> v32i8 {
-    static_assert_simm_bits!(IMM_S5, 5);
-    __lasx_xvslei_b(a, IMM_S5)
+pub fn lasx_xvslei_b<const IMM_S5: i32>(a: v32i8) -> v32i8 {
+    unsafe {
+        static_assert_simm_bits!(IMM_S5, 5);
```
folkertdev: can the assert be lifted out of the unsafe block?

folkertdev (commented):
Overall this looks good, I added some nits and questions inline

crates/stdarch-gen-loongarch/src/main.rs
heiher: Currently, only the memory-accessing SIMD intrinsics are unsafe. If new instructions are added in the future, we will make sure to handle them accordingly during updates.

crates/core_arch/src/loongarch64/lasx/generated.rs
heiher: All done. Thank you!

folkertdev:
the generated version of the file is not the same as the checked-in version (just some whitespace). Can you re-run locally and add the updated file?

heiher:
> the generated version of the file is not the same as the checked-in version (just some whitespace). Can you re-run locally and add the updated file?

Fixed. The generated code no longer needs to be reformatted. (https://github.com/rust-lang/stdarch/pull/1874/files#diff-1f7c8be4f6ea74d070d6b6b9201605541a6c52f74f7fa3c78c83c892b3d5dd7aL14)

PR #1878, opened 2025-07-16
`intrinsic-test`: Implemented DerefMut for ArmIntrinsicTest

madhav-madhusoodanan:
## Summary
1. Moved the return type of IntrinsicType::from_c to Rust<Self, String> from Result<Box<Self>, String>
2. Implemented DerefMut for ArmIntrinsicTest struct

## Context
This PR is part 2 of the changes that were originally made in the x86 extension PR rust-lang/stdarch#1814 for intrinsic-test, and is intended to unblock the speedup efforts that are being made in PRs rust-lang/stdarch#1856 , rust-lang/stdarch#1862 and rust-lang/stdarch#1863.

### DIFF
--- a/crates/intrinsic-test/src/arm/intrinsic.rs
+++ b/crates/intrinsic-test/src/arm/intrinsic.rs
@@ -2,7 +2,7 @@ use crate::common::argument::ArgumentList;
 use crate::common::indentation::Indentation;
 use crate::common::intrinsic::{Intrinsic, IntrinsicDefinition};
 use crate::common::intrinsic_helpers::{IntrinsicType, IntrinsicTypeDefinition, Sign, TypeKind};
-use std::ops::Deref;
+use std::ops::{Deref, DerefMut};
 
 #[derive(Debug, Clone, PartialEq)]
 pub struct ArmIntrinsicType(pub IntrinsicType);
@@ -12,6 +12,12 @@ impl Deref for ArmIntrinsicType {
 
     fn deref(&self) -> &Self::Target {
         &self.0
+    }
+}
+
+impl DerefMut for ArmIntrinsicType {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
     }
 }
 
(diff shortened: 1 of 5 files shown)


crates/intrinsic-test/src/arm/intrinsic.rs
```diff
@@ -15,6 +15,12 @@ impl Deref for ArmIntrinsicType {
     }
 }
 
+impl DerefMut for ArmIntrinsicType {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
+
```
folkertdev, line 23: Is this really worth it? 

In general it is my impression that this code base is much too generic already (dyn, bunch of traits), but then only had one concrete implementation. 

Deref is generally discouraged, why does it pull its weight here?

madhav-madhusoodanan: There are 3 levels of functionality for structs that implement `IntrinsicTypeDefinition`:
1. The basics (name, is_simd, is_ptr, etc) where the logic is strightforward and doesn't need to be implemented per architecture
2. The mid level (get_load_function, c_type, etc) where the logic may vary across architectures, and would need architecture-specific definition
3. The code generation part (generate_c_test_loop, format_c_main_template, etc) where the logic is again common across architectures:
    a. Define headers
    b. Define any typecasting functionality as per necessity
    c. Define the variable
    d. Define the list of values that would be passed as an argument to the intrinsic
    e. Define test loops
    f. Load the arguments, call the intrinsics and print the value
    
We figured that:
1. Any functionality that would be needed at Level 2 could use `Deref` and cleanly use the functions defined on the inner type.
2. Any functionality that would be needed at Level 3 could use Generics and associated traits to cleanly use the architecture-specific functionality

madhav-madhusoodanan: I look forward to learning how we can better design `intrinsic-test` so that it supports easy addition of other architectures too.

folkertdev: hmm ok. well I don't want to block this PR on it, it does fit in with the current design. 

Can you mark this as ready when it's ready?

r? folkertdev

madhav-madhusoodanan: Ohh yes, I thought this needed working, hence put it as a Draft.
I think at this stage, it is ready to be merged.

PR #1879, opened 2025-07-19
loongarch: Use unified data types for SIMD intrinsics

heiher:
Suggested by @sayantn

### DIFF
--- a/crates/core_arch/src/loongarch64/lasx/types.rs
+++ b/crates/core_arch/src/loongarch64/lasx/types.rs
@@ -1,33 +1,140 @@
 types! {
     #![unstable(feature = "stdarch_loongarch", issue = "117427")]
 
-    /// LOONGARCH-specific 256-bit wide vector of 32 packed `i8`.
-    pub struct v32i8(32 x pub(crate) i8);
+    /// 256-bit wide integer vector type, LoongArch-specific
+    ///
+    /// This type is the same as the `__m256i` type defined in `lasxintrin.h`,
+    /// representing a 256-bit SIMD register. Usage of this type typically
+    /// occurs in conjunction with the `lasx` target features for LoongArch.
+    ///
+    /// Internally this type may be viewed as:
+    ///
+    /// * `i8x32` - thirty two `i8` values packed together
+    /// * `i16x16` - sixteen `i16` values packed together
+    /// * `i32x8` - eight `i32` values packed together
+    /// * `i64x4` - four `i64` values packed together
+    ///
+    /// (as well as unsigned versions). Each intrinsic may interpret the
+    /// internal bits differently, check the documentation of the intrinsic
+    /// to see how it's being used.
+    ///
+    /// The in-memory representation of this type is the same as the one of an
+    /// equivalent array (i.e. the in-memory order of elements is the same, and
+    /// there is no padding); however, the alignment is different and equal to
+    /// the size of the type. Note that the ABI for function calls may *not* be
+    /// the same.
+    ///
+    /// Note that this means that an instance of `m256i` typically just means
+    /// a "bag of bits" which is left up to interpretation at the point of use.
+    ///
+    /// Most intrinsics using `m256i` are prefixed with `lasx_` and the integer
+    /// types tend to correspond to suffixes like "b", "h", "w" or "d".
+    pub struct m256i(4 x i64);
 
-    /// LOONGARCH-specific 256-bit wide vector of 16 packed `i16`.
-    pub struct v16i16(16 x pub(crate) i16);
+    /// 256-bit wide set of eight `f32` values, LoongArch-specific
+    ///
+    /// This type is the same as the `__m256` type defined in `lasxintrin.h`,
+    /// representing a 256-bit SIMD register which internally consists of
+    /// eight packed `f32` instances. Usage of this type typically occurs in
+    /// conjunction with the `lasx` target features for LoongArch.
+    ///
+    /// Note that unlike `m256i`, the integer version of the 256-bit registers,
+    /// this `m256` type has *one* interpretation. Each instance of `m256`
+    /// always corresponds to `f32x8`, or eight `f32` values packed together.
+    ///
+    /// The in-memory representation of this type is the same as the one of an
+    /// equivalent array (i.e. the in-memory order of elements is the same, and
+    /// there is no padding  between two consecutive elements); however, the
+    /// alignment is different and equal to the size of the type. Note that the
+    /// ABI for function calls may *not* be the same.
+    ///
+    /// Most intrinsics using `m256` are prefixed with `lasx_` and are
+    /// suffixed with "s".
+    pub struct m256(8 x f32);
 
-    /// LOONGARCH-specific 256-bit wide vector of 8 packed `i32`.
-    pub struct v8i32(8 x pub(crate) i32);
+    /// 256-bit wide set of four `f64` values, LoongArch-specific
+    ///
+    /// This type is the same as the `__m256d` type defined in `lasxintrin.h`,
+    /// representing a 256-bit SIMD register which internally consists of
+    /// four packed `f64` instances. Usage of this type typically occurs in
+    /// conjunction with the `lasx` target features for LoongArch.
+    ///
+    /// Note that unlike `m256i`, the integer version of the 256-bit registers,
+    /// this `m256d` type has *one* interpretation. Each instance of `m256d`
+    /// always corresponds to `f64x4`, or four `f64` values packed together.
+    ///
+    /// The in-memory representation of this type is the same as the one of an
+    /// equivalent array (i.e. the in-memory order of elements is the same, and
+    /// there is no padding); however, the alignment is different and equal to
+    /// the size of the type. Note that the ABI for function calls may *not* be
+    /// the same.
+    ///
+    /// Most intrinsics using `m256d` are prefixed with `lasx_` and are suffixed
+    /// with "d". Not to be confused with "d" which is used for `m256i`.
+    pub struct m256d(4 x f64);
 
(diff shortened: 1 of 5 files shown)


heiher:
Although type aliases won't break the build, they will trigger clippy wanrings. If the user's CI treats warnings as errors, this could cause the CI to fail.

```
warning: transmute from a type (`std::arch::loongarch64::m256i`) to itself
   --> src/simd/lines_bwd.rs:204:47
    |
204 |                 let tmp = lasx_xvpermi_q::<1>(T(sum), T(sum));
    |                                               ^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_transmute
    = note: `#[warn(clippy::useless_transmute)]` on by default
```

sayantn:
seeing that it is a *`clippy`* warning, not even rustc-builtin, I would say this is acceptable

crates/stdarch-gen-loongarch/src/main.rs
```diff
@@ -281,27 +402,27 @@ fn gen_bind_body(
             let fn_output = if out_t.to_lowercase() == "void" {
                 String::new()
             } else {
-                format!(" -> {}", type_to_rst(out_t, is_store))
+                format!(" -> {}", type_to_rst(out_t, is_store, true))
```
folkertdev: would using a custom enum instead of a `bool` be nicer here? just reading this it's hard to know what `true` means.

folkertdev (commented):
re. the transmute warning, that is fine, but it's probably nicest for users to `#[allow]` the lint (or even `#[expect]` if you can make that work)

crates/stdarch-gen-loongarch/src/main.rs
heiher: Good idea. Thank you!

heiher:
a gentle ping.

crates/core_arch/src/loongarch64/lasx/types.rs
```diff
@@ -1,33 +1,140 @@
...
+    /// This type is the same as the `__m256i` type defined by C intrinsic,
+    /// representing a 256-bit SIMD register. Usage of this type typically
+    /// corresponds to the `lasx` target features for LoongArch.
+    ///
+    /// Internally this type may be viewed as:
+    ///
+    /// * `i8x32` - thirty two `i8` variables packed together
+    /// * `i16x16` - sixteen `i16` variables packed together
+    /// * `i32x8` - eight `i32` variables packed together
+    /// * `i64x4` - four `i64` variables packed together
```
folkertdev:
```suggestion
    /// * `i8x32` - thirty-two `i8` values packed together
    /// * `i16x16` - sixteen `i16` values packed together
    /// * `i32x8` - eight `i32` values packed together
    /// * `i64x4` - four `i64` values packed together
```

crates/core_arch/src/loongarch64/lasx/types.rs
```diff
@@ -1,33 +1,140 @@
...
+    /// Note that this means that an instance of `m256i` typically just means
+    /// a "bag of bits" which is left up to interpretation at the point of use.
+    ///
+    /// Most intrinsics using `m256i` are prefixed with `lasx_` and the integer
+    /// types tend to correspond to suffixes like "b", "h", "w" or "d".
+    pub struct m256i(4 x i64);
 
-    /// LOONGARCH-specific 256-bit wide vector of 16 packed `i16`.
-    pub struct v16i16(16 x pub(crate) i16);
+    /// 256-bit wide set of eight `f32` types, LoongArch-specific
```
folkertdev:
```suggestion
    /// 256-bit wide set of eight `f32` values, LoongArch-specific
```

crates/core_arch/src/loongarch64/lasx/types.rs
```diff
@@ -1,33 +1,140 @@
...
+    ///
+    /// This type is the same as the `__m256` type defined by C intrinsic,
```
folkertdev: what is "C intrinsic"? you mean the relevant header file?

crates/core_arch/src/loongarch64/lasx/types.rs
```diff
@@ -1,33 +1,140 @@
...
+    /// representing a 256-bit SIMD register which internally is consisted of
```
folkertdev:
```suggestion
    /// representing a 256-bit SIMD register which internally consists of
```

crates/core_arch/src/loongarch64/lasx/types.rs
```diff
@@ -1,33 +1,140 @@
...
+    /// eight packed `f32` instances. Usage of this type typically corresponds
+    /// to the `lasx` target features for LoongArch.
```
folkertdev: that reads weird. Do you mean that to use this type, the `lasx` target feature must be enabled?

crates/core_arch/src/loongarch64/lasx/types.rs
```diff
@@ -1,33 +1,140 @@
...
+    ///
+    /// Note that unlike `m256i`, the integer version of the 256-bit registers,
+    /// this `m256` type has *one* interpretation. Each instance of `m256`
+    /// always corresponds to `f32x8`, or eight `f32` types packed together.
```
folkertdev:
```suggestion
    /// always corresponds to `f32x8`, or eight `f32` values packed together.
```

crates/core_arch/src/loongarch64/lasx/types.rs
```diff
@@ -1,33 +1,140 @@
...
+    /// alignment is different and equal to the size of the type. Note that the
+    /// ABI for function calls may *not* be the same.
+    ///
+    /// Most intrinsics using `m256` are prefixed with `lasx_` and are
+    /// suffixed with "s".
+    pub struct m256(8 x f32);
 
-    /// LOONGARCH-specific 256-bit wide vector of 8 packed `i32`.
-    pub struct v8i32(8 x pub(crate) i32);
+    /// 256-bit wide set of four `f64` types, LoongArch-specific
```
folkertdev:
```suggestion
    /// 256-bit wide set of four `f64` values, LoongArch-specific
```

crates/core_arch/src/loongarch64/lasx/types.rs
```diff
@@ -1,33 +1,140 @@
...
+    ///
+    /// This type is the same as the `__m256d` type defined by C intrinsic,
+    /// representing a 256-bit SIMD register which internally is consisted of
+    /// four packed `f64` instances. Usage of this type typically corresponds
+    /// to the `lasx` target features for LoongArch.
+    ///
```
folkertdev: just in general, if you update the text for f32x8, make sure this is updated too.

crates/core_arch/src/loongarch64/lasx/types.rs
heiher: Yes. https://github.com/llvm/llvm-project/blob/main/clang/lib/Headers/lasxintrin.h

crates/core_arch/src/loongarch64/lasx/types.rs
heiher: No. What I meant is that this type is typically used in conjunction with the `lasx` target features for LoongArch.

crates/core_arch/src/loongarch64/lasx/types.rs
heiher: All done. Thanks!

crates/core_arch/src/loongarch64/lasx/types.rs
folkertdev: Right, so then here and elsewhere I'd rephrase that as 

```suggestion
    /// This type is the same as the `__m256` type defined in `lasxintrin.h`,
```

crates/core_arch/src/loongarch64/lsx/types.rs
```diff
@@ -1,33 +1,140 @@
...
+pub(crate) struct __v4f32([f32; 4]);
+#[allow(non_camel_case_types)]
+#[repr(simd)]
+pub(crate) struct __v2f64([f64; 2]);
 
-    /// LOONGARCH-specific 128-bit wide vector of 2 packed `f64`.
-    pub struct v2f64(2 x pub(crate) f64);
-}
+// These type aliases are provided solely for transitional compatibility.
+// They are temporary and will be removed when appropriate.
```
folkertdev, line 110: you could annotate them with `#[deprecate]` (e.g. an example form elsewhere in stdarch):

```
#[deprecated(
    since = "1.29.0",
    note = "See issue #51810 - use inline assembly instead"
)]
```

That gives people some a nudge and some time to upgrade.

heiher: This will require updating each usage site; otherwise, the following error will be triggered:
```
error: use of type alias `core_arch::loongarch64::lsx::types::v16u8` that will be deprecated in future version 1.91.0: use m128i instead
```

folkertdev: When I try it in the playground I get a warning, not an error. 

But it's totally up to you whether you want to transition between the symbol names in this way. It's a nightly api, so you are technically allowed to remove symbols at any time.

heiher: I believe some user projects treat warnings as errors. This goes against the goal of preserving existing builds as much as possible. I would suggest not doing this for now, and instead removing the aliases entirely after some time has passed.

folkertdev: you can allow specific warnings on a per-crate basis, if you need more time to fix them. Anyway, either way is fine.

heiher: Okay. Thank you!

PR #1880, opened 2025-07-22
`intrinsic-test`: remove `lazy_static` dependency 

folkertdev:
we use `std::sync::LazyLock` now.


tgross35 (approved):
This change itself looks fine, but it looks like stdarch needs a sync

folkertdev:
Ah, right, that PR got merged.

PR #1883, opened 2025-07-23
Rustc pull update

Kobzol:
Merge ref '5a30e4307f05' from rust-lang/rust

Pull recent changes from https://github.com/rust-lang/rust via Josh. In particular, this PR removes `std_detect` from `stdarch` (https://github.com/rust-lang/rust/pull/143412).

Upstream ref: 5a30e4307f0506bed87eeecd171f8366fdbda1dc
Filtered ref: 59749e9f8c765d3021796a9fe0c188643c4b8d77

This merge was created using https://github.com/rust-lang/josh-sync.


Kobzol:
The formatting error was a result of a rebase, I think. But maybe formatting diverges between stdarch and rustc, in which case we'd need to later sync them up.

folkertdev:
Well, https://github.com/rust-lang/stdarch/blob/master/rustfmt.toml is an empty file that was last updated 4 years ago, so yeah the formatting is probably different.

Kobzol:
Right, it was just a result of a rebase though, I forgot that we don't format submodules and subtress for this very reason in rust-lang/rust (https://github.com/rust-lang/rust/blob/a27f3e3fd1e4d16160f8885b6b06665b5319f56c/rustfmt.toml#L34). So it should be fine.

folkertdev (approved):

PR #1884, opened 2025-07-24
`intrinsic-test`: Final code cleanup for the `arm` and `common` module

madhav-madhusoodanan:
## Summary
1. Changed from `IntrinsicType::target` (String) to `IntrinsicType::metadata` (HashMap<String, String>) for better support for differing architectures
2. Added `Constraint::Set(Vec<i64>)` for support for distinct constant argument values (which may be of types similar to the IMM type)

## Context
This PR is part 3 of the changes that were originally made in the x86 extension PR rust-lang/stdarch#1814 for intrinsic-test, and is intended to accomodate the changes made in PRs  rust-lang/stdarch#1862 and rust-lang/stdarch#1863.

cc: @Amanieu @folkertdev

### DIFF
--- a/crates/intrinsic-test/src/arm/argument.rs
+++ b/crates/intrinsic-test/src/arm/argument.rs
@@ -1,0 +1,12 @@
+use crate::arm::intrinsic::ArmIntrinsicType;
+use crate::common::argument::Argument;
+
+impl Argument<ArmIntrinsicType> {
+    pub fn type_and_name_from_c(arg: &str) -> (&str, &str) {
+        let split_index = arg
+            .rfind([' ', '*'])
+            .expect("Couldn't split type and argname");
+
+        (arg[..split_index + 1].trim_end(), &arg[split_index + 1..])
+    }
+}
--- a/crates/intrinsic-test/src/arm/json_parser.rs
+++ b/crates/intrinsic-test/src/arm/json_parser.rs
@@ -86,13 +86,16 @@ fn json_to_intrinsic(
         .into_iter()
         .enumerate()
         .map(|(i, arg)| {
-            let arg_name = Argument::<ArmIntrinsicType>::type_and_name_from_c(&arg).1;
+            let (type_name, arg_name) = Argument::<ArmIntrinsicType>::type_and_name_from_c(&arg);
             let metadata = intr.args_prep.as_mut();
             let metadata = metadata.and_then(|a| a.remove(arg_name));
             let arg_prep: Option<ArgPrep> = metadata.and_then(|a| a.try_into().ok());
             let constraint: Option<Constraint> = arg_prep.and_then(|a| a.try_into().ok());
+            let ty = ArmIntrinsicType::from_c(type_name, target)
+                .unwrap_or_else(|_| panic!("Failed to parse argument '{arg}'"));
 
-            let mut arg = Argument::<ArmIntrinsicType>::from_c(i, &arg, target, constraint);
+            let mut arg =
+                Argument::<ArmIntrinsicType>::new(i, String::from(arg_name), ty, constraint);
 
             // The JSON doesn't list immediates as const
             let IntrinsicType {
--- a/crates/intrinsic-test/src/arm/json_parser.rs
+++ b/crates/intrinsic-test/src/arm/json_parser.rs
@@ -79,7 +79,8 @@ fn json_to_intrinsic(
 ) -> Result<Intrinsic<ArmIntrinsicType>, Box<dyn std::error::Error>> {
     let name = intr.name.replace(['[', ']'], "");
 
-    let results = ArmIntrinsicType::from_c(&intr.return_type.value, target)?;
+    let mut results = ArmIntrinsicType::from_c(&intr.return_type.value)?;
+    results.set_metadata("target", target);
 
     let args = intr
         .arguments
@@ -91,8 +92,9 @@ fn json_to_intrinsic(
             let metadata = metadata.and_then(|a| a.remove(arg_name));
             let arg_prep: Option<ArgPrep> = metadata.and_then(|a| a.try_into().ok());
             let constraint: Option<Constraint> = arg_prep.and_then(|a| a.try_into().ok());
-            let ty = ArmIntrinsicType::from_c(type_name, target)
+            let mut ty = ArmIntrinsicType::from_c(type_name)
                 .unwrap_or_else(|_| panic!("Failed to parse argument '{arg}'"));
+            ty.set_metadata("target", target);
 
             let mut arg =
                 Argument::<ArmIntrinsicType>::new(i, String::from(arg_name), ty, constraint);
--- a/crates/intrinsic-test/src/arm/mod.rs
+++ b/crates/intrinsic-test/src/arm/mod.rs
@@ -69,7 +69,7 @@ impl SupportedArchitectureTest for ArmArchitectureTest {
 
         let (chunk_size, chunk_count) = chunk_info(self.intrinsics.len());
 
-        let cpp_compiler = compile::build_cpp_compilation(&self.cli_options).unwrap();
+        let cpp_compiler_wrapped = compile::build_cpp_compilation(&self.cli_options);
 
         let notice = &build_notices("// ");
         self.intrinsics
@@ -81,9 +81,11 @@ impl SupportedArchitectureTest for ArmArchitectureTest {
                 write_mod_cpp(&mut file, notice, c_target, platform_headers, chunk).unwrap();
 
                 // compile this cpp file into a .o file
-                let output = cpp_compiler
-                    .compile_object_file(&format!("mod_{i}.cpp"), &format!("mod_{i}.o"))?;
-                assert!(output.status.success(), "{output:?}");
+                if let Some(cpp_compiler) = cpp_compiler_wrapped.as_ref() {  
+                    let output = cpp_compiler
+                        .compile_object_file(&format!("mod_{i}.cpp"), &format!("mod_{i}.o"))?;
+                    assert!(output.status.success(), "{output:?}");
+                }
 
                 Ok(())
             })
@@ -99,21 +101,25 @@ impl SupportedArchitectureTest for ArmArchitectureTest {
         )
         .unwrap();
 
+        // This is done because `cpp_compiler_wrapped` is None when
+        // the --generate-only flag is passed
+        if let Some(cpp_compiler) = cpp_compiler_wrapped.as_ref() {
         // compile this cpp file into a .o file
-        info!("compiling main.cpp");
-        let output = cpp_compiler
-            .compile_object_file("main.cpp", "intrinsic-test-programs.o")
-            .unwrap();
-        assert!(output.status.success(), "{output:?}");
-
-        let object_files = (0..chunk_count)
-            .map(|i| format!("mod_{i}.o"))
-            .chain(["intrinsic-test-programs.o".to_owned()]);
-
-        let output = cpp_compiler
-            .link_executable(object_files, "intrinsic-test-programs")
-            .unwrap();
-        assert!(output.status.success(), "{output:?}");
+            info!("compiling main.cpp");
+            let output = cpp_compiler
+                .compile_object_file("main.cpp", "intrinsic-test-programs.o")
+                .unwrap();
+            assert!(output.status.success(), "{output:?}");
+    
+            let object_files = (0..chunk_count)
+                .map(|i| format!("mod_{i}.o"))
+                .chain(["intrinsic-test-programs.o".to_owned()]);
+    
+            let output = cpp_compiler
+                .link_executable(object_files, "intrinsic-test-programs")
+                .unwrap();
+            assert!(output.status.success(), "{output:?}");
+        }
 
         true
     }
--- a/crates/intrinsic-test/src/arm/mod.rs
+++ b/crates/intrinsic-test/src/arm/mod.rs
@@ -1,10 +1,11 @@
+mod argument;
 mod compile;
 mod config;
 mod intrinsic;
 mod json_parser;
 mod types;
 
-use std::fs::File;
+use std::fs::{self, File};
 
 use rayon::prelude::*;
 
@@ -72,6 +73,7 @@ impl SupportedArchitectureTest for ArmArchitectureTest {
         let cpp_compiler_wrapped = compile::build_cpp_compilation(&self.cli_options);
 
         let notice = &build_notices("// ");
+        fs::create_dir_all("c_programs").unwrap();
         self.intrinsics
             .par_chunks(chunk_size)
             .enumerate()
@@ -81,7 +83,7 @@ impl SupportedArchitectureTest for ArmArchitectureTest {
                 write_mod_cpp(&mut file, notice, c_target, platform_headers, chunk).unwrap();
 
                 // compile this cpp file into a .o file
-                if let Some(cpp_compiler) = cpp_compiler_wrapped.as_ref() {  
+                if let Some(cpp_compiler) = cpp_compiler_wrapped.as_ref() {
                     let output = cpp_compiler
                         .compile_object_file(&format!("mod_{i}.cpp"), &format!("mod_{i}.o"))?;
                     assert!(output.status.success(), "{output:?}");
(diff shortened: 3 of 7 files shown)


madhav-madhusoodanan:
I'm coming across a weird compilation issue on the CI: 

```bash
clang++: warning: argument unused during compilation: '-fuse-ld=lld' [-Wunused-command-line-argument]
mod_3.cpp:138:43: error: use of undeclared identifier 'vld1q_f16'
```

crates/core_arch/src/x86/sse2.rs
```diff
@@ -1281,15 +1281,6 @@ pub unsafe fn _mm_loadu_si128(mem_addr: *const __m128i) -> __m128i {
...
 /// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_maskmoveu_si128)
-///
-/// # Safety of non-temporal stores
-///
-/// After using this intrinsic, but before any other access to the memory that this intrinsic
-/// mutates, a call to [`_mm_sfence`] must be performed by the thread that used the intrinsic. In
-/// particular, functions that call this intrinsic should generally call `_mm_sfence` before they
-/// return.
-///
-/// See [`_mm_sfence`] for details.
```
folkertdev: this was merged today. Something must have gone wrong with git on your end.

crates/intrinsic-test/src/arm/argument.rs
```diff
@@ -0,0 +1,12 @@
...
+
+impl Argument<ArmIntrinsicType> {
+    pub fn type_and_name_from_c(arg: &str) -> (&str, &str) {
+        let split_index = arg
+            .rfind([' ', '*'])
+            .expect("Couldn't split type and argname");
+
+        (arg[..split_index + 1].trim_end(), &arg[split_index + 1..])
+    }
+}
```
folkertdev, line 12: does this really need its own file?

crates/intrinsic-test/src/arm/compile.rs
```diff
@@ -12,7 +12,7 @@ pub fn build_cpp_compilation(config: &ProcessedCli) -> Option<CppCompilation> {
         .set_opt_level("2")
         .set_cxx_toolchain_dir(config.cxx_toolchain_dir.as_deref())
         .set_project_root("c_programs")
-        .add_extra_flags(vec!["-ffp-contract=off", "-Wno-narrowing"]);
+        .add_extra_flags(vec!["-ffp-contract=off", "-Wno-narrowing", "-mfpu=neon"]);
```
folkertdev: why this change?

crates/intrinsic-test/src/arm/mod.rs
```diff
@@ -68,28 +69,31 @@ impl SupportedArchitectureTest for ArmArchitectureTest {
 
         let (chunk_size, chunk_count) = chunk_info(self.intrinsics.len());
 
-        let cpp_compiler = compile::build_cpp_compilation(&self.cli_options).unwrap();
+        let cpp_compiler_wrapped = compile::build_cpp_compilation(&self.cli_options);
```
folkertdev, line 73: we really should just panic if there is no compiler configured. What are we supposed to do in that scenario? We should not fail silently.

crates/intrinsic-test/src/arm/mod.rs
```diff
@@ -68,28 +69,31 @@ impl SupportedArchitectureTest for ArmArchitectureTest {
...
 
         let notice = &build_notices("// ");
+        fs::create_dir_all("c_programs").unwrap();
         self.intrinsics
             .par_chunks(chunk_size)
             .enumerate()
             .map(|(i, chunk)| {
                 let c_filename = format!("c_programs/mod_{i}.cpp");
-                let mut file = File::create(&c_filename).unwrap();
+                let mut file = fs::File::create(&c_filename).unwrap();
```
folkertdev: Let's just keep this as `File`, you can use `use std::fs::{self, File}` to import file and the `fs` module.

crates/intrinsic-test/src/arm/json_parser.rs
```diff
@@ -79,20 +79,25 @@ fn json_to_intrinsic(
...
         .into_iter()
         .enumerate()
         .map(|(i, arg)| {
-            let arg_name = Argument::<ArmIntrinsicType>::type_and_name_from_c(&arg).1;
-            let metadata = intr.args_prep.as_mut();
-            let metadata = metadata.and_then(|a| a.remove(arg_name));
-            let arg_prep: Option<ArgPrep> = metadata.and_then(|a| a.try_into().ok());
+            let (type_name, arg_name) = Argument::<ArmIntrinsicType>::type_and_name_from_c(&arg);
+            let ty = ArmIntrinsicType::from_c(type_name)
+                .unwrap_or_else(|_| panic!("Failed to parse argument '{arg}'"));
```
folkertdev: so, this line fails a bunch in CI, so the logic here is incorrect for this target.

crates/intrinsic-test/src/common/constraint.rs
```diff
@@ -5,13 +5,15 @@ use std::ops::Range;
...
 
 impl Constraint {
-    pub fn to_range(&self) -> Range<i64> {
+    pub fn to_vector(&self) -> Vec<i64> {
         match self {
-            Constraint::Equal(eq) => *eq..*eq + 1,
-            Constraint::Range(range) => range.clone(),
+            Constraint::Equal(eq) => vec![*eq],
+            Constraint::Range(range) => range.clone().collect::<Vec<i64>>(),
+            Constraint::Set(values) => values.clone(),
```
folkertdev: We can use this instead to avoid clones/allocations:

```rust
impl Constraint {
    /// Iterate over the values of this constraint.
    fn iter<'a>(&'a self) -> impl Iterator<Item = i64> + 'a {
        match self {
            Constraint::Equal(i) => std::slice::Iter::default().copied().chain(*i..*i + 1),
            Constraint::Range(range) => std::slice::Iter::default().copied().chain(range.clone()),
            Constraint::Set(items) => items.iter().copied().chain(std::ops::Range::default()),
        }
    }
}
```

crates/intrinsic-test/src/common/intrinsic_helpers.rs
```diff
@@ -121,7 +122,7 @@ pub struct IntrinsicType {
     /// A value of `None` can be assumed to be 1 though.
     pub vec_len: Option<u32>,
 
-    pub target: String,
+    pub metadata: HashMap<String, String>,
```
folkertdev, line 125: why is this a hashmap? Couldn't this be a struct instead? what values do you actually need to store in here for x86?

crates/intrinsic-test/src/common/constraint.rs
```diff
@@ -5,13 +5,15 @@ use std::ops::Range;
 pub enum Constraint {
     Equal(i64),
     Range(Range<i64>),
+    Set(Vec<i64>),
```
folkertdev: Can you add an example here of a set that is not a range (but that will actually come up)?

crates/intrinsic-test/src/arm/mod.rs
folkertdev: it could be 

```rust
let Some(cpp_compiler) = compile::build_cpp_compilation(&self.cli_options) else { 
    panic!("no C compiler configured");
}
```

I guess.

crates/intrinsic-test/src/common/constraint.rs
folkertdev: I'm also not super happy with the name "constraint" (probably made sense at the time, but now?) but we can resolve that later.

folkertdev (commented):
This'll need some work, and may benefit from splitting it up even further.

I'm not convinced about metadata being a hashmap. What do you actually need to store in there? (I'm really trying to avoid generalizing this codebase before there is a concrete need).

---

re 

```
clang++: warning: argument unused during compilation: '-fuse-ld=lld' [-Wunused-command-line-argument]
mod_3.cpp:138:43: error: use of undeclared identifier 'vld1q_f16'
```

the first part you can ignore, the second part is due to an `expect` that this PR introduces, and it is emitted so often that CI gives up.

crates/intrinsic-test/src/arm/compile.rs
madhav-madhusoodanan: I was debugging the cause of the issue `error: use of undeclared identifier 'vld1q_f16'` on the CI tool.
This specific change was only experimental.

crates/intrinsic-test/src/arm/mod.rs
madhav-madhusoodanan: There is a CLI option to only generate the C and Rust test files (without building or linking).
This is useful if I only need to get the test files without needing to setup the compilation or linking toolchains.

The relevant flag is `--generate-only`.

crates/intrinsic-test/src/common/constraint.rs
madhav-madhusoodanan: Example X86 intrinsic: https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_i32gather_pd&ig_expand=3720

The function definition is: 
```C
__m128d _mm_i32gather_pd (double const* base_addr, __m128i vindex, const int scale)
```
and `scale` can only take the values 1, 2, 4, and 8.

madhav-madhusoodanan: This is an x86 related change though, would it be recommended to keep it in the PR that will add x86 to the behavioural testing test suite?

crates/intrinsic-test/src/common/constraint.rs
madhav-madhusoodanan: You are right, this actually is present to support `IMM` type values, since Rust uses const generics to pass the value.

Example: 
```Rust
pub fn _mm256_blend_ps<const IMM8: i32>(a: __m256, b: __m256) -> __m256
```

crates/intrinsic-test/src/arm/mod.rs
madhav-madhusoodanan: Sure. 
I made this change, since I thought importing `{self, File}` would look awkward.

crates/core_arch/src/x86/sse2.rs
madhav-madhusoodanan: Let me look into this.

crates/intrinsic-test/src/common/constraint.rs
folkertdev: This is fine, can you change this type to something like:

```rust
/// Describes the values to test for a const generic parameter.
#[derive(Debug, PartialEq, Clone, Deserialize)]
pub enum Constraint {
    /// Test a single value.
    Equal(i64),
    /// Test a range of values, e.g. `0..16`.
    Range(Range<i64>),
    /// Test some specific values, e.g. `vec![1, 2, 4, 8]`.
    Set(Vec<i64>),
}
```

We might make `Equal` just a singleton set at some point too, maybe.

crates/intrinsic-test/src/arm/mod.rs
folkertdev: it's a bit weird at first but we do this all the time e.g. in the compiler.

crates/intrinsic-test/src/arm/mod.rs
folkertdev: ah, ok, can you add a comment mentioning that flag? 

(I think we should clean that up with a custom `enum Mode { Compile(PathBuf), GenerateOnly }` at some point)

madhav-madhusoodanan:
I was debugging this issue: `mod_3.cpp:138:43: error: use of undeclared identifier 'vld1q_f16'`.

The command that was being run was:
`clang++ -march=armv8.6-a+crypto+crc+dotprod+fp16 -O2 -ffp-contract=off -Wno-narrowing -mfpu=neon --target=armv7-unknown-linux-gnueabihf mod_3.cpp -c -o mod_3.o`.

I suspected that `armv8.6-a` and `--target=armv7-unknown-linux-gnueabihf` might be clashing (armv8-a vs armv7), so i removed the `-march` tag itself. 

This led me to the following error (showing a sample, there were more intrinsics): 
```bash
mod_3.cpp:861:43: error: use of undeclared identifier 'vld1_f16'
        float16x4_t a = cast<float16x4_t>(vld1_f16(&a_vals[i]));
                                          ^
mod_3.cpp:862:43: error: use of undeclared identifier 'vld1_f16'
        float16x4_t b = cast<float16x4_t>(vld1_f16(&b_vals[i]));
```
The intrinsic `vld1_f16` is present on v7, A64 and A32 architectures, so not like we can add it (and the others) to the exclusion list:

```JSON
  {
    "SIMD_ISA": "Neon",
    "name": "vld1_f16",
    "arguments": [
      "float16_t const * ptr"
    ],
    "return_type": {
      "value": "float16x4_t"
    },
    "Arguments_Preparation": {
      "ptr": {
        "register": "Xn"
      }
    },
    "Architectures": [
      "v7",
      "A32",
      "A64"
    ],
    "instructions": [
      [
        "LD1"
      ]
    ]
  },
```

folkertdev:
Can you isolate the issue into a godbolt.org snippet? I vaguely remember there is some `full-fp16` flag though I can't find much about it now.

madhav-madhusoodanan:
@folkertdev Here is a minimal reproduction: https://godbolt.org/z/e1r3x9xfo

folkertdev:
using `-fpu=neon-fp16` helps a bit. 

also I don't really see why these changes cause this issue? was this code not generated before?

folkertdev (commented):
I think at this point it's probably best to make separate PRs for

- the changes to `Constraint` (incorporating my comments)
- the changes to `crates/intrinsic-test/src/common/argument.rs` and related

I think that is the core here, any by having just those changes there should be fewer weird CI problems etc.

madhav-madhusoodanan:
Yeah you're right.
I'm looking at the minimal set of changes that could waive the undeclared identifier issue away.

crates/intrinsic-test/src/common/intrinsic_helpers.rs
madhav-madhusoodanan: X86 has a “TYPE” and “ETYPE” values for each argument of an intrinsic.

The “TYPE” tag is the type of the argument/return value as it appears in the C definition, while “ETYPE” gives information about how the argument will be used (eg: when using vector arguments, whether the operation would be U64 or FP16 based)

Both the information would be required, but it doesn’t make sense to create explicit struct members for it.

madhav-madhusoodanan:
Finally fixed that weird CI issue.
Had to commit from scratch

crates/intrinsic-test/src/common/intrinsic_helpers.rs
folkertdev: oh, I didn't see what type this was in. Hmm, shouldn't the target-specific intrinsic type hold such information?

I'll say I am a bit surprised by the target being a part of `IntrinsicType`, that doesn't seem right.

folkertdev (commented):
Can you make separate PRs for those two changes I mentioned (and maybe also the `--generate-only` changes) first? Assuming the commits are atomic that should be straightforward with some `giet cherry-pick` ing

madhav-madhusoodanan:
Sure, let me do it right away.

Ruanyx1823:
Hello, I'm a student interested in the rust compiler. I'd like to ask how exactly one can write an sve intrinsic. Or how to write the following C language in rust: // #include <arm_sve.h>

// void vec_add(const float *a,const float *b,float *c,int n){
//     for (int i=0; i<n; i+=svcntw()){
//         svbool_t pg = svwhilelt_b32(i,n);
//         svfloat32_t va = svld1(pg,&a[i]);
//         svfloat32_t vb = svld1(pg,&b[i]);
//         svfloat32_t vc = svadd_f32_m(pg,va,vb);
//         svst1(pg,&c[i],vc);
//}
//}

#include <stdio.h>
#include <stdlib.h>
#include <arm_sve.h>

void vec_add(const float* a, const float* b, float* c, int n) {
for (int i = 0;  i < n;  i += svcntw()) {
svbool_t pg = svwhilelt_b32(i, n);
svfloat32_t va = svld1(pg, &a[i]);
svfloat32_t vb = svld1(pg, &b[i]);
svfloat32_t vc = svadd_f32_m(pg, va, vb);
svst1(pg, &c[i], vc);
}
}

void print_array(const char* name, const float* arr, int n) {
printf("%s: [", name);
for (int i = 0;  i < n;  ++i) {
printf("%.1f", arr[i]);
if (i < n - 1) printf(", ");
}
printf("]\n");
}

int main() {
#ifdef __ARM_FEATURE_SVE
printf("SVE is supported! \n");
printf("SVE vector length: %d floats (%d bits)\n",
svcntw(), svcntw() * 32);
#else
printf("SVE not supported.\n");
return 1;
#endif

const int n = 16;
float a[n], b[n], c[n];

for (int i = 0;  i < n;  ++i) {
a[i] = (float)i;
b[i] = (float)(2 * i);
}

printf("\nBefore vector addition:\n");
print_array("A", a, n);
print_array("B", b, n);

vec_add(a, b, c, n);

printf("\nAfter vector addition:\n");
print_array("C = A+B", c, n);

int errors = 0;
for (int i = 0;  i < n;  ++i) {
if (c[i] ! = a[i] + b[i]) {
printf("Error at index %d: expected %.1f, got %.1f\n",
i, a[i] + b[i], c[i]);
errors++;
}
}

if (errors == 0) {
printf("\nAll results correct! \n");
}
else {
printf("\nFound %d errors! \n", errors);
}

return 0;
}

tgross35:
@Ruanyx1823 if your repo is relevant to developing Rust, please bring it to https://rust-lang.zulipchat.com/ rather than posting on a random issue. If it's a question about your own code, ask at https://users.rust-lang.org/ or https://discord.com/invite/rust-lang-community.

adamgemmell:
Has this patch been superceded?

adamgemmell:
> Has this patch been superceded?

Same question, I assume it effectively has anyway by this point ~~but wanted to ask before closing.~~ Let me know if not!

PR #1888, opened 2025-07-27
`intrinsic-test`: bringing back support for  --generate-only flag

madhav-madhusoodanan:
## Summary
1. Added support for the case when the `--generate-only` flag is passed. This helps with generating the Rust and C artifacts even if the compilation environment is not supported

## Context
This PR is part of the changes that were originally made in the x86 extension PR rust-lang/stdarch#1814 for intrinsic-test, and is intended to accomodate the changes made in PRs  rust-lang/stdarch#1862 and rust-lang/stdarch#1863.

cc: @folkertdev @Amanieu

### DIFF
--- a/crates/intrinsic-test/src/arm/mod.rs
+++ b/crates/intrinsic-test/src/arm/mod.rs
@@ -70,7 +70,7 @@ impl SupportedArchitectureTest for ArmArchitectureTest {
 
         let (chunk_size, chunk_count) = chunk_info(self.intrinsics.len());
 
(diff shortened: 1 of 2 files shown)


folkertdev:
r? folkertdev

folkertdev (approved):
Nice, I think we can do better (using some custom types), I'll look into that, but this at least unbreaks the workflow.

madhav-madhusoodanan:
Yes, that was my intention too.
A common interface for the `common` module so that we can both work in parallel, unblocked with the changes in each other's workflows.

PR #1890, opened 2025-07-29
Use an ugly name to help catch field projection

scottmcm:
Something's happening, based on https://github.com/rust-lang/rust/issues/144621, so go further than before by removing the visibility from the field and having definitions use a `from_array` method instead where needed.

This is draft because it's absolutely not going to work first try, but I'd like to borrow CI for fixing.

### DIFF
--- a/crates/core_arch/src/aarch64/neon/mod.rs
+++ b/crates/core_arch/src/aarch64/neon/mod.rs
@@ -425,7 +425,7 @@ pub fn vmovq_n_f64(value: f64) -> float64x2_t {
 #[cfg_attr(test, assert_instr(nop))]
 #[stable(feature = "neon_intrinsics", since = "1.59.0")]
 pub fn vget_high_f64(a: float64x2_t) -> float64x1_t {
-    unsafe { float64x1_t([simd_extract!(a, 1)]) }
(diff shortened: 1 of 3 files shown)


scottmcm:
Oh, the arm generator is completely different from when I was last here.  I have no idea how to add the `::from_array` calls...

folkertdev:
I fixed those here https://github.com/rust-lang/stdarch/commit/e3ba06b17f9b6dd36ec6b4e6812bb295ab345eda, you can cherry-pick + squash that if you think that looks good.

scottmcm:
Oh, no wonder I couldn't find it in the `Expression` stuff 🤦   Thanks for putting that together!

That said, it looks like the problem wasn't in here after all -- it was just a different thing that failed for all SIMD.  So I'm going to close this and not do the churn of updating everything, unless people feel that this would be better (in avoiding accidentally adding projections in future) enough to actually go through with doing it everywhere.  Thoughts welcome.

scottmcm:
Actually, wait, I can do this slow-stepped: add `from_array`, start using it in places, but not insist that *everything* move over all at once.  That's probably worth doing, at least.

PR #1892, opened 2025-08-02
`Intrinsic-test`: Updated the Constraint enum to support discrete values

madhav-madhusoodanan:
## Context
This PR is part of the changes that were originally made in the x86 extension PR rust-lang/stdarch#1814 for `intrinsic-test`, and is intended to unblock efforts to support other architectures too.

cc: @folkertdev @Amanieu

### DIFF
--- a/crates/intrinsic-test/src/common/constraint.rs
+++ b/crates/intrinsic-test/src/common/constraint.rs
@@ -1,17 +1,24 @@
 use serde::Deserialize;
 use std::ops::Range;
 
+/// Describes the values to test for a const generic parameter.
 #[derive(Debug, PartialEq, Clone, Deserialize)]
 pub enum Constraint {
+    /// Test a single value.
     Equal(i64),
+    /// Test a range of values, e.g. `0..16`.
     Range(Range<i64>),
+    /// Test discrete values, e.g. `vec![1, 2, 4, 8]`.
+    Set(Vec<i64>),
 }
 
 impl Constraint {
-    pub fn to_range(&self) -> Range<i64> {
+    /// Iterate over the values of this constraint.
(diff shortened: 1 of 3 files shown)


madhav-madhusoodanan:
@folkertdev I remember our [discussion](https://github.com/rust-lang/stdarch/pull/1884#discussion_r2229703995) on whether there might be a better name than "Constraint".

Essentially, this feature is present for arguments that have a restricted set of legal values.

Would a word like "Restriction" or so be more appropriate?

folkertdev:
I was thinking something more like `ConstArgument`? That is slightly rust-specific, but really these are just arguments, they just need to be known at compile-time.

Weirdly CI fails in `loongarch`, that's strange.

madhav-madhusoodanan:
> I was thinking something more like ConstArgument? That is slightly rust-specific, but really these are just arguments, they just need to be known at compile-time.

In this context though, would it be appropriate?
```Rust
/// An argument for the intrinsic.
#[derive(Debug, PartialEq, Clone)]
pub struct Argument<T: IntrinsicTypeDefinition> {
    /// The argument's index in the intrinsic function call.
    pub pos: usize,
    /// The argument name.
    pub name: String,
    /// The type of the argument.
    pub ty: T,
    /// Any constraints that are on this argument
    pub constraint: Option<Constraint>,
}
```

> Weirdly CI fails in loongarch, that's strange.

Let me try making another commit, perhaps this might be a one-off occurrence with the CI

folkertdev:
https://github.com/rust-lang/stdarch/pull/1893 fixes the loongarch issue

folkertdev:
> In this context though, would it be appropriate?

Hmm, let's just leave it as `Constraint` for now then and I'll think about it some more.

PR #1894, opened 2025-08-03
`intrinsic-test`: Adding x86 behavioural testing.

madhav-madhusoodanan:
## Context
This is a redo of PR https://github.com/rust-lang/stdarch/pull/1814, since a lot of details have changed with PRs https://github.com/rust-lang/stdarch/pull/1863, https://github.com/rust-lang/stdarch/pull/1862, https://github.com/rust-lang/stdarch/pull/1861, https://github.com/rust-lang/stdarch/pull/1852.

r? @folkertdev 
cc: @Amanieu


crates/intrinsic-test/src/x86/intrinsic.rs
```diff
@@ -0,0 +1,43 @@
+use crate::common::argument::ArgumentList;
+use crate::common::indentation::Indentation;
+use crate::common::intrinsic::{Intrinsic, IntrinsicDefinition};
+use crate::common::intrinsic_helpers::IntrinsicType;
+use std::ops::{Deref, DerefMut};
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct X86IntrinsicType(pub IntrinsicType);
```
madhav-madhusoodanan: @folkertdev as per our discussion in https://github.com/rust-lang/stdarch/pull/1884#discussion_r2229684661, would it make sense to add architecture-specific elements here?

crates/intrinsic-test/src/x86/mod.rs
```diff
@@ -0,0 +1,195 @@
...
+        intrinsics.sort_by(|a, b| a.name.cmp(&b.name));
+        Box::new(Self {
+            intrinsics: intrinsics,
+            cli_options: cli_options,
+        })
+    }
+
+    fn build_c_file(&self) -> bool {
+        let c_target = "aarch64";
+        let platform_headers = &["arm_neon.h", "arm_acle.h", "arm_fp16.h"];
```
folkertdev: is this just a direct copy right now? otherwise why does arm show up here?

crates/intrinsic-test/src/x86/types.rs
```diff
@@ -0,0 +1,249 @@
...
+                    Err(_) => {
+                        data.bit_len = match data.kind() {
+                            TypeKind::Char(_) => Some(8),
+                            TypeKind::BFloat => Some(16),
+                            TypeKind::Int(_) => Some(32),
+                            TypeKind::Float => Some(32),
+                            _ => None,
+                        };
+                    }
+                }
```
folkertdev, line 413: why are only some type kinds covered here? Maybe this could be a method on `TypeKind`?

crates/intrinsic-test/src/x86/compile.rs
```diff
@@ -0,0 +1,38 @@
...
+            "popcnt",
+            "adx",
+            "aes",
+        ])
+        .set_compiler(cpp_compiler)
+        .set_target(&config.target)
+        .set_opt_level("2")
+        .set_cxx_toolchain_dir(config.cxx_toolchain_dir.as_deref())
+        .set_project_root("c_programs")
+        .add_extra_flags(vec!["-ffp-contract=off", "-Wno-narrowing"]);
```
folkertdev: https://github.com/rust-lang/stdarch/pull/1896 just merged, so now the `vec!` calls here can be dropped (just pass an array)

crates/intrinsic-test/src/x86/mod.rs
madhav-madhusoodanan: Ohh thanks for the catch!
Yes I was following the structure of the ARM module, let me correct it.

ci/run.sh
```diff
@@ -181,6 +184,15 @@ case "${TARGET}" in
...
         ;;
+
+    x86_64-unknown-linux-gnu*)
+        CPPFLAGS="${TEST_CPPFLAGS}" RUSTFLAGS="${HOST_RUSTFLAGS}" RUST_LOG=warn \
+            cargo run "${INTRINSIC_TEST}" "${PROFILE}"  \
+            --bin intrinsic-test -- intrinsics_data/x86-intel.xml \
+            --runner "${TEST_RUNNER}" \
+            --cppcompiler "${TEST_CXX_COMPILER}" \
+            --target "${TARGET}"
+        ;;
```
folkertdev: we'll have to see how to do this exactly, but we want to split these out of the main CI job to speed it up

folkertdev (commented):
you should rebase on top of the upstream master branch instead of merging it in. That keeps the git history clean.

madhav-madhusoodanan:
Seems like the CI run at this point failed due to this error. I'll retry shortly:
```
#6 39.45   Could not connect to archive.ubuntu.com:80 (185.125.190.82), connection timed out Could not connect to archive.ubuntu.com:80 (185.125.190.83), connection timed out Could not connect to archive.ubuntu.com:80 (91.189.91.83), connection timed out Could not connect to archive.ubuntu.com:80 (185.125.190.81), connection timed out Could not connect to archive.ubuntu.com:80 (91.189.91.81), connection timed out Could not connect to archive.ubuntu.com:80 (185.125.190.36), connection timed out Could not connect to archive.ubuntu.com:80 (185.125.190.39), connection timed out Could not connect to archive.ubuntu.com:80 (91.189.91.82), connection timed out
#6 39.45   Unable to connect to archive.ubuntu.com:80:
#6 39.45 Fetched 126 kB in 39s (3208 B/s)
#6 39.45 Reading package lists...
#6 39.46 W: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/questing/InRelease  Unable to connect to archive.ubuntu.com:80:
#6 39.46 W: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/questing-updates/InRelease  Unable to connect to archive.ubuntu.com:80:
#6 39.46 W: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/questing-backports/InRelease  Unable to connect to archive.ubuntu.com:80:
#6 39.46 W: Some index files failed to download. They have been ignored, or old ones used instead.
```

folkertdev:
Can you rebase this (and remove the merge commits) sometime? That would make it a lot easier to see what has actually changed.

Also, why do you push to then have CI fail? Running this locally is a lot faster than having CI do it, because you can skip the earlier steps and run just the intrinsic tests.

crates/intrinsic-test/src/x86/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+                .collect::<Vec<_>>();
+
+            compare_outputs(
+                &intrinsics_name_list,
+                &self.cli_options.runner,
+                &self.cli_options.target,
+            )
+        } else {
+            true
+        }
```
folkertdev: this function is already defined by the trait, so it can be removed here.

folkertdev (commented):
oh, wait, it probably is up to date? it's just that xml file is _enormous_. that's crazy. Should we like, try to trim that down?

madhav-madhusoodanan:
@folkertdev I've been syncing using the below command whenever I'm updating from master:

```bash
git rebase master
```

Am I doing it correctly?

madhav-madhusoodanan:
About the XML file specification, may I ask what "trim down" means?

folkertdev:
depends a bit on your git setup, but it seems to work allright. I was just confused by the enormous number of lines changed, but that's all due to that XML file.

folkertdev:
What i mean is, could we process that XML file into a smaller XML file that only stores what we need? That would reduce the size of the repo (not sure how big that file is, maybe it compresses well?) and the speed of the intrinsic tests.

There are risks too, e.g. what we generate could go out of date with the official XML file. So maybe it's fine this way.

madhav-madhusoodanan:
Ahh, makes sense.

I think it might be best to keep the source of truth as unchanged as possible, since there is no direct way to obtain the sources sometimes.

For example, the XML file originally existed in the `stdarch-verify` crate, and I've had to check it by manually downloading the x86 reference (which comes in a folder containing HTML, CSS and JS files) at the link called [Download: Offline Intel® Intrinsics Guide](https://cdrdv2.intel.com/v1/dl/getContent/671338) in the Intel intrinsics reference [site](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=extract&cats=Swizzle)

The entire XML data can be found stored as a hardcoded string that is assigned to a variable in a JS file.

folkertdev:
Eww cursed. Allright, that's fine then

madhav-madhusoodanan:
My bad with the force-pushing here, I'll remember to run the check-style script locally before I push.

folkertdev:
This is looking good code-wise I think, now it just needs to work :)

```
Running `/intel-sde/sde64 -cpuid-in /checkout/ci/docker/x86_64-unknown-linux-gnu/cpuid.def -rtm-mode full -tsx -- target/release/intrinsic-test intrinsics_data/x86-intel.xml --runner '/intel-sde/sde64             -cpuid-in /checkout/ci/docker/x86_64-unknown-linux-gnu/cpuid.def             -rtm-mode full -tsx --' --cppcompiler clang++-19 --target x86_64-unknown-linux-gnu`
```

Looks like the command is malformed? Where does that inner `--runner` come from, that doesn't look right.

ci/run.sh
```diff
@@ -181,6 +184,15 @@ case "${TARGET}" in
             --linker "${CARGO_TARGET_AARCH64_BE_UNKNOWN_LINUX_GNU_LINKER}" \
             --cxx-toolchain-dir "${AARCH64_BE_TOOLCHAIN}"
         ;;
+
+    x86_64-unknown-linux-gnu*)
+        CPPFLAGS="${TEST_CPPFLAGS}" RUSTFLAGS="${HOST_RUSTFLAGS}" RUST_LOG=warn \
+            cargo run "${INTRINSIC_TEST}" "${PROFILE}"  \
+            --bin intrinsic-test -- intrinsics_data/x86-intel.xml \
+            --runner "${TEST_RUNNER}" \
```
madhav-madhusoodanan: @folkertdev It seems like the error comes from here, but I'm not sure how the outer runner command comes to be.

folkertdev: ah, right, that comes from 

https://github.com/rust-lang/stdarch/blob/818f0646de7232bc8b3592981aeccacae97422bd/ci/docker/x86_64-unknown-linux-gnu/Dockerfile#L14-L16

In other words, any `cargo run` is actually run with that runner. That is useful for the unit tests, but not here where we actually want to run the `intrinsic-test` binary on the host. On `aarch64` etc. we don't configure the runner.

It might work if you add `CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER=""` to this command?

madhav-madhusoodanan:
Hang on, it isn't supposed to work like this.
Let me check if there are errors that we are not catching.

ci/run.sh
madhav-madhusoodanan: Apparently `CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER` cannot have a zero size value, so I set it to `"true"` instead

crates/intrinsic-test/src/common/argument.rs
```diff
@@ -108,7 +108,7 @@ where
         for arg in self.iter().filter(|&arg| !arg.has_constraint()) {
             writeln!(
                 w,
-                "{indentation}const {ty} {name}_vals[] = {values};",
+                "{indentation}alignas(64) const {ty} {name}_vals[] = {values};",
```
folkertdev, line 118: interesting, I suppose we're casting values to simd vectors and then performing some sort of aligned read? In any case, can you leave a comment on why the alignment is required?

crates/intrinsic-test/src/common/gen_rust.rs
```diff
@@ -255,15 +255,15 @@ pub fn generate_rust_test_loop<T: IntrinsicTypeDefinition>(
 /// Generate the specializations (unique sequences of const-generic arguments) for this intrinsic.
 fn generate_rust_specializations<'a>(
     constraints: &mut impl Iterator<Item = impl Iterator<Item = i64>>,
-) -> Vec<Vec<u8>> {
+) -> Vec<Vec<i64>> {
```
folkertdev: these constraints are typically `i32` values in practice, right? is the full `i64` range needed?

crates/intrinsic-test/src/common/mod.rs
```diff
@@ -76,6 +77,14 @@ pub trait SupportedArchitectureTest {
...
                     let output = cpp_compiler
                         .compile_object_file(&format!("mod_{i}.cpp"), &format!("mod_{i}.o"))?;
+                    if !output.status.success() {
+                        io::stdout()
+                            .write_all(&output.stdout)
+                            .expect("Failed to write to stdout!");
+                        io::stderr()
+                            .write_all(&output.stderr)
+                            .expect("Failed to write to stderr!");
+                    }
```
folkertdev: the debug print that is part of the assert should already print the stdoud and stderr data. Is this to get better formatting?

crates/intrinsic-test/src/arm/mod.rs
```diff
@@ -31,7 +31,17 @@ impl SupportedArchitectureTest for ArmArchitectureTest {
...
+        "iostream",
+        "cstring",
+        "iomanip",
+        "sstream",
+        "cstddef",
+        "cstdint",
+        "arm_neon.h",
+        "arm_acle.h",
+        "arm_fp16.h",
+    ];
```
folkertdev: why are additional headers needed on arm?

ci/run.sh
folkertdev: that'll need a comment at least

crates/intrinsic-test/src/common/argument.rs
madhav-madhusoodanan: Really, I added this as a general precautionary measure.

Fortunately when I was working on x86, there were intrinsics that helped with unaligned reads (like `_mm_loadu_epi16` and the like), but I'm not sure if the same could be told for other architectures too.

madhav-madhusoodanan: But yes, I'll add a comment on the same.

crates/intrinsic-test/src/common/gen_rust.rs
madhav-madhusoodanan: Yes, you're right, I've only seen i32 values.
Fair suggestion, I'll change this.

crates/intrinsic-test/src/common/mod.rs
madhav-madhusoodanan: Oh yes, you're right.
I hadn't noticed the debug print of the assert, my bad.

crates/intrinsic-test/src/arm/mod.rs
madhav-madhusoodanan: There were a couple of headers that were part of [`common/gen_c.rs`](https://github.com/rust-lang/stdarch/pull/1894/files#diff-05c2e21e317958392f57fc77f950ec119e813e511fe5e1fd641950f804948a0aL109-L112) which I had brought here.

Oh, I think `cstddef` and `cstdint` won't be necessary in `arm` though.

crates/intrinsic-test/src/common/argument.rs
madhav-madhusoodanan: @folkertdev I wish to confer with you on this. I chose 64 so that it generalizes well across 16-bit, 32-bit and 64-bit alignment requirements.

Would 64 be a good value, or is it overkill?

folkertdev: I'd go with either 64 or remove it completely, any other inbetween value does not really make sense.

sayantn:
@madhav-madhusoodanan there is already a version of the `x86-intel.xml` file in `crates/stdarch-verify`, you can just move that file to `intrinsics-data` (and change the path in the `crates/stdarch-verify/tests/x86-intel.rs`)

crates/intrinsic-test/src/x86/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    fn create(cli_options: ProcessedCli) -> Self {
+        let intrinsics =
+            get_xml_intrinsics(&cli_options.filename).expect("Error parsing input file");
+
+        let mut intrinsics = intrinsics
+            .into_iter()
+            // Not sure how we would compare intrinsic that returns void.
+            .filter(|i| i.results.kind() != TypeKind::Void)
+            .filter(|i| i.results.kind() != TypeKind::BFloat)
```
sayantn, line 56: This can probably be dropped, x86 stdarch has the `bfloat16` newtype that takes the place of C `__bfloat16` (at least for now)

ci/run.sh
madhav-madhusoodanan: I read more on what "true" does.

Apparently, it simply exits successfully and does nothing else. I think this was the reason why the CI test for x86 passed, because it did not run in the first place.

Thank you so much for pointing out this one. I've changed it to be 'bash" instead.

folkertdev: still I think it's useful here to mention that you're resetting the runner, because we want to run this binary directly on the host

madhav-madhusoodanan: Yes, I've mentioned it in a comment right above this highlighted section.

By the way, even `bash` failed because it cannot execute binary files. I think `eval` might be a better command for this purpose.

madhav-madhusoodanan:
I'm consistently running into errors with ARM's generated C test files, like so:
```bash
mod_0.cpp:10319:51: error: assigning to 'const long *' from 'const poly64_t *' (aka 'const unsigned long *') converts between pointers to integer types with different sign
        poly64x2_t c = cast<poly64x2_t>(vld1q_s64(&c_vals[i]));
                                                  ^~~~~~~~~~
/usr/lib/llvm-14/lib/clang/14.0.6/include/arm_neon.h:8816:46: note: expanded from macro 'vld1q_s64'
  __ret = (int64x2_t) __builtin_neon_vld1q_v(__p0, 35); \
                                             ^~~~
```

Even after executing `intrinsic-test` at this [commit state](https://github.com/rust-lang/stdarch/pull/1894/commits/e32b078aeee92f88d188acd2c0696cdcf063ed00) when the ARM side passed successfully.

@folkertdev @sayantn has some sort of re-definition happened to `vld1q_s64` or something similar?

sayantn:
Seems like `-Wpointer-sign`, but it is an error for some reason, do we pass `-Werror` or `-Wpedantic` to clang++?

madhav-madhusoodanan:
No, we aren't passing either of those flags currently.

madhav-madhusoodanan:
Okay, seems like I wasn't passing `--target aarch64-unknown-linux-gnu` explicitly. No wonder.

madhav-madhusoodanan:
The error I'm running into now is:
```bash
thread 'main' (102234) panicked at src/common/mod.rs:91:14:
called `Result::unwrap()` on an `Err` value: "Error compiling mod_1.cpp: Os { code: 2, kind: NotFound, message: \"No such file or directory\" }"
```
The command that runs the `intrinsic-test` process is:
```bash
target/release/intrinsic-test intrinsics_data/x86-intel.xml --runner '/intel-sde/sde64             -cpuid-in /checkout/ci/docker/x86_64-unknown-linux-gnu/cpuid.def             -rtm-mode full -tsx --' --skip crates/intrinsic-test/missing_x86.txt --cppcompiler clang++-19 --target x86_64-unknown-linux-gnu
```

Running `ls -lhSa` on an x86 VM after the above command gives me this structure of `c_programs`:
```bash
total 235M
-rw-r--r-- 1 root root  48M Sep 28 08:42 mod_2.o
-rw-r--r-- 1 root root  39M Sep 28 08:31 mod_2.cpp
-rw-r--r-- 1 root root  37M Sep 28 08:31 mod_1.cpp
-rw-r--r-- 1 root root  35M Sep 28 08:40 mod_0.o
-rw-r--r-- 1 root root  28M Sep 28 08:31 mod_0.cpp
-rw-r--r-- 1 root root  26M Sep 28 08:39 mod_3.o
-rw-r--r-- 1 root root  23M Sep 28 08:31 mod_3.cpp
-rw-r--r-- 1 root root 1.2M Sep 28 08:42 intrinsic-test-programs.o
-rw-r--r-- 1 root root 789K Sep 28 08:42 main.cpp
drwxr-xr-x 5 root root 4.0K Sep 28 08:31 ..
drwxr-xr-x 2 root root  196 Sep 28 08:42 .
-rw-r--r-- 1 root root    0 Sep 28 08:31 mod_1-52d3beba.o.tmp
```

Seems like mod_1.cpp isn't being compiled.

However, further debugging gives me this command that was used to compile the same:
```bash
"clang++-19" CommandArgs { inner: ["-march=icelake-client", "-O2", "-ffp-contract=off", "-Wno-narrowing", "-mavx", "-mavx2", "-mavx512f", "-msse2", "-mavx512vl", "-mavx512bw", "-mavx512dq", "-mavx512cd", "-mavx512fp16", "-msha512", "-msm4", "-mavxvnni", "-mavx512bitalg", "-mavx512ifma", "-mavx512vbmi", "-mavx512vbmi2", "-mavx512vnni", "-mavx512vpopcntdq", "-ferror-limit=1000", "--target=x86_64-unknown-linux-gnu", "mod_1.cpp", "-c", "-o", "mod_1.o"] }
```

Simplifying this command gives:
```bash
clang++-19 -march=icelake-client -O2 -ffp-contract=off -Wno-narrowing -mavx -mavx2 -mavx512f -msse2 -mavx512vl -mavx512bw -mavx512dq -mavx512cd -mavx512fp16 -msha512 -msm4 -mavxvnni -mavx512bitalg -mavx512ifma -mavx512vbmi -mavx512vbmi2 -mavx512vnni -mavx512vpopcntdq -ferror-limit=1000 --target=x86_64-unknown-linux-gnu mod_1.cpp -c -o mod_1.o
```

Which runs without any issues and generates `mod_1.o`.

One other possibility where this command would fail to generate the object file is if it's been started by its parent process (which led to the creation of the `.tmp` file) but then had to be yielded before it could write to the same. I'll look into `rayon`'s internals to see if such a case may be possible.

@folkertdev Could there be other possibilities/details that I may be missing?

folkertdev:
Hmm not sure, rayon should not be able to move on until all of the modules have produced a .o file (that is what `par_chunks` and the subsequent `collect` do). But eh yeah if it's consistent then some additional logging might tell you more.

madhav-madhusoodanan:
@folkertdev Seems like the `clang++-19` was being passed as the value to the `--cppcompiler` flag, instead of `clang++`. That was the reason behind the `No such file or directory` error. Fixed it.

I ran into an issue with the process being killed (likely due to resource constraints since the `clang++` processes are running in parallel). Would it be okay if I increase the CPU and memory limits within the Github workflow file?

For now I've made it sequential so that I can step over that error, but the amount of time it's taking has got me reconsidering it.

folkertdev:
> I ran into an issue with the process being killed (likely due to resource constraints since the clang++ processes are running in parallel). Would it be okay if I increase the CPU and memory limits within the Github workflow file?

I don't think that is how it works? You just get a CI machine (typically with 4 cores) for some duration. Do you know how large these files get (how many intrinsics do you generate a test for)?

We definitely want to be running these processes in parallel.

madhav-madhusoodanan:
Ohh I see, I thought I could edit the resource allocation.
However, while running them in parallel I consistently run into the process being killed.

Let me experiment with keeping more number of `mod_i.cpp` and a worker pooling mechanism, seeing as we can't change the resource allocation.

folkertdev:
you could also try to just cut the total number of generated tests (e.g. to 10%) to see if that helps? 

By the looks of it the whole process is just _way_ too slow, and the runners eventually time out. So, you'll have to figure out how to make it faster. I think the crux here is what I plan to do: link all the c++ and rust code into one binary that generates, in effect, a large number of rust `#[test]` tests, and just execute that using the given runner.

But that is probably not feasible for you in the remaining time, so by just cutting the total number of tests (maybe using some random sampling, e.g. keep every 10th test) we'll be able to merge your code without completely tanking our CI times.

madhav-madhusoodanan:
Good idea @folkertdev. Let me first check whether processing smaller-sized chunks would be of any help.

madhav-madhusoodanan:
@sayantn was this change related to the change we made for some x86 intrinsics? Found this in the CI tests

```bash
Illegal instruction at address = 56435cedbf49: f2 0f 2b 00 0f ae f8 f2 0f 10 04 24 66 0f 2e 
Image name: /checkout/target/x86_64-unknown-linux-gnu/release/deps/core_arch-92d6a56b530b6fbf
Offset in image: 0x6cef49
If you believe your application should attempt to execute
this illegal instruction (and others that may be present),
Then use this knob: -emit-illegal-insts 0
and this error message will be avoided.
Use -print-exception-details to get more details.

SDE ERROR: Illegal instruction at address = 56435cedbf49: f2 0f 2b 00 0f ae f8 f2 0f 10 04 24 66 0f 2e 
Image name: /checkout/target/x86_64-unknown-linux-gnu/release/deps/core_arch-92d6a56b530b6fbf
Offset in image: 0x6cef49
If you believe your application should attempt to execute
this illegal instruction (and others that may be present),
Then use this knob: -emit-illegal-insts 0
and this error message will be avoided.
Use -print-exception-details to get more details.
```

sayantn:
I don't think I have seen this before, seems to be in `_mm_stream_sd`. I did recently update the implementation in #1929, but the CI did pass there (and locally). It shouldn't happen if we didn't change anything, might be spurious. Worth a retry in CI

Amanieu:
`_mm_stream_sd` and `_mm_stream_ss` seem to be in sse4a, not sse2.

sayantn:
Yea, and the CI failures looked like it got to sse4a and then failed (although I don't know how much we can trust output order, considering that the tests are parallelized)

Amanieu:
SSE4a is an AMD extension, so intel-sde may not support it. Does the intrinsic tester check for feature availability before using those features?

sayantn:
Iirc sde supports (most) amd extensions. It even has a cpu preset named `amd-future`. Currently we just enable the sse4a, vp2intersect, xop and tbm in the `future` cpu preset of sde

our testsuite uses `is_x86_feature_detected`, but it's mostly useless because we completely the cpuid outputs via the `-cpuid-in` argument to sde (the `ci/docker/x86_64-unknown-linux-gnu/cpuid.def` file). SDE parses that cpuid file to see what extensions it should emulate (it can also delegate to the host cpu if the host cpu has that instruction)

madhav-madhusoodanan:
@folkertdev I've added the ability to sample intrinsics. Currently it's implemented for x86 and about 5% of intrinsics are randomly chosen for testing.

This is intended as a temporary solution really, we can remove it once we restructure the C++/Rust testfiles to run all intrinsics in one go.

folkertdev:
Neat!

So, this does increase the longest CI job to 20 minutes, up from 12 minutes. I think we should try to make this its own CI job at least. I'm also not sure that we want to randomly sample the tests on each run. If we do hit a failure, that will be hard to reproduce. The status quo is no testing at all, so I think testing even a fixed subset is an improvement. 

> @madhav-madhusoodanan there is already a version of the `x86-intel.xml` file in `crates/stdarch-verify`, you can just move that file to `intrinsics-data` (and change the path in the `crates/stdarch-verify/tests/x86-intel.rs`)

did this happen? It still looks like the file is added again here.

madhav-madhusoodanan:
Ohh no, let me do that too.

madhav-madhusoodanan:
About the hit failure, hmm yes you have a point.
Let's do a naive partitioning in that case.

crates/intrinsic-test/src/common/compile_c.rs
```diff
@@ -119,7 +119,7 @@ impl CppCompilation {
         output: &str,
     ) -> std::io::Result<std::process::Output> {
         let mut cmd = clone_command(&self.0);
-        cmd.args([input, "-c", "-o", output]);
+        cmd.args([input, "-v", "-c", "-o", output]);
```
folkertdev, line 122: was this for debugging only? or is it generally useful?

crates/intrinsic-test/src/common/intrinsic_helpers.rs
```diff
@@ -129,9 +132,9 @@ impl IntrinsicType {
 
     pub fn inner_size(&self) -> u32 {
         if let Some(bl) = self.bit_len {
-            bl
+            cmp::max(bl, 8)
         } else {
-            unreachable!("")
+            unreachable!("{:#?}", self)
```
folkertdev: does `{self:#?}` not work?

crates/intrinsic-test/src/common/intrinsic_helpers.rs
```diff
@@ -283,6 +280,43 @@ impl IntrinsicType {
...
+                kind: TypeKind::Vector,
+                bit_len: Some(128 | 256 | 512),
+                simd_len,
+                vec_len,
+                ..
+            } => {
+                let (prefix, suffix) = match language {
+                    Language::Rust => ("[", "]"),
+                    Language::C => ("{", "}"),
+                };
```
folkertdev, line 293: these could be `char`s I think? So `('[', ']')` etc.

crates/intrinsic-test/src/common/compile_c.rs
madhav-madhusoodanan: I ran into `iostream not found` errors and wanted to know the include path of the header files while debugging. This would be helpful when adding new architectures by looking into the stdout and stderr of the compilation step.

crates/intrinsic-test/src/common/intrinsic_helpers.rs
madhav-madhusoodanan: It should work.

I must confess though, I wasn't aware that I could use this pattern until I noticed you use it elsewhere in the codebase 😅

madhav-madhusoodanan:
I'll need to rename the variables in `arm/config.rs` and `x86/config.rs`. I'll do that.

madhav-madhusoodanan:
Okay, so I've split the `intrinsic-test` run onto separate CI processes.
The `x86` run of the same is the longest among them and takes 10 mins.

Now the longest process among all the CI processes, is the `Test` CI process for `x86` which takes 12 mins.

ci/run.sh
```diff
@@ -92,6 +91,8 @@ fi
 case ${TARGET} in
     x86_64-unknown-linux-gnu)
         export STDARCH_DISABLE_ASSERT_INSTR=1
+        PATH="$PATH":"$(pwd)"/c_programs
+        export PATH
```
folkertdev: why is this needed? it really should not be I think?

madhav-madhusoodanan: Okay, so it turns out that for aarch64, `qemu`'s executable resolution algorithm looks into the current directory too.
However, `sde64` looks only at PATH.

This was intended to fix that discrepancy, but now that I think about it, it might be cleaner to use `./intrinsic-test-programs` in `compare_outputs.rs`.

crates/intrinsic-test/src/arm/config.rs
```diff
@@ -23,11 +34,26 @@ std::ostream& operator<<(std::ostream& os, poly128_t value) {
...
 #endif
+
+std::ostream& operator<<(std::ostream& os, float16_t value) {
+    uint16_t temp = 0;
+    memcpy(&temp, &value, sizeof(float16_t));
+    std::stringstream ss;
+    ss << "0x" << std::setfill('0') << std::setw(4) << std::hex << temp;
+    os << ss.str();
+    return os;
+}
```
madhav-madhusoodanan, line 47: @folkertdev do we need stringstream here? Seems to me that we can make this definition shorter.

crates/intrinsic-test/src/common/mod.rs
```diff
@@ -188,9 +196,13 @@ pub trait SupportedArchitectureTest {
     }
 }
 
-pub fn chunk_info(intrinsic_count: usize) -> (usize, usize) {
```
madhav-madhusoodanan, line 191: I'll need to remove this, since we aren't using this.
The CI processes on x86 broke bcoz of RAM resource constraints.

crates/intrinsic-test/src/x86/mod.rs
madhav-madhusoodanan: I'll remove it on a separate PR, I'm slightly afraid of breaking the current successful runs.

madhav-madhusoodanan:
Okay, so I've reverted that last one commit (that split the CI job) and the PATH update.
I'll make that PR as soon as we merge this one.

crates/intrinsic-test/src/arm/config.rs
folkertdev: My c++ is not good enough to know for sure. If there is something shorter that works I'm all for it

folkertdev:
Maybe this was not clear but my intention was to merge the ci changes separately first, them rebase this pr on top

madhav-madhusoodanan:
Ohh ohh, my bad.
Sure let me do that.

Created a PR for that: https://github.com/rust-lang/stdarch/pull/1941

madhav-madhusoodanan:
Turns out `std::hex` is sticky for the stream that it is applied on.
So stringstream's usage actually makes sense.

crates/intrinsic-test/src/common/mod.rs
```diff
@@ -96,7 +104,7 @@ pub trait SupportedArchitectureTest {
         // the --generate-only flag is passed
         if let Some(cpp_compiler) = cpp_compiler_wrapped.as_ref() {
             // compile this cpp file into a .o file
-            info!("compiling main.cpp");
+            trace!("compiling main.cpp");
             let output = cpp_compiler
```
adamgemmell, line 108: I'm having trouble running this locally without the intrinsic sampling on, I get an apparent segfault at this clang invocation. Were you seeing that too?

crates/intrinsic-test/src/common/argument.rs
```diff
@@ -106,11 +111,13 @@ where
         loads: u32,
     ) -> std::io::Result<()> {
         for arg in self.iter().filter(|&arg| !arg.has_constraint()) {
```
adamgemmell, line 113: I note that arguments of the same type seem to have identical `_vals` arrays. Not required for this but as a future optimisation we could just emit one array for each unique argument type

crates/intrinsic-test/src/x86/constraint.rs
```diff
@@ -0,0 +1,30 @@
+use crate::common::constraint::Constraint;
+
+pub fn map_constraints(imm_type: &String, imm_width: u32) -> Option<Constraint> {
+    if imm_width > 0 {
+        let max: i64 = 2i64.pow(imm_width);
```
adamgemmell, line 5: Some x86 intrinsics have a lot of possible immediate values, for example `_mm256_fpclass_pd_mask` has an immediate of width 8 giving 256*20 loop iterations just for that one intrinsic.

We never really saw so many constraint values on aarch64 - perhaps testing every possible constraint value isn't a reasonable strategy here

ci/run.sh
```diff
@@ -181,6 +186,22 @@ case "${TARGET}" in
...
+    x86_64-unknown-linux-gnu*)
+        # `CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER` is not necessary for `intrinsic-test`
+        # because the binary needs to run directly on the host.
+        # Hence the use of `env -u`.
+        env -u CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER \
+            CPPFLAGS="${TEST_CPPFLAGS}" RUSTFLAGS="${HOST_RUSTFLAGS}" \
+            RUST_LOG=warn RUST_BACKTRACE=1 \
+            cargo run "${INTRINSIC_TEST}" "${PROFILE}"  \
+            --bin intrinsic-test -- intrinsics_data/x86-intel.xml \
+            --runner "${TEST_RUNNER}" \
```
adamgemmell: I would suggest trying just this tool with `qemu-x86_64 -cpu=max` rather than intel SDE which I expect would run a lot quicker, potentially at the cost of less coverage

crates/intrinsic-test/src/common/argument.rs
folkertdev: Yeah that's something I want to work on (when this PR is merged)

ci/run.sh
folkertdev: `qemu` does not support (most of?) avx-512, which is where most of the complexity is. So we'll have to split that out.

folkertdev: which I'd rather not do in this PR

adamgemmell: Agreed, none of my comments should be blocking - this PR is useful without them

Maybe the situation is better today - these are the qemu cpu flags containing avx from 10.1 which is included in 25.10:

```
avx avx-ifma
avx-ne-convert avx-vnni avx-vnni-int16 avx-vnni-int8 avx10 avx10-128
avx10-256 avx10-512 avx2 avx512-4fmaps avx512-4vnniw avx512-bf16
avx512-fp16 avx512-vp2intersect avx512-vpopcntdq avx512bitalg avx512bw
avx512cd avx512dq avx512er avx512f avx512ifma avx512pf avx512vbmi
avx512vbmi2 avx512vl avx512vnni
```
  
  And here's what `lscpu` reports with `-cpu=max`
  ```
  Flags:                                   fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc art arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf tsc_known_freq pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 sdbg fma cx16 xtpr pdcm pcid dca sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm abm 3dnowprefetch cpuid_fault epb cat_l3 cat_l2 cdp_l3 intel_ppin cdp_l2 ssbd mba ibrs ibpb stibp ibrs_enhanced tpr_shadow flexpriority ept vpid ept_ad fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid cqm rdt_a avx512f avx512dq rdseed adx smap avx512ifma clflushopt clwb intel_pt avx512cd sha_ni avx512bw avx512vl xsaveopt xsavec xgetbv1 xsaves cqm_llc cqm_occup_llc cqm_mbm_total cqm_mbm_local split_lock_detect user_shstk avx_vnni avx512_bf16 wbnoinvd dtherm ida arat pln pts hwp hwp_act_window hwp_epp hwp_pkg_req vnmi avx512vbmi umip pku ospke waitpkg avx512_vbmi2 gfni vaes vpclmulqdq avx512_vnni avx512_bitalg avx512_vpopcntdq la57 rdpid bus_lock_detect cldemote movdiri movdir64b enqcmd fsrm md_clear serialize tsxldtrk pconfig arch_lbr ibt amx_bf16 avx512_fp16 amx_tile amx_int8 flush_l1d arch
  ```
  
QEMU is pretty quick to build to build a single backend from source if we wanted something newer than what's bundled with ubuntu too.

sayantn: Qemu-user supports avx512 with only the kvm backend, but as we run it in docker in CI, we don't have kvm. Idk about qemu-system though, it should be the same

crates/intrinsic-test/src/common/mod.rs
madhav-madhusoodanan: Can you show me the stdout/stderr? Perhaps I could help

madhav-madhusoodanan:
Just so that the review process is easier, would it make sense to break down the ARM and common changes into their own PRs?

A couple of them were dependent on the changes that arose to accommodate x86 in an attempt to maintain a clean separation of concerns.

folkertdev:
I'm very familiar with that arm code now, so I think we're good once the CI stuff is merged.

crates/intrinsic-test/src/x86/constraint.rs
madhav-madhusoodanan: I'd prefer keeping it comprehensible than risk edge-case issues not getting caught.

For example, we were able to find out issues with some corner cases with intrinsics such as `_kshift*` and `_mm_mask_cvtepi16_epi8 `.

For `_mm_mask_cvtepi16_epi8`, perhaps the current implementation in stdarch doesn't zero out the upper bits (will look into this and fix it in another PR)

[Pastebin of the diff of the intrinsic we obtained from a run of `intrinsic-test`](https://pastebin.com/y9PNENJb)

adamgemmell: That's fair enough, it's all part of the CI speed/coverage trade-off

folkertdev:
If you rebase I think we're good to go here.

madhav-madhusoodanan:
Just rebased and added `x86_64-unknown-linux-gnu` into `ci/intrinsic-test.sh` and the `intrinsic-test` CI pipeline

folkertdev (approved):
Allright, I think this looks good now. Thanks for the massive amount of work that went into this PR!

PR #1897, opened 2025-08-07
Fix CI after LLVM upgrade

nikic:
This fixes a number of issues after the LLVM upgrade:
 * Drop no longer needed feature gates (unrelated, just needed to work with current nightly)
 * Use intrinsics for some operations on s390x, as these are no longer recognized after https://github.com/rust-lang/rust/pull/145144 (fixed upstream by https://github.com/llvm/llvm-project/pull/153557).
 * Add missing avx512vl features to tests using assert_eq_m128h. This avoid a selection failure (github.com/llvm/llvm-project/issues/153570), but is also necessary for correctness.
 * Added some more avx512vl features to tests to work around another selection failure (https://github.com/llvm/llvm-project/issues/154492). These should get removed once the issue is fixed.
 * Adjusted immediate for vrndscalepd tests to ensure they actually use scaled rounding.

PR #1898, opened 2025-08-07
Implement arithmetic operation traits for x86 SIMD types

sayantn:
ACP: rust-lang/libs-team#628
Tracking Issue: rust-lang/rust#145066

Unresolved questions: Should we have `Not` and `Rem` implementations? At least `Not` seems quite useful.

I will send the PRs of other archs after this is merged, as they also use the common macro


sayantn:
CI is stuck on #1897 🤦🏽

folkertdev:
yeah that will likely take a couple of days at least

crates/core_arch/src/macros.rs
```diff
@@ -163,3 +163,61 @@ macro_rules! simd_extract {
     ($x:expr, $idx:expr $(,)?) => {{ $crate::intrinsics::simd::simd_extract($x, const { $idx }) }};
     ($x:expr, $idx:expr, $ty:ty $(,)?) => {{ $crate::intrinsics::simd::simd_extract::<_, $ty>($x, const { $idx }) }};
 }
+
+#[allow(unused)]
+macro_rules! impl_arith_op {
+    (__internal $op:ident, $intrinsic:ident $_:ident) => {
+        #[inline]
+        fn $op(self, rhs: Self) -> Self {
```
bjorn3, line 171: These are missing a `#[target_feature(enable = "...")]` matching the `_mm*` caller, so the MIR inliner will not inline them and LLVM may not inline them either.

sayantn: `simd_*` intrinsics call the *platform-independent* LLVM intrinsics/instructions. The conversion to ASM checks the available feature flags, and uses them as available. See https://github.com/rust-lang/libs-team/issues/628#issuecomment-3166386362

folkertdev:
Do you want to decide on those unresolved questions here, or later, or bounce it back to T-libs?

- `simd_rem` is not really used (I find 2 occurrences), suggesting that there is no instruction support for this operation and we should probably steer users away from it
- we don't have a `simd_not`, but `simd_xor(v, splat(all_one_bits))` should do something efficient

folkertdev:
Also @Amanieu merging this PR insta-stabilizes these instances, so just checking: that is intentional, right?

I feel like in that case we should at least implement the same operations for all stable simd types in the same rust release, i.e. at least make sure the aarch64 ones make it into the same release as the x86 ones.

Amanieu:
Any instantly-stable API needs a libs-api FCP, *even* if an ACP was accepted.

I can nominate this for libs-api discussion around Not/Rem, but I believe at least in the last discussion we decided against having `Rem` since there's no hardware support for it on any architecture.

Personally, I'm having second thoughts about having these trait implementations at all: I feel that in the long term what you really want to do is use the portable SIMD API most of the time and only occasionally dip into the target-specific intrinsics where needed. I also feel that these trait implementations can be a performance footgun because they may give you the impression that SIMD support is available for an operation (e.g. division) when it isn't actually supported by hardware.

sayantn:
Closing this because the ACP was rejected after further thought

PR #1899, opened 2025-08-10
Add testing for Arm64EC Windows

dpaoliello:
Add a CI job for Arm64EC Windows on the new Arm64 Windows 11 runners.

Removes the dedicated step to install rustup on the Arm64 Windows 11 runners as it's not included.


dpaoliello:
Had a clean build here: <https://github.com/rust-lang/stdarch/actions/runs/16865911732/job/47772786509?pr=1899>

Looks like s390x is currently broken in CI.

folkertdev:
Right, we're blocked on https://github.com/rust-lang/stdarch/pull/1897. The fix for our issue has been merged and cherry-picked into the next LLVM point release, but we'll have to wait a bit I guess before rust adopts that and we can use it.

folkertdev:
For context, is the lack of fp16 support something fundamental, or is it a temporary limitation (e.g. in msvc)?

tgross35:
> For context, is the lack of fp16 support something fundamental, or is it a temporary limitation (e.g. in msvc)?

Currently it's a crash in LLVM and MSVC doesn't support it. I'm hoping that https://github.com/llvm/llvm-project/pull/152843 is all that will be needed to make the scalar type work (if so, we can probably consider stabilizing it after we get LLVM22). If so, AIUI there shouldn't be any problem with these intrinsics.

folkertdev (approved):
Cool, this looks good! 

The line noise is a bit unfortunate, but it looks like we're on track to remove the `cfg`s in the future.

PR #1901, opened 2025-08-14
wasm: use `{read, write}_unaligned` methods

folkertdev:
This is shorter and more directly expresses what we want here.

(obviously blocked on getting CI working again)


tgross35:
This was added in cf1dbfa87cbb675ef0cb45523d471a27adde4ebb when read_unaligned definitely existed, @alexcrichton any reason not to do this?

alexcrichton (approved):
I suspect that was just my own ignorance for not using `{read,write}_unaligned`, agreed this is much better!

PR #1905, opened 2025-08-27
(Partially) Stabilize AVX512-FP16

sayantn:
Except for these 44 functions that explicitly use `f16` in the signature
 - `_mm{,256,512}_{set,set1,setr}_ph`
 - `_mm_set_sh`
 - `_mm{,256,512}_{load,loadu,store,storeu}_ph`
 - `_mm_{load,mask_load,maskz_load,store,mask_store}_sh`
 - `_mm{,256,512}_reduce_{add,mul,min,max}_ph`
 - `_mm{,256,512}_cvtsh_h`
 - `_mm{,256}_bcstnesh_ps`

Tracking issue: rust-lang/rust#127213

cc @tgross35


sayantn:
> It would probably be best to rename the `stdarch_x86_avx512fp16_2` or something like that rather than having the only difference be an underscore. But that can happen separately.

The other feature name is `stdarch_x86_avx512_f16` (not `fp16`), but I get your point lol

crates/core_arch/src/x86_64/avx512fp16.rs
```diff
@@ -32,7 +32,7 @@ pub fn _mm_cvti64_sh(a: __m128h, b: i64) -> __m128h {
 #[target_feature(enable = "avx512fp16")]
 #[cfg_attr(test, assert_instr(vcvtsi2sh, ROUNDING = 8))]
 #[rustc_legacy_const_generics(2)]
```
RalfJung, line 34: This should not be stabilized as-is. (@oli-obk @BoxyUwu at least that's my understanding of rustc_legacy_const_generics, please chime in if I am going off-track here)

BoxyUwU: Yeah IMO we should not be adding more uses of this attribute anywhere

RalfJung:
https://github.com/rust-lang/rust/issues/149654 has passed FCP so this should be ready to land from a t-lang perspective.

PR #1906, opened 2025-08-29
use `llvm.roundeven` on arm

folkertdev:
This did not work before (e.g. in https://github.com/rust-lang/stdarch/pull/1807), but I fixed the LLVM issue in https://github.com/llvm/llvm-project/pull/141786, which is now part of our LLVM.

Do we need to wait with this change for rust's minimum LLVM version to reach this version?

### DIFF
--- a/crates/stdarch-gen-arm/spec/neon/arm_shared.spec.yml
+++ b/crates/stdarch-gen-arm/spec/neon/arm_shared.spec.yml
@@ -2486,9 +2486,7 @@ intrinsics:
           name: "llvm.frinn.{neon_type}"
           links:
             - link: "llvm.roundeven.{neon_type}"
-              arch: aarch64,arm64ec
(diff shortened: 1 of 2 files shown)


crates/stdarch-gen-arm/spec/neon/arm_shared.spec.yml
```diff
@@ -2510,9 +2508,7 @@ intrinsics:
           name: "llvm.frinn.{neon_type}"
           links:
             - link: "llvm.roundeven.{neon_type}"
-              arch: aarch64,arm64ec
-            - link: "llvm.arm.neon.vrintn.{neon_type}"
-              arch: arm
+              arch: aarch64,arm64ec,arm
```
folkertdev, line 2511: the `arch` field is required, even though we now use the same code for all targets that this tool supports, so technically the cfg's could be dropped.

PR #1908, opened 2025-08-30
Make some remaining X86 intrinsics safe

sayantn:
- ADC/ADX
 - `_mm512_reduce_mul_ph` (missed)
 - `_bswap{,64}`
 - RDRAND/RDSEED
 - TBM
 - `_mm_prefetch` - It doesn't actually dereference the pointer argument, so it's safe
 - `_mm_{l,s,m}fence` - These can't introduce inconsistencies, as they are fences.
 - `_mm_pause` - It's like `std::hint::spin_loop`, so nothing unsafe here. Worst it can do is make the program slower

### DIFF
--- a/crates/core_arch/src/x86/sse.rs
+++ b/crates/core_arch/src/x86/sse.rs
@@ -1445,8 +1445,8 @@ pub fn _mm_move_ss(a: __m128, b: __m128) -> __m128 {
 #[target_feature(enable = "sse")]
 #[cfg_attr(test, assert_instr(sfence))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
-pub unsafe fn _mm_sfence() {
-    sfence()
+pub fn _mm_sfence() {
+    unsafe { sfence() }
 }
 
 /// Gets the unsigned 32-bit value of the MXCSR control and status register.
--- a/crates/core_arch/src/x86/sse.rs
+++ b/crates/core_arch/src/x86/sse.rs
@@ -1897,11 +1897,13 @@ pub const _MM_HINT_ET1: i32 = 6;
 #[cfg_attr(test, assert_instr(prefetchnta, STRATEGY = _MM_HINT_NTA))]
 #[rustc_legacy_const_generics(1)]
 #[stable(feature = "simd_x86", since = "1.27.0")]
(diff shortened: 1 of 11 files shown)


Amanieu:
The prefetch intrinsics can also be made safe.

sayantn:
It has a pointer parameter, can it still be made safe?

Also, what about TSC, CPUID and the fence instructions?

Amanieu:
These intrinsics are all safe to call as long as the required target features are available.

@rfcbot merge

sayantn:
@BurntSushi @m-ou-se @the8472 small ping for the rfc checkbox

crates/core_arch/src/x86/sse.rs
```diff
@@ -1897,11 +1897,13 @@ pub const _MM_HINT_ET1: i32 = 6;
...
-pub unsafe fn _mm_prefetch<const STRATEGY: i32>(p: *const i8) {
+pub fn _mm_prefetch<const STRATEGY: i32>(p: *const i8) {
     static_assert_uimm_bits!(STRATEGY, 3);
     // We use the `llvm.prefetch` intrinsic with `cache type` = 1 (data cache).
     // `locality` and `rw` are based on our `STRATEGY`.
-    prefetch(p, (STRATEGY >> 2) & 1, STRATEGY & 3, 1);
+    unsafe {
+        prefetch(p, (STRATEGY >> 2) & 1, STRATEGY & 3, 1);
+    }
 }
```
tgross35, line 1909: Seeing a safe function that takes a raw pointer is a bit surprising. It may be good to note in the documentation that this is safe because it does not interact with the AM and does not trap on invalid pointers (per https://llvm.org/docs/LangRef.html#llvm-prefetch-intrinsic).

Amanieu:
Restarting FCP to reflect updated team membership.

@rfcbot cancel
@rfcbot merge

Amanieu:
Needs a rebase.

crates/core_arch/src/x86/sse.rs
Amanieu: @sayantn Could you add a line in the docs to explain why this is safe despite taking a raw pointer?

sayantn:
@Amanieu rebased

@tgross35 added the comment on `_mm_prefetch`, could you check it pls

PR #1909, opened 2025-08-30
RISC-V: Lower requirements of `clmul` and `clmulh`

a4lg:
They don't need full "Zbc" extension but only its subset: the "Zbkc" extension.

Since the compiler implies `zbkc` from `zbc`, it's safe to use `#[target_feature(enable = "zbkc")]`.

### DIFF
--- a/crates/core_arch/src/riscv_shared/zb.rs
+++ b/crates/core_arch/src/riscv_shared/zb.rs
@@ -68,7 +68,7 @@ pub fn orc_b(rs: usize) -> usize {
 ///
 /// Section: 2.11
 #[unstable(feature = "riscv_ext_intrinsics", issue = "114544")]
-#[target_feature(enable = "zbc")]
+#[target_feature(enable = "zbkc")]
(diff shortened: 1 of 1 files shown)


a4lg:
Updating the documentation would be better but considering many intrinsics refer older, separate documentation (before being merged into the main RISC-V ISA Manual), that would be better to be a separate PR.

a4lg:
Hmm, CI is failing for an unrelated reason.

sayantn:
It's just that the `qemu-user-static` package was removed (replaced with `qemu-user`), and LA64 CI uses it (idk why it doesn't just use `qemu-user`)

sayantn:
Could you rebase thee branch pls, #1912 fixed the CI failure

a4lg:
Thanks for letting me know. Rebased and force-pushed.

a4lg:
Merge failure occurred on RISC-V (RV32) but it seems it's just a connection issue.
Could you add this to the merge queue again?

PR #1912, opened 2025-09-02
use `qemu-user` instead of `qemu-user-static` for loongarch CI

folkertdev:
This fixes CI, I'm not sure what broke really, but this is more consistent with our other Dockerfiles as well.

cc @heiher


sayantn:
Iirc the `qemu-user-static` package was deprecated, idk why the LA64 ci was using it.

heiher:
> Iirc the `qemu-user-static` package was deprecated, idk why the LA64 ci was using it.

From the logs, it looks like apt ended up installing `qemu-user` instead of `qemu-user-static`, which is likely due to the latter being deprecated.

heiher (approved):
LGTM. Thanks!

PR #1913, opened 2025-09-06
s390x: use the new `u128::funnel_shl`

folkertdev:
r? @sayantn 

btw the tracking issue mentions `wrapping_funnel_{shl,shr}` but those are not actually implemented (yet?).


crates/core_arch/src/s390x/vector.rs
```diff
@@ -3533,21 +3522,21 @@ mod sealed {
                     #[target_feature(enable = "vector")]
                     unsafe fn vec_sld<const C: u32>(self, b: Self) -> Self {
                         static_assert_uimm_bits!(C, 4);
-                        transmute(funnel_shl_u128(transmute(self), transmute(b), const { C as u128 * 8 }))
+                        transmute(u128::funnel_shl(transmute(self), transmute(b), const { C  * 8 }))
```
sayantn: Are these `const` blocks needed here? Is LLVM not propagating `const` status?

folkertdev: that's a bit of an artifact of using the s390x intrinsic originally, where you had to manually make sure that you passed a const or else llvm would panic. It looks like they can be removed now, everything still works without them (as expected, really).

PR #1914, opened 2025-09-07
`intrinsic-test`: make `build_{c, rust}_file` generic

folkertdev:
Most of the tricky logic of `SupportedArchitectureTest` is now in the trait, making it a lot easier, and shorter, to define new variants. That should make the x86 work a lot simpler.

best reviewed commit-by-commit I think.

cc @madhav-madhusoodanan

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,9 +13,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.19"
+version = "0.6.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


sayantn:
LGTM, cc @madhav-madhusoodanan waiting for your confirmation as you have been working on `intrinsic-test`

madhav-madhusoodanan:
Really useful commits @folkertdev, particularly moving `print_results_c` and removing the `IntrinsicDefinition` trait, and placing the arch-specific constants back in configs so that they don't have to be moved around.

Looks good to me too.

Issue #1904, opened 2025-08-27
`vreinterpret` on big endian swaps byte order


usamoi:
Consider the following code:

```rust
#![allow(improper_ctypes_definitions)]

use core::arch::aarch64::*;

#[inline(never)]
#[unsafe(no_mangle)]
#[target_feature(enable = "neon")]
extern "C" fn f(x: uint8x16_t) -> uint16x8_t {
    vreinterpretq_u16_u8(x)
}

fn main() {
    unsafe {
        let src = [0_u8, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
        let x: uint8x16_t = core::mem::transmute(src);
        let y: [u8; 16] = std::mem::transmute(f(x));
        dbg!(y);
    }
}
```

`cargo +nightly run --target aarch64_be-unknown-linux-gnu -Zbuild-std --release`:

```
[src/main.rs:17:9] y = [
    1,
    0,
    3,
    2,
    5,
    4,
    7,
    6,
    9,
    8,
    11,
    10,
    13,
    12,
    15,
    14,
]
```

`llvm-objdump -d`:

```
0000000000028870 <f>:
   28870: 4e200800     	rev64	v0.16b, v0.16b
   28874: 6e004000     	ext	v0.16b, v0.16b, v0.16b, #0x8
   28878: 4e201800     	rev16	v0.16b, v0.16b
   2887c: 4e200800     	rev64	v0.16b, v0.16b
   28880: 6e004000     	ext	v0.16b, v0.16b, v0.16b, #0x8
   28884: d65f03c0     	ret
```

`vreinterpret` should not generate instructions, but it does, and it also swaps the byte order of each element.

Clang does not generate any instructions with `--target=aarch64_be-linux-gnu`: https://godbolt.org/z/Erse7GnEK

sayantn:
This seems to be an oversight in the implementation (this also means the tests are probably doing something wrong too, they compare output with clang, it shouldn't have been passing for aarch64_be)

adamgemmell:
These intrinsics were doing some unnecessary swizzling but we're not sure if removing this has changed their behaviour. Our tests pass before and after the above fix and they appear to be working correctly.

Can I confirm that you're running on a BE platform and that the above fix works for you @usamoi ? (it hasn't made it into nightly rust yet, happy to wait until it does)

usamoi:
[...]

PR #1919, opened 2025-09-11
Remove big-endian swizzles from `vreinterpret`

sayantn:
Fixes: rust-lang/stdarch#1904 

This is the correct behavior according to the [Intrinsics Guide](https://developer.arm.com/architectures/instruction-sets/intrinsics/vreinterpret_s16_s8), which specifies that these should compile to `NOP`. Also, GCC and Clang matches this behavior.

I am a bit surprised that this didn't get caught in the AArch64_BE `intrinsic_test` run!

### DIFF
--- a/crates/core_arch/src/aarch64/neon/generated.rs
+++ b/crates/core_arch/src/aarch64/neon/generated.rs
@@ -21477,7 +21477,6 @@ pub fn vrecpxh_f16(a: f16) -> f16 {
 #[doc = "Vector reinterpret cast operation"]
 #[doc = "[Arm's documentation](https://developer.arm.com/architectures/instruction-sets/intrinsics/vreinterpret_f64_f16)"]
 #[inline]
-#[cfg(target_endian = "little")]
 #[target_feature(enable = "neon,fp16")]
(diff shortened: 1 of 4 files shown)


folkertdev:
cc @Jamesbarford @adamgemmell

adamgemmell:
I'm also really confused why the intrinsic-test tool doesn't catch this. Printing out the values you can see that the intrinsic does indeed get tested and it matches C. At the same time this PR clearly changes their behaviour but still passes the test tool

folkertdev:
I've confirmed locally that the rust `intrinsic-test-programs`  binary emits the same output on `master` and on this branch, but also that the generated assembly is in fact different (as expected based on the source code). So, eh, weird? 

Just confirming with the Arm folks: the implementation in this PR is the correct one, right?

Jamesbarford:
> I've confirmed locally that the rust `intrinsic-test-programs` binary emits the same output on `master` and on this branch, but also that the generated assembly is in fact different (as expected based on the source code). So, eh, weird?
> 
> Just confirming with the Arm folks: the implementation in this PR is the correct one, right?

Looking through how it is implemented in `arm_neon.h`, there is no shuffling. This, to me, looks to be correct

PR #1921, opened 2025-09-13
RISC-V: Improvements of inline assembly uses

a4lg:
This commit performs various improvements (better register allocation, less register clobbering on the worst case and better readability) of RISC-V inline assembly use cases.

Note that it does not change the `p` module (which defines the "P" extension draft instructions but very likely to change).

1.  Use `lateout` as possible.  
    Unlike `out(reg)` and `in(reg)` pair, `lateout(reg)` and `in(reg)` can share the same register because they state that the late-output register is written after all the reads are performed. It can improve register allocation.
2.  Add `preserves_flags` option as possible.  
    While RISC-V doesn't have _regular_ condition codes, RISC-V inline assembly in the Rust language assumes that some registers (mainly vector state registers) may be overwritten by default. By adding `preserves_flags` to the intrinsics corresponding instructions without overwriting them, it can minimize register clobbering on the worst case.
3.  Use trailing semicolon.  
    As `asm!` declares an action and it doesn't return a value by itself, it would be better to have trailing semicolon to denote that an `asm!` call is effectively a statement.
4.  Make most of `asm!` calls multi-lined.  
    `rustfmt` makes some simple (yet long) `asm!` calls multi-lined but it does not perform formatting of complex `asm!` calls with inputs and/or outputs.  To keep consistency, it makes most of the `asm!` calls multi-lined.

### DIFF
--- a/crates/core_arch/src/riscv64/mod.rs
+++ b/crates/core_arch/src/riscv64/mod.rs
@@ -20,7 +20,12 @@ pub use zk::*;
 #[unstable(feature = "riscv_ext_intrinsics", issue = "114544")]
 pub unsafe fn hlv_wu(src: *const u32) -> u32 {
     let value: u32;
-    asm!(".insn i 0x73, 0x4, {}, {}, 0x681", out(reg) value, in(reg) src, options(readonly, nostack));
+    asm!(
+        ".insn i 0x73, 0x4, {}, {}, 0x681",
+        lateout(reg) value,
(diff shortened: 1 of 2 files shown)


a4lg:
r? @sayantn

Fixed the style issues (resulted in a Check Style CI failure) and performed similar changes (not automatically performed by `rustfmt`) for consistency.  As a result, this PR makes most of the `asm!` calls multi-lined.

PR #1923, opened 2025-09-20
intrinsic-test: test intrinsics with patched `core_arch`

usamoi:
I'm surprised to find that `intrinsic-test` is testing `core::arch`, not `core_arch::arch`. This caused the failure that should have been detected in rust-lang/stdarch#1919 to go unnoticed until https://github.com/rust-lang/rust/pull/146709.

I guess the compiler's previous behavior was most likely correct, so we probably need to revert rust-lang/stdarch#1919. However, I really can't find any specification that describes the behavior of ARM NEON in big-endian mode.

cc @sayantn @adamgemmell @folkertdev @Jamesbarford

### DIFF
--- a/crates/core_arch/src/aarch64/neon/generated.rs
+++ b/crates/core_arch/src/aarch64/neon/generated.rs
@@ -10306,7 +10306,7 @@ pub fn vfmad_lane_f64<const LANE: i32>(a: f64, b: f64, c: float64x1_t) -> f64 {
 #[unstable(feature = "stdarch_neon_f16", issue = "136306")]
 #[cfg(not(target_arch = "arm64ec"))]
 pub fn vfmah_f16(a: f16, b: f16, c: f16) -> f16 {
-    unsafe { fmaf16(b, c, a) }
+    fmaf16(b, c, a)
 }
 #[doc = "Floating-point fused multiply-add to accumulator"]
 #[doc = "[Arm's documentation](https://developer.arm.com/architectures/instruction-sets/intrinsics/vfmah_lane_f16)"]
@@ -23259,7 +23259,7 @@ pub fn vrndaq_f64(a: float64x2_t) -> float64x2_t {
 #[cfg(not(target_arch = "arm64ec"))]
 #[cfg_attr(test, assert_instr(frinta))]
 pub fn vrndah_f16(a: f16) -> f16 {
-    unsafe { roundf16(a) }
+    roundf16(a)
 }
 #[doc = "Floating-point round to integral, to nearest with ties to away"]
 #[doc = "[Arm's documentation](https://developer.arm.com/architectures/instruction-sets/intrinsics/vrndh_f16)"]
@@ -23269,7 +23269,7 @@ pub fn vrndah_f16(a: f16) -> f16 {
 #[cfg(not(target_arch = "arm64ec"))]
 #[cfg_attr(test, assert_instr(frintz))]
 pub fn vrndh_f16(a: f16) -> f16 {
-    unsafe { truncf16(a) }
+    truncf16(a)
 }
 #[doc = "Floating-point round to integral, using current rounding mode"]
 #[doc = "[Arm's documentation](https://developer.arm.com/architectures/instruction-sets/intrinsics/vrndi_f16)"]
@@ -23450,7 +23450,7 @@ pub fn vrndmq_f64(a: float64x2_t) -> float64x2_t {
 #[cfg(not(target_arch = "arm64ec"))]
 #[cfg_attr(test, assert_instr(frintm))]
 pub fn vrndmh_f16(a: f16) -> f16 {
-    unsafe { floorf16(a) }
+    floorf16(a)
 }
 #[doc = "Floating-point round to integral, to nearest with ties to even"]
 #[doc = "[Arm's documentation](https://developer.arm.com/architectures/instruction-sets/intrinsics/vrndn_f64)"]
@@ -23581,7 +23581,7 @@ pub fn vrndpq_f64(a: float64x2_t) -> float64x2_t {
 #[cfg(not(target_arch = "arm64ec"))]
 #[cfg_attr(test, assert_instr(frintp))]
 pub fn vrndph_f16(a: f16) -> f16 {
-    unsafe { ceilf16(a) }
+    ceilf16(a)
 }
 #[doc = "Floating-point round to integral exact, using current rounding mode"]
 #[doc = "[Arm's documentation](https://developer.arm.com/architectures/instruction-sets/intrinsics/vrndx_f16)"]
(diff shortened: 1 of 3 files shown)


folkertdev:
The problem analysis is correct here, I think: that test should be using the local build of stdarch.

But we did all agree that https://github.com/rust-lang/stdarch/pull/1919 makes a correct change according to the documentation. I'll see if I can actually compare the generated assembly for clang and rustc for some of the failing functions.

usamoi:
> I'll see if I can actually compare the generated assembly for clang and rustc for some of the failing functions.

I checked a simplified version of generated assembly of `run_vreinterpret_u16_u8`. This code

```c
__attribute__((noinline)) void run_vreinterpret_u16_u8(uint8_t *a_vals,
                                                      uint16_t *b_vals) {
  vst1_u16(b_vals, vreinterpret_u16_u8(vld1_u8(a_vals)));
}
```

compiles to

```
0000000000400b8c <_Z23run_vreinterpret_u16_u8PhPt>:
  400b8c:	0c40701f 	ld1	{v31.8b}, [x0]
  400b90:	0c00743f 	st1	{v31.4h}, [x1]
  400b94:	d65f03c0 	ret
```

with gcc

and

```
00000000000106a4 <run_vreinterpret_u16_u8>:
   106a4:	0c407000 	ld1	{v0.8b}, [x0]
   106a8:	0e201800 	rev16	v0.8b, v0.8b
   106ac:	0c007020 	st1	{v0.8b}, [x1]
   106b0:	d65f03c0 	ret
```

with clang.

According to https://developer.arm.com/documentation/102159/0400/Load-and-store---data-structures,

> Element size also affects endianness handling. In general, if you specify the correct element size to the load and store instructions, bytes are read from memory in the appropriate order. This means that the same code works on little-endian systems and big-endian systems.

Reversing the byte order is reasonable. Moreover, GCC indeed did not generate any instructions for `vreinterpret`.

folkertdev:
Right, I think we just missed some? e.g. `vreinterpretq_u8_s32` fails in CI, and it still has its custom shuffling implementation while it should really just be a `transmute` based on what clang does https://godbolt.org/z/r8eY41vsb.

folkertdev:
Ah, no, I was using an old branch, and got confused. I actually think it might be an ABI difference that bites us here, consider:

```rust
use core::arch::aarch64::{int32x4_t, uint8x16_t, vreinterpretq_u8_s32};

#[no_mangle]
pub unsafe extern "C" fn c(x: int32x4_t) -> uint8x16_t {
    vreinterpretq_u8_s32(x)
}

#[no_mangle]
pub unsafe fn rust(x: int32x4_t) -> uint8x16_t {
    vreinterpretq_u8_s32(x)
}
```

```llvm
; Function Attrs: mustprogress nofree norecurse nosync nounwind willreturn memory(none)
define <16 x i8> @c(<4 x i32> %x) unnamed_addr #0 !dbg !6 {
start:
    #dbg_value(<4 x i32> %x, !32, !DIExpression(), !33)
    #dbg_value(<4 x i32> %x, !34, !DIExpression(), !39)
  %_0 = bitcast <4 x i32> %x to <16 x i8>, !dbg !41
  ret <16 x i8> %_0, !dbg !42
}

; Function Attrs: mustprogress nofree norecurse nosync nounwind willreturn memory(argmem: readwrite)
define void @rust(ptr dead_on_unwind noalias noundef writable writeonly sret([16 x i8]) align 16 captures(none) dereferenceable(16) initializes((0, 16)) %_0, ptr dead_on_return noalias noundef readonly align 16 captures(none) dereferenceable(16) %x) unnamed_addr #1 !dbg !43 {
start:
    #dbg_declare(ptr %x, !45, !DIExpression(), !46)
    #dbg_declare(ptr %x, !47, !DIExpression(), !50)
  %0 = load <4 x i32>, ptr %x, align 16, !dbg !52
  store <4 x i32> %0, ptr %_0, align 16, !dbg !52
  ret void, !dbg !53
}
```

with the rust ABI a load and store are performed, while with `extern "C"` the function just becomes a transmute.

sayantn:
As @usamoi's test with gcc and clang shows, they have different behavior for this intrinsic. The [ARM spec](https://developer.arm.com/architectures/instruction-sets/intrinsics/vreinterpret_s16_s8) explicitly says that it should generate `NOP`. But we see that only the GCC one adheres to this. I think clang has a bug here, not us.

usamoi:
> As @usamoi's test with gcc and clang shows, they have different behavior for this intrinsic. The [ARM spec](https://developer.arm.com/architectures/instruction-sets/intrinsics/vreinterpret_s16_s8) explicitly says that it should generate `NOP`. But we see that only the GCC one adheres to this. I think clang has a bug here, not us.

It should be emphasized that although the assembly generated by GCC does not contain `vrev`, it still reverses the byte order, on QEMU.

usamoi:
I guess `ld1` and `st1` reverse the byte order on big endian based on the element size. `clang` loads the vector of bytes, instead of the vector of halfwords, so it generates a `vrev` to emulate it.

Edit: see also https://llvm.org/docs/BigEndianNEON.html

sayantn:
But in C, `vreinterpretq` and a simple transmute (`*(T*)&`) generates the same code, the `rev`s we see are merely because of the `vld`/`vst`s, not the `vreinterpret`. https://godbolt.org/z/ax4hK587d

edit: the c++ code, using `reinterpret_cast`, also gives the same output https://godbolt.org/z/Gh5aGvhvh

edit: I believe the error is in the generated rust files, we print the output of rust using `transmute` and array indexing, whereas C uses `vget_lane`

edit: it seems that `vgetq_lane` `bswap`s the elements https://godbolt.org/z/8rzTj3WM9

usamoi:
```rust
#[unsafe(no_mangle)]
extern "C" fn run_vreinterpret_u16_u8(a_vals: *mut u8, b_vals: *mut u16) {
    use core::arch::aarch64::*;
    unsafe {
        vst1_u16(b_vals, vreinterpret_u16_u8(vld1_u8(a_vals)));
    }
}
```

```
0000000000028ce8 <run_vreinterpret_u16_u8>:
   28ce8: f9400008     	ldr	x8, [x0]
   28cec: f9000028     	str	x8, [x1]
   28cf0: d65f03c0     	ret
```

It's asm generated by `1.92.0-nightly (dd7fda570 2025-09-20)`. We can see that it generated `ldr` and `str` instead of `ld1` and `st1`. This is incorrect.

sayantn:
Ok, so the problem seems to be in `vld1` and `vst1`. They simply `transmute` and store, where they should `bswap` first. But the test harness would also need to be changed, because we use `transmute` to generate the Rust output, whereas we should use `vst1`.

edit: Just realized, if that is the case, why has been our test suite passing at all??

folkertdev:
it is probably cleanest to revert https://github.com/rust-lang/stdarch/pull/1919, merge this PR, and then properly fix the intrinsics? unless the fix is simple?

sayantn:
> it is probably cleanest to revert #1919, merge this PR, and then properly fix the intrinsics? unless the fix is simple?

Atp that seems to be the best route, this would probably need more discussions. I am opening a pr to revert #1919, and reopened #1904

sayantn:
@usamoi could you add the revert of #1919 here? The PR requires this test fix

usamoi:
> @usamoi could you add the revert of #1919 here? The PR requires this test fix

Done.

PR #1924, opened 2025-09-22
Revert "Remove big-endian swizzles from `vreinterpret`"

sayantn:
Reverts rust-lang/stdarch#1919. See rust-lang/stdarch#1923, we would need more discussions for this


sayantn:
We also need to do the work of #1923 here.

Also, wtf github, why did it make the branch in upstream 🤦🏽‍♂️ ?

Amanieu:
It might be easier to just add the revert to #1923.

sayantn:
That also works.

PR #1925, opened 2025-09-25
Make X86 intrinsics `const`

sayantn:
This is draft because it requires rust-lang/rust#146568, and one more PR after that which will implement the remaining SIMD intrinsics in const-eval.

Pending:
 - `cvtepi*_epi*`
 - `cvtt` (I need to check if these can be done without invoking UB)
 - `testz` etc (check if `simd_reduce_or` gives good codegen)


crates/core_arch/src/x86/test.rs
```diff
@@ -1,115 +1,228 @@
 //! Utilities used in testing the x86 intrinsics
 
+use crate::core_arch::assert_eq_const as assert_eq;
 use crate::core_arch::x86::*;
 use std::mem::transmute;
 
 #[track_caller]
 #[target_feature(enable = "sse2")]
-pub unsafe fn assert_eq_m128i(a: __m128i, b: __m128i) {
+#[rustc_const_unstable(feature = "stdarch_const_helpers", issue = "none")]
```
Amanieu, line 9: Is this actually needed? These functions aren't exported and shouldn't need stability attributes.

crates/core_arch/src/x86/test.rs
```diff
@@ -1,115 +1,228 @@
...
+
+make_ct_rt! {
+    // SAFETY: we can use simple float equality because when this should only be used in const
+    // context where Intel peculiarities don't appear
+
+    #[track_caller]
+    #[target_feature(enable = "sse2")]
+    pub unsafe fn assert_eq_m128d(a: __m128d, b: __m128d) {
+        ct: unsafe {
+            assert_eq!(transmute::<_, [f64; 2]>(a), transmute::<_, [f64; 2]>(b))
```
Amanieu, line 61: Since this is only used in tests, could we always us the ct version here? That would simplify the code.

crates/core_arch/src/x86/test.rs
sayantn: Yes, unfortunately const-stability attributes are needed even for private functions

crates/core_arch/src/x86/test.rs
sayantn: The ct version doesn't print the left and right arguments, because debug-printing is not const yet. I wanted to print the values in rt at least, so had 2 separate impls. 

Also one more problem is the ct version uses simple float equality, which differs from intel's definition for denormals and nans. In the functions we actually call the const version, we don't have such problems because those have standard float conventions.

PR #1926, opened 2025-09-26
Update memchr to 2.7.6

neuschaefer:
memchr 2.7.6 contains a bugfix for aarch64_be


neuschaefer:
One possible concern, not about this update specifically, but about the interaction of stdarch and memchr generally, is that the recent work in https://github.com/rust-lang/stdarch/issues/1904 seems like it might make this change in memchr (https://github.com/BurntSushi/memchr/pull/188) invalid.

neuschaefer:
CI is segfaulting on x86_64-pc-windows-gnu, which seems spurious

PR #1928, opened 2025-10-01
Use SIMD intrinsics whereever possible

sayantn:
Also, remove uses of some deprecated LLVM intrinsics

FIXME: `_mm256_test{z,c}_{ps,pd}` doesn't optimize well in the testsuite for some reason, it works on godbolt 😕

### DIFF
--- a/crates/core_arch/src/x86/avx.rs
+++ b/crates/core_arch/src/x86/avx.rs
@@ -1983,7 +1983,10 @@ pub fn _mm256_unpacklo_ps(a: __m256, b: __m256) -> __m256 {
 #[cfg_attr(test, assert_instr(vptest))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub fn _mm256_testz_si256(a: __m256i, b: __m256i) -> i32 {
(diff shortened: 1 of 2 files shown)


sayantn:
`macos-15` runners are failing for some reason, is this a GH issue?

Amanieu:
I restarted the runners and it seems to be working fine now. Do you want this merged now or try to find a solution for `_mm256_test{z,c}_{ps,pd}`?

sayantn:
let's merge this for now. That can be another PR later (the `testnzc` intrinsics will also need some attention)

sayantn:
Just discovered that the `s{l, r}l{,v}` intrinsics are wrongly implemented here, they should zero on shift amount overflow (like `unbounded_shl`), but do wrap on overflow (like `wrapping_shl`). I will send a pr to revert those

Edit: weirdly, the rotate instructions do wrap on overflow

RalfJung:
Nice, this means we can remove the implementation of vperm2 from Miri. :)
Are there any similar recent patches / more implementations we can remove now?

sayantn:
There should be quite a few, we have removed uses of many llvm intrinsics recently

PR #1929, opened 2025-10-01
Fixes for non-temporal intrinsics

sayantn:
- Use inline asm for `_mm_stream_{ss,sd}`
 - Add `_mm_sfence` calls in tests, without it the tests are technically UB

### DIFF
--- a/crates/core_arch/src/x86/avx.rs
+++ b/crates/core_arch/src/x86/avx.rs
@@ -4291,6 +4291,7 @@ mod tests {
         let a = _mm256_setr_epi64x(1, 2, 3, 4);
         let mut r = _mm256_undefined_si256();
         _mm256_stream_si256(ptr::addr_of_mut!(r), a);
+        _mm_sfence();
         assert_eq_m256i(r, a);
     }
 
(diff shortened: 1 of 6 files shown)


Amanieu:
Can you add a comment explaining why inline asm is used here?

sayantn:
Shouldn't the implementation of `stream_load` intrinsics use `_mm_sfence`? Because afaiu `stream_load` is just `stream`, but we are loading the data onto the stack after the move, so the user doesn't need to call `_mm_sfence`. But doesn't this mean that *we* need to call it, otherwise the stack variable might end up (partially) uninitalized?

Amanieu:
My understanding is that x86 memory model weirdness only applies to non-temporal stores, not non-temporal loads.

sayantn:
SDM says

> MOVNTDQA loads a double quadword from the source operand (second operand) to the destination operand (first
operand) using a non-temporal hint if the memory source is WC (write combining) memory type

and 

> Because the WC protocol uses a weakly-ordered memory consistency model, a fencing operation implemented with
a MFENCE instruction should be used in conjunction with MOVNTDQA instructions if multiple processors might use
different memory types for the referenced memory locations or to synchronize reads of a processor with writes by other agents in the system.

so it seems like this can be UB if some very specific conditions are met (I am not too good with memory types). Do we not support such things (like using different memory types for same chunk in different processors)? It still seems like the second condition (syncing with other agents) might be a vulnerability

Amanieu:
Memory type is an attribute that the OS sets for a range of physical memory. Normal memory uses the WB protocol which has the expected memory ordering semantics. Essentially we don't consider WC memory as normal memory for the Rust abstract machine, so the condition doesn't apply here.

PR #1930, opened 2025-10-02
`core_arch::x86` : Fix the implementation of `_kshift` instructions

madhav-madhusoodanan:
# Summary
The `_kshiftri_mask32`, `_kshiftri_mask64`, `_kshiftli_mask32` and `_kshiftli_mask64` intrinsics in `core_arch::x86` module do not handle a critical edge case when the shift amount exceeds the bit-length of the input argument. 

EDIT: updated the 8-bit and 16-bit versions of the same.

# Current behaviour
For `_kshiftri_mask32` and `_kshiftli_mask32` intrinsics, when the shift amount (passed as a const-generic argument) exceeds 32, the shift amount applied to the argument becomes `shift % 32`.

Similar is the case with the 8-bit, 16-bit and 64-bit variants of the same.

[Godbolt link for the 32-bit variant with minimal replication](https://godbolt.org/z/oKea3d55f).
[Godbolt link for the 64-bit variant with minimal replication](https://godbolt.org/z/EbGhh7eo8).

# Expected behaviour
When the shift amount exceeds 32 (for the 32-bit versions) or 64 (for the 64-bit versions) the result becomes zero, in line with Intel's [documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_kshiftli_mask32&ig_expand=3943) on the same.

Similar is the case with 8-bit and 16-bit versions.

# Fix
Use `unbounded_shr()` and `unbounded_shl()` functions instead of `>>` and `<<` operations.

[Godbolt link that shows the corrected version of the 32-bit implementation](https://godbolt.org/z/x1jfh1G9n).
[Godbolt link that shows the corrected version of the 64-bit implementation](https://godbolt.org/z/EbGhh7eo8)

r? @sayantn 
cc: @folkertdev

### DIFF
--- a/crates/core_arch/src/x86/avx512bw.rs
+++ b/crates/core_arch/src/x86/avx512bw.rs
@@ -10440,7 +10440,7 @@ pub fn _kortestz_mask64_u8(a: __mmask64, b: __mmask64) -> u8 {
 #[rustc_legacy_const_generics(1)]
 #[stable(feature = "stdarch_x86_avx512", since = "1.89")]
 pub fn _kshiftli_mask32<const COUNT: u32>(a: __mmask32) -> __mmask32 {
-    a << COUNT
+    a.unbounded_shl(COUNT)
 }
 
 /// Shift the bits of 64-bit mask a left by count while shifting in zeros, and store the least significant 32 bits of the result in k.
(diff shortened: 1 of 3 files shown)


sayantn:
There are 8- and 16-bit versions too in avx512dq, check them too pls

madhav-madhusoodanan:
Done @sayantn

sayantn:
Thanks, I have opened #1931 to fix the other shift-related bugs

PR #1931, opened 2025-10-02
Fix mistake in #1928

sayantn:
The x86 shift intrinsics do an overflowing shift (i.e. 0 in case of shift larger than type size), instead of wrapping shift (i.e. shift amount is taken modulo type size). So, we can't use `simd_sh{l,r}` here.

Also, `simd_funnel_sh{l,r}` doc says its UB for the shift amount to be greater than or equal to the type size. So, and them with `T::BITS - 1` to ensure soundness (LLVM is clever enough to sees through it, there is no codegen impact)


sayantn:
wtf? random segfault??

edit: what has github been doing recently? Seeing so many spurious failures!

madhav-madhusoodanan:
x86-64_pc-windows-msvc has been giving segfaults randomly, been noticing it too frequently when building - https://github.com/rust-lang/stdarch/pull/1894

sayantn:
Weird, but expected from msvc lol

Edit: this can possibly be something wrong on our side too, one test can be UBing and somehow only msvc is able to raise a segfault

PR #1933, opened 2025-10-04
[DO NOT MERGE][EXPERIMENT] check `x86_64-pc-windows-msvc` CI

sayantn:
It has been failing fairly often, just want to be sure that everything is OK

edit: CI be like

image

edit: ok so it's probably `test_xsave`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,39 +4,39 @@ on:
   merge_group:
 
 jobs:
-  style:
-    name: Check Style
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v4
-    - name: Install Rust
-      run: rustup update nightly --no-self-update && rustup default nightly
-    - run: ci/style.sh
+  # style:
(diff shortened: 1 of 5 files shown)


sayantn:
seems like our xsave buffer is too small. We have a fixed-size 2560 byte buffer, but SDM (Vol 1, section 13.1-13.5) specifies that the position and size of xsave components are not fixed in general, and require querying CPUID for the exact size and position. Even doing a naive computation, with assuming packed arrangement, I am getting 3636 bytes without AMX, and with AMX present it skyrockets to 11892 bytes.

So, ideally we should query CPUID for each xsave component we are going to retrieve, and allocate *dynamically* as needed (I couldn't see any mention of a "largest" possible xsave area size in SDM)

As for why it is only showing up just now, idk. Maybe GH updated their runner machines, so they are more capable now (meaning more components *will* actually be stored in the xsave area, because `xsave` only stores the data for areas which are requested *and* are actually available via `XCR0`)

PR #1934, opened 2025-10-04
Fix xsave segfaults

sayantn:
It turns out that MSVC was right, and we were wrong (such a weird thing to say).

Our `xsave` tests used a fixed-size 2560 byte buffer for saving all register data. But the problem is that the required size for xsave is not fixed, it depends on the CPU. This piece of code is ancient, and when it was written the authors calculated this limit based on the typical size. But currently, the maximum size required for xsave can be as high as 11892 bytes (heck on my laptop it is 2696 bytes, if anyone wants to try it out the code is in [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=98d5ba7f35621ded6d71e191ae5cb999)), for our SDE run it is 11008 bytes (which also explains why these tests were disabled in the `x86_64-unknown-linux-gnu` run, due to such a huge discrepancy in size, it always generated a segfault). So in a way, MSVC helped us detect this bug.

See rust-lang/stdarch#1933 for some more context

As for why we are only seeing this now, if the CPU only supports up to AVX, 2560 bytes is enough space, but my guess is that GH introduced some better (windows?) runners, with AVX512 or AMX support, and both of those are capable of skyrocketing the XSAVE area size.

I will still rerun the CI a few times to ensure no weird random behavior is snooping around.

### DIFF
--- a/crates/core_arch/src/x86/xsave.rs
+++ b/crates/core_arch/src/x86/xsave.rs
@@ -160,28 +160,38 @@ pub unsafe fn _xrstors(mem_addr: *const u8, rs_mask: u64) {
 }
 
 #[cfg(test)]
+pub(crate) use tests::XsaveArea;
+
+#[cfg(test)]
 mod tests {
-    use std::{fmt, prelude::v1::*};
+    use std::boxed::Box;
 
     use crate::core_arch::x86::*;
     use stdarch_test::simd_test;
 
-    #[repr(align(64))]
     #[derive(Debug)]
-    struct XsaveArea {
-        // max size for 256-bit registers is 800 bytes:
-        // see https://software.intel.com/en-us/node/682996
-        // max size for 512-bit registers is 2560 bytes:
-        // FIXME: add source
-        data: [u8; 2560],
+    pub(crate) struct XsaveArea {
+        data: Box<[AlignedArray]>,
     }
 
+    #[repr(align(64))]
+    #[derive(Copy, Clone, Debug)]
+    struct AlignedArray([u8; 64]);
+
     impl XsaveArea {
-        fn new() -> XsaveArea {
-            XsaveArea { data: [0; 2560] }
+        #[target_feature(enable = "xsave")]
+        pub(crate) fn new() -> XsaveArea {
+            // `CPUID.(EAX=0DH,ECX=0):ECX` contains the size required to hold all supported xsave
+            // components. `EBX` contains the size required to hold all xsave components currently
+            // enabled in `XCR0`. We are using `ECX` to ensure enough space in all scenarios
+            let CpuidResult { ecx, .. } = unsafe { __cpuid(0x0d) };
+
+            XsaveArea {
+                data: vec![AlignedArray([0; 64]); ecx.div_ceil(64) as usize].into_boxed_slice(),
+            }
         }
-        fn ptr(&mut self) -> *mut u8 {
-            self.data.as_mut_ptr()
+        pub(crate) fn ptr(&mut self) -> *mut u8 {
+            self.data.as_mut_ptr().cast()
         }
     }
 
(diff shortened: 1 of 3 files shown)


sayantn:
something still seems wrong, the segfaulting rate is now 100% lol, imma try it again in #1933

sayantn:
i'm just dumb, made the `Box` aligned to 64 bytes instead of the inner slice 😑

crates/core_arch/src/x86/xsave.rs
```diff
@@ -159,29 +159,40 @@ pub unsafe fn _xrstors(mem_addr: *const u8, rs_mask: u64) {
...
     impl XsaveArea {
-        fn new() -> XsaveArea {
-            XsaveArea { data: [0; 2560] }
+        #[target_feature(enable = "xsave")]
+        pub(crate) fn new() -> XsaveArea {
+            // `CPUID.(EAX=0DH,ECX=0):ECX` contains the size required to hold all supported xsave
+            // components. `EBX` contains the size required to hold all xsave components currently
+            // enabled in `XCR0`.
+            // FIXME: which one to use?
+            let CpuidResult { ecx, .. } = unsafe { __cpuid(0x0d) };
```
folkertdev, line 187: You want to leave this fixme for later? And default to `ecx` because that should be enough space no matter what?

crates/core_arch/src/x86_64/xsave.rs
```diff
@@ -126,39 +126,20 @@ pub unsafe fn _xrstors64(mem_addr: *const u8, rs_mask: u64) {
...
         let m = 0xFFFFFFFFFFFFFFFF_u64; //< all registers
         let mut a = XsaveArea::new();
         let mut b = XsaveArea::new();
 
-        xsave::_xsave64(a.ptr(), m);
-        xsave::_xrstor64(a.ptr(), m);
-        xsave::_xsave64(b.ptr(), m);
+        _xsave64(a.ptr(), m);
+        _xrstor64(a.ptr(), m);
+        _xsave64(b.ptr(), m);
```
folkertdev, line 142: you're re-enabling 4 functions `xsave,xsaveopt,xsave64,xsaveopt64`, but there are only 3 tests here. Also shouldn't there be some sort of assert? or is this just a test that the functions work at all?

folkertdev: Ah looking at it again all 4 functions do occur in these tests. Still, I'm not sure what this is actually testing beyond that these functions don't just segfault/sigill.

crates/core_arch/src/x86/xsave.rs
sayantn: Yeah, that's the question, if we default to ecx then can some user of `xsetbv`change `xcr0` to enable more features than supported? I don't think it is possible, but want to be sure

crates/core_arch/src/x86_64/xsave.rs
sayantn: There were `assert`s before, they were removed for some reason, can't remember why. I'll check if I can add them back

sayantn: looking back on #1672, I don't think adding back the `assert`s is a good idea. The compiler can insert some other calls in between `_xrstor` and the second `_xsave`, which can change the register states. These tests should just be there to ensure the intrinsics don't segfault/sigill

crates/core_arch/src/x86/xsave.rs
sayantn: i guess using ecx is always safe, because the cpu doesn't allow anyone setting an unsupported `xcr0` bit (afaik it allows unsetting a xcr0 bit, but `_xsetbv` is supervisor-mode only)

PR #1935, opened 2025-10-05
Make `cpuid` safe and update docs

sayantn:
Counterproposal to rust-lang/rust#146560


traviscross:
Proposing FCP on this (in fact, even nominating it) is blocked on:

- https://github.com/rust-lang/team/pull/2007

crates/core_arch/src/x86/cpuid.rs
```diff
@@ -28,10 +28,19 @@ pub struct CpuidResult {
...
 ///
-/// The highest-supported leaf value is returned by the first tuple argument of
-/// [`__get_cpuid_max(0)`](fn.__get_cpuid_max.html). For leaves containing
-/// sub-leaves, the second tuple argument returns the highest-supported
-/// sub-leaf value.
+/// There are 2 types of information leafs - basic leafs (with `leaf < 0x8000000`)
+/// and extended leafs (with `leaf >= 0x80000000`). The highest supported basic and
+/// extended leafs can be obtained by calling CPUID with `0` and `0x80000000`,
+/// respectively, and reading the value in the `EAX` register. If the leaf supports
+/// more than one sub-leafs, then the procedure of obtaining the highest supported
```
RalfJung:
```suggestion
/// more than one sub-leaf, then the procedure of obtaining the highest supported
```

crates/core_arch/src/x86/cpuid.rs
```diff
@@ -28,10 +28,19 @@ pub struct CpuidResult {
 /// Returns the result of the `cpuid` instruction for a given `leaf` (`EAX`)
 /// and `sub_leaf` (`ECX`).
 ///
-/// The highest-supported leaf value is returned by the first tuple argument of
-/// [`__get_cpuid_max(0)`](fn.__get_cpuid_max.html). For leaves containing
-/// sub-leaves, the second tuple argument returns the highest-supported
-/// sub-leaf value.
+/// There are 2 types of information leafs - basic leafs (with `leaf < 0x8000000`)
+/// and extended leafs (with `leaf >= 0x80000000`). The highest supported basic and
```
RalfJung:
```suggestion
/// and extended leaves (with `leaf >= 0x80000000`). The highest supported basic and
```
Does this not use the usual plural spelling of leaf?

sayantn: it should, I'm just a lousy speller

traviscross:
@sayantn, if you could please update the PR description with the details of the stabilization being proposed, the motivation, etc., that would be helpful to us in considering this nomination and for proposing FCP.

joshtriplett:
:+1: for this. One clarification:

> processors lower than i386

For the avoidance of ambiguity: lower than i586, which is lower than the current 32-bit x86 targets in Rust.

traviscross:
Thanks @sayantn.  Le'ts propose to do it.

@rfcbot fcp merge

tmandry:
We discussed this in the lang team meeting today and the agreement was that if we add support for an older target like i586, we can make this unsafe and require target_feature for the intrinsic on that target. These intrinsics are already platform-specific, so it's natural to have a more target-specific definition that changes the safety of the intrinsic.

@rfcbot reviewed

nikomatsakis:
@rfcbot reviewed

sayantn:
Any updates/concerns on this @Amanieu @BurntSushi @the8472 ? Seems like this got more traction than the counterproposal in r-l/r

crates/core_arch/src/x86/cpuid.rs
```diff
@@ -28,12 +28,21 @@ pub struct CpuidResult {
 /// Returns the result of the `cpuid` instruction for a given `leaf` (`EAX`)
 /// and `sub_leaf` (`ECX`).
 ///
-/// The highest-supported leaf value is returned by the first tuple argument of
-/// [`__get_cpuid_max(0)`](fn.__get_cpuid_max.html). For leaves containing
-/// sub-leaves, the second tuple argument returns the highest-supported
-/// sub-leaf value.
+/// There are 2 types of information leaves - basic leaves (with `leaf < 0x8000000`)
```
tgross35: Optional nit: in formal writing it's usually preferred to spell out small numbers as "two" rather than using digits

sayantn:
@traviscross @joshtriplett can this be merged now that the RFC is complete?

traviscross:
> can this be merged now that the [FCP] is complete?

Yes; this is OK to go forward as far as lang and libs-api are concerned.  It now has an approving review from @tgross35, so it seems this is OK to merge.

jethrogb:
> SGX environments can't use CPUID. Or more accurately, they can't trust the output of CPUID.

The latter sentence is incorrect. The former is correct. Executing the CPUID instruction in SGX results in an #UD exception.

> It should also be mentioned that SGX has been deprecated by intel for quite some time now.

This is also incorrect. SGX is not deprecated and [Intel's newest processors](https://www.intel.com/content/www/us/en/products/sku/245097/intel-xeon-6725p-processor-192m-cache-3-70-ghz/specifications.html) continue to have SGX.

The SGX target maintainers would've appreciated a heads up for this change.

tgross35:
In that case we could either panic or return zeros on SGX when this is called, rather than allowing a crash. I expect zeros would be nicer in more cases so code just progresses as if target features aren't available when run on SGX, rather than needing to special case

@jethrogb if one of these sounds reasonable, would you mind putting up a PR?

jethrogb:
Panicking should be fine, it's not really more or less safe than triggering #UD.

Pretending features aren't available could cause security issues, for example if code is checking whether AES-NI is supported and falling back to an AES implementation that's not constant time.

sayantn:
That is *not* a secure implementation, an implementation of AES should always fall back on constant-time implementations (e.g using bitslicing). We cannot do anything about such code -- in general any secure program should behave identically in presence or absence of any target features

Also an alternate approach, can we just `#[cfg(not(target_env = "sgx"))]`it?

raoulstrackx:
You mean doing something like:
```
    if cfg!(not(target_env = "sgx")) {
        ...
    } else {
        panic!("Instruction not allowed in SGX");
    }
```
I think that's fine.

sayantn:
No I meant more like `#[cfg(target_env = "sgx")]` the whole `cpuid` function, but now I realize that might be breaking. If panicking at runtime is acceptable, I can put a PR

raoulstrackx:
Yes `#[cfg(target_env = "sgx")]` the whole `cpuid` function is backwords incompatible. Panicking at runtime is fine. Thanks you!

mkroening:
@joshtriplett wrote in https://github.com/rust-lang/rust/pull/146560#issuecomment-3357108322:

> That said, counterproposal: rather than a target feature, can we transition to cpuid being universally safe (using the mechanisms we normally use to move something from unsafe to safe without provoking widespread warnings about unused unsafe)?

This currently triggers `unused_unsafe` (see https://github.com/rust-osdev/x86_64/pull/575). Is this intentional?

RalfJung:
@mkroening could you file an issue for that? Comments on closed PRs get lost very easily.

mkroening:
Sure! I opened https://github.com/rust-lang/stdarch/issues/1990. I hope this is the right repo and not the Rust one. :)

PR #1938, opened 2025-10-09
Implement fjcvtzs under the name __jcvt like the C intrinsic

linkmauve:
This instruction is only available when the `jsconv` target_feature is available, so on ARMv8.3 or higher.

It is used e.g. by [Ruffle](https://github.com/ruffle-rs/ruffle/pull/21780) to speed up its conversion from `f64` to `i32`, or by any JS engine probably.

I got redirected to this repository from https://github.com/rust-lang/rust/pull/147517

### DIFF
--- a/crates/stdarch-gen-arm/spec/neon/aarch64.spec.yml
+++ b/crates/stdarch-gen-arm/spec/neon/aarch64.spec.yml
@@ -63,6 +63,9 @@ neon-unstable-f16: &neon-unstable-f16
 neon-unstable-feat-lut: &neon-unstable-feat-lut
   FnCall: [unstable, ['feature = "stdarch_neon_feat_lut"', 'issue = "138050"']]
 
+aarch64-unstable-jscvt: &aarch64-unstable-jscvt
+  FnCall: [unstable, ['feature = "stdarch_aarch64_jscvt"', 'issue = "147555"']]
+
 # #[cfg(target_endian = "little")]
 little-endian: &little-endian
   FnCall: [cfg, ['target_endian = "little"']]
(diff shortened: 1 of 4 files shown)


crates/stdarch-gen-arm/spec/neon/aarch64.spec.yml
```diff
@@ -63,6 +63,9 @@ neon-unstable-f16: &neon-unstable-f16
 neon-unstable-feat-lut: &neon-unstable-feat-lut
   FnCall: [unstable, ['feature = "stdarch_neon_feat_lut"', 'issue = "138050"']]
 
+aarch64-jsconv-stable: &aarch64-jsconv-stable
+  FnCall: [stable, ['feature = "stdarch_aarch64_jsconv"', 'since = "1.92.0"']]
+
```
folkertdev: new functions should be marked as `unstable`. To actually stabilize them requires a `T-libs-api` approval, which generally happens after a function has been unstable for some time.

folkertdev: That means you'd need to introduce a new feature, which can be done by just picking a new feature name here, e.g. `"stdarch_aarch64_jcvt"`  (this is the only intrinsic for this target feature, based on https://github.com/llvm/llvm-project/blob/main/clang/lib/Headers/arm_acle.h#L663). You should also create a tracking issue for that feature in the main rust repo, e.g. like https://github.com/rust-lang/rust/issues/138050)

folkertdev: Fleshing this out a bit further, the cycle is roughly

- merge the change here
- wait for a sync with `rust-lang/rust`
- wait for a new nightly build of rust
- use this feature on nightly and generally give it some time to see if any issues pop up (unlikely in this case)
- request stabilization

linkmauve: Thanks for the review!

I’ve now opened #147555 and pushed a new unstable revision, under the feature `stdarch_aarch64_jscvt` (and as said in the commit message, this name is debatable).

folkertdev:
you also need to enable your new feature here 

https://github.com/rust-lang/rust/blob/9725c4baacef19345e13f91b27e66e10ef5592ae/library/stdarch/crates/intrinsic-test/src/arm/config.rs#L124-L125

for the automated testing to work

linkmauve:
> you also need to enable your new feature here
> 
> https://github.com/rust-lang/rust/blob/9725c4baacef19345e13f91b27e66e10ef5592ae/library/stdarch/crates/intrinsic-test/src/arm/config.rs#L124-L125
> 
> for the automated testing to work

Oops, thanks for the hint.

PR #1940, opened 2025-10-10
Disable AMD-specific intrinsic tests in the SDE run

sayantn:
Somehow these tests succeeded all this time, when SDE doesn't actually support AMD extensions at all

r? @Amanieu


ci/docker/x86_64-unknown-linux-gnu/Dockerfile
```diff
@@ -14,3 +14,6 @@ RUN tar -xJf sde.tar.xz --strip-components=1 -C intel-sde
 ENV CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER="/intel-sde/sde64 \
             -cpuid-in /checkout/ci/docker/x86_64-unknown-linux-gnu/cpuid.def \
             -rtm-mode full -tsx --"
+# SDE doesn't support AMD extensions
+# FIXME: find a way to test these
+ENV STDARCH_TEST_SKIP_FEATURE="sse4a,tbm,xop"
```
Amanieu, line 19: Is this still needed with the modified cpuid? We should automatically skip tests for which we don't have the features.

sayantn: It will skip, but then it will fail due to `STDARCH_TEST_EVERYTHING`

ci/docker/x86_64-unknown-linux-gnu/cpuid.def
```diff
@@ -59,5 +59,4 @@
 80000008 ******** => 00003028 00000200 00000200 00000000
 
 # This file was copied from intel-sde/misc/cpuid/future/cpuid.def, and modified to
-# use "AuthenticAMD" as the vendor and the support for `XOP`, `SSE4a`, `TBM`,
-# `AVX512_VP2INTERSECT` and the VEX variants of AVX512 was added in the CPUID.
+# add support for `AVX512_VP2INTERSECT`
```
RalfJung, line 62: Distributing a modified version of this file violates its license. We should delete this from the repo ASAP.

RalfJung: Ah looks like this was discussed on [Zulip](https://rust-lang.zulipchat.com/#narrow/channel/231349-licensing/topic/Intel.20Simplified.20Software.20License/with/613853671).

PR #1941, opened 2025-10-18
CI: Separate `intrinsic-test` from the other CI tests

madhav-madhusoodanan:


.github/workflows/main.yml
```diff
@@ -248,6 +248,74 @@ jobs:
...
+        target:
+        # Dockers that are run through docker on linux
+        - tuple: arm-unknown-linux-gnueabihf
+          os: ubuntu-latest
+        - tuple: armv7-unknown-linux-gnueabihf
+          os: ubuntu-latest
+        - tuple: aarch64-unknown-linux-gnu
+          os: ubuntu-latest
+        - tuple: aarch64_be-unknown-linux-gnu
+          os: ubuntu-latest
```
folkertdev: I think you should be able to structure these rules like here

https://github.com/trifectatechfoundation/libzstd-rs-sys/blob/054f99b1f625fc584022253edf99699c877cf044/.github/workflows/checks.yaml#L38-L65

To reduce the duplication

.github/workflows/main.yml
```diff
@@ -248,6 +248,74 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v4
+    - name: Install Rust
+      run: |
+        rustup update nightly --no-self-update
+        rustup default nightly
+      shell: bash
+    - run: rustup target add ${{ matrix.target.tuple }}
+      shell: bash
+      if: matrix.build_std == ''
```
folkertdev: just 
```suggestion
      if: matrix.build_std
```

should work I think?

.github/workflows/main.yml
```diff
@@ -248,6 +248,74 @@ jobs:
...
+    - run: |
+        rustup component add rust-src
+        echo "CARGO_UNSTABLE_BUILD_STD=std" >> $GITHUB_ENV
+      shell: bash
+      if: matrix.build_std != ''
```
folkertdev:
```suggestion
      if: ${{ (matrix.build_std || false) == false }}
```

is the best I can think of

ci/intrinsic-test.sh
```diff
@@ -0,0 +1,114 @@
...
+
+set -ex
+
+: "${TARGET?The TARGET environment variable must be set.}"
+
+# Tests are all super fast anyway, and they fault often enough on travis that
+# having only one thread increases debuggability to be worth it.
+#export RUST_BACKTRACE=full
+#export RUST_TEST_NOCAPTURE=1
+#export RUST_TEST_THREADS=1
```
folkertdev: we are not (currently) running `RUST_TEST`, so I think this can be removed?

ci/intrinsic-test.sh
```diff
@@ -0,0 +1,114 @@
...
+        export RUSTFLAGS="${RUSTFLAGS} -Ctarget-feature=+zk,+zks,+zbb,+zbc"
+        ;;
+esac
+
+echo "RUSTFLAGS=${RUSTFLAGS}"
+echo "OBJDUMP=${OBJDUMP}"
+echo "STDARCH_DISABLE_ASSERT_INSTR=${STDARCH_DISABLE_ASSERT_INSTR}"
+echo "STDARCH_TEST_EVERYTHING=${STDARCH_TEST_EVERYTHING}"
+echo "STDARCH_TEST_SKIP_FEATURE=${STDARCH_TEST_SKIP_FEATURE}"
+echo "STDARCH_TEST_SKIP_FUNCTION=${STDARCH_TEST_SKIP_FUNCTION}"
```
folkertdev: I think these are not needed?

ci/intrinsic-test.sh
```diff
@@ -0,0 +1,114 @@
...
+echo "PROFILE=${PROFILE}"
+
+INTRINSIC_TEST="--manifest-path=crates/intrinsic-test/Cargo.toml"
+
+# Test targets compiled with extra features.
+case ${TARGET} in
+
+    x86_64* | i686*)
+        export STDARCH_DISABLE_ASSERT_INSTR=1
+        ;;
```
folkertdev: this can likely wait until the x86 PR?

.github/workflows/main.yml
madhav-madhusoodanan: Ohh you mean, combining the `target` and `include` sections?

ci/intrinsic-test.sh
madhav-madhusoodanan: You're right, they aren't needed

.github/workflows/main.yml
madhav-madhusoodanan: I think the intention of the `build_std` variable is to mention whether `std` for a target should be built (true) or not (false).

I think it might be better to set `build_std` for each target to either true or false, and then just keep it as `matrix.build_std` so that we future-proof its meaning.

What do you think?

folkertdev: I think you still need to handle the case of it not being specified somehow.

.github/workflows/main.yml
folkertdev: basically, yes

.github/workflows/main.yml
```diff
@@ -248,6 +248,65 @@ jobs:
...
+
+        include:
+        - target:
+            tuple: aarch64-unknown-linux-gnu
+            os: ubuntu-latest
+
+        - target:
+            tuple: aarch64_be-unknown-linux-gnu
+            os: ubuntu-latest
+          build_std: true
```
folkertdev: I think the indentation like 

```yml
        include:
          - target: "x86_64-unknown-linux-gnu"
            rust: stable
            os: ubuntu-latest
            codecov: true
```

works better? Right now no tests actually run.

madhav-madhusoodanan: I think indentation is needed here because rustup needs to exclude adding the targets (and build std) if `build_std` is true.

In the steps, `matrix.build_std` is being checked (hence the indentation being on the same level as `target`), while`target` and `os` are being referred to by `matrix.target.tuple` and `matrix.target.os` respectively.

Wouldn't it break this logic then? 

Actually, now I have a question about whether `build_std` needs to be under `target`.

folkertdev: Let me play with it a bit and I'll try to force-push something

madhav-madhusoodanan: Feel free to!

folkertdev: Allright, I makde my changes, which seem to now work. Do the changes seem correct to you, or did I remove something important?

madhav-madhusoodanan: This is perfect, taking `ubuntu-latest` out of the matrix.
Looks good to me.

I'm just concerned about `build_std` being on the same level as `target` though.

folkertdev: it's not, it is just metadata for a particular target.

PR #1946, opened 2025-10-25
`intrinsic-test`: dedup constants

folkertdev:
deduplicates the input constants and places them in the global scope. For now the inputs are still emitted once per lib file, we could decide to lift them to some shared location eventually.

I've also tried to make the formatting a bit better. It's not perfect, it doesn't need to be, but at least the emitted rust code is a bit more readable now.


sayantn:
@folkertdev are you done with this? Lgtm

folkertdev:
Yes it's done

PR #1948, opened 2025-10-26
Forward `TEST_SAMPLE_INTRINSICS_PERCENTAGE`

folkertdev:
This is useful when iterating on `intrinsic-test` so you don't need to wait for all intrinsics to get tested.

I'm also enabling more logs, that seems useful to just kind of see what it's doing, and less of a problem now that the intrinsic tests are in their own CI actions.

### DIFF
--- a/crates/intrinsic-test/src/x86/mod.rs
+++ b/crates/intrinsic-test/src/x86/mod.rs
@@ -47,7 +47,7 @@ impl SupportedArchitectureTest for X86ArchitectureTest {
             get_xml_intrinsics(&cli_options.filename).expect("Error parsing input file");
 
         intrinsics.sort_by(|a, b| a.name.cmp(&b.name));
-        intrinsics.dedup();
+        intrinsics.dedup_by(|a, b| a.name == b.name);
 
         let sample_percentage: usize = cli_options.sample_percentage as usize;
         let sample_size = (intrinsics.len() * sample_percentage) / 100;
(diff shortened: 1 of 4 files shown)


crates/intrinsic-test/src/x86/mod.rs
```diff
@@ -44,12 +43,16 @@ impl SupportedArchitectureTest for X86ArchitectureTest {
...
             get_xml_intrinsics(&cli_options.filename).expect("Error parsing input file");
 
+        intrinsics.sort_by(|a, b| a.name.cmp(&b.name));
+        intrinsics.dedup();
+
         let sample_percentage: usize = cli_options.sample_percentage as usize;
+        let sample_size = (intrinsics.len() * sample_percentage) / 100;
 
-        let mut intrinsics = intrinsics
+        let intrinsics = intrinsics
```
folkertdev, line 55: @madhav-madhusoodanan does this look right to you? CI failed on this before complaining it cannot find `mod_0.o` which I think can only happen if we somehow filter out all intrinsics.

folkertdev: Ah, got it. We have duplicate definitions of a couple of intrinsics. e.g.

```
	<intrinsic name="_mm256_dpbusds_epi32" tech="AVX_ALL">
	<intrinsic name="_mm256_dpbusds_epi32" tech="AVX-512">
```

So we really need to dedup by the name.

sayantn: this is an intel weirdness, these functions can be called with either avxvnni or avx512vnni + avx512vl (similar to the, now revoked, avx10.1-256 proposal). C handles them by just using macros, and we just don't support the avx variants (there's `_mm256_dpbusds_avx_epi32` that does the exact same thing, so we support just that)

PR #1949, opened 2025-10-28
Amends CI and tests for arm intrinsics to enable tests in debug mode

CrooseGit:
Previously, the CI has only been running the tests for the arm intrinsic in release mode.
When the CI is modified to fix this, it is discovered that the tests do not build in debug for arm targets.

This PR amends the CI, test tool, and the intrinsics themselves to address the above issues.

### DIFF
--- a/crates/core_arch/src/arm_shared/neon/generated.rs
+++ b/crates/core_arch/src/arm_shared/neon/generated.rs
@@ -9183,6 +9183,7 @@ pub fn vcvtq_u32_f32(a: float32x4_t) -> uint32x4_t {
 #[doc = "Dot product arithmetic (indexed)"]
 #[doc = "[Arm's documentation](https://developer.arm.com/architectures/instruction-sets/intrinsics/vdot_lane_s32)"]
 #[inline]
+#[cfg(target_endian = "little")]
 #[cfg_attr(target_arch = "arm", target_feature(enable = "v8"))]
 #[target_feature(enable = "neon,dotprod")]
 #[cfg_attr(all(test, target_arch = "arm"), assert_instr(vsdot, LANE = 0))]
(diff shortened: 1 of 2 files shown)


ci/intrinsic-test.sh
```diff
@@ -83,22 +83,24 @@ esac
...
         CPPFLAGS="${TEST_CPPFLAGS}" RUSTFLAGS="${HOST_RUSTFLAGS}" RUST_LOG=warn \
-            cargo run "${INTRINSIC_TEST}" "${PROFILE}" \
+            cargo run "${INTRINSIC_TEST}" --release  \
             --bin intrinsic-test -- intrinsics_data/arm_intrinsics.json \
             --runner "${TEST_RUNNER}" \
             --cppcompiler "${TEST_CXX_COMPILER}" \
             --skip "${TEST_SKIP_INTRINSICS}" \
-            --target "${TARGET}"
+            --target "${TARGET}" \
+            --profile "${PROFILE}"
```
folkertdev: Combining both `--release` and `--profile` looks suspicious. What does that do?

crates/intrinsic-test/src/common/compare.rs
```diff
@@ -20,8 +25,13 @@ pub fn compare_outputs(intrinsic_name_list: &Vec<String>, runner: &str, target:
                 .current_dir("c_programs")
                 .output();
 
+            let profile_dir = match profile {
+                "dev" => "debug",
+                _ => "release",
+            };
```
folkertdev, line 31: I think we should explicitly match `release` here and just panic on anything else to make future debugging easier.

CrooseGit: That makes sense, will do.
Thanks

ci/intrinsic-test.sh
CrooseGit: Fair point, it not obvious.
It runs the intrinsic test tool itself in release mode, and then passes the `$PROFILE` env-var as an argument to the tool. The tool then compiles and runs a series of rust test programs with the specified profile, comparing the outputs against the equivalent C programs.
I could add a comment explaining this?
Thanks

CrooseGit:
@folkertdev 
> Can you give some background on why the tests did not pass in debug mode? I'd also personally like to merge the changes to the intrinsics separately from changes to the CI infrastructure

The offending intrinsics were passing arguments to the linked llvm intrinsics that were then failing LLVMs internal assertions. LLVM requires certain arguments to be const for some of its intrinsics, and in debug mode these arguments were not const. However, in release due to things being inlined and / or being optimised away the relevant arguments became const anyway. 
RE separate PRs, that can be arranged. Currently I am trying to fix the big endian CI failures, but once that is sorted, then will do.
Thanks

CrooseGit:
@folkertdev 
Now I have fixed the ARM intrinsic issues in debug, I have split the PR like you requested.
The CI changes must be merged after the intrinsic changes otherwise the tests will just fail, so here here is part one of the split PRs https://github.com/rust-lang/stdarch/pull/1964

CrooseGit:
Split in two:
- https://github.com/rust-lang/stdarch/pull/1964
- https://github.com/rust-lang/stdarch/pull/1967

PR #1951, opened 2025-11-02
improve `_mm256_permute2f128` tests

folkertdev:
This came up in https://github.com/rust-lang/rust/pull/148402, the zeroing behavior was untested, and the old inputs had repeating patterns that made it harder to figure out how the output is based on the input.


RalfJung:
This makes these tests exactly the same as the Miri tests with your changes in https://github.com/rust-lang/rust/pull/148402, right?

folkertdev:
Yes, exactly.

PR #1952, opened 2025-11-04
`intrinsic-test`: Coalescing all intrinsic tests into a single run of the test-file binaries

madhav-madhusoodanan:


madhav-madhusoodanan:
@folkertdev I was able to run 30% of the intrinsics and get a 17-minute CI runtime.
I'll be exploring some optimizations on how to reduce it further.

ci/intrinsic-test.sh
```diff
@@ -75,7 +75,7 @@ case ${TARGET} in
         TEST_CXX_COMPILER="clang++"
         TEST_RUNNER="${CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER}"
         TEST_SKIP_INTRINSICS=crates/intrinsic-test/missing_x86.txt
-        : "${TEST_SAMPLE_INTRINSICS_PERCENTAGE:=5}"
+        : "${TEST_SAMPLE_INTRINSICS_PERCENTAGE:=30}"
```
folkertdev: please adjust this so that it is not the longest CI job. I think the next-longest job is 11 minutes now.

crates/intrinsic-test/src/common/compare.rs
```diff
@@ -11,85 +11,103 @@ fn runner_command(runner: &str) -> Command {
...
-            let c = std::str::from_utf8(&c.stdout)
-                .unwrap()
-                .to_lowercase()
-                .replace("-nan", "nan");
-            let rust = std::str::from_utf8(&rust.stdout)
-                .unwrap()
-                .to_lowercase()
-                .replace("-nan", "nan");
+    let c_output_map = c
+        .split("############")
```
folkertdev: can you put this marker into a `const` and use that instead of repeating the literal characters?

crates/intrinsic-test/src/common/compare.rs
```diff
@@ -11,85 +11,103 @@ fn runner_command(runner: &str) -> Command {
...
-            let (c, rust) = match (c, rust) {
-                (Ok(c), Ok(rust)) => (c, rust),
-                a => panic!("{a:#?}"),
-            };
+                .output()
+        },
+    );
+    let (c, rust) = match (c, rust) {
+        (Ok(c), Ok(rust)) => (c, rust),
+        a => panic!("{a:#?}"),
```
folkertdev: let's call this "failure" or something a bit more descriptive

folkertdev (commented):
Very encouraging results!

ci/intrinsic-test.sh
madhav-madhusoodanan: Sure, let me dial it down to 20%.

madhav-madhusoodanan:
The longest process is the C++/Rust binary runs (which run on a single thread each, in parallel)
This takes about 7 mins.

I think the next optimization would be parallelizing the binary runs themselves.

folkertdev:
@madhav-madhusoodanan is this done (it needs a rebase)?

madhav-madhusoodanan:
Apologies for the delay @folkertdev.
I still am implementing the binary run parallelization, which I shall do in another PR perhaps.

For now let's merge this once I rebase

PR #1953, opened 2025-11-05
Use generic SIMD masked load/stores for avx512 masked load/stores

sayantn:
Due to recently merged rust-lang/rust#147355, we can now use `simd_masked_load/store` for masked avx512 load/stores. This reduces the number of used LLVM intrinsics a lot.

also removes use of LLVM intrinsics for AVX maskload/maskstore intrinsics. cc @RalfJung now these can also be removed from Miri


sayantn:
seems like it's not in nightly yet, making it draft for now

PR #1955, opened 2025-11-10
Use SIMD intrinsics for vector shifts

sayantn:
Retrying rust-lang/stdarch#1928, this time actually checking for overflow. LLVM sees through this, and eliminates the branch/selects https://godbolt.org/z/WsYvfjPas

cc @RalfJung this also removes uses of a few intrinsics from avx2

### DIFF
--- a/crates/core_arch/src/x86/avx2.rs
+++ b/crates/core_arch/src/x86/avx2.rs
@@ -2786,7 +2786,12 @@ pub fn _mm256_bslli_epi128<const IMM8: i32>(a: __m256i) -> __m256i {
 #[cfg_attr(test, assert_instr(vpsllvd))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub fn _mm_sllv_epi32(a: __m128i, count: __m128i) -> __m128i {
-    unsafe { transmute(psllvd(a.as_i32x4(), count.as_i32x4())) }
+    unsafe {
+        let count = count.as_u32x4();
+        let no_overflow: u32x4 = simd_lt(count, u32x4::splat(u32::BITS));
+        let count = simd_select(no_overflow, count, u32x4::ZERO);
+        simd_select(no_overflow, simd_shl(a.as_u32x4(), count), u32x4::ZERO).as_m128i()
+    }
 }
 
 /// Shifts packed 32-bit integers in `a` left by the amount
@@ -2799,7 +2804,12 @@ pub fn _mm_sllv_epi32(a: __m128i, count: __m128i) -> __m128i {
 #[cfg_attr(test, assert_instr(vpsllvd))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub fn _mm256_sllv_epi32(a: __m256i, count: __m256i) -> __m256i {
-    unsafe { transmute(psllvd256(a.as_i32x8(), count.as_i32x8())) }
+    unsafe {
+        let count = count.as_u32x8();
+        let no_overflow: u32x8 = simd_lt(count, u32x8::splat(u32::BITS));
+        let count = simd_select(no_overflow, count, u32x8::ZERO);
+        simd_select(no_overflow, simd_shl(a.as_u32x8(), count), u32x8::ZERO).as_m256i()
+    }
 }
 
 /// Shifts packed 64-bit integers in `a` left by the amount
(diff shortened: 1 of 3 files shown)


crates/core_arch/src/x86/avx2.rs
```diff
@@ -2893,7 +2913,11 @@ pub fn _mm_srav_epi32(a: __m128i, count: __m128i) -> __m128i {
 #[cfg_attr(test, assert_instr(vpsravd))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub fn _mm256_srav_epi32(a: __m256i, count: __m256i) -> __m256i {
-    unsafe { transmute(psravd256(a.as_i32x8(), count.as_i32x8())) }
+    unsafe {
+        let count = count.as_i32x8();
+        let overflow: i32x8 = simd_ge(count, i32x8::splat(32));
+        simd_select(overflow, i32x8::ZERO, simd_shr(a.as_i32x8(), count)).as_m256i()
+    }
 }
```
eduardosm, line 2935: I don't think this is correct, since this is an arithmetic shift and overflow should result in all-zeros or all-ones depending on the sign bit.

eduardosm:
`simd_shl` and `simd_shr` are documented to be UB on overflow, but the implementations of this PR will unconditionally call these intrinsics.

crates/core_arch/src/x86/avx2.rs
sayantn: Thanks a lot for pointing it out, yes `srav` intrinsics should fill sign-bit on overflow. I would modify it

sayantn:
Ah, so I need to add a `simd_and` to make it safe. Thanks for pointing it out

sayantn:
It looks like adding that one `simd_and` messes up the whole LLVM optimization. Strictly speaking, this is *fine* as long as we are only in LLVM backend, LLVM generates `poison` values if the shift value overflows and we are filtering them out in the next step. But obviously that's not a good argument, as the main motive for this change is to make it more portable across backends. I will look into what needs to be done

edit: https://godbolt.org/z/99zrnb5Gj

edit2: somehow using `simd_select(good, count, T::splat(dummy_value))` works. It lets LLVM collapse the 2 `select`s, while also keeping the `simd_shl` call safe from Rust (https://godbolt.org/z/cc4EKfeb7)

sayantn:
https://godbolt.org/z/cafKWsf6x I think these should do the job, @eduardosm could you check please?

edit: seems to pass miri on overflow [playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=ef495a3ecafdd6d8bb7335a68f35c8d2)

crates/core_arch/src/x86/avx2.rs
```diff
@@ -2778,7 +2778,12 @@ pub fn _mm256_bslli_epi128<const IMM8: i32>(a: __m256i) -> __m256i {
 #[cfg_attr(test, assert_instr(vpsllvd))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub fn _mm_sllv_epi32(a: __m128i, count: __m128i) -> __m128i {
-    unsafe { transmute(psllvd(a.as_i32x4(), count.as_i32x4())) }
+    unsafe {
+        let count = count.as_u32x4();
+        let good: u32x4 = simd_lt(count, u32x4::splat(32));
```
folkertdev: maybe 

```suggestion
        let good: u32x4 = simd_lt(count, u32x4::splat(u32::BITS));
```

to make it really clear where that number comes from?

crates/core_arch/src/x86/avx2.rs
```diff
@@ -2791,7 +2796,12 @@ pub fn _mm_sllv_epi32(a: __m128i, count: __m128i) -> __m128i {
 #[cfg_attr(test, assert_instr(vpsllvd))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
 pub fn _mm256_sllv_epi32(a: __m256i, count: __m256i) -> __m256i {
-    unsafe { transmute(psllvd256(a.as_i32x8(), count.as_i32x8())) }
+    unsafe {
+        let count = count.as_u32x8();
+        let good: u32x8 = simd_lt(count, u32x8::splat(32));
```
folkertdev: also maybe 
```suggestion
        let no_overflow: u32x8 = simd_lt(count, u32x8::splat(32));
```

eduardosm:
> https://godbolt.org/z/cafKWsf6x I think these should do the job, @eduardosm could you check please?

That looks good to me 👍🏻

sayantn:
Thanks @eduardosm :love:, then it should be good to merge

RalfJung:
I think this may have to be reverted, it doesn't codegen the way these intrinsics should: https://github.com/rust-lang/rust/issues/159831

PR #1958, opened 2025-11-11
add logic tests for ternarylogic

folkertdev:
previously the output would just always be all zeroes.

this came up in https://github.com/rust-lang/miri/pull/4678

r? @sayantn


sayantn:
I would LOVE if we can test all 256 of the choices, but ig that's a job for `intrinsic-test`. thanks for the PR

folkertdev:
Ralf did have some further testing suggestions, I'll add those to a follow-up then. But yeah exhaustive is overkill here

PR #1959, opened 2025-11-11
improve ternary logic tests

folkertdev:
follow-up to https://github.com/rust-lang/stdarch/pull/1958

r? @sayantn when Ralf approves


folkertdev:
@RalfJung does this look good to you, if so I'll copy the tests over. Here we actually get CI, though I also did run these tests on actual hardware.

RalfJung:
Looks great, thanks. :)

PR #1960, opened 2025-11-14
aarch64: Remove withdrawn FEAT_TME

maurer:
ARM has withdrawn FEAT_TME
https://developer.arm.com/documentation/102105/lb-05/

LLVM has also dropped support for enabling the feature.

llvm/llvm-project#167687


folkertdev:
Given that LLVM drops support I think our only option is to remove these unstable intrinsics.

cc our arm maintainers @JamieCunliffe @adamgemmell @Jamesbarford if you have thoughts.

adamgemmell:
> Given that LLVM drops support I think our only option is to remove these unstable intrinsics.

I agree, looks like they're to be completely removed from the [ACLE](https://github.com/ARM-software/acle/pull/422) too. If it turns out someone was making use of them that can be a future conversation.

PR #1961, opened 2025-11-17
correct signedness of pmadd arguments

folkertdev:


sayantn:
It doesn't really matter when linking to LLVM intrinsics (LLVM only had `i8`, the signedness is determined by the operation). But we can do this just for the sake of our understanding. Will it be possible for you to check for other such inconsistencies? I believe there should be a lot more of them

folkertdev:
These I just found looking at https://github.com/rust-lang/miri/pull/4705, I don't currently want to spend the time to look for more cases really.

sayantn:
Ok np, should I merge this then?

folkertdev:
Yes, if you think it looks good.

PR #1962, opened 2025-11-20
Use declarative attribute macros instead of procedural macros for better compile times

sayantn:
Currently, we use 2 procedural macros, `assert_instr` and `simd_test`, for testing. It is convenient, but the problem is that it significantly slows down compilation. Seeing that these macros are pretty small, I thought of using declarative attribute macros (rust-lang/rust#143547) instead

The diffcount is huge due to all uses of `simd_test` being changed, like

```
#[simd_test(enable = "avx,gfni")]
```
got changed to
```
#[simd_test("avx", "gfni")]
```

The reason being that we cannot do string splitting etc in declarative macros

### DIFF
--- a/crates/stdarch-test/src/simd_test.rs
+++ b/crates/stdarch-test/src/simd_test.rs
@@ -36,7 +36,7 @@ macro_rules! simd_test {
     attr($($feat:tt),*) ($( #[$meta:meta] )* $(unsafe)? fn $name:ident () $body:block) => {
         #[allow(non_snake_case)]
         #[test]
-        $( #[$meta] )* 
+        $( #[$meta] )*
         fn $name () {
             let mut missing_features = ::std::vec::Vec::new();
             $(


bjorn3:
> The reason being that we cannot do string splitting etc in declarative macros

Wouldn't a split at runtime work fine here?

sayantn:
No, we need to pass the target features to the `std::arch` detection macros, and they only accept literals

sayantn:
for some reason this is actually a little bit slower on my pc. Does anyone know what's going on? I thought proc-macros are supposed to be slower than decl macros in general

joshtriplett:
@sayantn I'm interested in following upon this. What process did you follow to measure it? Did you count the time to compile the proc macro itself?

sayantn:
Essentially I just ran `cargo clean` and then `time cargo test -p core_arch` before and after the change, it turned out that the time taken for this is a bit higher. Also, the CI time is a bit higher too

PR #1964, opened 2025-11-27
Fixes issue with arm intrinsics not compiling in dev profile

CrooseGit:
- This patch begins by fixing build failures that occurred due to failing LLVM assertions when building arm neon intrinsics in debug.
- It then addresses some differences the intrinsic tests flagged between C and rust on big endian aarch64 caused by an issue in in both [clang](https://github.com/llvm/llvm-project/issues/166190) and rust.
- Finally it reworks some of the changes from previous step as they were causing instr assert test failures on armv7 in release due to an [llvm bug](https://github.com/llvm/llvm-project/pull/169337).

This is part 1/2 of this [PR](https://github.com/rust-lang/stdarch/pull/1949) which was asked to be split in two.


folkertdev (approved):
Nice, thanks!

PR #1965, opened 2025-11-27
s390x: use `simd_shuffle!` macro

folkertdev:


crates/core_arch/src/s390x/vector.rs
```diff
@@ -5307,11 +5307,13 @@ pub unsafe fn vec_search_string_until_zero_cc<T: sealed::VectorSearchString>(
 #[inline]
 #[target_feature(enable = "vector-enhancements-1")]
 #[unstable(feature = "stdarch_s390x", issue = "135681")]
-// FIXME: this emits `vflls` where `vldeb` is expected
-// #[cfg_attr(all(test, target_feature = "vector-enhancements-1"), assert_instr(vldeb))]
+// NOTE: `vflls` and `vldeb` are equivalent; our disassmbler prefers vflls.
+#[cfg_attr(
+    all(test, target_feature = "vector-enhancements-1"),
+    assert_instr(vflls)
+)]
```
folkertdev, line 5314: @uweigand 

just checking, is that correct? this implementation produces `vldeb` on godbolt, so I really think it's the disassembler that is picking `vflls` here.

crates/core_arch/src/s390x/vector.rs
```diff
@@ -5322,11 +5324,7 @@ pub unsafe fn vec_doublee(a: vector_float) -> vector_double {
 // #[cfg_attr(all(test, target_feature = "vector-enhancements-1"), assert_instr(vledb))]
 pub unsafe fn vec_floate(a: vector_double) -> vector_float {
     let truncated: f32x2 = simd_as(a);
-    simd_shuffle(
-        truncated,
-        truncated,
-        const { u32x4::from_array([0, 0, 1, 1]) },
-    )
+    simd_shuffle!(truncated, truncated, [0, 0, 1, 1])
```
folkertdev, line 5327: and here, `vec_floate` does not specify what happens to the odd elements. clang + llvm end up using poison values, but rust does not have those. So, is there anything that can be done?

this is not all that important of course, but it's a bit of a wart.

crates/core_arch/src/s390x/vector.rs
uweigand: Yes, the two are equivalent; in fact the binary machine code is identical, it's just different assembler mnemonics for the same instruction.  Both `vldeb v1, v2` and `vflls v1, v2` are simply extended mnemonics for the "base" instruction `vfll v1, v2, 0, 0`  (see the PoP chapter 24 under VECTOR FP LOAD LENGTHENED).  It's a bit unfortunate to have two mnemonics for the same thing, but it's due to historical reasons (`vflls` is the more recent one, and probably should be preferred).

crates/core_arch/src/s390x/vector.rs
uweigand: Yes, it's deliberately left unspecified what happens to the odd elements.   This matches the behavior of the underlying VECTOR FP LOAD ROUNDED instruction, which also states: "The data in the odd elements of the first operand is unpredictable."  You could mask it to some defined value, but at extra runtime cost of course.

folkertdev: Right, well there is nothing we can do then. There is always inline assembly if someone really needs to emit the exact instruction.

PR #1968, opened 2025-11-28
Format `f16` values correctly in intrinsic-test

sayantn:
Fix some x86 intrinsics too

 - `_mm_{u}comineq_sh` (these should be unordered, i.e. should return true if either operand is nan)
 - `_mm_mask_cvt{epi16_epi8, pd_ps, pd_epi32}` (top 64 bits should be 0)

### DIFF
--- a/crates/intrinsic-test/missing_x86.txt
+++ b/crates/intrinsic-test/missing_x86.txt
@@ -51,29 +51,17 @@ _mm_sm3rnds2_epi32
 # SDE ERROR: Cannot execute XGETBV with ECX != 0
 _xgetbv
 
-# Miscellaneous issues that can be fixed first
-_kshiftli_mask16
-_kshiftli_mask32
-_kshiftli_mask64
-_kshiftli_mask8
-_kshiftri_mask16
-_kshiftri_mask32
-_kshiftri_mask64
-_kshiftri_mask8
+# top bits are undefined, unclear how to test these
 _mm256_castsi128_si256
-_mm256_extract_epi16
-_mm256_extract_epi8
 _mm512_castsi128_si512
 _mm512_castsi256_si512
-# _mm512_conj_pch
+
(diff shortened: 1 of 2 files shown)


sayantn:
r? @folkertdev 

Just lmk before merging, will need to remove the last commit

folkertdev:
cc @tgross35 was there some caveat to this approach?

tgross35:
Not sure about the rest but just regarding formatting, these were added at a time that we didn't have `Debug` for f16

folkertdev:
Right just double-checking that there is not some edge case where we'd get incorrect outputs by upcasting to `f32`.

crates/intrinsic-test/missing_x86.txt
```diff
@@ -594,311 +23,45 @@ _rotwl
...
-_kshiftri_mask8
+# top bits are undefined, unclear how to test these
 _mm256_castsi128_si256
-_mm256_extract_epi16
-_mm256_extract_epi8
 _mm512_castsi128_si512
 _mm512_castsi256_si512
-# _mm512_conj_pch
+
+# Clang bug
```
folkertdev, line 59: Do we document what clang bug(s) these are?

sayantn: yeah I was thinking about putting some clang PRs to correct these. in summary,

 - `_mm{256}_extract_epi{8,16}` does `(int) (unsigned {char,short})`, but afaik C does sign-extension whenever the *target type* is signed, so this is getting sign-extended, whereas Intel specifies that it should zero-extend.
 - `_mm512_mask_reduce_{max,min}_{ps,pd}` uses `{f32,f64}::{MIN,MAX}` in place of masked-out values, but Intel specifies it should use `{f32,f64}::{NEG_INFINITY,INFINITY}`

PR #1973, opened 2025-12-09
thread, grid, and block dim/idx can only return non-negative values

ZuseZ4:
Based on the previous discussion in
https://rust-lang.zulipchat.com/#narrow/channel/422870-t-compiler.2Fgpgpu-backend/topic/Return.20type.20of.20NVPTX.20index.20and.20dimension.20intrinsics

Also matching amdgpu: https://doc.rust-lang.org/nightly/core/arch/amdgpu/fn.workitem_id_x.html

r? @workingjubilee


ZuseZ4:
ping @workingjubilee

sayantn:
As @workingjubilee was fine with the changes, I think it is fine to merge this

PR #1974, opened 2025-12-10
Mark the neon intrinsics as inline(always).

JamieCunliffe:
Mark the neon intrinics as `inline(always)` now that we can apply the attribute at the call site and perform the checks to ensure that inlining would be correct. See tracking issue rust-lang/rust#145574 for more info.

This will trigger some warnings on code that is using intrinsics without having enabled the correct target feature in the function they are being called from.


sayantn:
CI seems to be failing for some reason, I don't see why rustc is complaining that the lint doesn't exist - that pr was merged 3 months ago! 😕 

edit: it might be because the rustc PR didn't add the lint to the `HardwiredLints` section in `declare_lint_pass!`

edit: just noticed that the PR was merged 3 days ago, retrying CI

edit: seems like we found an ICE

JamieCunliffe:
The ICE was fixed in rust-lang/rust#150805 and the CI now passes.

sayantn:
@JamieCunliffe thanks, the only concerning thing is that we are using `inline_always_missing_target_features` indiscriminately, imo that kind of defeats the purpose of the feature. Could you confirm that it is still needed even after #1987 is merged?

JamieCunliffe:
Given that enables fp16 globally for the armv7 targets it should be OK with those, it also failed on some other targets. I see there have been some other changes regarding fp16 target features. I'll do a push without that for a CI run.

sayantn:
Looks like the CI passes. LGTM, let's get this merged

folkertdev:
Are we sure we want to expose this to users? The inlining behavior is not conditional on the feature being enabled, the tracking issue https://github.com/rust-lang/rust/issues/145574 shows that inlining may be unsound. So, maybe we should revert this for now?

PR #1976, opened 2025-12-14
Add amdgpu intrinsics

Flakebi:
Add intrinsics for the amdgpu architecture.

I’m not sure how to add/run CI (`ci/run.sh` fails for me e.g. for
nvptx because `core` cannot be found), but I checked that it compiles
without warnings with
```bash
$ CARGO_BUILD_RUSTFLAGS=-Ctarget-cpu=gfx900 cargo check --target amdgcn-amd-amdhsa -Zbuild-std=core
```

Tracking issue: https://github.com/rust-lang/rust/issues/149988

### DIFF
--- a/ci/dox.sh
+++ b/ci/dox.sh
@@ -15,6 +15,15 @@ dox() {
 
   cargo clean --target "${1}"
 
+  if [ "${1}" == "amdgcn-amd-amdhsa" ]; then
+    if [ "$CI" != "" ]; then
+      rustup component add rust-src
+    fi
+    export CARGO_UNSTABLE_BUILD_STD=core
+    # amdgpu needs a target-cpu, any is fine
+    export RUSTFLAGS="${RUSTFLAGS} -Ctarget-cpu=gfx900"
+  fi
+
   cargo build --verbose --target "${1}" --manifest-path crates/core_arch/Cargo.toml
   cargo doc --verbose --target "${1}" --manifest-path crates/core_arch/Cargo.toml
 }
@@ -33,6 +42,7 @@ if [ -z "$1" ]; then
   #dox mips64-unknown-linux-gnuabi64
   dox wasm32-unknown-unknown
   dox nvptx64-nvidia-cuda
+  dox amdgcn-amd-amdhsa
 else
   dox "${1}"
 fi
--- a/crates/core_arch/src/amdgpu/mod.rs
+++ b/crates/core_arch/src/amdgpu/mod.rs
@@ -1,0 +1,1053 @@
+//! amdgpu intrinsics
+//!
+//! The reference is the [LLVM amdgpu guide] and the [LLVM implementation].
+//! The order of intrinsics here follows the order in the [LLVM implementation].
+//!
+//! [LLVM amdgpu guide]: https://llvm.org/docs/AMDGPUUsage.html#llvm-ir-intrinsics
+//! [LLVM implementation]: https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/IR/IntrinsicsAMDGPU.td
+
+#[allow(improper_ctypes)]
+unsafe extern "unadjusted" {
+    #[link_name = "llvm.amdgcn.workitem.id.x"]
+    safe fn llvm_workitem_id_x() -> u32;
+    #[link_name = "llvm.amdgcn.workitem.id.y"]
+    safe fn llvm_workitem_id_y() -> u32;
+    #[link_name = "llvm.amdgcn.workitem.id.z"]
+    safe fn llvm_workitem_id_z() -> u32;
+
+    #[link_name = "llvm.amdgcn.workgroup.id.x"]
+    safe fn llvm_workgroup_id_x() -> u32;
+    #[link_name = "llvm.amdgcn.workgroup.id.y"]
+    safe fn llvm_workgroup_id_y() -> u32;
+    #[link_name = "llvm.amdgcn.workgroup.id.z"]
+    safe fn llvm_workgroup_id_z() -> u32;
+
+    #[link_name = "llvm.amdgcn.groupstaticsize"]
+    safe fn llvm_groupstaticsize() -> u32;
+    #[link_name = "llvm.amdgcn.dispatch.id"]
+    safe fn llvm_dispatch_id() -> u64;
+
+    #[link_name = "llvm.amdgcn.wavefrontsize"]
+    safe fn llvm_wavefrontsize() -> u32;
+
+    #[link_name = "llvm.amdgcn.s.barrier"]
+    safe fn llvm_s_barrier();
+    #[link_name = "llvm.amdgcn.s.barrier.signal"]
+    fn llvm_s_barrier_signal(barrier_type: i32);
+    #[link_name = "llvm.amdgcn.s.barrier.signal.isfirst"]
+    fn llvm_s_barrier_signal_isfirst(barrier_type: i32) -> bool;
+    #[link_name = "llvm.amdgcn.s.barrier.wait"]
+    fn llvm_s_barrier_wait(barrier_type: i16);
+    #[link_name = "llvm.amdgcn.s.get.barrier.state"]
+    fn llvm_s_get_barrier_state(barrier_type: i32) -> u32;
+    #[link_name = "llvm.amdgcn.wave.barrier"]
+    safe fn llvm_wave_barrier();
+    #[link_name = "llvm.amdgcn.sched.barrier"]
+    fn llvm_sched_barrier(mask: u32);
+    #[link_name = "llvm.amdgcn.sched.group.barrier"]
+    fn llvm_sched_group_barrier(mask: u32, size: u32, sync_id: u32);
+
+    #[link_name = "llvm.amdgcn.s.sleep"]
+    safe fn llvm_s_sleep(count: u32);
+
+    #[link_name = "llvm.amdgcn.s.sethalt"]
+    safe fn llvm_s_sethalt(value: u32) -> !;
+
+    #[link_name = "llvm.amdgcn.s.getpc"]
+    safe fn llvm_s_getpc() -> i64;
+
+    #[link_name = "llvm.amdgcn.mbcnt.lo"]
+    safe fn llvm_mbcnt_lo(value: u32, init: u32) -> u32;
+    #[link_name = "llvm.amdgcn.mbcnt.hi"]
+    safe fn llvm_mbcnt_hi(value: u32, init: u32) -> u32;
+
+    #[link_name = "llvm.amdgcn.ballot"]
+    safe fn llvm_ballot(b: bool) -> u64;
+
+    #[link_name = "llvm.amdgcn.inverse.ballot"]
+    safe fn llvm_inverse_ballot(value: u64) -> bool;
+
+    #[link_name = "llvm.amdgcn.wave.reduce.umin"]
+    safe fn llvm_wave_reduce_umin(value: u32, strategy: u32) -> u32;
+    #[link_name = "llvm.amdgcn.wave.reduce.min"]
+    safe fn llvm_wave_reduce_min(value: i32, strategy: u32) -> i32;
+    #[link_name = "llvm.amdgcn.wave.reduce.umax"]
+    safe fn llvm_wave_reduce_umax(value: u32, strategy: u32) -> u32;
+    #[link_name = "llvm.amdgcn.wave.reduce.max"]
+    safe fn llvm_wave_reduce_max(value: i32, strategy: u32) -> i32;
+    #[link_name = "llvm.amdgcn.wave.reduce.add"]
+    safe fn llvm_wave_reduce_add(value: u32, strategy: u32) -> u32;
+    #[link_name = "llvm.amdgcn.wave.reduce.and"]
+    safe fn llvm_wave_reduce_and(value: u32, strategy: u32) -> u32;
+    #[link_name = "llvm.amdgcn.wave.reduce.or"]
+    safe fn llvm_wave_reduce_or(value: u32, strategy: u32) -> u32;
+    #[link_name = "llvm.amdgcn.wave.reduce.xor"]
+    safe fn llvm_wave_reduce_xor(value: u32, strategy: u32) -> u32;
+
+    // The following intrinsics can have multiple sizes
+
+    #[link_name = "llvm.amdgcn.readfirstlane.i32"]
+    safe fn llvm_readfirstlane_u32(value: u32) -> u32;
+    #[link_name = "llvm.amdgcn.readfirstlane.i64"]
+    safe fn llvm_readfirstlane_u64(value: u64) -> u64;
+    #[link_name = "llvm.amdgcn.readlane.i32"]
+    fn llvm_readlane_u32(value: u32, lane: u32) -> u32;
+    #[link_name = "llvm.amdgcn.readlane.i64"]
+    fn llvm_readlane_u64(value: u64, lane: u32) -> u64;
+    #[link_name = "llvm.amdgcn.writelane.i32"]
+    fn llvm_writelane_u32(value: u32, lane: u32, default: u32) -> u32;
+    #[link_name = "llvm.amdgcn.writelane.i64"]
+    fn llvm_writelane_u64(value: u64, lane: u32, default: u64) -> u64;
+
+    #[link_name = "llvm.amdgcn.endpgm"]
+    safe fn llvm_endpgm() -> !;
+
+    #[link_name = "llvm.amdgcn.update.dpp.i32"]
+    fn llvm_update_dpp(
+        old: u32,
+        src: u32,
+        dpp_ctrl: u32,
+        row_mask: u32,
+        bank_mask: u32,
+        bound_control: bool,
+    ) -> u32;
+
+    #[link_name = "llvm.amdgcn.s.memrealtime"]
+    safe fn llvm_s_memrealtime() -> u64;
+
+    #[link_name = "llvm.amdgcn.ds.permute"]
+    fn llvm_ds_permute(lane: u32, value: u32) -> u32;
+    #[link_name = "llvm.amdgcn.ds.bpermute"]
+    fn llvm_ds_bpermute(lane: u32, value: u32) -> u32;
+    #[link_name = "llvm.amdgcn.perm"]
+    fn llvm_perm(src0: u32, src1: u32, selector: u32) -> u32;
+
+    // gfx10
+    #[link_name = "llvm.amdgcn.permlane16.i32"]
+    fn llvm_permlane16_u32(
+        old: u32,
+        src0: u32,
+        src1: u32,
+        src2: u32,
+        fi: bool,
+        bound_control: bool,
+    ) -> u32;
+
+    // gfx10
+    #[link_name = "llvm.amdgcn.permlanex16.i32"]
+    fn llvm_permlanex16_u32(
+        old: u32,
+        src0: u32,
+        src1: u32,
+        src2: u32,
+        fi: bool,
+        bound_control: bool,
+    ) -> u32;
+
+    #[link_name = "llvm.amdgcn.s.get.waveid.in.workgroup"]
+    safe fn llvm_s_get_waveid_in_workgroup() -> u32;
+
+    // gfx11
+    #[link_name = "llvm.amdgcn.permlane64.i32"]
+    fn llvm_permlane64_u32(value: u32) -> u32;
+
+    // gfx12
+    #[link_name = "llvm.amdgcn.permlane16.var"]
+    fn llvm_permlane16_var(old: u32, src0: u32, src1: u32, fi: bool, bound_control: bool) -> u32;
+
+    // gfx12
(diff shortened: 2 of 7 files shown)


sayantn:
I think it can be "tested" the same way we do `core::arch::nvptx` (this is also a similar `NO_RUN`, `NO_STD` target). For those targets, we just test in CI that it builds. You can just look at how we handle `nvptx64-nvidia-cuda` in .github/workflows/main.yml. Also you would need to add this target to ci/dox.sh to check that the documentation builds

Flakebi:
Thanks, I tried to replicate what’s there for nvptx + adding `-Ctarget-cpu` (as amdgpu doesn’t compile without, it has no generic cpu) + build-std=core.
I still can’t build ci/run-docker.sh or ci/run.sh locally (ci/dox.sh seems to work if I remove s390x), but I think I got it working now.
Sorry for the many force-pushes!

The diff to my first push from when opening this PR is here: https://github.com/rust-lang/stdarch/compare/b3f5bdae0efbdd5f7297d0225623bd31c7fe895b..e1028110e77561574bfb7ea349154d46b5ea7b86

ci/dox.sh
```diff
@@ -15,6 +15,14 @@ dox() {
 
   cargo clean --target "${1}"
 
+  if [ "${1}" == "amdgcn-amd-amdhsa" ]; then
+    if [ "$CI" != "" ]; then
+      rustup component add rust-src
+    fi
+    export CARGO_UNSTABLE_BUILD_STD=core,alloc
```
sayantn: I don't think we need to build `alloc`, `core_arch` is `no_std`, `no_alloc`

crates/core_arch/src/amdgpu/mod.rs
```diff
@@ -0,0 +1,818 @@
...
+#[inline]
+#[unstable(feature = "stdarch_amdgpu", issue = "149988")]
+pub unsafe fn update_dpp(
+    old: u32,
+    src: u32,
+    dpp_ctrl: u32,
+    row_mask: u32,
+    bank_mask: u32,
+    bound_control: bool,
+) -> u32 {
```
sayantn: `llvm.amdgcn.update.dpp` should have the last 4 arguments immediate, i.e. it should use const generics

```suggestion
pub unsafe fn update_dpp<
    const DPP_CTRL: u32,
    const ROW_MASK: u32,
    const BANK_MASK: u32,
    const BOUND_CONTROL: bool
>(old: u32, src: u32) -> u32 {
```

crates/core_arch/src/amdgpu/mod.rs
```diff
@@ -0,0 +1,818 @@
...
+#[inline]
+#[unstable(feature = "stdarch_amdgpu", issue = "149988")]
+pub unsafe fn permlane16_u32(
+    old: u32,
+    src0: u32,
+    src1: u32,
+    src2: u32,
+    fi: bool,
+    bound_control: bool,
+) -> u32 {
```
sayantn:
```suggestion
pub unsafe fn permlane16_u32<const FI: bool, const bound_control: bool>(
    old: u32,
    src0: u32,
    src1: u32,
    src2: u32,
) -> u32 {
```

crates/core_arch/src/amdgpu/mod.rs
sayantn, line 1: I have commented on several intrinsics that they should use const-generics, as the LLVM intrinsics use `ImmArg`. There are a few more, just ensure that the Immediate arguments use const-generics. I checked by temporarily making all the intrinsics `#[inline(never)]`, and running `ci/run.sh` (this ensures that the functions are actually codegened)

 - `update.dpp`
 - `wave.reduce.{or,and,max,min,xor,umax,umin}`
 - `permlane{x}16{.var}`
 - `permlane{16,32}.swap`
 - `sched.barrier`
 - `sched.group.barrier`
 - `s.barrier.wait`
 - `s.barrier.signal`
 - `s.barrier.signal.isfirst`
 - `s.sleep`
 - `s.sethalt`

also, if an Immediate argument has some kind of restrictions (e.g. it needs to be in range `0..10`, then you can check that using the `static_assert` macros in )

sayantn (changes requested):
Thanks, I have put some comments

sayantn:
Also I have noticed that Clang provides a different set of intrinsics than these ([`amdgpuintrin.h`](https://clang.llvm.org/doxygen/amdgpuintrin_8h_source.html) and [`gpuintrin.h`](https://clang.llvm.org/doxygen/gpuintrin_8h_source.html)). Is this divergence intentional? We normally stick to following the C conventions in stdarch

Flakebi:
Thanks for the review!

Interesting, I thought `core_arch` is more akin to the `__builtin_*` family of functions in clang.
The clang `__gpu` intrinsics are definitely useful, though they are (to no surprise) missing low-level, hardware-specific intrinsics (e.g. `wave_id`, `sethalt` and buffer/image intrinsics, which I can’t add yet because they need simd type support in rustc).

I do plan to write a common `gpu` module for amdgpu and nvptx as well.
Maybe this one should follow `gpuintrin.h`?
We could e.g. have
```rust
mod amdgpu {
    mod gpu {
        use super::*;
        pub fn block_id_x() -> u32 {
            workitem_id_x()
        }
    }
}
// Same for nvptx

mod gpu {
    #[cfg(target_arch = "amdgpu")]
    pub use amdgpu::gpu::*;
    #[cfg(target_arch = "nvptx64")]
    pub use nvptx::gpu::*;

    // + more intrinsics as in gpuintrin.h
}
```

sayantn:
The analogue to `__builtin` functions in C are `#[rust_intrinsic]` functions and linked LLVM intrinsics. They are generally not exposed to users (in stable) as they are tightly linked to the compiler implementation.

If there are some interesting platform-specific intrinsics, we can add them, but generally we follow GCC and Clang in regards to intrinsics.

Yea, a common `gpu` module can be added following `gpuintrin.h`.

Flakebi:
Thanks for the review, that’s a nice case for const generics! I fixed all your inline comments.

Also thanks for linking the clang intrinsics, I tend to forget about those.
As far as I am aware, the moste used APIs for GPUs are CUDA/HIP/SYCL on the compute side and DirectX/Vulkan (through HLSL/GLSL) on the graphics side. Although clang is a building block for these, the clang intrinsics are most often not used directly. Therefore I’m rather hestitant to follow clang for GPUs.

As common GPU intrinsics is a larger topic, I started a Discourse post here (please add your thoughts!): https://internals.rust-lang.org/t/naming-gpu-things-in-the-rust-compiler-and-standard-library/23833
Though, I think progress on this doesn’t need to block amdgpu intrinsics (note that I don’t have a plan to stabilize amdgpu intrinsics, stabilizing generic GPU intrinsics seems more desirable).

My planned intrinsic “stack” is roughly as follows, from high-level to low-level:
1. Rust exposes (experimental) generic gpu intrinsics, e.g. at `core::gpu`, naming tbd, see the Discourse post
2. These are implemented for amdgpu and nvptx by calling target-specific intrinsics from `core::arch::amdgpu/nvptx`
3. `core::arch::amdgpu/nvptx` intrinsics expose (not on stable) their respective LLVM intrinsics, naming can follow the LLVM intrinsic names

We _could_ also use clang names for `core::arch::amdgpu/nvptx`, then we would end up with three set of names:
1. Generic `core::gpu` intrinsics: Rust-specific names
2. `core::arch::amdgpu/nvptx`: clang names
3. LLVM intrinsics (used in `core::arch`): LLVM names

I would prefer following LLVM’s intrinsic names in `core::arch::amdgpu/nvptx` instead of clang’s names.
What do you think?

Note for my future self: I got run.sh to work with `NORUN=1 NOSTD=1 TARGET=amdgcn-amd-amdhsa CARGO_UNSTABLE_BUILD_STD=core ci/run.sh`.

crates/core_arch/src/amdgpu/mod.rs
```diff
@@ -0,0 +1,822 @@
+//! amdgpu intrinsics
+//!
+//! The reference is the [LLVM amdgpu guide] and the [LLVM implementation].
+//! The order of intrinsics here follows the order in the [LLVM implementation].
+//!
+//! [LLVM amdgpu guide]: https://llvm.org/docs/AMDGPUUsage.html#llvm-ir-intrinsics
+//! [LLVM implementation]: https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/IR/IntrinsicsAMDGPU.td
+
+#[allow(improper_ctypes)]
+unsafe extern "unadjusted" {
```
sayantn, line 10: is there some specific reason you only made *some* of the (both LLVM and Rust) intrinsics safe? Normally we make an intrinsic unsafe only if it can cause memory unsafety, and I don't see how something like `s.barrier.signal` is `unsafe`, where `s.barrier` is safe?

crates/core_arch/src/amdgpu/mod.rs
```diff
@@ -0,0 +1,822 @@
...
+    fn llvm_s_barrier_signal(barrier_type: i32);
+    #[link_name = "llvm.amdgcn.s.barrier.signal.isfirst"]
+    fn llvm_s_barrier_signal_isfirst(barrier_type: i32) -> bool;
+    #[link_name = "llvm.amdgcn.s.barrier.wait"]
+    fn llvm_s_barrier_wait(barrier_type: u16);
+    #[link_name = "llvm.amdgcn.s.barrier.leave"]
+    fn llvm_s_barrier_leave(barrier_type: u16);
+    #[link_name = "llvm.amdgcn.s.get.barrier.state"]
+    fn llvm_s_get_barrier_state(barrier_type: i32) -> u32;
+    #[link_name = "llvm.amdgcn.s.wave.barrier"]
```
sayantn:
```suggestion
    #[link_name = "llvm.amdgcn.wave.barrier"]
```
typo (probably also rename the intrinsics to `(llvm_)wave_barrier`

crates/core_arch/src/amdgpu/mod.rs
```diff
@@ -0,0 +1,822 @@
...
+    safe fn llvm_wave_reduce_umax(value: u32, strategy: u32) -> u32;
+    #[link_name = "llvm.amdgcn.wave.reduce.fmax"]
+    safe fn llvm_wave_reduce_fmax(value: f32, strategy: u32) -> f32;
+    #[link_name = "llvm.amdgcn.wave.reduce.max"]
+    safe fn llvm_wave_reduce_max(value: i32, strategy: u32) -> i32;
+    #[link_name = "llvm.amdgcn.wave.reduce.add"]
+    safe fn llvm_wave_reduce_add(value: u32, strategy: u32) -> u32;
+    #[link_name = "llvm.amdgcn.wave.reduce.fadd"]
+    safe fn llvm_wave_reduce_fadd(value: f32, strategy: u32) -> f32;
+    #[link_name = "llvm.amdgcn.wave.reduce.and"]
```
sayantn, line 80: Q: I see that LLVM also defined `fsub` and `sub` versions of these intrinsics, why aren't they included here?

crates/core_arch/src/amdgpu/mod.rs
```diff
@@ -0,0 +1,822 @@
...
+        src2: u32,
+        fi: bool,
+        bound_control: bool,
+    ) -> u32;
+
+    #[link_name = "llvm.amdgcn.s.get.waveid.in.workgroup"]
+    safe fn llvm_s_get_waveid_in_workgroup() -> u32;
+
+    // gfx11
+    #[link_name = "llvm.amdgcn.permlane64"]
```
sayantn:
```suggestion
    #[link_name = "llvm.amdgcn.permlane64.i32"]
```
the type parameter should be specified

crates/core_arch/src/amdgpu/mod.rs
```diff
@@ -0,0 +1,822 @@
...
+/// - 0x0020: VMEM read instructions may be scheduled across `sched_barrier`.
+/// - 0x0040: VMEM write instructions may be scheduled across `sched_barrier`.
+/// - 0x0080: All DS instructions may be scheduled across `sched_barrier`.
+/// - 0x0100: All DS read instructions may be scheduled across `sched_barrier`.
+/// - 0x0200: All DS write instructions may be scheduled across `sched_barrier`.
+/// - 0x0400: All Transcendental (e.g. V_EXP) instructions may be scheduled across `sched_barrier`.
+#[inline]
+#[unstable(feature = "stdarch_amdgpu", issue = "149988")]
+pub unsafe fn sched_barrier<const MASK: u32>() {
+    static_assert_uimm_bits!(MASK, 11);
```
sayantn, line 321: Is this check enough? It seems like `MASK` can *only* take values mentioned above, not any linear combination of them -- in that case we should also check `MASK.is_power_of_two()` or something like that. Also nit, it might be better to make some constants for these values, but I'll leave it to your discretion

crates/core_arch/src/amdgpu/mod.rs
sayantn, line 1: For some reason when I tried to build this with all `#[inline]`s replaced by `#[inline(never)]` (so that the functions actually codegen), I get a segfault for `permlane64_u32`, and LLVM selection errors for `s.get.waveid.in.workgroup` and `wave.id`. This is probably due to missing target features or something similar, I noticed that you made comments that these are only available in `gfx12` and higher (probably, my knowledge of GPUs is nonexistent), but we are building with `target-cpu=gfx900`.

sayantn (changes requested):
Thanks, left some more comments

sayantn:
I don't have any strong preferences, if you feel that the lower-level LLVM intrinsics are more useful that just having the clang intrinsics, go ahead with that -- we can change it, remove it, or add the clang intrinsics too if required (as long as they are unstable)

crates/core_arch/src/amdgpu/mod.rs
Flakebi: Yes, I tried to follow the Rust safe/unsafe principles.
Every intrinsic that is marked `safe` will not cause undefined behavior or memory safety issues, not matter with which arguments it is called. They could still cause other issues, like deadlocks if `s.barrier` is used incorrectly, but this is still “safe” in the Rust sense.
Intrinsics that are `unsafe` can cause undefined behavior, either in the compiler/LLVM or in the hardware when calling them with incorrect arguments. E.g. `readlane` or `permlane` can read registers from inactive threads (which I guess is either directly UB in LLVM or `poison`, which can cause UB further down the line).
For `s.barrier.signal`, I’m not sure what the allowed `BARRIER_TYPE`s are. What happens when passing an invalid number for the type seems to be currently unspecified (the documentation says “only for non-named barriers”) and could very well become undefined behavior with a future LLVM change.

crates/core_arch/src/amdgpu/mod.rs
Flakebi: I couldn’t make much sense of the `sub` variants :smile:
The reduce intrinsics do some accumulation/reduction over the 32 or 64 threads in a wave.
This can be neatly implemented as a parallel reduce (i.e. a binary tree where first thread 0+1→0, 2+3→2, … are added, then 0+2, 4+6, etc.; LLVM currently doesn’t do this… but that’s a different problem).

Doing an `add` reduce of multiple values (over a wave here) makes sense to me. `add` is an associative operation.
Doing a `sub` doesn’t make sense to me, because it’s not associative. The order of operations matters, so what is a parallel sub reduce supposed to do? (Also, is the very first value kept positive or subtracted from 0?)
I looked into the LLVM implementation, it seems to do a sum (add reduce) and then multiply by -1. If someone needed that functionality, I think it would be much clearer to write exactly that, `reduce.add * -1`. It would also be as efficient as the `reduce.sub` intrinsic as it doesn’t do anything different in the implementation.

crates/core_arch/src/amdgpu/mod.rs
Flakebi: I see it used like a proper bitmask here (15/0xf), so I think it’s fine, but thanks for checking!
https://github.com/llvm/llvm-project/blob/main/llvm/test/CodeGen/AMDGPU/llvm.amdgcn.sched.barrier.ll#L16

crates/core_arch/src/amdgpu/mod.rs
Flakebi: Ah yes, I can reproduce this.
There are quite a few “target-cpu”s (gfx numbers). Every generation of GPUs adds four or more of them.
E.g. for the RX 90xx series, there’s gfx1200 (RX 9060 (XT)), gfx1201 (RX 9070 (XT)), gfx1250 and gfx1251 (for APUs).
The full list is available here: https://llvm.org/docs/AMDGPUUsage.html#processors

The HPC/datacenter cards (MI 100, MI 250, … in case you’ve heard the names) are building upon gfx9, so new releases are gfx90a, gfx942, etc.

The amdgpu LLVM backend has about one target feature per intrinsic and a large list of target features for every generation and variant.
I’m hesitant to introduce such things as target-features to Rust as it quickly becomes a maintenance burden.

Flakebi:
Thanks for taking such a thorough look!
I fixed the (s.)wave.barrier and permlane64.i32 and answered your questions inline. I think the “works only on some target-cpus” is the most difficult one. I’m thinking, as long as it’s unstable, this state is fine. But that’s because I have no good idea to fix this, so if you or someone else has an idea, please let me know.

crates/core_arch/src/amdgpu/mod.rs
sayantn: ah, that makes sense. Also, (just mentioning, no relation to this PR) named barriers seem to be a `targetext` type in LLVM, specifically `target("amdgcn.named.barrier", 0)`, and those types are normally very difficult to access from Rust, but the intrinsics use a `addrspace(3)` pointer for that, and I think we would need more compiler support for them (currently cg_llvm mostly treats all pointers as `addrspace(0)` by default, support for other address spaces is quite small)

crates/core_arch/src/amdgpu/mod.rs
sayantn: I see, we can tackle this problem later. Just for the sake of testing, is there some `target-cpu` that has like all (or most) of the target features set, then we can just use that in tests

crates/core_arch/src/amdgpu/mod.rs
sayantn: I discovered some more problems with the intrinsics, `wave.reduce.f{add,sub,max,min}` and `llvm.s.barrier.leave` are not in the LLVM release branch yet, they were added like 1 month ago. So they will fail to compile once someone actually monomorphises the functions. We should add them when the changes actually propagates to the rust fork (~5 months probably), otherwise it will cause link failures

Also, you probably forgot to add intrinsics for the `add` and `fadd` ones

crates/core_arch/src/amdgpu/mod.rs
```diff
@@ -0,0 +1,822 @@
...
+pub unsafe fn s_barrier_wait<const BARRIER_TYPE: u16>() {
+    unsafe { llvm_s_barrier_wait(BARRIER_TYPE) }
+}
+
+/// Leave a specific barrier type.
+///
+/// Only for non-named barriers.
+#[inline]
+#[unstable(feature = "stdarch_amdgpu", issue = "149988")]
+pub unsafe fn s_barrier_leave<const BARRIER_TYPE: u16>() {
```
sayantn:
```suggestion
pub unsafe fn s_barrier_leave<const BARRIER_TYPE: i32>() {
```

Flakebi:
I fixed your comments and added a tests mod with instructions to check that generating the instructions works.
I changed the default target-cpu to gfx12, that seems like the one that supports most intrinsics (it doesn’t support `s_memrealtime`, `s_get_waveid_in_workgroup`, `permlane16_swap` and `permlane32_swap`).

crates/core_arch/src/amdgpu/mod.rs
```diff
@@ -0,0 +1,1053 @@
...
+
+// Functions to generate code, used to check that the intrinsics build.
+// Marked as no_mangle, so the compiler does not remove the functions.
+// To test, uncomment the `#[cfg(test)]` line below and run
+// NORUN=1 NOSTD=1 TARGET=amdgcn-amd-amdhsa CARGO_UNSTABLE_BUILD_STD=core ci/run.sh
+//
+// Note that depending on the target-cpu set in run.sh, some of these intrinsics are not available
+// and compilation fails with `Cannot select: intrinsic %llvm.amdgcn...`.
+// Uncomment these intrinsics to check.
+#[cfg(test)]
```
sayantn, line 802:
```suggestion
#[cfg(debug_assertions)]
```
We can't do `cargo test` or `cargo build --tests` because this target doesn't support `bin` crates. So this is probably better, this will also ensure that it is checked in CI (we do a `--profile=dev` run in CI)

sayantn (approved):
LGTM, just one nit

Edit: sorry, my mistake. No other problems. Thanks ❤️

crates/core_arch/src/amdgpu/mod.rs
sayantn: Oh wait, no. We cannot run it in CI, because it requires manual intervention to check which ones to disable.

sayantn: Oh wait, no. We cannot run it in CI, because it requires manual intervention to check which ones to disable.

PR #1977, opened 2025-12-16
Panic if cpuid is called with SGX

sayantn:
Corrects mistake in rust-lang/stdarch#1935, and panics if CPUID is called on SGX. 

cc @raoulstrackx @jethrogb @rust-lang/libs-api


raoulstrackx:
Looks good to me! Thank you!

PR #1979, opened 2025-12-17
Fix broken AMD Developer's Manual links

nikokajula:
AMD moved its technical manuals from support.amd.com to docs.amd.com,
causing the existing links to redirect to the documentation hub landing
page. Updated the links to point to the current developer manual location.

### DIFF
--- a/crates/core_arch/src/x86/abm.rs
+++ b/crates/core_arch/src/x86/abm.rs
@@ -12,8 +12,8 @@
 //! [Wikipedia][wikipedia_bmi] provides a quick overview of the instructions
 //! available.
 //!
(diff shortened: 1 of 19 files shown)


nikokajula:
I also decided to fix the intel documentation. Since there was an issue created for that, but it seems you were faster. Did the force-push mess up anything?
edit: doesn't seem like it.

folkertdev:
Well this merged the earlier commit, without the intel changes https://github.com/rust-lang/stdarch/commit/cff412f7ef9bd8bdef43a30b9a29237d25155873.

 I  think you can just rebase and open a new PR with those additional changes.

PR #1981, opened 2025-12-26
x86: remove "unsafe" from tests that do not need it

eduardosm:
Some test utility functions (e.g., `assert_eq_m128`) have been refactored so they are not unsafe. Many test functions (`test_mm_*`) do not call any unsafe function (besides needing target features) anymore, so they do not need to be `unsafe fn`.

### DIFF
--- a/crates/core_arch/src/x86/test.rs
+++ b/crates/core_arch/src/x86/test.rs
@@ -1,225 +1,122 @@
 //! Utilities used in testing the x86 intrinsics
 
 use crate::core_arch::assert_eq_const as assert_eq;
+use crate::core_arch::simd::*;
 use crate::core_arch::x86::*;
 use std::mem::transmute;
 
 #[track_caller]
-#[target_feature(enable = "sse2")]
 #[rustc_const_unstable(feature = "stdarch_const_helpers", issue = "none")]
-pub(crate) const unsafe fn assert_eq_m128i(a: __m128i, b: __m128i) {
-    assert_eq!(transmute::<_, [u64; 2]>(a), transmute::<_, [u64; 2]>(b))
-}
-
-#[track_caller]
-#[target_feature(enable = "avx")]
-#[rustc_const_unstable(feature = "stdarch_const_helpers", issue = "none")]
-pub(crate) const unsafe fn assert_eq_m256i(a: __m256i, b: __m256i) {
-    assert_eq!(transmute::<_, [u64; 4]>(a), transmute::<_, [u64; 4]>(b))
+pub(crate) const fn assert_eq_m128i(a: __m128i, b: __m128i) {
+    assert_eq!(a.as_u32x4(), b.as_u32x4());
 }
 
 #[track_caller]
 #[rustc_const_unstable(feature = "stdarch_const_helpers", issue = "none")]
-pub(crate) const unsafe fn assert_eq_m512i(a: __m512i, b: __m512i) {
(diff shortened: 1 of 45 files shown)


crates/core_arch/src/x86/test.rs
```diff
@@ -1,225 +1,122 @@
...
-            }
-
-            $crate::intrinsics::const_eval_select(($($param),*), ct, rt)
-        }
-    )*}
-}
-
-make_ct_rt! {
-    // SAFETY: we can use simple float equality because when this should only be used in const
-    // context where Intel peculiarities don't appear
```
eduardosm, line 55: I don't think Intel peculiarities matter much here.

sayantn: it kinda does, the docs of `const_eval_select` say that

>Therefore, when using this intrinsic anywhere that can be reached from stable, it is crucial that the end-to-end behavior of the stable const fn is the same for both modes of execution.

Fortunately, as it turns out, `EQ_OQ` mode (`true` only if `a == b`, with no `#IA` signalling on QNaN) behaves exactly the same as `a == b` (= `a.partial_cmp(&b) == Some(Equal)`), except for some `MXCSR` flags which I am not too worried about. And Intel uses this exact mode for the `cmpeq` intrinsics.

crates/core_arch/src/x86/test.rs
```diff
@@ -1,225 +1,122 @@
...
-        };
-        rt: unsafe {
-            let r = _mm_cmpeq_ps(a, b);
-            if _mm_movemask_ps(r) != 0b1111 {
-                panic!("{:?} != {:?}", a, b);
-            }
-        };
-    }
+#[track_caller]
+#[rustc_const_unstable(feature = "stdarch_const_helpers", issue = "none")]
```
sayantn, line 27: these annotations can be removed, as we are not using any intrinsics here

sayantn (commented):
you did leave some of the `unsafe` annotations (mostly) on functions that do load/store. Although the intrinsics themselves are unsafe because they take a raw pointer argument, we can still make the tests safe.

crates/core_arch/src/x86/test.rs
eduardosm: The `assert_eq_const` macro used in those functions uses `const_eval_select`, which requires the annotation.

eduardosm:
> you did leave some of the `unsafe` annotations (mostly) on functions that do load/store. Although the intrinsics themselves are unsafe because they take a raw pointer argument, we can still make the tests safe.

Yes, indeed, since `target_feature` no longer requires it, we can make all test functions safe `fn`s and use instead `unsafe` blocks where needed. And the final goal would be to make `#[simd_test]` reject `unsafe fn`s (like `#[test]` does).

However, since this PR is already as large as it is, I wanted first do the trivial cases (in most cases, it is just removing the `unsafe` keyword) and handle the rest in a follow up PRs.

crates/core_arch/src/x86/test.rs
eduardosm: That what I was hoping for, avoiding `const_eval_select` and using directly the `assert_eq_const` macro (which uses `const_eval_select` to produce a simpler panic message in const-eval).

crates/core_arch/src/x86/test.rs
sayantn: Ohh, I completely forgot that I named the const-safe version `assert_eq` as well :man_facepalming:

sayantn (approved):

PR #1985, opened 2026-01-01
Use LLVM intrinsics for `madd` intrinsics

usamoi:
https://github.com/rust-lang/rust/issues/150560

r? @sayantn

### DIFF
--- a/crates/core_arch/src/x86/sse2.rs
+++ b/crates/core_arch/src/x86/sse2.rs
@@ -210,14 +210,20 @@ pub const fn _mm_avg_epu16(a: __m128i, b: __m128i) -> __m128i {
 #[target_feature(enable = "sse2")]
 #[cfg_attr(test, assert_instr(pmaddwd))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
-#[rustc_const_unstable(feature = "stdarch_const_x86", issue = "149298")]
-pub const fn _mm_madd_epi16(a: __m128i, b: __m128i) -> __m128i {
-    unsafe {
-        let r: i32x8 = simd_mul(simd_cast(a.as_i16x8()), simd_cast(b.as_i16x8()));
-        let even: i32x4 = simd_shuffle!(r, r, [0, 2, 4, 6]);
-        let odd: i32x4 = simd_shuffle!(r, r, [1, 3, 5, 7]);
-        simd_add(even, odd).as_m128i()
-    }
+pub fn _mm_madd_epi16(a: __m128i, b: __m128i) -> __m128i {
+    // It's a trick used in the Adler-32 algorithm to perform a widening addition.
+    //
+    // ```rust
+    // #[target_feature(enable = "sse2")]
+    // unsafe fn widening_add(mad: __m128i) -> __m128i {
+    //     _mm_madd_epi16(mad, _mm_set1_epi16(1))
+    // }
+    // ```
+    //
+    // If we implement this using generic vector intrinsics, the optimizer
+    // will eliminate this pattern, and `pmaddwd` will no longer be emitted.
+    // For this reason, we use x86 intrinsics.
+    unsafe { transmute(pmaddwd(a.as_i16x8(), b.as_i16x8())) }
 }
 
 /// Compares packed 16-bit integers in `a` and `b`, and returns the packed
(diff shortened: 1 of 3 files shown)


folkertdev:
Can you add something like this 

```rust
    /// This is a trick used in the adler32 algorithm to get a widening addition. The
    /// multiplication by 1 is trivial, but must not be optimized out because then the vpmaddwd
    /// instruction is no longer selected.
    #[target_feature(enable = "avx2")]
    #[cfg_attr(test, assert_instr(vpmaddwd))]
    unsafe fn _mm256_madd_epi16_mul_one(mad: __m256i) -> __m256i {
        let one_v = _mm256_set1_epi16(1);
        _mm256_madd_epi16(mad, one_v)
    }
```

So that we remember in the future why we can't just use that generic implementation?

usamoi:
I have added the following comments:

```rust
// It's a trick used in the Adler-32 algorithm to perform a widening addition.
//
// ```rust
// #[target_feature(enable = "sse2")]
// unsafe fn widening_add(mad: __m128i) -> __m128i {
//     _mm_madd_epi16(mad, _mm_set1_epi16(1))
// }
// ```
//
// If we implement this using generic vector intrinsics, the optimizer
// will eliminate this pattern, and `pmaddwd` will no longer be emitted.
// For this reason, we use x86 intrinsics.
```

crates/core_arch/src/x86/sse2.rs
```diff
@@ -210,14 +210,20 @@ pub const fn _mm_avg_epu16(a: __m128i, b: __m128i) -> __m128i {
...
+    // unsafe fn widening_add(mad: __m128i) -> __m128i {
+    //     _mm_madd_epi16(mad, _mm_set1_epi16(1))
+    // }
+    // ```
+    //
+    // If we implement this using generic vector intrinsics, the optimizer
+    // will eliminate this pattern, and `pmaddwd` will no longer be emitted.
+    // For this reason, we use x86 intrinsics.
+    unsafe { transmute(pmaddwd(a.as_i16x8(), b.as_i16x8())) }
 }
```
folkertdev, line 227: we could consider using https://doc.rust-lang.org/std/intrinsics/fn.const_eval_select.html so we don't loose all of the `const` stuff. Up to @sayantn though, I don't have full context on what we'd like to be `const fn` currently.

usamoi: `#[target_feature]` functions do not implement the `Fn` traits, while `const_eval_select` restricts `FnOnce`. So this does not seem feasible.

sayantn: We can still do it, the inner function which will be invoked doesn't need to have the correct target features if marked `#[inline]` (`#[inline(always)]` will probably be better) [Godbolt](https://godbolt.org/z/Yqb6Ecesn). But I don't think it is *that* important to make this `const` right now. I think a better approach will be fixing the LLVM bug and make this truly const in future, but this should be enough for the patch

sayantn:
The Miri test suite might start failing now, since the implementation of these intrinsics were removed. cc @RalfJung

folkertdev:
> The Miri test suite might start failing now, since the implementation of these intrinsics were removed.

Do you happen to know where that was changed? It turns out zlib-rs uses 

```rust
    #[link_name = "llvm.x86.avx512.pmaddw.d.512"]
    fn vpmaddwd(a: i16x32, b: i16x32) -> i32x16;
```

So our CI started failing now that miri is no longer able to execute that instruction.

usamoi:
> > The Miri test suite might start failing now, since the implementation of these intrinsics were removed.
> 
> Do you happen to know where that was changed? It turns out zlib-rs uses
> 
> ```rust
>     #[link_name = "llvm.x86.avx512.pmaddw.d.512"]
>     fn vpmaddwd(a: i16x32, b: i16x32) -> i32x16;
> ```
> 
> So our CI started failing now that miri is no longer able to execute that instruction.

It's https://github.com/rust-lang/rust/commit/8d597aa36528dada3cbd9bcfec889c2da6ecaaac. See https://github.com/rust-lang/rust/pull/150639#issuecomment-3707051747. Reverted in https://github.com/rust-lang/rust/pull/150639/changes/5e4168b162ba9286161226b315a2d3632f31e804. It's never implemented for AVX512.

RalfJung:
Yeah, https://github.com/rust-lang/rust/commit/5e4168b162ba9286161226b315a2d3632f31e804 fixed this for every case where this was ever supported by Miri. Maybe the avx512 variant was accidentally supported and we didn't even realize.^^

folkertdev:
Right, miri would have been able to execute the implementation that is removed by this PR, and we happened to rely on that. I'll see if i can add support for the avx512 variant to miri then to tide us over.

PR #1986, opened 2026-01-01
Clarify documentation around "undefined" bits in intrinsics.

veluca93:


crates/core_arch/src/x86/avx.rs
```diff
@@ -2960,7 +2964,11 @@ pub const fn _mm256_castps128_ps256(a: __m128) -> __m256 {
 }
 
 /// Casts vector of type __m128d to type __m256d;
-/// the upper 128 bits of the result are undefined.
+/// the upper 128 bits of the result are indeterminate.
+///
+/// In the Intel documentation, the upper bits are declared to be `undefined`.
+/// This is not equivalent to [`mem::MaybeUninit`], as these bits are non-deterministically
+/// set to some valid value. In practice, this is typically equivalent to [`mem::zeroed`].
```
RalfJung:
```suggestion
/// In the Intel documentation, the upper bits are declared to be "undefined".
/// This is not equivalent to [`mem::MaybeUninit`]; instead, these bits are non-deterministically
/// set to some valid value. In practice, this is typically equivalent to [`mem::zeroed`].
```

PR #1987, opened 2026-01-01
Use a generic type to implement SIMD types in `core_arch::simd` and avoid `unsafe fn` in more tests

eduardosm:


crates/core_arch/src/macros.rs
```diff
@@ -135,6 +135,22 @@ macro_rules! types {
...
+            }
+        }
+
+        $(#[$stability])+
+        impl crate::convert::From<$name> for crate::core_arch::simd::Simd<$elem_type, $len> {
+            #[inline(always)]
+            fn from(simd: $name) -> Self {
+                unsafe { crate::mem::transmute(simd) }
+            }
+        }
```
eduardosm, line 153: This is now easy to include thanks to the generic `core_arch::simd::Simd` type.

crates/core_arch/src/simd.rs
```diff
@@ -16,129 +16,209 @@ pub(crate) const unsafe fn simd_imin<T: Copy>(a: T, b: T) -> T {
...
+    }
+}
+
+// `#[derive(Clone)]` causes ICE "Projecting into SIMD type core_arch::simd::Simd is banned by MCP#838"
+impl<T: SimdElement, const N: usize> Clone for Simd<T, N> {
+    #[inline]
+    fn clone(&self) -> Self {
+        *self
+    }
+}
```
eduardosm, line 89: I don't know why the ICE with `#[derive(Clone)]` did not happen with non-generic types.

crates/core_arch/src/arm_shared/neon/mod.rs
```diff
@@ -5546,11 +5789,19 @@ mod tests {
...
-        [13_f16, 14., 15., 16.]));
+        #[cfg_attr(target_arch = "arm", simd_test(enable = "neon,fp16"))]
+        #[cfg_attr(not(target_arch = "arm"), simd_test(enable = "neon"))]
+        fn test_vcombine_f16() {
+            let a = f16x4::from_array([3_f16, 4., 5., 6.]);
+            let b = f16x4::from_array([13_f16, 14., 15., 16.]);
+            let e = f16x8::from_array([3_f16, 4., 5., 6., 13_f16, 14., 15., 16.]);
+            let c = f16x8::from(vcombine_f16(a.into(), b.into()));
+            assert_eq!(c, e);
+        }
```
eduardosm, line 5804: No longer uses the `test_vcombine` due to the need of specifying "fp16" depending on ARM version.

ci/run.sh
```diff
@@ -40,7 +40,7 @@ case ${TARGET} in
 	export RUSTFLAGS="${RUSTFLAGS} -C llvm-args=-fast-isel=false"
 	;;
     armv7-*eabihf | thumbv7-*eabihf)
-        export RUSTFLAGS="${RUSTFLAGS} -Ctarget-feature=+neon"
+        export RUSTFLAGS="${RUSTFLAGS} -Ctarget-feature=+neon,+fp16"
```
eduardosm, line 43: Since `fp16` cannot be checked at runtime (at least `is_arm_feature_detected` does not support it), we enable it at compile time.

crates/core_arch/src/simd.rs
sayantn: this is normal, the `derive`d `Clone` implementation tries to do the clone elementwise, projecting to the SIMD type in the process. The compiler explicitly rejects any SIMD type projection via an ICE.

The `derive`d implementation looks like this
```rust
fn clone(&self) -> Self {
    Self(self.0.clone())
              ^ SIMD type field projection => ICE
}
```

edit: oh I just got the problem, why does the same ICE not occur for non-generic versions. I imagine this is due to optimizations, the compiler can eliminate the field projection for non-generic versions by doing a bitwise `Copy`, but for generic `T`, the compiler cannot ensure that `Copy` and `Clone` behave the same way. For example, the following code is perfectly valid

```rust
#[derive(Copy)]
struct Foo(u32);

impl Clone for Foo {
    fn clone(&self) -> Self {
        println!("Cloning");
        *self
    }
}
```

although I am just guessing.

crates/core_arch/src/simd.rs
```diff
@@ -16,129 +16,209 @@ pub(crate) const unsafe fn simd_imin<T: Copy>(a: T, b: T) -> T {
...
+
+#[rustc_const_unstable(feature = "stdarch_const_helpers", issue = "none")]
+impl<T: SimdElement, const N: usize> const core::cmp::PartialEq for Simd<T, N> {
```
sayantn:
```suggestion
impl<T: SimdElement, const N: usize> const crate::cmp::PartialEq for Simd<T, N> {
```
and similarly in other places. It is better to use `crate` because then it will use the in-tree libcore instead of the previous stage while bootstrapping

crates/core_arch/rustfmt.toml
sayantn, line 1: We probably shouldn't delete this file, otherwise user `rustfmt` configurations will cause the code to be wrongly formatted. We can just leave this as empty

crates/core_arch/src/simd.rs
```diff
@@ -16,129 +16,209 @@ pub(crate) const unsafe fn simd_imin<T: Copy>(a: T, b: T) -> T {
...
-        pub(crate) struct $id([$elem_type; $len]);
+        pub(crate) type $id = Simd<$elem_type, $len>;
 
-        #[allow(clippy::use_self)]
         impl $id {
-            /// A value of this type where all elements are zeroed out.
-            pub(crate) const ZERO: Self = unsafe { crate::mem::zeroed() };
-
             #[inline(always)]
             pub(crate) const fn new($($param_name: $elem_type),*) -> Self {
```
sayantn, line 154: TBH is there any reason to even keep this function, callers can just throw in a couple `[`-`]`s and use `from_array`. This makes the file very long for no good reason.

crates/core_arch/src/simd.rs
```diff
@@ -16,129 +16,209 @@ pub(crate) const unsafe fn simd_imin<T: Copy>(a: T, b: T) -> T {
...
+}
+
+impl<T: SimdElement, const N: usize> core::fmt::Debug for Simd<T, N> {
+    #[inline]
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        debug_simd_finish(f, "Simd", self.as_array())
+    }
+}
+
+impl<const N: usize> Simd<f32, N> {
```
sayantn, line 120: please don't forget the youngest child `f16`

crates/core_arch/src/simd.rs
eduardosm: I agree we should get rid of `new` and just use `from_array`. However, a quick search for `[iu](8|16|32|64)x(2|4|8|16)::new` throws 5000+ results, so I will not be doing it now.

sayantn: Ouch, I guess most of then is in arm/aarch64, as it doesn't have a convenient way to create vectors from its elements (like `_mm_set_ps` etc in x86)

sayantn:
The changes LGTM, although the `simd_test` changes seem a bit hacky, I guess we can live with it if it enables us to have a simpler life

PR #1988, opened 2026-01-02
Change `repr(packed)` struct to `repr(C, packed)`

eduardosm:
`repr(packed)` structs do not have a well-defined layout


sayantn:
nice catch, thanks

PR #1991, opened 2026-01-10
remove fp16 target feature from some vreinterpret intrinsics

usamoi:
Usage of them is accepted by clang without `fullfp16` (https://godbolt.org/z/dhffa4YT9). It seems that they are missing in https://github.com/rust-lang/stdarch/pull/1978.

cc @adamgemmell, in case I misunderstood

### DIFF
--- a/crates/core_arch/src/x86/avx512fp16.rs
+++ b/crates/core_arch/src/x86/avx512fp16.rs
@@ -23993,16 +23993,16 @@ mod tests {
 
     #[simd_test(enable = "avx512fp16,avx512vl")]
     const fn test_mm256_reduce_mul_ph() {
-        let a = _mm256_set1_ph(2.0);
+        let a = _mm256_set1_ph(1.2);
         let r = _mm256_reduce_mul_ph(a);
-        assert_eq!(r, 65536.0);
+        assert_eq!(r, 18.5);
     }
 
     #[simd_test(enable = "avx512fp16")]
     const fn test_mm512_reduce_mul_ph() {
-        let a = _mm512_set1_ph(2.0);
+        let a = _mm512_set1_ph(1.2);
         let r = _mm512_reduce_mul_ph(a);
-        assert_eq!(r, 16777216.0);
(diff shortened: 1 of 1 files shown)


adamgemmell:
Looks good to me, thank you!

crates/core_arch/src/arm_shared/neon/generated.rs
```diff
@@ -45037,13 +45136,14 @@ pub fn vreinterpretq_f16_p64(a: poly64x2_t) -> float16x8_t {
...
 #[cfg(target_endian = "big")]
+#[target_feature(enable = "neon")]
 #[cfg_attr(target_arch = "arm", target_feature(enable = "v8"))]
 #[cfg_attr(all(test, target_arch = "arm"), assert_instr(nop))]
 #[cfg_attr(
     all(test, any(target_arch = "aarch64", target_arch = "arm64ec")),
     assert_instr(nop)
 )]
-#[target_feature(enable = "neon,fp16")]
+#[cfg_attr(target_arch = "arm", target_feature(enable = "fp16"))]
```
Amanieu: I don't believe the `fp16` target feature is required on 32-bit either.

usamoi: This is what was done in https://github.com/rust-lang/stdarch/pull/1978, for stores and loads.
@adamgemmell, do you have ideas?

adamgemmell: vreinterprets are NOPs, so there's no issues with the instruction not being available on 32 bit platforms. If it builds and the tests pass then it's fine to remove for both sets of targets

usamoi:
`fp16` is removed on arm now.

The second commit is for passing CI. https://github.com/rust-lang/rust/pull/151529

folkertdev:
For context: this includes a fix for miri failing to build `std`, see https://rust-lang.zulipchat.com/#narrow/channel/269128-miri/topic/Miri.20test-libstd.20Failure.20.282026-01.29/near/570520551

PR #1992, opened 2026-01-11
remove `impl Neg` on s390x/powerpc vector types

folkertdev:
For other targets we don't have such instances. We specifically decided against implementing arithmetic operators on these low-level vector types. The `impl` has been replaced by a `vec_neg` function.

Over on https://www.ibm.com/docs/en/zos/2.4.0?topic=functions-defining-vector-built-in-from-operators it is suggested to define `vec_neg`. We already implement the other `define`s on that page, so this seems like a logical extension. 

On s390x `vec_neg` is also implemented for unsigned integer vector types https://godbolt.org/z/ddadsrn3q. `vec_neg` is not implemented for the `bool` types, see https://godbolt.org/z/TdKr7enhb).

On powerpc `vec_neg` is also implemented for the unsigned integer _and_ bool vector types https://godbolt.org/z/faMbx45K3.

I plan to propose stabilization of the s390x vector types in the near future.

### DIFF
--- a/crates/core_arch/src/powerpc/altivec.rs
+++ b/crates/core_arch/src/powerpc/altivec.rs
@@ -364,14 +364,45 @@ unsafe extern "C" {
     fn vrfin(a: vector_float) -> vector_float;
 }
 
-impl_neg! { i8x16 : 0 }
-impl_neg! { i16x8 : 0 }
-impl_neg! { i32x4 : 0 }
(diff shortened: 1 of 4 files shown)


sayantn:
LGTM, we shouldn't have such ops implementations only for some archs - that creates a non-uniformity in stdarch.

PR #1993, opened 2026-01-13
Stabilize `__jcvt`

folkertdev:
implementation: https://github.com/rust-lang/stdarch/pull/1938
tracking issue: https://github.com/rust-lang/rust/issues/147555

I think we should stabilize this intrinsic. The `jsconv` feature itself has been stable since `1.60.0`. The intrinsic works as expected in ruffle, see https://github.com/ruffle-rs/ruffle/pull/22138#issuecomment-3743856346.

# API surface

The `stdarch_aarch64_jscvt` feature contains one function:

> Floating-point JavaScript convert to signed fixed-point, rounding toward zero

```rust
#[target_feature(enable = "jsconv")]
pub fn __jcvt(a: f64) -> i32;
```

### DIFF
--- a/crates/core_arch/src/aarch64/neon/generated.rs
+++ b/crates/core_arch/src/aarch64/neon/generated.rs
@@ -49,7 +49,7 @@ pub fn __crc32d(crc: u32, data: u64) -> u32 {
 #[inline]
 #[target_feature(enable = "jsconv")]
 #[cfg_attr(test, assert_instr(fjcvtzs))]
-#[unstable(feature = "stdarch_aarch64_jscvt", issue = "147555")]
+#[stable(feature = "stdarch_aarch64_jscvt", since = "CURRENT_RUSTC_VERSION")]
 pub fn __jcvt(a: f64) -> i32 {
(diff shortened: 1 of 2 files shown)


folkertdev:
@adamgemmell (or someone else from arm), could you give the changes to `intrinsics_data/arm_intrinsics.json` from https://github.com/rust-lang/stdarch/pull/1938 a once-over just to make sure that is all correct?

adamgemmell:
The JSON looks good to me. This intrinsic isn't part of Neon but it's fine for the JSON to say that to avoid having to edit some of stdarch's tests.

As was pointed out on the original rust PR the link to our documentation 404s, but I'll see if we can get this added to that page.

traviscross:
@rfcbot fcp merge lang

(Doing the lang-side of the FCP from https://github.com/rust-lang/rust/issues/147555; will check the lang boxes already checked there.)

traviscross:
cc @RalfJung

nikomatsakis:
@rfcbot reviewed

folkertdev:
The other FCP is over as well.

PR #1997, opened 2026-01-21
use `simd_extract_dyn` for `extract`

folkertdev:
blocked on https://github.com/rust-lang/rust/pull/151453

Seems nicer to make the `dyn` part explicit, and also this does generate slightly better code for some targets.


folkertdev:
r? sayantn

PR #1999, opened 2026-01-22
arch: Add Hexagon HVX instructions

androm3da:
Note: this depends on [#151500](https://github.com/rust-lang/rust/pull/151500)
Tracking issue [#151523](https://github.com/rust-lang/rust/issues/151523)

### DIFF
--- a/crates/stdarch-gen-hexagon/src/main.rs
+++ b/crates/stdarch-gen-hexagon/src/main.rs
@@ -312,8 +312,13 @@ fn read_header(crate_dir: &Path) -> Result<String, String> {
     println!("Reading HVX header from: {}", header_path.display());
     println!("  (LLVM version: {})", LLVM_VERSION);
 
-    std::fs::read_to_string(&header_path)
-        .map_err(|e| format!("Failed to read header file {}: {}", header_path.display(), e))
+    std::fs::read_to_string(&header_path).map_err(|e| {
+        format!(
+            "Failed to read header file {}: {}",
+            header_path.display(),
+            e
+        )
+    })
 }
 
 /// Parse a C function prototype to extract return type and parameters


androm3da:
cc @folkertdev

I'm still working out how to get the CI to execute this, I should be able to make it work with `qemu-hexagon` linux-userspace emu, which is hopefully similar to how other targets do it.

folkertdev:
Yeah I figured that would be a bit of a challenge, so more efficient to do that in parallel.

folkertdev (commented):
General note: this looks largely machine-generated? We already have `stdarch-gen-loongarch` and `stdarch-gen-arm`, so it's possible to check that in.

androm3da:
> General note: this looks largely machine-generated? We already have `stdarch-gen-loongarch` and `stdarch-gen-arm`, so it's possible to check that in.

Ok, will add the generation script here too.

crates/core_arch/src/hexagon/hvx.rs
```diff
@@ -0,0 +1,7536 @@
...
+// Target feature: hvxv60
+// ============================================================================
+
+/// `Rd32=vextract(Vu32,Rs32)`
+///
+/// Instruction Type: LD
+/// Execution Slots: SLOT0
+#[inline]
+#[cfg_attr(target_arch = "hexagon", target_feature(enable = "hvxv60"))]
+#[unstable(feature = "stdarch_hexagon", issue = "none")]
```
folkertdev: For most targets we like adding a test that the intrinsic does in fact produce the expected instruction, e.g. 

```
// avx512bw
#[cfg_attr(test, assert_instr(vpabsw))]

// wasm
#[cfg_attr(test, assert_instr(v128.load8x8_s))]
```

especially if you are machine-generating these that should hopefully be straightforward.

crates/core_arch/src/hexagon/hvx.rs
```diff
@@ -0,0 +1,7536 @@
...
+    extractw(vu, rs)
+}
+
+/// `Vd32=hi(Vss32)`
+///
+/// Instruction Type: CVI_VA
+/// Execution Slots: SLOT0123
+#[inline]
+#[cfg_attr(target_arch = "hexagon", target_feature(enable = "hvxv60"))]
+#[unstable(feature = "stdarch_hexagon", issue = "none")]
```
folkertdev: before we merge this you can create a tracking issue in the main repo, see e.g. https://github.com/rust-lang/rust/issues/117427

crates/core_arch/src/hexagon/hvx.rs
```diff
@@ -0,0 +1,7536 @@
...
+
+/// `Vd32=vxor(Vu32,Vv32)`
+///
+/// Instruction Type: CVI_VA
+/// Execution Slots: SLOT0123
+#[inline]
+#[cfg_attr(target_arch = "hexagon", target_feature(enable = "hvxv60"))]
+#[unstable(feature = "stdarch_hexagon", issue = "none")]
+pub unsafe fn q6_v_vxor_vv(vu: HvxVector, vv: HvxVector) -> HvxVector {
+    vxor(vu, vv)
```
folkertdev: it's a bit hard to deduce from the names, but I'm assuming this corresponds to `simd_xor`

https://doc.rust-lang.org/nightly/std/intrinsics/simd/fn.simd_xor.html

We generally prefer using functions from `crate::intrinsics::simd` (starting with `crate` is deliberate, using `core::` can cause trouble when bootstrapping). It's much easier for us to then know what a function does, and as a bonus Miri will be able to run the intrinsics too.

When using `intrinsics::simd` it is even more important to check that these functions emit the right instruction. Sometimes additional optimizations are needed in LLVM to make that work. 

I think it's fine to do this gradually, but for some of the simpler operations it's probably possible to do it from the start.

androm3da: > for some of the simpler operations it's probably possible to do it from the start.

Sure - can do!

androm3da:
> > General note: this looks largely machine-generated? We already have `stdarch-gen-loongarch` and `stdarch-gen-arm`, so it's possible to check that in.
> 
> Ok, will add the generation script here too.

Ah - okay, I see that these are different from what I've implemented.  I have a python script that takes a C/C++ toolchain header file as a source-of-truth.  So @folkertdev  should I implement that instead as a crate that parses a yaml file?

folkertdev:
The input is up to you, `arm` uses an elaborate yaml setup, `loongarch` uses a simpler text format derived from header files. A custom format makes it easier to override an implementation with the `intrinsics::simd` one. Maybe you can use some sort of special comment to include that in the header?

We would prefer having rust over python though, hopefully that is straightforward to port?

androm3da:
> We would prefer having rust over python though, hopefully that is straightforward to port?

It is, yeah.

> Maybe you can use some sort of special comment to include that in the header?

Sorry, not sure if I follow.  Add comments in the C header file used as an input?  Or ... ?

folkertdev:
Taking `q6_v_vxor_vv` as an example, you'd need some way to indicate that instead of following the normal path picking some LLVM intrinsic,  it should use `simd_xor`. For `arm` we update its yaml file and re-run the generator. `loongarch` does not yet use `intrinsics::simd`, so it also needs to figure this out.

Maybe this was unclear, but currently the workflow is that we don't manually update the generated rust code, but instead modify the inputs to the generator and re-run it. That is convenient when new instructions are added, and it is usually simpler to edit the input than the output.

Anyhow, none of that is official, that's just what we have. So if there are good reasons to deviate from it that is something we can discuss.

androm3da:
> some way to indicate that instead of following the normal path picking some LLVM intrinsic

Ah okay I gotcha now.  So my plan was to just include a hardcoded list in the script/rust program of the intrinsics that do map to the architecture-independent ones and the exceptions get lowered to the architecture-specific ones.

folkertdev:
Cool, that should work

folkertdev:
doing a quick close/open here to see what CI says now that the features are available

folkertdev:
hmm

```
    Compiling panic_abort v0.0.0 (/rust/lib/rustlib/src/rust/library/panic_abort)
   Compiling cfg-if v1.0.4
   Compiling rustc-demangle v0.1.26
   Compiling unwind v0.0.0 (/rust/lib/rustlib/src/rust/library/unwind)
   Compiling rustc-literal-escaper v0.0.7
error: symbol 'fma' is already defined

error: could not compile `compiler_builtins` (lib) due to 1 previous error
```

is there some issue with hexagon in `compiler_builtins`? Maybe it just wasn´t considered in some `cfg` construction?

androm3da:
> 
> is there some issue with hexagon in `compiler_builtins`? Maybe it just wasn´t considered in some `cfg` construction?

This `fma` issue was a bug in `compiler_builtins` and I've since fixed it - or so I thought.

I have been noodling over the design of this PR and have several local changes.  None of them were intended to address the `fma` symbol issue but I'll push those and then I'll work on addressing that issue.

folkertdev:
Weird. We're using latest nightly here, maybe something was not synchronized?

androm3da:
> Weird. We're using latest nightly here, maybe something was not synchronized?

Yeah, it's not obvious to me yet.  This was fixed a while ago (https://github.com/rust-lang/compiler-builtins/pull/682).  I'll need to spend some time trying to reproduce it and maybe something just needs to update an explicit compiler-builtins dependency to a newer release.

folkertdev:
Oh, yes, it's building `compiler_builtins v0.1.160`, which is the latest, but it's 8 months old. I'm not sure (yet) why that dependency is built though, `cargo tree` does not show it.

androm3da:
> Oh, yes, it's building `compiler_builtins v0.1.160`, which is the latest, but it's 8 months old. I'm not sure (yet) why that dependency is built though, `cargo tree` does not show it.

`compiler_builtins-v0.1.160` has that fix, so I'm not sure if that's the issue.

tgross35:
> Oh, yes, it's building `compiler_builtins v0.1.160`, which is the latest, but it's 8 months old. I'm not sure (yet) why that dependency is built though, `cargo tree` does not show it.

Note that we don't use crates.io c-b anymore, it should only ever be coming via the submodule which is actively updated. Think this will still technically show as 0.1.160 though.

crates/core_arch/src/hexagon/hvx.rs
```diff
@@ -0,0 +1,8488 @@
...
+//! ## Vector Types
+//!
+//! HVX supports different vector lengths depending on the configuration:
+//! - 128-byte mode: `HvxVector` is 1024 bits (128 bytes)
+//! - 64-byte mode: `HvxVector` is 512 bits (64 bytes)
+//!
+//! This implementation targets 128-byte mode by default. To change the vector
+//! length mode, use the appropriate target feature when compiling:
+//! - For 128-byte mode: `-C target-feature=+hvx-length128b`
+//! - For 64-byte mode: `-C target-feature=+hvx-length64b`
```
folkertdev: This doesn't work unless you also pass `-Zbuild-std`. Normally when you compile a program, you're using a pre-built `std` that uses the default configuration.

folkertdev: On other targets, the types and intrinsics are actually available unconditionally. It is just UB to call the intrinsics that need a target feature when that feature is not present.

e.g. on x86_64 the namespacing is in the name `_mm` vs `_mm256` etc. Here maybe we should namespace with a module, so have a 64 and 128-byte module, both of which are available unconditionally, but of course need the target feature to actually use the intrinsics. 

but maybe there are additional details that I'm missing?

androm3da: > Any luck with figuring out why the build fails?

Sorry, I haven't spent enough time looking into this in the last few days.  I'll dig in and sort it out.

> This doesn't work unless you also pass `-Zbuild-std`. Normally when you compile a program, you're using a pre-built `std` that uses the default configuration.

Not for hexagon, though.  Since its artifacts aren't included, `-Zbuild-std` is a fairly convenient way to resolve that dependency.

> maybe there are additional details that I'm missing?

Yeah - it's kinda inconvenient that the ISA was designed this way.  But it was.  The semantics of the instructions change based on the system mode.  Either the system is in 64-byte mode, or it's in 128-byte mode.  The register size depends on the mode.  In practice, modern parts only support 128-byte mode.  But several DSPs were made with only 64-byte mode, and several with support for both modes.  So we're stuck with the LLVM backend looking and behaving the way it does in order to target all the various Hexagon HVX coprocs out there.  You'll see C/C++ code handle these intrinsics and their types by considering the `__HVX_LENGTH__` preprocessor definition that's emitted.  If there's a better way to handle it for rust than how it's being done here, I'm happy to consider another approach.

I'll start by documenting the above details in the crate's user documentation so that it's less confusing for all involved.

androm3da: Hmm but now re-reading what you've said I think I misunderstood your suggestion.

> Here maybe we should namespace with a module, so have a 64 and 128-byte module, both of which are available unconditionally, but of course need the target feature to actually use the intrinsics.

Yeah sorry for the confusion -- this should be do-able.

androm3da: Rebased and squashed the commits for this PR, leaving the 64b/128b crate split as its own commit to make it easier to see what changed from the previous commit set.

> Any luck with figuring out why the build fails?

Tsk!  https://github.com/rust-lang/compiler-builtins/pull/1066 -- dunno why I thought the symbol would be private but that was the issue.

examples/gaussian.rs
```diff
@@ -0,0 +1,358 @@
...
+//!         cargo +nightly build --bin gaussian -p stdarch_examples \
+//!         --target hexagon-unknown-linux-musl \
+//!         -Zbuild-std -Zbuild-std-features=llvm-libunwind
+//!
+//! To run under QEMU:
+//!
+//!     qemu-hexagon -L <sysroot>/target/hexagon-unknown-linux-musl \
+//!         target/hexagon-unknown-linux-musl/debug/gaussian
+
+#![cfg_attr(target_arch = "hexagon", feature(stdarch_hexagon))]
```
folkertdev, line 37: you could use 

```
#![target_arch = "hexagon"]
```

so that it applies to the module as a whole. Is there any particular reason not to do that, it seems like it would clean up a bunch of the conditionals.

androm3da: will do

androm3da: While making this update I remembered that there's an outstanding compiler issue that I should fix so that the example can be at parity with the intrinsics in the C/C++ reference code.  

I'll make a rustc PR for that one and revisit this after.

androm3da: Opened https://github.com/rust-lang/rust/pull/152552

I will put the `HvxVectorPair`s back into the `gaussian.rs` when that lands.

folkertdev:
I think that fix should be available now?

Another thing I thought of: we check in CI that the generated and checked-in version agree

https://github.com/rust-lang/stdarch/blob/0f754628fd9de8eb882a512281dc0067740708c9/.github/workflows/main.yml#L300-L320

So the hexagon generator should be checked there too.

androm3da:
> I think that fix should be available now?
> 
> Another thing I thought of: we check in CI that the generated and checked-in version agree
...
> So the hexagon generator should be checked there too.

Yeah - IIUC this is already present in this PR, though, right?

```
@@ -318,6 +325,10 @@ jobs:
       run: |
         cargo run --bin=stdarch-gen-loongarch --release -- crates/stdarch-gen-loongarch/lasx.spec
         git diff --exit-code
+    - name: Check hexagon
+      run: |
+        cargo run -p stdarch-gen-hexagon --release
+        git diff --exit-code
 
   conclusion:
```

crates/core_arch/src/mod.rs
```diff
@@ -320,6 +320,19 @@ pub mod arch {
...
+
+    /// Platform-specific intrinsics for the `hexagon` platform.
+    ///
+    /// This module provides intrinsics for the Qualcomm Hexagon DSP architecture,
+    /// including the Hexagon Vector Extensions (HVX).
+    ///
+    /// See the [module documentation](../index.html) for more details.
+    #[cfg(any(target_arch = "hexagon", doc))]
+    #[doc(cfg(target_arch = "hexagon"))]
+    #[unstable(feature = "stdarch_hexagon", issue = "none")]
```
folkertdev: you missed adding the issue number here I think

crates/stdarch-gen-hexagon/src/main.rs
```diff
@@ -0,0 +1,1715 @@
...
+    println!("Downloading HVX header from: {}", url);
+
+    let response = ureq::get(&url)
+        .call()
+        .map_err(|e| format!("Failed to download header: {}", e))?;
+
+    response
+        .into_string()
+        .map_err(|e| format!("Failed to read response: {}", e))
+}
```
folkertdev: I think it would be better to check in the header file? Currently any change to the header file would break our CI.

androm3da: It's bound to that particular git tag, so it wouldn't have changed under normal circumstances.  But I agree that it should be more robust, I'll add the file.

crates/core_arch/src/mod.rs
androm3da: Fixed, thanks

crates/stdarch-gen-hexagon/src/main.rs
androm3da: Fixed

androm3da:
Thanks for your patience and support @folkertdev 

Let me know if you think we should move ahead or if it needs more work.

androm3da:
> I think we're good. We just did a sync with `rust-lang/rust`, so it may be a couple of days before this makes it over (unless you'd like to use it immediately).

No hurry, thanks!

PR #2003, opened 2026-01-26
use `simd_splat`

folkertdev:
using https://github.com/rust-lang/rust/pull/151346


folkertdev:
r? sayantn

sayantn:
While we're at it, could you also change the x86 `set1` functions to use these as well

sayantn:
We have to use things like `i32x4::splat(a).as_m128i`, because `__m128i` etc are defined as `i64xN`

folkertdev:
github is having issues https://www.githubstatus.com/incidents/90hj03y5tj3c ...

PR #2007, opened 2026-01-29
Update expected instructions for LLVM 22

nikic:


nikic:
First i686 difference is due to this: https://llvm.godbolt.org/z/oGW1d66zY

Codegen for `_mm256_set_pd` has changed (which is fine, it does not correspond to a specific instruction), though llvm-mca seems to think that the new form is slower. (I have some doubts that it's correct about that though.)

nikic:
Second i686 difference: https://llvm.godbolt.org/z/41Td7oaqq

Codegen for `_mm_set_ps` changed, though here llvm-mca agrees that the new form is better.

nikic:
The aarch64 failures are all variants of this: https://llvm.godbolt.org/z/PnGdxY3rf

Where zip2 is now used instead of a lane move.

folkertdev:
https://clang.godbolt.org/z/jhPP8hdPq

for powerpc I think the new output, actually emitting `vrfin`, is correct.

.github/workflows/main.yml
```diff
@@ -40,6 +40,7 @@ jobs:
     name: Test
     runs-on: ${{ matrix.target.os }}
     strategy:
+      fail-fast: false
```
folkertdev: Did you intend to keep this in? It's probably fine, though usually not needed.

nikic: Nope, this was just for testing. Dropped it.

nikic:
I filed https://github.com/llvm/llvm-project/issues/178632 to check on the vperm2f128 change, as that seems the only potentially problematic one.

folkertdev:
Nice, thanks!

PR #2009, opened 2026-01-30
test the `vld1*` functions

folkertdev:
This adds tests for the `vld1` functions using a roundtrip. 

It also changes the generated code for unsigned `vld1` for big-endian targets. The unsigned version called the signed version, but then applied an additional shuffle. I see no reason that signedness would require a shuffle, so this is probably a quirk in the generator.

cc @adamgemmell if I'm missing anything

### DIFF
--- a/crates/core_arch/src/aarch64/neon/mod.rs
+++ b/crates/core_arch/src/aarch64/neon/mod.rs
@@ -1006,400 +1006,148 @@ mod tests {
         };
     }
 
-    #[simd_test(enable = "neon,fp16")]
-    #[cfg(not(target_arch = "arm64ec"))]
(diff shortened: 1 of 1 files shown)


Amanieu:
I'm generally happy with the changes, but the large amount of repetition is unfortunate. Is there a way we could reduce this using a macro?

folkertdev:
I cleaned it up a bunch

adamgemmell:
Thank you very much for working on this, looks good to me. There are still tests missing for all f16 loads but this is a significant chunk of work and I don't see a reason to block either this patch or the other one on that

adamgemmell:
Just checked and we should be able to get tests in for any intrinsics missing them by the end of this quarter.

PR #2014, opened 2026-02-01
Revert "Use LLVM intrinsics for `madd` intrinsics"

folkertdev:
This reverts commit 32146718741ee22ff0d54d21b9ab60353014c980.

The commit is from https://github.com/rust-lang/stdarch/pull/1985, which itself reverted this original change.

Use `intrinsics::simd` for `madd` (again), now that https://github.com/llvm/llvm-project/pull/174149 should make this optimize properly.

### DIFF
--- a/crates/core_arch/src/x86/avx2.rs
+++ b/crates/core_arch/src/x86/avx2.rs
@@ -4669,6 +4669,16 @@ mod tests {
         assert_eq_m256i(r, e);
     }
 
+    #[target_feature(enable = "avx2")]
+    #[cfg_attr(test, assert_instr(vpmaddwd))]
+    unsafe fn test_mm256_madd_epi16_mul_one(mad: __m256i) -> __m256i {
+        // This is a trick used in the adler32 algorithm to get a widening addition. The
+        // multiplication by 1 is trivial, but must not be optimized out because then the vpmaddwd
+        // instruction is no longer selected. The assert_instr verifies that this is the case.
+        let one_v = _mm256_set1_epi16(1);
+        _mm256_madd_epi16(mad, one_v)
+    }
+
     #[simd_test(enable = "avx2")]
     const fn test_mm256_inserti128_si256() {
         let a = _mm256_setr_epi64x(1, 2, 3, 4);


folkertdev:
See also https://godbolt.org/z/8z53f6WPE

r? sayantn

crates/core_arch/src/x86/avx2.rs
```diff
@@ -1813,20 +1813,24 @@ pub const fn _mm256_inserti128_si256<const IMM1: i32>(a: __m256i, b: __m128i) ->
...
+
+/// This is a trick used in the adler32 algorithm to get a widening addition. The
+/// multiplication by 1 is trivial, but must not be optimized out because then the vpmaddwd
+/// instruction is no longer selected. The assert_instr verifies that this is the case.
+#[target_feature(enable = "avx2")]
+#[cfg_attr(test, assert_instr(vpmaddwd))]
+unsafe fn test_mm256_madd_epi16_mul_one(mad: __m256i) -> __m256i {
+    let one_v = _mm256_set1_epi16(1);
+    _mm256_madd_epi16(mad, one_v)
 }
```
sayantn: Please just append this to the `test_mm256_madd_epi16` function. Or if you really want to keep it as a separate function, just move it to the `tests` module

folkertdev: The test checks that the instruction is emitted, so it can't just be added to that existing test. Locally it didn't run when I up it with the other tests, but on CI it does, so I've moved it now.

usamoi:
This seems like it would break the usage of base64. https://godbolt.org/z/q6e1ne568 [^1][^2]

I have some concerns about this kind of change. It might introduce reliance on hard-to-predict optimizer behavior and lead to performance issues that are difficult to detect.

[^1]: https://github.com/ozgb/bs64/blob/ef033b373cb08d76808da9242fe57e89b2eff002/src/avx2/mod.rs#L136
[^2]: https://github.com/aki-akaguma/basexx/blob/536976ff8aefb8299e0dba335a34198c6ed62d79/src/base64/base64_avx2.rs#L158

folkertdev:
Interesting, I think it uses a shift instead of a multiply and that fools the pattern match. Yeah that needs a fix. 

Do you have any other cases that don't optimize?

I'll revert this, and then add separate tests for the adler32 and base64 patterns.

usamoi:
> Do you have any other cases that don't optimize?

Can't find more on GitHub. https://github.com/search?q=_mm_madd_epi16+NOT+is%3Afork+language%3ARust&type=code

If not limited to real-world cases, there are also cases like `_mm256_madd_epi16(v, _mm256_set1_epi16(-1))` that might make sense.

PR #2015, opened 2026-02-01
wasm: use `intrinsics::simd` for dot product

folkertdev:
see https://godbolt.org/z/efe8EvoYd

However this fails locally for me, not sure why yet, let's see what CI says.


folkertdev:
Blocked on https://github.com/llvm/llvm-project/issues/179143 I guess. Our tests use `extern "C"` and pass the vectors in registers. Then a bitcast is needed, and that defeats the pattern recognizer.

usamoi:
This change seems to be facing the same issue as #2014. https://godbolt.org/z/h1hxvM54x

PR #2016, opened 2026-02-01
wasm: use `intrinsics::simd` for extadd

folkertdev:
See https://godbolt.org/z/6Worb476x


folkertdev:
blocked on https://github.com/llvm/llvm-project/issues/179145 (will be part of LLVM 23)

folkertdev:
no idea if this works here

@rustbot blocked

PR #2018, opened 2026-02-02
Revert "Revert "Use LLVM intrinsics for `madd` intrinsics""

folkertdev:
Reverts rust-lang/stdarch#2014

While LLVM 22 fixes the adler32 pattern of a useless multiply by one, it does not handle a pattern found in base64 where a shift is used instead of a multiply.


folkertdev:
linking https://github.com/llvm/llvm-project/pull/179326

Issue #2019, opened 2026-02-03
[AArch64] stdarch unconditionally uses AtomicI64 in RCPC3 intrinsics, breaking targets with `max-atomic-width = 0`


Snehal-Reddy:
The recent addition of RCPC3 intrinsics (specifically `vldap1_lane_s64`, `vldap1q_lane_s64`, `vstl1_lane_s64`, and `vstl1q_lane_s64`) in stdarch causes compilation failures when building core for AArch64 targets that disable 64-bit atomics (e.g., custom targets with "max-atomic-width": 0).

The core library conditionally defines AtomicI64 based on #[cfg(target_has_atomic_load_store = "64")]. However, the generated AArch64 NEON intrinsics module uses AtomicI64 unconditionally within the RCPC3  functions, leading to resolution errors when the target does not support 64-bit atomics.

### Reproduction Steps

1. Create a custom target specification file aarch64-no-atomic.json that disables atomics:

  ```
   {
       "arch": "aarch64",
       "data-layout":
   "e-m:e-p270:32:32-p271:32:32-p272:64:64-i8:8:32-i16:16:32-i64:64-i128:128-n32:64-S128-Fn32",
       "llvm-target": "aarch64-unknown-none",
       "max-atomic-width": 0,
       "atomic-cas": false,
       "linker-flavor": "gnu-lld",
       "linker": "rust-lld",
       "panic-strategy": "abort",
       "target-pointer-width": 64
   }
  ```
2. Attempt to build core using build-std with a recent Nightly toolchain (e.g., 1.95 or later):

  ```
  
     cargo +nightly build \
         -Z build-std=core \
         --target aarch64-no-atomic.json \
         --release
  ```

 Compilation fails with errors indicating AtomicI64 cannot be found:

```

   error[E0433]: failed to resolve: could not find `AtomicI64` in `atomic`
        --> library/core/src/../../stdarch/crates/core_arch/src/aarch64/neon/generated.rs:12863:43
         |
   12863 |     let atomic_src = crate::sync::atomic::AtomicI64::from_ptr(ptr as *mut i64);
         |                                           ^^^^^^^^^ could not find `AtomicI64` in `atomic`

   error[E0425]: cannot find type `AtomicI64` in module `crate::sync::atomic`
        --> library/core/src/../../stdarch/crates/core_arch/src/aarch64/neon/generated.rs:27193:55
         |
[...]

PR #2020, opened 2026-02-03
aarch64: Guard RCPC3 intrinsics with target_has_atomic = "64"

Snehal-Reddy:
The `vldap1` and `vstl1` RCPC3 intrinsics introduced in standard library unconditionally use `AtomicI64`. This breaks builds on target that do not support 64-bit atomics, such as `aarch64-unknown-none` with `max-atomic-width` set to 0. 

This commit adds a `#[cfg(target_has_atomic = "64")]` guard to these intrinsics. 

fixes rust-lang/stdarch#2019 

Open to other suggestions as well!


sayantn:
@adamgemmell can you confirm that this is not breaking. I don't think it is, as we don't have any arm targets with `max_atomic_width=0` (afaik), but still just want to confirm

adamgemmell:
No aarch64 targets, no (max_atomic_width has previously been used to disable atomics on some targets [here](https://github.com/rust-lang/rust/pull/149241)). Looks like a safe change to unbreak a custom target to me

sayantn:
@adamgemmell thanks :heart:! Then I am ok with merging this

PR #2022, opened 2026-02-05
disable s390x vector intrinsics if softfloat is enabled

fneddy:
we will add an explicit incompatibility of softfloat and vector feature in rutsc s390x-unknown-none-softfloat target specification. Therefore we need to disable vector intrinsics here to be able to compile core for this target.

=> https://github.com/rust-lang/rust/pull/151154

### DIFF
--- a/crates/core_arch/src/s390x/mod.rs
+++ b/crates/core_arch/src/s390x/mod.rs
@@ -2,6 +2,11 @@
 
 pub(crate) mod macros;
 
+/// the float and vector registers overlap therefore we cannot use any vector
+/// extensions if softfloat is enabled.
+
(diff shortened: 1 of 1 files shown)


folkertdev:
cc @uweigand you're probably following along already, just making sure you see this. 

All of  the vector types/functions are unstable currently, but them not being available with `softfloat` is something we should bring up in an eventual stabilization report.

uweigand:
> cc @uweigand you're probably following along already, just making sure you see this.
> 
> All of the vector types/functions are unstable currently, but them not being available with `softfloat` is something we should bring up in an eventual stabilization report.

Agreed.   Just for context, with GCC and clang we have the same situation where vector types and intrinsics are not available with `softfloat`.

PR #2023, opened 2026-02-14
use `intrinsics::simd` for `vmull_*`

folkertdev:
We already do the same for e.g. addition, looks like multiplication was missed.

### DIFF
--- a/crates/stdarch-gen-arm/spec/neon/arm_shared.spec.yml
+++ b/crates/stdarch-gen-arm/spec/neon/arm_shared.spec.yml
@@ -6507,13 +6507,10 @@ intrinsics:
       - ["s16", int16x4_t, int32x4_t]
       - ["s32", int32x2_t, int64x2_t]
     compose:
-      - LLVMLink:
-          name: "smull.{neon_type[1]}"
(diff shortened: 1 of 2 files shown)


folkertdev:
r? sayantn

crates/stdarch-gen-arm/spec/neon/arm_shared.spec.yml
```diff
@@ -6507,13 +6507,15 @@ intrinsics:
...
-              arch: arm
+      - Let:
+          - a
+          - '{neon_type[2]}'
+          - FnCall: [simd_cast, [a]]
+      - Let:
+          - b
+          - '{neon_type[2]}'
+          - FnCall: [simd_cast, [b]]
+      - FnCall: ['simd_mul', [a, b]]
```
sayantn: I think you can just use `simd_mul(simd_cast(a), simd_cast(b))` without the explicit types, it can be inferred from the return type

folkertdev: Yup, that's nicer

PR #2026, opened 2026-02-15
use `intrinsics::simd` for vpadd

folkertdev:
Only neon, arm does not optimize currently https://godbolt.org/z/d5MrM6ada

annoyingly 32-bit arm actually does optimize this, it's just aarch64 that doesn't. Fix in https://github.com/llvm/llvm-project/pull/181549.


Amanieu:
LGTM but needs a rebase

PR #2025, opened 2026-02-14
use `intrinsics::simd` for aarch64 deinterleaving loads

folkertdev:
Hitting https://github.com/llvm/llvm-project/issues/181514 for some `ld2` cases.

PR #2029, opened 2026-02-17
test interleaving load/store roundtrip

folkertdev:
Well, turns out some of these were extremely broken on `aarch64_be`

### DIFF
--- a/crates/core_arch/src/aarch64/neon/generated.rs
+++ b/crates/core_arch/src/aarch64/neon/generated.rs
@@ -11962,27 +11962,11 @@ pub unsafe fn vld2q_p64(a: *const p64) -> poly64x2x2_t {
 #[doc = "## Safety"]
 #[doc = "  * Neon intrinsic unsafe"]
 #[inline(always)]
-#[cfg(target_endian = "little")]
 #[target_feature(enable = "neon")]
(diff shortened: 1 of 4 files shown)


folkertdev:
cc @adamgemmell but really all this does is add more tests. The failures i encountered while making the changes indicate that this is all actually fairly well-tested.

folkertdev:
actually 

r? sayantn 

because we should probably merge this before https://github.com/rust-lang/stdarch/pull/2025

sayantn:
@folkertdev lgtm, but I am not too familiar with the semantics of ARMBE loads. This PR changes a lot of intrinsic implementations for ARMBE, and it would be nice if we could confirm that this is the correct behavior

folkertdev:
Yeah, fair. I'm not too worried here because of the intrinsic tests that do end up exercising the functions that changed, but let's see.

folkertdev:
Oh also note that I'm only actually touching the unsigned variants of these functions. They used to defer to the signed variants, but applied additional shuffling. Logically, that just does not make sense.

sayantn:
That's a nice point, at least we are making it more consistent. I don't have a problem with merging this rn, as this will unblock some other PRs.

PR #2032, opened 2026-02-18
use `intrinsics::simd` for interleaving store

folkertdev:


folkertdev:
The `fp16` variants sadly don't work on `arm` currently, so I've skipped those for now.

PR #2033, opened 2026-02-19
x86: use `simd::intrinsics` for saturating packs

okaneco:
- Use `simd::intrinsics` for `sse2`, `sse41`, `avx2`, `avx512bw`

All but one of the implementations make use of `simd_shuffle`. Some `avx512` intrinsics call the lower-target-feature intrinsics but with additional masking capability which caused some trial and error figuring out how to make the optimizer happy. Saturating packing instructions are essentially shuffles and LLVM can recognize a lot of these patterns by now.

Combined with masked stores, instruction tests routinely failed unless using shuffles which is probably the lack of being able to see through the clamping and truncating as in the truncating conversion stores issue. This same strategy could probably be used to get more of the saturating masked truncation instructions to pass.

`_mm_packs_epi32` was the single case that failed to optimize at all unless I wrote it without a shuffle.


sayantn:
It lgtm too, just one point - all the intrinsics can now be marked `const` (just remember to mark the corresponding tests `const` too)

folkertdev:
That's better as a follow-up I think

PR #2034, opened 2026-02-19
 x86: use `simd::intrinsics` for bitshifts 

okaneco:
- Use intrinsics in `sse2`, `avx2`, `avx512f`, and `avx512bw`
- Mark intrinsics and tests that call these functions as const

Nothing notable about the implementation.

Intrinsics take `count` as the shift argument with the shift being a `u64` in the low 64-bits of an `__m128i`. The shift amount must be less than the bit width of the type being shifted.

### DIFF
--- a/crates/core_arch/src/x86/sse2.rs
+++ b/crates/core_arch/src/x86/sse2.rs
@@ -577,8 +577,17 @@ pub const fn _mm_slli_epi16<const IMM8: i32>(a: __m128i) -> __m128i {
 #[target_feature(enable = "sse2")]
 #[cfg_attr(test, assert_instr(psllw))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
-pub fn _mm_sll_epi16(a: __m128i, count: __m128i) -> __m128i {
-    unsafe { transmute(psllw(a.as_i16x8(), count.as_i16x8())) }
+#[rustc_const_unstable(feature = "stdarch_const_x86", issue = "149298")]
+pub const fn _mm_sll_epi16(a: __m128i, count: __m128i) -> __m128i {
+    let shift = count.as_u64x2().as_array()[0];
+    unsafe {
+        if shift >= 16 {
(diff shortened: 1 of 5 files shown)


crates/core_arch/src/x86/sse2.rs
```diff
@@ -577,8 +577,17 @@ pub const fn _mm_slli_epi16<const IMM8: i32>(a: __m128i) -> __m128i {
 #[target_feature(enable = "sse2")]
 #[cfg_attr(test, assert_instr(psllw))]
 #[stable(feature = "simd_x86", since = "1.27.0")]
-pub fn _mm_sll_epi16(a: __m128i, count: __m128i) -> __m128i {
-    unsafe { transmute(psllw(a.as_i16x8(), count.as_i16x8())) }
+#[rustc_const_unstable(feature = "stdarch_const_x86", issue = "149298")]
+pub const fn _mm_sll_epi16(a: __m128i, count: __m128i) -> __m128i {
+    let shift = count.as_u64x2().as_array()[0];
+    unsafe {
+        if shift >= 16 {
```
Amanieu, line 584: Can you check that LLVM is actually able to eliminate this check and fold it into the `psllw` instruction? It might still be keeping the check and only doing a `psllw` on the else branch.

okaneco: Good point, I assumed the instruction threshold tests were low enough to catch that.

But it looks like neither left nor right shifts fold into the instructions we want. Branches for the left shifts and cmov for the right shifts.
https://rust.godbolt.org/z/rzsehj5b8

folkertdev: Can you open an LLVM issue for this? (those don't always get fixed, but as it stands it basically certain that nothing will change)

okaneco: Sure, reported https://github.com/llvm/llvm-project/issues/192221

PR #2035, opened 2026-02-20
x86: Followup to add const for pack intrinsics and tests

okaneco:
Adds const to the functions and tests rewritten with intrinsics in https://github.com/rust-lang/stdarch/pull/2033 as mentioned here https://github.com/rust-lang/stdarch/pull/2033#issuecomment-3928028941


folkertdev (approved):
r? sayantn if you have notes, but this looks straightforward

sayantn:
Lgtm too, thanks ❤️

PR #2063, opened 2026-03-16
Redesign `intrinsic-test` to use simple comparison

sayantn:
Currently `intrinsic-test` prints the outputs and then compares the outputs manually. This PR uses a different approach -- generate C wrappers for the intrinsics, link to them from Rust, and then just use simple rust tests to compare outputs

It is much easier to review commit-by-commit


sayantn:
```
---- test_vdupq_n_f16 stdout ----

thread 'test_vdupq_n_f16' (2187) panicked at mod_0/src/lib.rs:13773:17:
assertion `left == right` failed: 
  left: [NiceF16(0.0), NiceF16(0.0), NiceF16(0.0), NiceF16(0.0), NiceF16(0.0), NiceF16(0.0), NiceF16(0.0), NiceF16(0.0)]
 right: [NiceF16(0.0), NiceF16(0.0), NiceF16(0.0), NiceF16(0.0), NiceF16(1.43e-5), NiceF16(0.0), NiceF16(-50430.0), NiceF16(1.79e-5)]
```

This seems weird (`left` is the Rust output, `right` is the C one, and `NiceF16` is a wrapper which implements `PartialEq` as `a == b || (a.is_nan() && b.is_nan())`). This looks like ABI-related issue. For reference, the declaration looks like
```rust
unsafe extern "C" {
    fn vdup_n_f16_wrapper(value: f16) -> float16x4_t;
}
```
In fact most `f16` tests fail in armv7. @folkertdev can you help?

Edit:

To work around this issue I have modified the tool to communicate with C via pointers (e.g. the C wrapper for `_mm_add_ps` looks like `void _mm_add_ps_wrapper(__m128 *dst, const __m128* a, const __m128* b)`. This fixed the AArch64 and ARMv7 problems, but now the AArch64BE tests are failing, because apparently C and Rust have different pointer load semantics for matrix-like vectors (e.g. `uint64x2x2_t`) https://godbolt.org/z/j1d16z1P9

@Amanieu is this intended behavior or a bug?

sayantn:
Btw the time gains are significant, it reduces the Arm and aarch64 times to 2-3 minutes, and the full x86 run (we did 20% previously) to around 12 mins for release and 17 mins for dev

folkertdev:
Great work. Quick sanity check on `f16`, perhaps we're still using LLVM 21 to compile the C? If LLVM 21 is used (and also on windows apparently still with LLVM 22) then on some targets the ABI is inconsistent.

sayantn:
@folkertdev ooh, that makes sense. I don't particularly care about windows, but we are using LLVM20 in the CI. I can change it to use the build from kernel.org

folkertdev:
I'm seeing `clang-18` here even https://triage.rust-lang.org/gha-logs/rust-lang/stdarch/67182959392?pr=2063. I'm not sure what the best solution is really. You could ask T-infra if they have ideas.

sayantn:
yeah, but I can use the LLVM github builds or the kernel.org builds

tgross35:
Can f16 tests just be gated with `#[cfg(target_has_reliable_f16)]`? That's likely easier than working around the Windows and old LLVM failures.

sayantn:
@tgross35 the f16 tests are mostly fine now. More concerning is that a lot of tests are failing in all 3 arm archs, e.g. `vzipq`. The C version seems to return all zeros

edit: sorry, my mistake, they are still failing in ARMv7. I will gate them against the flag

folkertdev:
With LLVM 22 `f16` should work on `armv7` though?

tgross35:
FTZ/DAZ-related perhaps?

sayantn:
> FTZ/DAZ-related perhaps?

I don't really think so, the outputs seem completely distinct.

I noticed that `vzipq` etc was failing so I tried out the assemblies.
In aarch64_be-unknown-linux-gnu,
```rust
use core::arch::aarch64::*;

#[unsafe(no_mangle)]
pub unsafe extern "C" fn foo(dst: *mut uint8x16x2_t, a: *const uint8x16_t, b: *const uint8x16_t) {
    unsafe {
        *dst = vzipq_u8(*a, *b);
    }
}
```
produces
```asm
foo:
        ld1 { v0.16b }, [x1]
        ld1 { v1.16b }, [x2]
        add x8, x0, #16
        zip1 v2.16b, v0.16b, v1.16b
        zip2 v0.16b, v0.16b, v1.16b
        st1 { v2.16b }, [x0]
        st1 { v0.16b }, [x8]
        ret
```

But the C code seemingly has different behavior on GCC and clang https://godbolt.org/z/T3YnrejjG

@adamgemmell can you help in this?

adamgemmell:
I'm not sure it will fix your issue but the difference in instructions comes from the fact that in arm_neon.h, they reverse every vector before and after the operation on big endian. It's not always actually necessary so we only do it if it's broken without it - however, the intrinsic test tool doesn't detect the difference in behaviour because both arguments it picks are identical.

e.g.:
```
| a               | 15 | 14 | 13 | ... | 2  | 1  | 0  |
| b               | 31 | 30 | 29 | ... | 18 | 17 | 16 |
| a = rev(a)      | 0  | 1  | 2  | ... | 13 | 14 | 15 |
| b = rev(b)      | 16 | 17 | 18 | ... | 29 | 30 | 31 |
| ret = zip(a, b) | 0  | 16 | 1  | ... | 30 | 15 | 31 |
| rev(ret)        | 31 | 15 | 30 | ... | 1  | 16 | 0  |
```

adamgemmell:
You can try adding `big_endian_inverse: true` to the definition in the yaml, regenerating to see if it changes anything. We can probably offset the pointers to the values array for arguments slightly to ensure they're different

Also I don't actually see vzipq_u8 on the [latest CI run](https://github.com/rust-lang/stdarch/actions/runs/23181252828/job/67354518123?pr=2063), why is that?

sayantn:
> Also I don't actually see vzipq_u8 on the [latest CI run](https://github.com/rust-lang/stdarch/actions/runs/23181252828/job/67354518123?pr=2063), why is that?

I have no idea, I can confirm that locally the test is generated and run.

> You can try adding big_endian_inverse: true to the definition in the yaml, regenerating to see if it changes anything. We can probably offset the pointers to the values array for arguments slightly to ensure they're different

I will check. Thanks

Edit: @adamgemmell adding `big_endian_inverse: true` to the functions seem to work. The only question remains is that is it the correct behavior, or is clang buggy here

Edit2: sorry, `vzipq_u8` is not getting tested even locally. I will look into it

adamgemmell:
None of the unsigned variants of vzipq seem to be seen there, weird.

I'd quite like to know why this patch detects the difference - when I looked locally the codegen of the tests seemed very similar

sayantn:
Yeah I fixed the test not being included, I used `/` instead of `div_ceil` when computing the chunk sizes, so the last module wasn't getting included 😅

sayantn:
The x86 failures are very weird. The `fma` ones are just rounding errors (even in GCC), so something related to FTZ/DAZ probably. But the `fixupimm` tests are weirder, they fail even in gcc, so something is probably wrong.

edit: we did have a bug in the `fixupimm` implementations! I am surprised that even `intrinsic-test` didn't catch it yet!

davidtwco (approved):
I've worked with @sayantn to fix some of the AArch64 failures from this (and they have links to patches that should resolve the remaining ones) - I think this is a great change and would love to see it land.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2063/ef18e28f8d2ca0c20ab7b053786b84af14b0e20a..185efd734e6b3a38f40f5ade6959dc4133c835aa)*

crates/stdarch-gen-arm/spec/neon/arm_shared.spec.yml
```diff
@@ -1751,14 +1764,15 @@ intrinsics:
...
...md_shuffle!(a, b, [3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18]), 4 => simd_shuffle!(a, b, [4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19]), 5 => simd_shuffle!(a, b, [5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20]), 6 => simd_shuffle!(a, b, [6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21]), 7 => simd_shuffle!(a, b, [7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22]), 8 => simd_shuffle!(a, b, [8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23]), 9 => simd_shuffle!(a, b, [9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24]), 10 => simd_shuffle!(a, b, [10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25]), 11 => simd_shuffle!(a, b, [11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26]), 12 => simd_shuffle!(a, b, [12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27]), 13 => simd_shuffle!(a, b, [13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28]), 14 => simd_shuffle!(a, b, [14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29]), 15 => simd_shuffle!(a, b, [15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30]), _ => unreachable_unchecked(), }']
```
folkertdev: any chance we can lift this pattern out into a normal function that gets imported and used?

folkertdev (commented):
I'm very excited for it too!

Overall I do think it would help to extract what is feasible into separate PRs. E.g. some of the dockerfile and yml changes at least seem like they could be merged separately.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2063/ef18e28f8d2ca0c20ab7b053786b84af14b0e20a..185efd734e6b3a38f40f5ade6959dc4133c835aa)*

sayantn:
@folkertdev @davidtwco I will be splitting the commits into a few PRs, e.g. some CI changes and some changes to stdarch-gen-arm that can be landed independently. I have already put up 5 PRs, I will see if I can split more

sayantn:
@Amanieu @folkertdev sorry for the inconvenience, I will split up just one more PR (last one I promise). Figured out how to split the `vcmla` changes

@rustbot author

crates/intrinsic-test/src/common/gen_rust.rs
```diff
@@ -4,18 +4,50 @@ use super::indentation::Indentation;
...
+            fn eq(&self, other: &Self) -> bool {
+                self.0 == other.0 || (self.0.is_nan() && other.0.is_nan())
+            }
+        }
+
+        impl Eq for $wrapper {}
+    )*}
+}
+
+make_nice!(NiceF16(f16), NiceF32(f32), NiceF64(f64));
```
folkertdev: What is "nice" here? is it really just partialeq wrapper? Maybe there is a better name for that? Or just document that that is the goal.

crates/intrinsic-test/src/common/mod.rs
```diff
@@ -38,15 +38,17 @@ pub trait SupportedArchitectureTest {
...
-        let (chunk_size, chunk_count) = manual_chunk(self.intrinsics().len(), 400);
+        let (chunk_size, _chunk_count) = manual_chunk(self.intrinsics().len());
 
         std::fs::create_dir_all("c_programs").unwrap();
         self.intrinsics()
             .par_chunks(chunk_size)
             .enumerate()
             .map(|(i, chunk)| {
-                let c_filename = format!("c_programs/wrapper_{i}.cpp");
+                let c_filename = format!("c_programs/wrapper_{i}.c");
```
folkertdev, line 51: oh are we just C now (no c++ any more)?

crates/intrinsic-test/missing_x86.txt
```diff
@@ -49,8 +49,19 @@ _umwait
 _xgetbv
 
 # top bits are undefined, unclear how to test these
+_mm256_castph128_ph256
+_mm256_castps128_ps256
+_mm256_castpd128_pd256
```
folkertdev, line 54: and we just didn't notice before?

folkertdev (approved):
Some questions/notes but overall this looks good and is very exciting!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2063/7b11ce8eaeb7ae156279fca18b30944099c20c71..abcaa7ebe14a9ebc68358039454d49eab7a87148)*

crates/intrinsic-test/src/common/gen_rust.rs
sayantn: yes, it is just wrapping PartialEq to deal with NaNs, I couldn't come up with a nice name so just named it `Nice` lol. Any suggestions for the name?

crates/intrinsic-test/src/common/mod.rs
sayantn: yea, C++ was used previously to take advantage of some templates, but C typically has much better compile times, and linking to C++ from Rust is weird af

crates/intrinsic-test/missing_x86.txt
sayantn: Actually not sure how, maybe the C compiler constant-folding some `extract`s to 0 (it was UB so the compiler could do anything)

crates/intrinsic-test/src/common/gen_rust.rs
folkertdev: idk, `CustomPartialEqF16` etc?

crates/intrinsic-test/src/common/mod.rs
folkertdev: yes, this is great!

crates/intrinsic-test/src/common/gen_rust.rs
sayantn: Decided to go with `NanEqF32` and added a small comment

PR #2072, opened 2026-04-08
hexagon: Add scalar DSP intrinsics

androm3da:
Add scalar (non-HVX) intrinsic wrappers for Hexagon DSP operations covering arithmetic, multiply, shift, saturate, compare, and floating-point instructions, mapping to llvm.hexagon.* LLVM builtins.

Includes the stdarch-gen-hexagon-scalar code generator crate which produces scalar.rs from the LLVM hexagon_protos.h header file, and a CI step to verify the generated output stays in sync.

### DIFF
--- a/crates/core_arch/src/hexagon/scalar.rs
+++ b/crates/core_arch/src/hexagon/scalar.rs
@@ -1,0 +1,12181 @@
+//! Hexagon scalar intrinsics
+//!
+//! This module provides intrinsics for scalar (non-HVX) Hexagon DSP operations,
+//! including arithmetic, multiply, shift, saturate, compare, and floating-point
+//! operations.
+//!
+//! [Hexagon V68 Programmer's Reference Manual](https://docs.qualcomm.com/doc/80-N2040-45)
+//!
+//! ## Naming Convention
+//!
+//! Function names preserve the original Q6 naming case because the convention
+//! uses case to distinguish register types:
+//! - `P` (uppercase) = 64-bit register pair (`Word64`)
+//! - `p` (lowercase) = predicate register (`Byte`)
+//!
+//! For example, `Q6_P_and_PP` operates on 64-bit pairs while `Q6_p_and_pp`
+//! operates on predicate registers.
+//!
+//! ## Architecture Versions
+//!
+//! Most scalar intrinsics are available on all Hexagon architectures.
+//! Some intrinsics require specific architecture versions (v60, v62, v65,
+//! v66, v67, v68, or v67+audio) and carry
+//! `#[target_feature(enable = "v68")]` (or the appropriate version).
+//! Enable these with `-C target-feature=+v68` or by setting the target CPU
+//! via `-C target-cpu=hexagonv68`.
+//!
+//! Each version includes all features from previous versions.
+
+#![allow(non_snake_case)]
+
+#[cfg(test)]
+use stdarch_test::assert_instr;
+
+// LLVM intrinsic declarations for Hexagon scalar operations
+#[allow(improper_ctypes)]
+unsafe extern "unadjusted" {
+    #[link_name = "llvm.hexagon.A2.abs"]
+    fn hexagon_A2_abs(_: i32) -> i32;
+    #[link_name = "llvm.hexagon.A2.absp"]
+    fn hexagon_A2_absp(_: i64) -> i64;
+    #[link_name = "llvm.hexagon.A2.abssat"]
+    fn hexagon_A2_abssat(_: i32) -> i32;
+    #[link_name = "llvm.hexagon.A2.add"]
+    fn hexagon_A2_add(_: i32, _: i32) -> i32;
+    #[link_name = "llvm.hexagon.A2.addh.h16.hh"]
+    fn hexagon_A2_addh_h16_hh(_: i32, _: i32) -> i32;
+    #[link_name = "llvm.hexagon.A2.addh.h16.hl"]
+    fn hexagon_A2_addh_h16_hl(_: i32, _: i32) -> i32;
+    #[link_name = "llvm.hexagon.A2.addh.h16.lh"]
+    fn hexagon_A2_addh_h16_lh(_: i32, _: i32) -> i32;
+    #[link_name = "llvm.hexagon.A2.addh.h16.ll"]
+    fn hexagon_A2_addh_h16_ll(_: i32, _: i32) -> i32;
+    #[link_name = "llvm.hexagon.A2.addh.h16.sat.hh"]
+    fn hexagon_A2_addh_h16_sat_hh(_: i32, _: i32) -> i32;
+    #[link_name = "llvm.hexagon.A2.addh.h16.sat.hl"]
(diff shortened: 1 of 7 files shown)


androm3da:
cc @quic-mliebel

crates/core_arch/src/hexagon/scalar.rs
```diff
@@ -0,0 +1,10901 @@
...
+//! operations.
+//!
+//! [Hexagon V68 Programmer's Reference Manual](https://docs.qualcomm.com/doc/80-N2040-45)
+//!
+//! ## Naming Convention
+//!
+//! Function names preserve the original Q6 naming case because the convention
+//! uses case to distinguish register types:
+//! - `P` (uppercase) = 64-bit register pair (`Word64`)
+//! - `p` (lowercase) = predicate register (`Byte`)
```
androm3da, line 14: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2072/5c1c65dfaafacf702be35a452dccb3a8eb10e437..b9092c479359429ad524d9f0c412b6460b994e5b)*

Because of this, I think it makes sense to revisit the HVX intrinsics as case-matching too (instead of squashing to lowercase).

It's not stable yet and hasn't been available for long, so it doesn't seem too bad to change it now.  If it helps we could alias the lowercase version for ~some time, but that's not my preference.

quic-mliebel:
Should the scalar intrinsics have cfg_attr like HVX intrinsics have?

For example:

```
#[cfg_attr(target_arch = "hexagon", target_feature(enable = "hvxv79"))]
#[cfg_attr(test, assert_instr(vfneg_f8))]
```

The generator could generate those lines.

androm3da:
> Should the scalar intrinsics have cfg_attr like HVX intrinsics have?
> 
> For example:
> 
> ```
> #[cfg_attr(target_arch = "hexagon", target_feature(enable = "hvxv79"))]
> #[cfg_attr(test, assert_instr(vfneg_f8))]
> ```
> 
> The generator could generate those lines.

Some of them do have a dependency, yes.  I'll add this.

crates/core_arch/src/hexagon/scalar.rs
androm3da: #2074

quic-mliebel:
> > Should the scalar intrinsics have cfg_attr like HVX intrinsics have?
> > For example:
> > ```
> > #[cfg_attr(target_arch = "hexagon", target_feature(enable = "hvxv79"))]
> > #[cfg_attr(test, assert_instr(vfneg_f8))]
> > ```
> >   
> > The generator could generate those lines.
> 
> Some of them do have a dependency, yes. I'll add this.

What about assert_instr?

androm3da:
> > > Should the scalar intrinsics have cfg_attr like HVX intrinsics have?
> > > For example:
> > > ```
> > > #[cfg_attr(target_arch = "hexagon", target_feature(enable = "hvxv79"))]
> > > #[cfg_attr(test, assert_instr(vfneg_f8))]
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > The generator could generate those lines.
> > 
> > 
> > Some of them do have a dependency, yes. I'll add this.
> 
> What about assert_instr?

yeah - we need that.  Will fix it.

And we need target features to gate some of these intrinsic too, I'll fix the compiler to provide those.

androm3da:
>  we need target features to gate some of these intrinsic too, I'll fix the compiler to provide those.

https://github.com/rust-lang/rust/pull/155014

.github/workflows/main.yml
```diff
@@ -331,6 +331,10 @@ jobs:
       run: |
         cargo run -p stdarch-gen-hexagon --release
         git diff --exit-code
+    # - name: Check hexagon scalar
+    #   run: |
+    #     cargo run -p stdarch-gen-hexagon-scalar --release
+    #     git diff --exit-code
```
Amanieu: Why is this being added here if it is just commented out?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2072/095646219d35769f013c4a54d843a982c87d979c..c6fe9373f067c9f9f363459c46a9e135a59a1232)*

androm3da:
> > we need target features to gate some of these intrinsic too, I'll fix the compiler to provide those.
> 
> [rust-lang/rust#155014](https://github.com/rust-lang/rust/pull/155014)

Now using the target features and static assertions regarding immediate value ranges.

.github/workflows/main.yml
androm3da: Sorry!  Should've made a comment about that one.

https://github.com/rust-lang/rust/issues/154686 caused CI failures here and was still pending the fix.

IIUC https://github.com/rust-lang/rust/pull/155332 landed the `libc` fix just recently.  So on one hand, it might make sense to enable this CI.   OTOH, there was some discussion regarding "does it make sense to have a tier 3 target in stdarch CI?" and I suspect that maybe the conclusion was "no"?  @folkertdev 

In the short term, maybe we'll omit hexagon build/test from stdarch CI.  In the coming ~weeks I will make an MCP to promote `hexagon-unknown-linux-musl` to tier 2.  I imagine that will take some time to converge, after that we could revisit this.

So, with respect to `.github/workflows/main.yml` -- I'll just remove the commented-out changes for now so there's no confusion?

quic-mliebel (approved):

.github/workflows/main.yml
Amanieu: That's fine, but eventually it would be nice to have this tested in CI. The general policy for tier 3 targets is that we'll just disable the CI if any issue comes up, and it's up to the target maintainers to submit a follow-up PR to fix and re-enable CI for that target.

This can be done in a separate PR.

crates/core_arch/src/hexagon/scalar.rs
```diff
@@ -0,0 +1,12181 @@
...
+}
+
+/// `Rxx32+=mpyu(Rs32.l,Rt32.l):<<1`
+///
+/// Instruction Type: M
+/// Execution Slots: SLOT23
+#[inline(always)]
+#[cfg_attr(test, assert_instr(mpyu))]
+#[unstable(feature = "stdarch_hexagon", issue = "151523")]
+pub unsafe fn Q6_P_mpyuac
```
taiki-e: This and 55 similar code cause compile errors.
```
  error: cannot use `#[inline(always)]` with `#[target_feature]`
       --> /home/runner/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/../../stdarch/crates/core_arch/src/hexagon/scalar.rs:11428:1
        |
  11428 | #[inline(always)]
        | ^^^^^^^^^^^^^^^^^
        |
        = note: See this issue for full discussion: https://github.com/rust-lang/rust/issues/145574
  ```

https://github.com/taiki-e/atomic-maybe-uninit/actions/runs/26321777148/job/77492694467#step:4:935

PR #2075, opened 2026-04-09
loongarch: Avoid constant folding in tests to ensure SIMD coverage

heiher:
Use `black_box` on SIMD intrinsic inputs to prevent the compiler from constant folding SIMD operations, ensuring the corresponding SIMD instructions are actually emitted and covered by tests.

### DIFF
--- a/crates/core_arch/src/loongarch64/lasx/tests.rs
+++ b/crates/core_arch/src/loongarch64/lasx/tests.rs
@@ -5,6 +5,7 @@ use crate::{
     core_arch::{loongarch64::*, simd::*},
     mem::transmute,
 };
+use std::hint::black_box;
 use stdarch_test::simd_test;
 
 #[simd_test(enable = "lasx")]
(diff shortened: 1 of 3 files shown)


crates/core_arch/src/loongarch64/lasx/tests.rs
```diff
@@ -5,17 +5,18 @@ use crate::{
...
     mem::transmute,
 };
+use std::hint::black_box;
 use stdarch_test::simd_test;
 
 #[simd_test(enable = "lasx")]
 unsafe fn test_lasx_xvsll_b() {
     let a = i8x32::new(
-        -111, -98, 47, -106, -82, -72, -70, 0, 110, -61, -20, 36, 41, -103, 42, 95, 15, -11,
-        -25, -5, 40, -63, 56, -39, 43, 127, 86, 75, -48, -32, 72, 69,
```
sayantn, line 14: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2075/c80c8f49ae9f2053ce990411da343f2ab2744d60..4bdc28134ac005b8156076d57f05551cca04a9c6)*

that's weird, why did this formatting change?? The changes only added `black_box`!

heiher: Did the new rustfmt version change the default settings? All I did was regenerate and run rustfmt.

heiher: This was caused by a change in rustfmt's max-width. I've reformatted the code, and the diff should be clean now.

PR #2076, opened 2026-04-10
Remove `simd_ty!` and `simd_m_ty!` macros

eduardosm:
For `SimdM`, since there were only a few uses, I replaced all uses of `new` with `from_array`.

For `Simd`, since there are thousands of uses (searching for `[iuf][0-9]+x[0-9]+::new` throws 6k results), I manually re-implemented `new` for 1-, 2-, 4-, 8-, 16- and 32-element vectors using generic `impl`s:

```rust
impl<T: SimdElement> Simd<T, 4> {
    #[inline]
    pub(crate) const fn new(x0: T, x1: T, x2: T, x3: T) -> Self {
        Self([x0, x1, x2, x3])
    }
}
```

For 64-element vectors, I just replaced the single use of `new` I found with `from_array`.


folkertdev:
Neat, thanks!

PR #2078, opened 2026-04-11
loongarch: Use `intrinsics::simd` for selected LSX/LASX intrinsics

heiher:
This change migrates a subset of LSX/LASX intrinsics to the portable `intrinsics::simd` interface.

Only straightforward mappings are converted in this patch. Intrinsics that require more complex transformations or currently result in suboptimal code generation are intentionally left unchanged and will be migrated incrementally in follow-up patches.


crates/stdarch-gen-loongarch/src/main.rs
```diff
@@ -90,6 +91,14 @@ impl TargetFeature {
     }
 }
 
+fn generic_intrinsics() -> HashSet<&'static str> {
```
folkertdev: At least I have been calling the functions from `intrinsics::simd`  "portable intrinsics", in that they have the same behavior across targets.

Generic has a similar meaning, but that word is already used all over the compiler code base for generic types.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2078/c80c8f49ae9f2053ce990411da343f2ab2744d60..3715c6915c16f2a3c06df07c0bad258ad538bbc3)*

heiher: Sounds good to me to call them "portable intrinsics" here.

heiher:
The recent changes are just minor impl tweaks.

For the intrinsics newly marked as portable, Is it better to include them in this patch, or put them in a separate PR?

crates/core_arch/src/loongarch64/lasx/portable.rs
```diff
@@ -0,0 +1,201 @@
...
+impl_vvv!("lasx", lasx_xvdiv_wu, simd_div, m256i, u32x8);
+impl_vvv!("lasx", lasx_xvdiv_du, simd_div, m256i, u64x4);
+impl_vvv!("lasx", lasx_xvmod_b, simd_rem, m256i, i8x32);
+impl_vvv!("lasx", lasx_xvmod_h, simd_rem, m256i, i16x16);
+impl_vvv!("lasx", lasx_xvmod_w, simd_rem, m256i, i32x8);
+impl_vvv!("lasx", lasx_xvmod_d, simd_rem, m256i, i64x4);
+impl_vvv!("lasx", lasx_xvmod_bu, simd_rem, m256i, u8x32);
+impl_vvv!("lasx", lasx_xvmod_hu, simd_rem, m256i, u16x16);
+impl_vvv!("lasx", lasx_xvmod_wu, simd_rem, m256i, u32x8);
+impl_vvv!("lasx", lasx_xvmod_du, simd_rem, m256i, u64x4);
```
sayantn, line 96: According to [this reference](https://jia.je/unofficial-loongarch-intrinsics-guide/viewer/?q=xvdiv), it should return 0 if the divisor is 0, but `simd_rem` and `simd_div` doesn't allow 0 as a divisor at all

crates/core_arch/src/loongarch64/lasx/portable.rs
```diff
@@ -0,0 +1,201 @@
...
+impl_vvv!("lasx", lasx_xvnor_v, simdl_nor, m256i, u8x32);
+impl_vvv!("lasx", lasx_xvxor_v, simd_xor, m256i, u8x32);
+impl_vvv!("lasx", lasx_xvfadd_s, simd_add, m256, f32x8);
+impl_vvv!("lasx", lasx_xvfadd_d, simd_add, m256d, f64x4);
+impl_vvv!("lasx", lasx_xvfsub_s, simd_sub, m256, f32x8);
+impl_vvv!("lasx", lasx_xvfsub_d, simd_sub, m256d, f64x4);
+impl_vvv!("lasx", lasx_xvfmul_s, simd_mul, m256, f32x8);
+impl_vvv!("lasx", lasx_xvfmul_d, simd_mul, m256d, f64x4);
+impl_vvv!("lasx", lasx_xvfdiv_s, simd_div, m256, f32x8);
+impl_vvv!("lasx", lasx_xvfdiv_d, simd_div, m256d, f64x4);
```
sayantn, line 110: same problems here,`simd_div` doesn't allow 0 divisors

crates/core_arch/src/loongarch64/lasx/portable.rs
```diff
@@ -0,0 +1,201 @@
...
+impl_vuv!("lasx", lasx_xvslli_w, simd_shl, m256i, i32x8);
+impl_vuv!("lasx", lasx_xvslli_d, simd_shl, m256i, i64x4);
+impl_vuv!("lasx", lasx_xvsrai_b, simd_shr, m256i, i8x32);
+impl_vuv!("lasx", lasx_xvsrai_h, simd_shr, m256i, i16x16);
+impl_vuv!("lasx", lasx_xvsrai_w, simd_shr, m256i, i32x8);
+impl_vuv!("lasx", lasx_xvsrai_d, simd_shr, m256i, i64x4);
+impl_vuv!("lasx", lasx_xvsrli_b, simd_shr, m256i, u8x32);
+impl_vuv!("lasx", lasx_xvsrli_h, simd_shr, m256i, u16x16);
+impl_vuv!("lasx", lasx_xvsrli_w, simd_shr, m256i, u32x8);
+impl_vuv!("lasx", lasx_xvsrli_d, simd_shr, m256i, u64x4);
```
sayantn: `simd_shl` and `simd_shr` doesn't allow the shift amount to be more that the bit width of the element, but the reference says it should behave as wrapping (i.e. `shift % bit_len`). This should be a straightforward wrapper around `simd_sh{l,r}`

crates/core_arch/src/loongarch64/lsx/portable.rs
```diff
@@ -0,0 +1,207 @@
...
+impl_vvv!("lsx", lsx_vdiv_wu, simd_div, m128i, u32x4);
+impl_vvv!("lsx", lsx_vdiv_du, simd_div, m128i, u64x2);
+impl_vvv!("lsx", lsx_vmod_b, simd_rem, m128i, i8x16);
+impl_vvv!("lsx", lsx_vmod_h, simd_rem, m128i, i16x8);
+impl_vvv!("lsx", lsx_vmod_w, simd_rem, m128i, i32x4);
+impl_vvv!("lsx", lsx_vmod_d, simd_rem, m128i, i64x2);
+impl_vvv!("lsx", lsx_vmod_bu, simd_rem, m128i, u8x16);
+impl_vvv!("lsx", lsx_vmod_hu, simd_rem, m128i, u16x8);
+impl_vvv!("lsx", lsx_vmod_wu, simd_rem, m128i, u32x4);
+impl_vvv!("lsx", lsx_vmod_du, simd_rem, m128i, u64x2);
```
sayantn, line 96: same problems here

crates/core_arch/src/loongarch64/lsx/portable.rs
```diff
@@ -0,0 +1,207 @@
...
+impl_vuv!("lsx", lsx_vslli_w, simd_shl, m128i, i32x4);
+impl_vuv!("lsx", lsx_vslli_d, simd_shl, m128i, i64x2);
+impl_vuv!("lsx", lsx_vsrai_b, simd_shr, m128i, i8x16);
+impl_vuv!("lsx", lsx_vsrai_h, simd_shr, m128i, i16x8);
+impl_vuv!("lsx", lsx_vsrai_w, simd_shr, m128i, i32x4);
+impl_vuv!("lsx", lsx_vsrai_d, simd_shr, m128i, i64x2);
+impl_vuv!("lsx", lsx_vsrli_b, simd_shr, m128i, u8x16);
+impl_vuv!("lsx", lsx_vsrli_h, simd_shr, m128i, u16x8);
+impl_vuv!("lsx", lsx_vsrli_w, simd_shr, m128i, u32x4);
+impl_vuv!("lsx", lsx_vsrli_d, simd_shr, m128i, u64x2);
```
sayantn, line 135: and here

crates/core_arch/src/loongarch64/simd.rs
```diff
@@ -0,0 +1,415 @@
+//! LoongArch64 SIMD helpers
+
+pub(super) const trait SimdL: Sized {
+    type Elem;
+
+    unsafe fn splat(v: i64) -> Self;
+}
```
sayantn, line 7: currently we represent all these `core_arch::simd` types as monomorphizations of a single `Simd<T, N>` struct, so you might be able to shorten this impl list a bit. If that's not feasible, you can use a macro to implement these, they are all identical

sayantn (changes requested):
Some small changes required for soundness, otherwise LGTM.
Also sorry for the late review

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2078/0414f7c9d9867eb98baeae592a88ab303eb0ed1b..5a234facda86ea7af17e5c60bea7bf535bd0d247)*

crates/core_arch/src/loongarch64/lasx/portable.rs
heiher: I checked the official SIMD reference, and for integer division (including mod), the result when the divisor is 0 isn't guaranteed to be `0`. It's actually considered UB.

crates/core_arch/src/loongarch64/lasx/portable.rs
heiher: This is the same as in the code shown here:

https://github.com/rust-lang/stdarch/blob/0414f7c9d9867eb98baeae592a88ab303eb0ed1b/crates/core_arch/src/x86/sse.rs#L104-L115

crates/core_arch/src/loongarch64/lsx/portable.rs
heiher: A static assert exists for the number of shift immediate bits.

crates/core_arch/src/loongarch64/simd.rs
heiher: Thank you!

sayantn (approved):

PR #2079, opened 2026-04-11
Fix MIPS tests

eduardosm:
Tests failed to build due to use of legacy const generics syntax, out-of-bounds immediates and incorrect name of const generics in `assert_instr`

### DIFF
--- a/crates/core_arch/src/mips/msa.rs
+++ b/crates/core_arch/src/mips/msa.rs
@@ -1407,7 +1407,7 @@ pub unsafe fn __msa_addv_d(a: v2i64, b: v2i64) -> v2i64 {
 ///
 #[inline]
 #[target_feature(enable = "msa")]
-#[cfg_attr(test, assert_instr(addvi.b, imm5 = 0b10111))]
+#[cfg_attr(test, assert_instr(addvi.b, IMM5 = 0b10111))]
 #[rustc_legacy_const_generics(1)]
 #[unstable(feature = "stdarch_mips", issue = "111198")]
 pub unsafe fn __msa_addvi_b<const IMM5: i32>(a: v16i8) -> v16i8 {
@@ -1423,7 +1423,7 @@ pub unsafe fn __msa_addvi_b<const IMM5: i32>(a: v16i8) -> v16i8 {
 ///
 #[inline]
 #[target_feature(enable = "msa")]
-#[cfg_attr(test, assert_instr(addvi.h, imm5 = 0b10111))]
+#[cfg_attr(test, assert_instr(addvi.h, IMM5 = 0b10111))]
 #[rustc_legacy_const_generics(1)]
 #[unstable(feature = "stdarch_mips", issue = "111198")]
 pub unsafe fn __msa_addvi_h<const IMM5: i32>(a: v8i16) -> v8i16 {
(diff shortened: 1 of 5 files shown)


crates/core_arch/src/mips/msa.rs
```diff
@@ -9602,7 +9602,7 @@ mod tests {
             103, -126, 103, -126
         );
 
-        assert_eq!(r, mem::transmute(__msa_addvi_b(mem::transmute(a), 67)));
+        assert_eq!(r, mem::transmute(__msa_addvi_b::<3>(mem::transmute(a))));
```
folkertdev, line 9605: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2079/dd561eb18296300fc901d083bc2023c63f19bacb..d9ca05fe50c7ec4e468ba1c5e9c30213aba9e94c)*

what's happening here?

folkertdev: ah `3 == 67 % 64`, and this function uses `static_assert_uimm_bits!(IMM5, 5);`

folkertdev (approved):

sayantn:
For some reason I can't run the mips tests, `ci/run-docker.sh mips64-unknown-linux-gnuabi64` runs into a few errors. E.g. the `static_assert_simm_bits` assertions all fail (even with this patch, and the failures don't make any sense, e.g it says 5 is not between -17 and 16)

folkertdev:
I think some of the assert logic might be buggy (across targets)?

https://godbolt.org/z/qP74MWvK1

in any case the `bit_width` version expresses what we mean more accurately.

folkertdev:
Ah, it's signed and that is weird because you need a bit for the sign. Definitely a poor error message in any case.

eduardosm:
I had checked `assert_instr` tests with `cargo check --release`, which does not check the static asserts. Then I ran tests with `cargo test` (debug mode), which does not compile `assert_instr` at all, so I missed those failures.

Many `assert_instr` are failing anyways, and I will not be fixing them for now.

eduardosm:
Windows job failing, unrelated I believe 

```
  = note: C:/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/15.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: D:\a\stdarch\stdarch\target\x86_64-pc-windows-gnu\debug\deps\stdarch_gen_arm-c2d4a4c482dd3e81.9ul22rteqt6x41b1oy83npkbi.1831grd.rcgu.o:9ul22rteqt6x41b1oy83npkbi:(.debug_gdb_scripts+0x0): multiple definition of `.weak.__rustc_debug_gdb_scripts_section__.default'; D:\a\stdarch\stdarch\target\x86_64-pc-windows-gnu\debug\deps\stdarch_gen_arm-c2d4a4c482dd3e81.1nzalzaaigcyndp95lkud74oo.1831grd.rcgu.o:1nzalzaaigcyndp95lkud74oo:(.debug_gdb_scripts+0x0): first defined here
```

folkertdev:
yes it's unrelated https://github.com/rust-lang/rust/issues/155277

folkertdev:
> Many `assert_instr` are failing anyways, and I will not be fixing them for now.

Fair enough. Did anything in particular cause you to look at this?

(the changes look fine to me, so I think we should just merge this when CI  works again with the next nightly)

eduardosm:
> Did anything in particular cause you to look at this?

I was going to continue reducing the use of `unsafe fn` in `simd_test` functions (https://github.com/rust-lang/stdarch/pull/1981, https://github.com/rust-lang/stdarch/pull/1987, https://github.com/rust-lang/stdarch/pull/1989, https://github.com/rust-lang/stdarch/pull/2006), but I rather have tests compiling before making more changes.

folkertdev:
We removed the CI jobs for these targets, but probably should just still have them but make them not blocking for PRs? Then at least we know when they are broken and can occasionally fix them.

PR #2082, opened 2026-04-15
provide a reminder on the signedness of a failed `assert_simm_bits`

folkertdev:
tripped me up when looking at that MIPS PR

r? sayantn


sayantn:
Ubuntu servers seem broken 💔

PR #2083, opened 2026-04-15
Update arm_intrinsics.json

adamgemmell:
Most of the JSON changes intended for this PR were already merged as part of the SVE intrinsics patch. Additionally this removes some SVE intrinsic definitions that we haven't yet implemented the types for.


folkertdev:
It probably makes sense to split this up a bit?

Also it's fine to just ignore some tests with miri if it doesn't support them yet, though after the tests are merged maybe we can turn those lane loads/stores into something portable?

adamgemmell:
Happy split this into multiple PRs if you prefer, but the commits are already atomic and well-defined

PR #2086, opened 2026-04-17
Add more manual aarch64 load/store intrinsic tests

adamgemmell:
This was originally larger, but a fair number of the intrinsics I wrote tests for were already tested under different test names!

### DIFF
--- a/crates/core_arch/src/aarch64/neon/mod.rs
+++ b/crates/core_arch/src/aarch64/neon/mod.rs
@@ -1037,6 +1037,7 @@ mod tests {
     macro_rules! wide_store_load_roundtrip_fp16 {
         ($( $name:ident $args:tt);* $(;)?) => {
             $(
+                #[cfg_attr(miri, ignore)]
                 #[simd_test(enable = "neon,fp16")]
                 #[cfg(not(target_arch = "arm64ec"))]
                 unsafe fn $name() {
@@ -1055,13 +1056,13 @@ mod tests {
         test_vld1q_f16_x3(f16, 24, float16x8x3_t, vst1q_f16_x3, vld1q_f16_x3);
         test_vld1q_f16_x4(f16, 32, float16x8x4_t, vst1q_f16_x4, vld1q_f16_x4);
 
-        test_vld2_f16_x2(f16, 8, float16x4x2_t, vst2_f16, vld2_f16);
-        test_vld2_f16_x3(f16, 12, float16x4x3_t, vst3_f16, vld3_f16);
-        test_vld2_f16_x4(f16, 16, float16x4x4_t, vst4_f16, vld4_f16);
+        test_vld2_f16(f16, 8, float16x4x2_t, vst2_f16, vld2_f16);
+        test_vld3_f16(f16, 12, float16x4x3_t, vst3_f16, vld3_f16);
+        test_vld4_f16(f16, 16, float16x4x4_t, vst4_f16, vld4_f16);
 
-        test_vld2q_f16_x2(f16, 16, float16x8x2_t, vst2q_f16, vld2q_f16);
-        test_vld3q_f16_x3(f16, 24, float16x8x3_t, vst3q_f16, vld3q_f16);
-        test_vld4q_f16_x4(f16, 32, float16x8x4_t, vst4q_f16, vld4q_f16);
+        test_vld2q_f16(f16, 16, float16x8x2_t, vst2q_f16, vld2q_f16);
+        test_vld3q_f16(f16, 24, float16x8x3_t, vst3q_f16, vld3q_f16);
+        test_vld4q_f16(f16, 32, float16x8x4_t, vst4q_f16, vld4q_f16);
     }
 
     macro_rules! wide_store_load_roundtrip_aes {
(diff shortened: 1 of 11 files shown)


crates/core_arch/src/arm_shared/neon/mod.rs
```diff
@@ -5814,6 +5847,97 @@ mod tests {
...
+                    lane_wide_store_load_roundtrip! $args;
+                }
+            )*
+        };
+    }
+
+    macro_rules! lane_wide_store_load_roundtrip_fp16 {
+        ($( $name:ident $args:tt);* $(;)?) => {
+            $(
+                #[simd_test(enable = "neon,fp16")]
```
adamgemmell, line 5880: Any idea why these f16 tests pass on miri but not the regular ones above? They're all implemented with llvm builtins and aren't shimmed in miri as far as I can see

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2086/2d6c6f385711a7ae191cf103504d940814ce1114..4aba691a239010775bcbc692324655e8274cb76c)*

folkertdev: what works and what doesn't work?

adamgemmell: Ah, got there - the simd_test macro was silently filtering out these tests under miri because the fp16 feature wasn't enabled. I've added brought some target features from a recent cpu in the github actions job, which causes these to fail, and then ignored them again 😄

crates/core_arch/src/aarch64/neon/mod.rs
```diff
@@ -1037,6 +1037,7 @@ mod tests {
     macro_rules! wide_store_load_roundtrip_fp16 {
         ($( $name:ident $args:tt);* $(;)?) => {
             $(
+                #[cfg_attr(miri, ignore)]
```
RalfJung, line 1040: Why is this disabled in Miri now? Please always add comments explaining that.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2086/17b587bd82f3a9d3759c798eb22808752fc2b517..2d49c7b69aaaf2647558348a2d74ff2c894c3420)*

crates/core_arch/src/aarch64/neon/mod.rs
```diff
@@ -1195,101 +1196,149 @@ mod tests {
...
+            let r: $vec_ty = $load::<$idx>(tmp.as_ptr().cast(), a);
+            let out: [$elem_ty; $len] = transmute(r);
+            assert_eq!(out, vals);
+        };
+    }
+
+    macro_rules! lane_wide_store_load_roundtrip_neon {
+        ($( $name:ident $args:tt);* $(;)?) => {
+            $(
+                #[cfg_attr(miri, ignore)]
```
RalfJung, line 1311: Same question here

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2086/17b587bd82f3a9d3759c798eb22808752fc2b517..2d49c7b69aaaf2647558348a2d74ff2c894c3420)*

crates/core_arch/src/arm_shared/neon/mod.rs
```diff
@@ -5814,6 +5847,98 @@ mod tests {
...
+            let out: [$elem_ty; $len] = transmute(r);
+            assert_eq!(out, vals);
+        };
+    }
+
+    // Most of these are implemented with builtins, which miri can't handle
+    macro_rules! lane_wide_store_load_roundtrip_neon {
+        ($( $name:ident $args:tt);* $(;)?) => {
+            $(
+                #[cfg_attr(miri, ignore)]
```
RalfJung, line 5867: and here

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2086/17b587bd82f3a9d3759c798eb22808752fc2b517..2d49c7b69aaaf2647558348a2d74ff2c894c3420)*

crates/core_arch/src/arm_shared/neon/mod.rs
```diff
@@ -5814,6 +5847,98 @@ mod tests {
...
+                    lane_wide_store_load_roundtrip! $args;
+                }
+            )*
+        };
+    }
+
+    macro_rules! lane_wide_store_load_roundtrip_fp16 {
+        ($( $name:ident $args:tt);* $(;)?) => {
+            $(
+                #[cfg_attr(miri, ignore)]
```
RalfJung, line 5879: and here

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2086/17b587bd82f3a9d3759c798eb22808752fc2b517..2d49c7b69aaaf2647558348a2d74ff2c894c3420)*

crates/core_arch/src/arm_shared/neon/mod.rs
RalfJung: Ah this one has a comment a few lines up (but far enough away that it's not visible when grepping).

> Most of these are implemented with builtins, which miri can't handle

Would be good to have that comment on the same line as the `cfg_attr`, and mention which builtins. And then I'm still confused because surely we are missing *tons* of builtins tested in this file, why do just these two tests get an annotation?

RalfJung: I think I figured it out, and added that to https://github.com/rust-lang/stdarch/pull/2116

PR #2087, opened 2026-04-17
Fix lane-swapping on pairwise big endian Neon intrinsic

adamgemmell:
This PR also includes some minor improvements to intrinsic-test

### DIFF
--- a/crates/intrinsic-test/src/common/argument.rs
+++ b/crates/intrinsic-test/src/common/argument.rs
@@ -175,9 +175,10 @@ where
     pub fn load_values_c(&self, indentation: Indentation) -> String {
         self.iter()
             .filter(|&arg| !arg.has_constraint())
-            .map(|arg| {
+            .enumerate()
+            .map(|(idx, arg)| {
                 format!(
-                    "{indentation}{ty} {name} = cast<{ty}>({load}(&{name}_vals[i]));\n",
+                    "{indentation}{ty} {name} = cast<{ty}>({load}(&{name}_vals[i+{idx}]));\n",
                     ty = arg.to_c_type(),
                     name = arg.generate_name(),
                     load = if arg.is_simd() {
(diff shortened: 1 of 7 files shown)


crates/intrinsic-test/src/common/argument.rs
```diff
@@ -30,8 +30,7 @@ where
     }
 
     pub fn to_c_type(&self) -> String {
-        let prefix = if self.ty.constant { "const " } else { "" };
-        format!("{prefix}{}", self.ty.c_type())
+        self.ty.c_type()
```
folkertdev, line 33: you've now removed the `"const "` prefix in both cases (here and in ` c_type`), is that right?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2087/0d9e26a1d1db09b3f5865bdbf3d47dceb11bf95e..9b2717a367cd02813aba10cb911ab2d2ec9608e1)*

crates/intrinsic-test/src/common/argument.rs
```diff
@@ -176,9 +175,10 @@ where
     pub fn load_values_c(&self, indentation: Indentation) -> String {
         self.iter()
             .filter(|&arg| !arg.has_constraint())
-            .map(|arg| {
+            .enumerate()
+            .map(|(idx, arg)| {
                 format!(
-                    "{indentation}{ty} {name} = cast<{ty}>({load}(&{name}_vals[i]));\n",
+                    "{indentation}{ty} {name} = cast<{ty}>({load}(&{name}_vals[i+{idx}]));\n",
```
folkertdev, line 181: to be clear, not using `idx` before was a bug right?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2087/0d9e26a1d1db09b3f5865bdbf3d47dceb11bf95e..9b2717a367cd02813aba10cb911ab2d2ec9608e1)*

crates/intrinsic-test/src/common/argument.rs
adamgemmell: Yep, it should be handled by generate_c_constraint_blocks() now, and this avoids a warning about `const const` appearing in the output

crates/intrinsic-test/src/common/argument.rs
adamgemmell: Doing it this way (ensuring that two inputs to the intrinsic aren't always equal) provides a bit of extra coverage and should slightly reduce runtime. Just an oversight that we never did it this way originally

PR #2088, opened 2026-04-17
use `deinterleaving_load!` for `vld2` implementations

folkertdev:
This works now that https://github.com/rust-lang/rust/pull/155005 is merged


folkertdev:
but `aarch64_be` is weird apparently https://github.com/rust-lang/rust/issues/155459

PR #2089, opened 2026-04-19
Add AMX-AVX512 BF16 intrinsics

sayantn:


crates/core_arch/src/x86_64/amx.rs
```diff
@@ -480,6 +480,72 @@ pub unsafe fn _tile_cvtrowps2phli<const TILE: i32, const ROW: i32>() -> __m512h
...
 }
 
+/// Moves a row from a tile register to a zmm register, converting the packed single-precision (32-bit)
+/// floating-point elements to packed BF16 (16-bit) floating-point elements. The resulting
+/// 16-bit elements are placed in the high 16-bits within each 32-bit element of the returned vector.
+#[inline]
+#[rustc_legacy_const_generics(0)]
+#[target_feature(enable = "amx-avx512,avx10.2")]
+#[cfg_attr(
+    all(test, any(target_os = "linux", target_env = "msvc")),
```
folkertdev: what's up with windows gnu?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2089/a989a693dc2265a3915216fd50dc7330b4f4abf9..0d61ee304df41d9d0c82ba2dcaf0edfd39253238)*

folkertdev (approved):

crates/core_arch/src/x86_64/amx.rs
sayantn: iirc it had some problems decoding the amx instructions, didn't check recently though. let me just check in CI once

sayantn: ah, so only apple-darwin fails now. let me reduce the conditions a bit

PR #2092, opened 2026-04-24
mark `vstl1_*` functions as unsafe

### LINKED ISSUE #155751: Safe Neon store-release intrinsics marked safe but should be unsafe
The aarch64 neon store-release intrinsics `vstl1_lane_s64, vstl1q_lane_s64, vstl1_lane_u64, vstl1q_lane_u64, vstl1_lane_f64, vstl1q_lane_f64, vstl1_lane_p64, and vstl1q_lane_p64` are all marked `safe` @ https://github.com/rust-lang/rust/blob/acb65f36a0697070a7083c8733f49e7efdbaff61/library/stdarch/crates/core_arch/src/aarch64/neon/generated.rs#L25430

These intrinsics perform unconstrained memory writes via a provided raw pointer, and should be marked unsafe. They allow for writing an arbitrary 8-byte memory region via a raw pointer provided via a safe function call.

ex:
```rust
pub fn vstl1_lane_f64<const LANE: i32>(ptr: *mut f64, val: float64x1_t) {
    static_assert!(LANE == 0);
    unsafe { vstl1_lane_s64::<LANE>(ptr as *mut i64, transmute(val)) }
}
```

folkertdev:
fixes https://github.com/rust-lang/rust/issues/155751

these functions write to a raw pointer, and so are clearly unsafe to use

### DIFF
--- a/crates/stdarch-gen-arm/spec/neon/aarch64.spec.yml
+++ b/crates/stdarch-gen-arm/spec/neon/aarch64.spec.yml
@@ -4459,7 +4459,9 @@ intrinsics:
     doc: "Store-Release a single-element structure from one lane of one register."
     arguments: ["ptr: {type[0]}", "val: {neon_type[1]}"]
     static_defs: ["const LANE: i32"]
-    safety: safe
+    safety:
+      unsafe:
+        - pointer_write: ptr
     attr:
       - FnCall: [target_feature, ['enable = "neon,rcpc3"']]
       - FnCall: [cfg_attr, [{FnCall: [all, [test, {FnCall: [not, ['target_env= "msvc"']]}]]}, {FnCall: [assert_instr, [stl1, 'LANE = 0']]}]]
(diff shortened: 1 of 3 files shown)


crates/stdarch-gen-arm/spec/neon/aarch64.spec.yml
```diff
@@ -4459,7 +4459,8 @@ intrinsics:
     doc: "Store-Release a single-element structure from one lane of one register."
     arguments: ["ptr: {type[0]}", "val: {neon_type[1]}"]
     static_defs: ["const LANE: i32"]
-    safety: safe
+    safety:
+      unsafe: [neon]
```
folkertdev: @adamgemmell using `neon` here is not right, this is unsafe in its own right. I'm not sure if there are further safety conditions though, the link to the documentation that is generated for these functions does not work.

I also haven't quite reverse-engineered the right yaml to use to get a better message here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2092/ebbd66773c84b3fcfc44ab36782e27f2753d63a7..e60a0fb1cecf65c5ac0fc0dfd5ea70ccb8b81bc9)*

adamgemmell: Only SVE intrinsics have custom unsafety messages - originally all neon intrinsics were unsafe with the same message, and it looks like we never added proper messages when making the majority of them safe. I think just the pointer dereference is the only relevant safety condition here.

In my opinion, neon is fine and is in keeping with the rest of the neon intrinsics. If you're up for it a proper message would involve adding a variant of "UnsafetyComment" in `intrinsic.rs` and implementing Display for it (the Dereference variant which you'd otherwise use is specific to SVE as it makes reference to predicates)

sayantn (approved):
Sorry for the late review

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2092/ebbd66773c84b3fcfc44ab36782e27f2753d63a7..e850a3ba1e9749cf503d70fa862619cb5538daff)*

PR #2098, opened 2026-04-30
Various fixes in `stdarch_gen_arm`

sayantn:


sayantn:
Apparently these fixes have to come with the redesigned `intrinsic-test`

PR #2100, opened 2026-05-02
Porting some intrinsics to stdarch-gen-arm

sayantn:
These intrinsics were implemented in `aarch64/neon/mod.rs`, so this PR ports them to stdarch-gen-arm. All the edits are by @davidtwco, I have just cherry-picked from his branch and putting up the PR


davidtwco (approved):
I authored these changes so unsurprisingly they look good to me

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2100/ef18e28f8d2ca0c20ab7b053786b84af14b0e20a..be38e34eb56345b39c764117710c1a4403ffd0c7)*

Amanieu:
I checked the GCC headers and all of these (vget*, vcombine*) are available on 32-bit ARM as well. So everything here should go into arm_shared and be implemented for all types.

sayantn:
@Amanieu but the official ARM reference says that it should be only available in A64 (e.g. [vcopy_lane_f64](https://developer.arm.com/architectures/instruction-sets/intrinsics/#q=vcopy_lane_f64))

edit: I see that only the `f64` versions are A64-only, the `u64`, `s64` and `p64` versions are all available in A32 (with `vget_lane_p64` available in v7 too)

davidtwco:
I've went over all the additions and made the following changes:

- `vcopy_lane_{p,u,s,f}64`
  - I've kept these in `aarch64`
- `vcopy_laneq_{p,u,s,f}64`
  - I've kept these in `aarch64`
- `vget_{high,low}_f64`
  - I've kept these in `aarch64`
- `vget_{high,low}_p64`
  - I've moved these to `arm_shared`
- `vgetq_lane_f64`
  - I've kept these in `aarch64`
- `vcombine_f64`
  - I've kept these in `aarch64`

I'm happy to revisit some of these in future given GCC makes them available on A32 too, but where there's a discrepancy between GCC headers and the [Arm intrinsics documentation](https://developer.arm.com/architectures/instruction-sets/intrinsics/), our `arm_intrinsics.json` matches the documentation, and so the intrinsics we make available for A32 would go untested. We could change `arm_intrinsics.json` so they get tested but we derive this from some internal data and just overwrite it whenever we make changes, and I'd like to avoid having to keep track of tweaks we've made.

I can make sure we follow-up on this and check whether our source of truth needs updated internally, then updating `arm_intrinsics.json` upstream to match, and then moving any intrinsics to `arm_shared` that need to be moved.

PR #2101, opened 2026-05-02
Revert changing number of passes depending on number of parameters

sayantn:
this causes buffer overflow in some x86 intrinsics with more than 20 parameters, e.g. _mm512_set_ph, so I decided to go to wrapping

cc @adamgemmell as this was added in rust-lang/stdarch#2087

### DIFF
--- a/crates/intrinsic-test/src/common/argument.rs
+++ b/crates/intrinsic-test/src/common/argument.rs
@@ -1,5 +1,6 @@
 use super::cli::Language;
 use super::constraint::Constraint;
+use super::gen_rust::PASSES;
 use super::indentation::Indentation;
 use super::intrinsic_helpers::IntrinsicTypeDefinition;
 
@@ -178,7 +179,7 @@ where
             .enumerate()
             .map(|(idx, arg)| {
                 format!(
-                    "{indentation}{ty} {name} = cast<{ty}>({load}(&{name}_vals[i+{idx}]));\n",
+                    "{indentation}{ty} {name} = cast<{ty}>({load}(&{name}_vals[(i+{idx}) % {PASSES}]));\n",
                     ty = arg.to_c_type(),
                     name = arg.generate_name(),
                     load = if arg.is_simd() {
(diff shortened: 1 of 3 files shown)


log.log
folkertdev, line 1: what is this file for?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2101/ef18e28f8d2ca0c20ab7b053786b84af14b0e20a..362994b2205caeea1cda833688433d996dc8be90)*

sayantn: oh sorry, that's my temporary log file, accidentally got committed

crates/intrinsic-test/src/common/argument.rs
```diff
@@ -178,7 +179,7 @@ where
             .enumerate()
             .map(|(idx, arg)| {
                 format!(
-                    "{indentation}{ty} {name} = cast<{ty}>({load}(&{name}_vals[i+{idx}]));\n",
+                    "{indentation}{ty} {name} = cast<{ty}>({load}(&{name}_vals[(i+{idx}) % {PASSES}]));\n",
```
adamgemmell, line 182: Thank you for spotting this, I didn't consider intrinsics with a large number of parameters

#PASSES is fine but is slightly shorter than the full values vector length and will mean that some interesting float values are never loaded.

I think the source vector length is `(number of lanes) * (number of vectors) + (passes) -1`. This is from `src/common/argument.rs:gen_arglists_rust` and `populate_random` - or you could modify those functions to make the source vectors longer by the number of arguments so the sliding window fits again

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2101/ef18e28f8d2ca0c20ab7b053786b84af14b0e20a..362994b2205caeea1cda833688433d996dc8be90)*

sayantn: we cannot start at more than `PASSES` because the load functions take `num_lanes * num_vectors` elements, so we can at max start at `PASSES-1`. That's why I made it mod `PASSES`

adamgemmell: I understand now, ignore my comment!

PR #2105, opened 2026-05-07
Some more changes in stdarch-gen-arm

sayantn:
attempting to make push some more changes to stdarch-gen-arm standalone before rust-lang/stdarch#2063. Again, all commits are by @davidtwco.

r? @folkertdev


folkertdev (approved):
Looks fine, and I'm happy to unblock the main PR. Some things we could follow up on

- `crate::mem::zeroed()` can probably be replaced by `vec_ty::ZERO`
- functions like `vreinterpret_u8_s8` probably don't need a separate big-endian implementation (in most cases?)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2105/b5c858b217c30313175f57896306ca2b61819f53..8f84a30c0d82148f49cf0eeccec268fc233264c1)*

sayantn:
I tried to put the `vreinterpret` changes here, CI was not too happy with it, I will try again in a different PR

folkertdev:
Perhaps it should use `simd_as` instead of a `transmute`?

PR #2106, opened 2026-05-07
Use LLVM intrinsics for `f16` to `{i,u}16` intrinsics

sayantn:
again, the main edit is by @davidtwco, I have also updated the clang version here, this was buggy in older clangs

r? @folkertdev

### DIFF
--- a/ci/docker/aarch64-unknown-linux-gnu/Dockerfile
+++ b/ci/docker/aarch64-unknown-linux-gnu/Dockerfile
@@ -10,8 +10,14 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
   qemu-user \
   make \
   file \
-  clang \
-  lld
+  xz-utils \
+  wget
+
+RUN wget https://mirrors.edge.kernel.org/pub/tools/llvm/files/llvm-22.1.4-x86_64.tar.gz -O llvm.tar.xz
+RUN mkdir llvm
+RUN tar -xvf llvm.tar.xz --strip-components=1 -C llvm
+
+ENV PATH="/llvm/bin:$PATH"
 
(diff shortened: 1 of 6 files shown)


ci/intrinsic-test.sh
```diff
@@ -49,30 +49,30 @@ case ${TARGET} in
     aarch64-unknown-linux-gnu*)
         TEST_CPPFLAGS="-fuse-ld=lld -I/usr/aarch64-linux-gnu/include/ -I/usr/aarch64-linux-gnu/include/c++/9/aarch64-linux-gnu/"
         TEST_SKIP_INTRINSICS=crates/intrinsic-test/missing_aarch64.txt
-        TEST_CXX_COMPILER="clang++"
```
davidtwco, line 52: This does require that you locally symlink a LLVM build to `/llvm` if you use `ci/intrinsic-test.sh` locally, which isn't ideal (and depending on permissions, might not be possible)

davidtwco (commented):
LGTM - I was obviously involved in some of the changes so I'm fine with those parts, but this should probably be reviewed by someone else too 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2106/f0a39a71c915945d53ebcc8e45d9c786584843d6..b857fee2213bcdcd0db386578240d7634811da7d)*

ci/intrinsic-test.sh
sayantn: hm, I didn't think about that. Another possibility can be simply just adding `/llvm/bin/` to `PATH` in the `Dockerfile` and using simply `clang++` here

davidtwco: I think that would be best, alternatively reading an `LLVM_PATH` environment variable that defaults to `/llvm` but can be overridden

PR #2107, opened 2026-05-08
gen-arm: Change definitions of some intrinsics

sayantn:
Changes the implementations of `vmlal`, `vmlsl`, `vmull`, `vfmlal` and `vfmlsl` to using ARM intrinsics rather than portable intrinsics, which will not require a separate big-endian implementation

r? @folkertdev


davidtwco (approved):
LGTM

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2107/f0a39a71c915945d53ebcc8e45d9c786584843d6..47563140999f1f3995332cc93e6c25f125db59a3)*

folkertdev:
In practice you're still using portable intrinsics I believe? Like `vdup_laneq_f16` etc just use a `simd_shuffle`.

sayantn:
yes, we are still using portable SIMD intrinsics, but now they are deeper in the call tree, so we don't need to put `big_endian_inverse`s everywhere

PR #2110, opened 2026-05-08
Remove the `aes` feature from many arm intrinsics

### LINKED ISSUE #2109: Should the `p64` functions be gated under `aes`?
Currently we gate a lot of `p64` functions under `aes`, but I couldn't spot any pattern on when we add `aes`. We have `aes` in the following functions

 - `vld1{q}_p64{_x2,_x3,_x4}`
 - `vld{1,2,3,4}{q}{,_dup, _lane}_p64`
 - `vmull{_high}_p64`
 - `vs{r,l}i{q}_n_p64`
 - `vst1{q}_p64`
 - `vst1{q}_p64_{x2,x3,x4}`
 - `vst{2,3,4}{_lane, q_lane,}_p64`
 - `vcreate_p64`
 - `vreinterpret_*_p{64,128}`
 - `vreinterpret_p{64,128}_*`
 - `vset{q}_lane_p64`

whereas, the following functions, even though they have a `p64` argument, don't have `aes`

 - `vceq{z}{q}_p64`
 - `vcopy{q}_lane{q}_p64`
 - `vdup{q}_lane{q}_p64`
 - `vextq_p64`
 - `vldap1{q}_lane_p64`
 - `vtrn{1,2}q_p64`,
 - `vtst{q}_p64`
 - `vuzp{1,2}q_p64`
 - `vzip{1,2}q_p64`
 - `vadd{q}_p64`
 - `vget_{high,low}_p64`
 - `vget{q}_lane_p64`
 - `vreinterpret{q}_{f32,f64,s64,u64}_p64`

The only reference of this in ACLE I could find was the fact that GCC gates _all_ `p64` functions and types under `target ("fpu=crypto-neon-fp-armv8")`, but afaik this does not translate to the Rust feature `aes`. I believe the closest would be the LLVM codegen feature `crypto`.

@rustbot ping arm

adamgemmell:
Fixes https://github.com/rust-lang/stdarch/issues/2109

The aes rust feature represents FEAT_AES and FEAT_PMULL - from the [architecture reference manual](https://developer.arm.com/documentation/ddi0487/mb/-Part-A-Arm-Architecture-Introduction-and-Overview/-Chapter-A2-A-profile-Architecture-Extensions/-A2-2-Armv8-A-architecture-extensions/-A2-2-1-The-Armv8-0-architecture-extension?lang=en), these just add 4 new aes* instructions, and the p64 variants of PMULL/PMULL2.

Clang's header doesn't seem to gate any poly types on the aes feature, and many of the intrinsics with the aes feature are nops or just load/stores, for which the aes feature shouldn't matter.  Accordingly, I've removed the aes feature from all intrinsics except the vaes* intrinsics, and ones using the p64 variants of PMULL[2] instructions


folkertdev:
This changes public, stable signatures. I don't think it can be observed (it's just removing a constraint), but we should be careful. 

Is there any way to validate that the new versions in fact work? I suspect that our CI just has the aes feature available anyway, so if any of these functions did still generate an `aes` instruction we'd not notice. Maybe qemu can be configured to just have `neon`, and not `aes` locally?

adamgemmell:
> This changes public, stable signatures. I don't think it can be observed (it's just removing a constraint), but we should be careful.

A few cases I can think of where removing a target feature could cause a breaking change:

- It's an evil one that changes the ABI - aes is a pretty simple one that just allows the generation of some instruction variants.
- It changes inlining behaviour - not technically breaking, but annoying, particularly with simd intrinsics. With the AArch64/ARM LLVM backends in particular, I expect that removing this feature will just allow the function to be inlined into more contexts rather than fewer.
- Having to revert this patch, of course, could be disruptive if required.

> Is there any way to validate that the new versions in fact work? I suspect that our CI just has the aes feature available anyway, so if any of these functions did still generate an aes instruction we'd not notice. Maybe qemu can be configured to just have neon, and not aes locally?

I did test this locally with some qemu cpus that don't support the crypto extension and it worked. I wanted to push it to CI to demonstrate it working but it's quite annoying to do so

If we rely on LLVM, it will fail to generate any instructions that require the aes feature when it's not available. intrinsic-test does not put target features on its test wrappers nor build with -Ctarget-feature, so those intrinics are running without the aes feature present.

Also, sayantn's concern at https://github.com/rust-lang/stdarch/issues/2109#issuecomment-4409143857 regarding GCC isn't fully resolved yet

PR #2121, opened 2026-05-19
intrinsic-test: sve type parsing and constraints

davidtwco:
First initial changes for SVE support in `intrinsic-test`:

- Simplifies `ArmIntrinsicType`, removing the unused `target` field
- Updates `parse_intrinsic_type` to be able to parse SVE types, updating `simd_len` to be able to represent both `Scalable` and `Fixed(u32)` length vectors, with minimal changes to keep the users of this field working even if they'll need further changes to actually support the `Scalable` case properly
- Updates `Constraint` to support additional constraint kinds that we'll want to test for SVE, and adds logic to insert these constraints where appropriate.

This patch doesn't actually test any SVE intrinsics yet, so none of this additional logic is hit yet, but I've confirmed that it does the right things locally when the filter for Neon intrinsics is removed.

### DIFF
--- a/crates/intrinsic-test/src/arm/types.rs
+++ b/crates/intrinsic-test/src/arm/types.rs
@@ -42,7 +42,7 @@ impl IntrinsicTypeDefinition for ArmIntrinsicType {
             simd_len,
             vec_len,
             ..
-        } = &self.data
+        } = **self
         {
             let quad = if simd_len.unwrap_or(1) * bl > 64 {
                 "q"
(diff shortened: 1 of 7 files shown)


crates/intrinsic-test/src/arm/types.rs
```diff
@@ -71,67 +80,97 @@ impl IntrinsicTypeDefinition for ArmIntrinsicType {
...
+
+    let bit_len = match (bit_len, kind) {
+        (None, TypeKind::SvPattern | TypeKind::SvPrefetchOp | TypeKind::Int(_)) => Some(32),
+        (None, TypeKind::Bool) => Some(8),
+        _ => bit_len,
+    };
+
+    // Continuing the previous examples..
+    // A) second_len=Some(4)
+    // B) second_len=Some(2)
```
adamgemmell: B should be Some(4)

crates/intrinsic-test/src/arm/types.rs
```diff
@@ -71,67 +80,97 @@ impl IntrinsicTypeDefinition for ArmIntrinsicType {
...
+    // C) second_len=None
+    let second_len = parts.next().map(|part| {
+        part.parse::<u32>()
+            .expect("failed to parse second part of type")
+    });
+
+    // Continuing the previous examples..
+    // A) third_len=Some(2)
+    // B) third_len=None
```
adamgemmell: A should be None, B should be Some(2)

adamgemmell (changes requested):
LGTM after fixing these comments

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2121/43350f210fb2a2fcd4d46ab870c8080bd673b5dd..f91b1837f295762b527dd2620798f0261098ad12)*

crates/intrinsic-test/src/arm/types.rs
davidtwco: Fixed

crates/intrinsic-test/src/arm/types.rs
davidtwco: Fixed

adamgemmell (approved):

PR #2122, opened 2026-05-19
intrinsic-test: run rustfmt on generated sources

davidtwco:
This allows for simplification of the intrinsic-test generator as it no longer needs to maintain readable formatting.


sayantn:
Can we maybe do this in the intrinsic-test output, like gen-arm does?

davidtwco:
> Can we maybe do this in the intrinsic-test output, like gen-arm does?

Sure, I've changed it to work that way

PR #2124, opened 2026-05-19
intrinsic-test: add lots of documentation

davidtwco:
When I had tried to make quick changes to `intrinsic-test` in the past, I found it quite an unintuitive codebase - lots of the code generation is split over multiple functions that make it hard to see the big picture of what is generated (particularly without running the tool and looking at what is generated), and lots of functions are confusingly named. Prior to doing any refactorings in pursuit of adding SVE testing, I wanted to first document as much of the codebase as I could to make it easier to understand - I'll follow-up with renames and refactorings.


crates/intrinsic-test/src/common/values.rs
adamgemmell, line 19: These lines are very weird, I bet they're dead code 🤔 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2124/391642bd9732d6358169111fbc98661253324a89..770f7ceb9caad2e7c6f7bfe107398a010ff16c09)*

adamgemmell:
LGTM - David & I have the same employer, but I'm going to merge this as it's just doc comments without functionality

davidtwco:
> LGTM - David & I have the same employer, but I'm going to merge this as it's just doc comments without functionality

I'm fine with this when the changes aren't likely to be controversial and you haven't been involved in writing the changes.

PR #2125, opened 2026-05-20
Add GCC run to intrinsic-test

sayantn:
Let's try to add a GCC run to intrinsic-test now that it is much easier to do. Also add a ICX run for x86_64 because why not!

Currently disables the GCC run on ARM32, because GCC has a lot of miscompilations and missing intrinsics in ARM32.

Also, weirdly exactly 6 intrinsics have rounding errors in release mode, in all 3 compilers!


ci/intrinsic-test.sh
```diff
@@ -2,7 +2,29 @@
...
+        export CC="${CLANG_PATH}"
+        CC_KIND=clang
+        ;;
+    gcc)
+        export CC="${GCC_PATH}"
+        CC_KIND=gcc
+        ;;
+    icx)
+        export CC="${ICX_PATH}"
+        CC_KIND=clang
```
folkertdev: is this right? Would benefit from a comment if it is.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2125/d94057907402f34929dfb4e87363fb95bbfe021c..af99dd8f67dc6d297f274016abd6a495a0a262d2)*

crates/intrinsic-test/src/arm/mod.rs
```diff
@@ -29,11 +29,14 @@ impl SupportedArchitectureTest for ArmArchitectureTest {
     const PLATFORM_RUST_DEFINITIONS: &str = config::PLATFORM_RUST_DEFINITIONS;
     const PLATFORM_RUST_CFGS: &str = config::PLATFORM_RUST_CFGS;
 
-    fn arch_flags(&self) -> Vec<&str> {
-        vec!["-march=armv8.6a+crypto+crc+dotprod+fp16"]
+    fn arch_flags(&self, cli_options: &ProcessedCli) -> Vec<&str> {
+        match cli_options.cc_kind {
+            CompilerKind::Clang => vec!["-march=armv8.6a+crypto+crc+dotprod+fp16"],
+            CompilerKind::Gcc => vec!["-march=armv8.6-a+crypto+crc+dotprod+fp16+sha3+sm4"],
+        }
```
folkertdev: maybe comment on the difference here?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2125/d94057907402f34929dfb4e87363fb95bbfe021c..af99dd8f67dc6d297f274016abd6a495a0a262d2)*

crates/intrinsic-test/src/common/cli.rs
```diff
@@ -22,13 +22,24 @@ pub struct Cli {
...
     #[arg(long, default_value_t = 100u8)]
     pub sample_percentage: u8,
+
+    /// Argument style of the C compiler
+    #[arg(long)]
+    pub cc_kind: CompilerKind,
+}
+
+#[derive(Copy, Clone, clap::ValueEnum)]
+pub enum CompilerKind {
```
folkertdev: Would `enum CcArgStyle` be clearer?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2125/d94057907402f34929dfb4e87363fb95bbfe021c..af99dd8f67dc6d297f274016abd6a495a0a262d2)*

ci/intrinsic-test.sh
sayantn: Yes, `icx` has clang-style arguments, but `icx-cl` has msvc-style arguments. Will put a comment

crates/intrinsic-test/src/arm/mod.rs
sayantn: The main difference is the arch name, gcc has one extra `-`. Will put a comment

crates/intrinsic-test/src/common/cli.rs
sayantn: Nice name, thanks. Will change

PR #2130, opened 2026-05-21
intrinsic-test fixes for low intrinsic, high core counts

adamgemmell:
When the intrinsic count is low (such as when adding filtering for debugging) and the core count is high, sometimes the chunk count returned by `manual_chunk()` didn't match the number of modules created, which meant that the `rust_programs/Cargo.toml` referred to modules that didn't exist.

e.g. 100 intrinsics and 80 cores:
chunk_size = 100.div_ceil(80) = 2,
chunk count = 80

However, in practice, with a chunk_size of 2 `Iter::chunks` will only return 50 chunks. The fix is to calculate the real number of chunks with another div ceil, such as `100.div_ceil(2) = 50`. There shouldn't be any difference when the intrinsic count is high enough for this to not matter.

In addition, this patch also adds the `--tests` argument to `cargo test` in order to not attempt running doc tests, which is a slowdown more noticeable with low intrinsic count and higher core count.

### DIFF
--- a/ci/intrinsic-test.sh
+++ b/ci/intrinsic-test.sh
@@ -57,4 +57,4 @@ case "${TARGET}" in
         ;;
 esac
 
-cargo test --manifest-path=rust_programs/Cargo.toml --target "${TARGET}" --profile "${PROFILE}"
+cargo test --manifest-path=rust_programs/Cargo.toml --target "${TARGET}" --profile "${PROFILE}" --tests
--- a/crates/intrinsic-test/src/common/mod.rs
+++ b/crates/intrinsic-test/src/common/mod.rs
@@ -42,11 +42,11 @@ pub trait SupportedArchitectureTest {
     fn arch_flags(&self) -> Vec<&str>;
 
     fn generate_c_file(&self) {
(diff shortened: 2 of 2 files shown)


ci/intrinsic-test.sh
```diff
@@ -57,4 +57,4 @@ case "${TARGET}" in
         ;;
 esac
 
-cargo test --manifest-path=rust_programs/Cargo.toml --target "${TARGET}" --profile "${PROFILE}"
```
sayantn, line 60: Is this just remnant from local trials?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2130/d94057907402f34929dfb4e87363fb95bbfe021c..8ab07a2e6dafcc72f270747f273c5cd520b62c40)*

adamgemmell: No, adding --tests avoids attempting to run doc tests on each mod_n crate

sayantn: Oh ok I see. Thanks

sayantn (approved):

PR #2136, opened 2026-05-24
Several fixes to the loongarch CRC intrinsics

Gelbpunkt:
The first commit makes the CRC intrinsics exclusive to 64-bit LoongArch, since they are exclusive to LA64 in Clang and GCC (see https://github.com/llvm/llvm-project/blob/main/llvm/lib/Target/LoongArch/LoongArchISelLowering.cpp#L4797).

The second commit aligns the CRC intrinsic definitions for 8 and 16-bit wide data with Clang and GCC, which both use `char` and `short` rather than `int` for the data parameter (see https://github.com/llvm/llvm-project/blob/main/clang/lib/Headers/larchintrin.h#L60 and https://github.com/gcc-mirror/gcc/blob/master/gcc/config/loongarch/larchintrin.h#L152). The ARM CRC intrinsics are handled identically in stdarch already: there, the LLVM intrinsic also takes an `i32` but we cast from `u8` or `u16`  (https://doc.rust-lang.org/stable/src/core/stdarch/crates/core_arch/src/arm_shared/neon/generated.rs.html#63).

### DIFF
--- a/crates/core_arch/src/loongarch64/mod.rs
+++ b/crates/core_arch/src/loongarch64/mod.rs
@@ -22,8 +22,20 @@ pub fn rdtime_d() -> (i64, isize) {
 
 #[allow(improper_ctypes)]
 unsafe extern "unadjusted" {
+    #[link_name = "llvm.loongarch.crc.w.b.w"]
+    fn __crc_w_b_w(a: i32, b: i32) -> i32;
+    #[link_name = "llvm.loongarch.crc.w.h.w"]
+    fn __crc_w_h_w(a: i32, b: i32) -> i32;
+    #[link_name = "llvm.loongarch.crc.w.w.w"]
+    fn __crc_w_w_w(a: i32, b: i32) -> i32;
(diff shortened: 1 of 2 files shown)


Gelbpunkt:
cc @heiher

folkertdev (approved):
LGTM but needs the target maintainer's approval too

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2136/329aeddeba01f6fe8ff398cf6c03efbd29a7d058..5a29b38ae466bad419e7031e42abb524590adc3f)*

heiher (approved):
LGTM. Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2136/329aeddeba01f6fe8ff398cf6c03efbd29a7d058..5a29b38ae466bad419e7031e42abb524590adc3f)*

PR #2137, opened 2026-05-25
Stabilize LoongArch CRC Intrinsics

heiher:
implementation: https://github.com/rust-lang/stdarch/pull/1688
tracking issue: https://github.com/rust-lang/rust/issues/156908

Following a proposal from @folkertdev and downstream interest from zlib-rs (https://github.com/trifectatechfoundation/zlib-rs/pull/511), this PR proposes stabilizing the LoongArch CRC intrinsics.

Since the API only operates on scalar integer types, there are no major ABI concerns around stabilization.

## API surface

The `stdarch_loongarch_crc` feature exposes 8 functions:

```rust
/// Calculate the CRC value using the IEEE 802.3 polynomial (0xEDB88320)
pub fn crc_w_b_w(a: i8, b: i32) -> i32

/// Calculate the CRC value using the IEEE 802.3 polynomial (0xEDB88320)
pub fn crc_w_h_w(a: i16, b: i32) -> i32

/// Calculate the CRC value using the IEEE 802.3 polynomial (0xEDB88320)
pub fn crc_w_w_w(a: i32, b: i32) -> i32

/// Calculate the CRC value using the IEEE 802.3 polynomial (0xEDB88320)
pub fn crc_w_d_w(a: i64, b: i32) -> i32 

/// Calculate the CRC value using the Castagnoli polynomial (0x82F63B78)
pub fn crcc_w_b_w(a: i8, b: i32) -> i32

/// Calculate the CRC value using the Castagnoli polynomial (0x82F63B78)
pub fn crcc_w_h_w(a: i16, b: i32) -> i32

/// Calculate the CRC value using the Castagnoli polynomial (0x82F63B78)
pub fn crcc_w_w_w(a: i32, b: i32) -> i32

/// Calculate the CRC value using the Castagnoli polynomial (0x82F63B78)
pub fn crcc_w_d_w(a: i64, b: i32) -> i32
```

### DIFF
--- a/crates/core_arch/src/loongarch64/mod.rs
+++ b/crates/core_arch/src/loongarch64/mod.rs
@@ -62,56 +62,56 @@ unsafe extern "unadjusted" {
 
 /// Calculate the CRC value using the IEEE 802.3 polynomial (0xEDB88320)
 #[inline(always)]
-#[unstable(feature = "stdarch_loongarch", issue = "117427")]
+#[stable(feature = "stdarch_loongarch_crc", since = "CURRENT_RUSTC_VERSION")]
 pub fn crc_w_b_w(a: i8, b: i32) -> i32 {
     unsafe { __crc_w_b_w(a as i32, b) }
 }
 
 /// Calculate the CRC value using the IEEE 802.3 polynomial (0xEDB88320)
 #[inline(always)]
-#[unstable(feature = "stdarch_loongarch", issue = "117427")]
+#[stable(feature = "stdarch_loongarch_crc", since = "CURRENT_RUSTC_VERSION")]
 pub fn crc_w_h_w(a: i16, b: i32) -> i32 {
     unsafe { __crc_w_h_w(a as i32, b) }
 }
 
 /// Calculate the CRC value using the IEEE 802.3 polynomial (0xEDB88320)
 #[inline(always)]
-#[unstable(feature = "stdarch_loongarch", issue = "117427")]
+#[stable(feature = "stdarch_loongarch_crc", since = "CURRENT_RUSTC_VERSION")]
 pub fn crc_w_w_w(a: i32, b: i32) -> i32 {
     unsafe { __crc_w_w_w(a, b) }
 }
 
 /// Calculate the CRC value using the IEEE 802.3 polynomial (0xEDB88320)
 #[inline(always)]
-#[unstable(feature = "stdarch_loongarch", issue = "117427")]
+#[stable(feature = "stdarch_loongarch_crc", since = "CURRENT_RUSTC_VERSION")]
 pub fn crc_w_d_w(a: i64, b: i32) -> i32 {
     unsafe { __crc_w_d_w(a, b) }
 }
 
 /// Calculate the CRC value using the Castagnoli polynomial (0x82F63B78)
 #[inline(always)]
-#[unstable(feature = "stdarch_loongarch", issue = "117427")]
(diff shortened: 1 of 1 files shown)


folkertdev:
Nice, can you open a new tracking issue for the new feature you just added (roughly similar to https://github.com/rust-lang/rust/issues/117224, but copy the whole api surface section from your post here, that's convenient).

I'll then nominate that tracking issue for T-libs-api to look at, and they can start the stabilization process. Pointing out the similarity to https://github.com/rust-lang/rust/issues/117215 is also useful, as well as the fact that these intrinsics do not require a target feature

folkertdev:
And also perhaps clarify that these are only defined on `loongarch64`, not on `loongarch32`

heiher:
@folkertdev A new tracking issue: https://github.com/rust-lang/rust/issues/156908

Gelbpunkt:
Thanks for proposing this! I agree it makes sense to stabilize these intrinsics and I'm happy with them now that the signatures match Clang's, but I'm not 100% sure about the names yet. In Clang, all of these have an `__` prefix, e.g. `__crc_w_b_w` (https://github.com/llvm/llvm-project/blob/main/clang/lib/Headers/larchintrin.h#L58-L104). *If* it is desired to match Clang in that regard as well, that should be changed before stabilization.

heiher:
> Thanks for proposing this! I agree it makes sense to stabilize these intrinsics and I'm happy with them now that the signatures match Clang's, but I'm not 100% sure about the names yet. In Clang, all of these have an `__` prefix, e.g. `__crc_w_b_w` (https://github.com/llvm/llvm-project/blob/main/clang/lib/Headers/larchintrin.h#L58-L104). _If_ it is desired to match Clang in that regard as well, that should be changed before stabilization.

We previously [discussed](https://rust-lang.zulipchat.com/#narrow/channel/208962-t-libs.2Fstdarch/topic/Should.20LoongArch64.20SIMD.20intrinsics.20adopt.20x86-style.20unified.20t/near/526993467) the naming of the Rust intrinsics and intentionally removed the double underscores.

RalfJung:
> now that the signatures match Clang's,

So these really should be *signed* integers? That seems a bit silly... in clang it doesn't really matter as C will silently convert it anyway.

Gelbpunkt:
> So these really should be _signed_ integers? That seems a bit silly... in clang it doesn't really matter as C will silently convert it anyway.

Clang defines the AArch64 CRC intrinsics with unsigned integer types (see https://github.com/llvm/llvm-project/blob/main/clang/lib/Headers/arm_acle.h#L627), while the LoongArch CRC intrinsics are defined with signed integer types (https://github.com/llvm/llvm-project/blob/main/clang/lib/Headers/larchintrin.h#L60). stdarch currently matches both definitions. I don't think it should diverge from Clang's headers

Gelbpunkt:
From a user perspective, the unsigned types would be nicer to work with, but is that really worth diverging from Clang and GCC here? In the end it doesn't matter as long as the number of bits is correct, the sign (or lack thereof) has no meaning here.

heiher:
> From a user perspective, the unsigned types would be nicer to work with, but is that really worth diverging from Clang and GCC here? In the end it doesn't matter as long as the number of bits is correct, the sign (or lack thereof) has no meaning here.

Makes sense.

folkertdev:
FCP is complete https://github.com/rust-lang/rust/issues/156908#issuecomment-4697127152, so this is good to go.

PR #2138, opened 2026-05-26
hexagon-gen: add optional output-dir argument

xonx4l:
This PR adds an optional output-directory argument to `stdarch-gen-hexagon` and `stdarch-gen-hexagon-scalar` while preserving the default current behaviour .

r? folkertdev


folkertdev:
cc @androm3da 

We need this for running these tests in `rust-lang/rust` CI where these files are mounted as read-only, and we want to write the output to a writable location.

PR #2139, opened 2026-05-28
endianness shuffles on loads/stores on `aarch64_be`

folkertdev:
With clang I get 

https://godbolt.org/z/Mcvebd79G

```c
// test_vpmaxq_u8.c
#include <arm_neon.h>
#include <stdint.h>
#include <stdio.h>
#include <stdlib.h>

uint8x16_t wrapper(uint8x16_t a, uint8x16_t b) {
    return vpmaxq_u8(a, b);
}

void wrapper2(uint8_t *a, uint8_t *b, uint8_t *out) {
    uint8x16_t x = vld1q_u8(a);
    uint8x16_t y = vld1q_u8(b);
    
    uint8x16_t tmp = vpmaxq_u8(x, y);

    vst1q_u8(out, tmp);
}
```

gives 

```asm
wrapper:
        rev64   v1.16b, v1.16b
        rev64   v0.16b, v0.16b
        ext     v1.16b, v1.16b, v1.16b, #8
        ext     v0.16b, v0.16b, v0.16b, #8
        rev64   v0.16b, v0.16b
        rev64   v1.16b, v1.16b
        ext     v0.16b, v0.16b, v0.16b, #8
        ext     v1.16b, v1.16b, v1.16b, #8
        umaxp   v0.16b, v0.16b, v1.16b
        rev64   v0.16b, v0.16b
        ext     v0.16b, v0.16b, v0.16b, #8
        rev64   v0.16b, v0.16b
        ext     v0.16b, v0.16b, v0.16b, #8
        ret

wrapper2:
        ld1     { v0.16b }, [x0]
        ld1     { v1.16b }, [x1]
        umaxp   v0.16b, v0.16b, v1.16b
        st1     { v0.16b }, [x2]
        ret
```

while with `cargo 1.98.0-nightly (fbb61be30 2026-05-26)` this rust code 

```rust
use core::arch::aarch64::*;

#[unsafe(no_mangle)]
extern "C" fn wrapper(a: uint8x16_t, b: uint8x16_t) -> uint8x16_t {
    unsafe {
        return vpmaxq_u8(a, b);
    }
}

#[unsafe(no_mangle)]
extern "C" fn wrapper2(a: *const u8, b: *const u8, out: *mut u8) {
    unsafe {
        let x: uint8x16_t = vld1q_u8(a);
        let y: uint8x16_t = vld1q_u8(b);

        let tmp = vpmaxq_u8(x, y);

        vst1q_u8(out, tmp);
    }
}
```

emits 

```asm
wrapper:
	rev64 v1.16b, v1.16b
	rev64 v0.16b, v0.16b
	ext v1.16b, v1.16b, v1.16b, #8
	ext v0.16b, v0.16b, v0.16b, #8
	rev64 v0.16b, v0.16b
	rev64 v1.16b, v1.16b
	ext v0.16b, v0.16b, v0.16b, #8
	ext v1.16b, v1.16b, v1.16b, #8
	umaxp v0.16b, v0.16b, v1.16b
	rev64 v0.16b, v0.16b
	ext v0.16b, v0.16b, v0.16b, #8
	rev64 v0.16b, v0.16b
	ext v0.16b, v0.16b, v0.16b, #8
	ret

wrapper2:
	ld1 { v0.16b }, [x0]
	ld1 { v1.16b }, [x1]
	rev64 v0.16b, v0.16b
	rev64 v1.16b, v1.16b
	ext v0.16b, v0.16b, v0.16b, #8
	ext v1.16b, v1.16b, v1.16b, #8
	umaxp v0.16b, v0.16b, v1.16b
	rev64 v0.16b, v0.16b
	ext v0.16b, v0.16b, v0.16b, #8
	st1 { v0.16b }, [x2]
	ret
```

So the `wrapper` is the same, but the loads and stores mess it up somehow.

---

So, this is pretty nasty. We also get instruction assertion errors like

```
---- core_arch::aarch64::neon::generated::assert_vst1q_u32_str stdout ----
disassembly for stdarch_test_shim_vst1q_u32_str: 
	 0: rev64 v0.4s, v0.4s
	 1: ext v0.16b, v0.16b, v0.16b, #8
	 2: rev64 v0.4s, v0.4s
	 3: ext v0.16b, v0.16b, v0.16b, #8
	 4: st1 {v0.4s}, [x0]
	 5: ret
```

So, I'm not sure what to do about that. We can just test for the `st*` prefix? Or drop the test altogether?

r? adamgemmell

### DIFF
--- a/crates/core_arch/src/aarch64/neon/generated.rs
+++ b/crates/core_arch/src/aarch64/neon/generated.rs
@@ -9924,6 +9924,7 @@ pub fn vgetq_lane_f64<const IMM5: i32>(a: float64x2_t) -> f64 {
 #[doc = "## Safety"]
 #[doc = "  * Neon intrinsic unsafe"]
 #[inline]
+#[cfg(target_endian = "little")]
 #[target_feature(enable = "neon,fp16")]
 #[cfg_attr(test, assert_instr(ldr))]
 #[unstable(feature = "stdarch_neon_f16", issue = "136306")]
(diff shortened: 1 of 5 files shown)


sayantn:
I think the ABI of passing vectors in A64BE is different in CLang and GCC. For example, https://godbolt.org/z/jedE96h6b both the versions look exactly the same. I think this is why we had to use pointers in intrinsic-test for vectors.

Also, a modified version of intrinsic-test passes with this change, with both clang and gcc

PR #2140, opened 2026-05-29
loongarch: Fix visibility and legacy const generic index

heiher:


crates/core_arch/src/loongarch64/lasx/portable.rs
```diff
@@ -69,7 +69,7 @@ const unsafe fn simd_pickod_h<T: Copy>(a: T, b: T) -> T {
 
 #[inline(always)]
 #[rustc_const_unstable(feature = "stdarch_const_helpers", issue = "none")]
```
adamgemmell, line 71: Happy with this PR, but another point here is that functions that are not exported from the crate do not need to be marked as unstable

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2140/65d15014c5bc1ff9459a5d6bdb2bff17de69bdb7..dd0dc620e69d5614088e29c36b9d00d723e44434)*

adamgemmell (approved):

PR #2141, opened 2026-05-29
Add `_mm_clflushopt` intrinsic

### LINKED ISSUE #1456: Support for `_mm_clflushopt`
We don't have a `core::arch::{x86_64,x86}` intrinsic for [`_mm_clflushopt`](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=clflushopt&ig_expand=769). I think this would need a new `clflushopt` target feature.

I don't mind implementing this.

jlpetz:
Fixes rust-lang/stdarch#1456

adds the `_mm_clflushopt` intrinsic for the
`CLFLUSHOPT` instruction (weakly-ordered cache-line flush). Until now stdarch
exposed only the SSE2 `_mm_clflush`, so there was no way to emit `clflushopt`
from an intrinsic.

Gated on the `clflushopt` target feature added in rust-lang/rust#157098
(tracking issue rust-lang/rust#157096). **CI will be red until that compiler
change merges and syncs into the pinned toolchain here** — the
`#[target_feature(enable = "clflushopt")]` won't be recognized before then.

Modeled on the existing `_mm_clflush` / `adx` intrinsics: links
`llvm.x86.clflushopt`, `assert_instr(clflushopt)` codegen test, and a
`simd_test` runtime test. The Intel-verification false-match exemption for
`_mm_clflush` is extended to cover `_mm_clflushopt` (both take `*const u8`).

### DIFF
--- a/crates/core_arch/missing-x86.md
+++ b/crates/core_arch/missing-x86.md
@@ -70,12 +70,6 @@
 </p></details>
 
 
-<details><summary>["CLFLUSHOPT"]</summary><p>
-
-  * [x] [`_mm_clflushopt`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_clflushopt)
-</p></details>
-
-
 <details><summary>["CLWB"]</summary><p>
 
   * [ ] [`_mm_clwb`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_clwb)
--- a/crates/core_arch/missing-x86.md
+++ b/crates/core_arch/missing-x86.md
@@ -72,7 +72,7 @@
 
 <details><summary>["CLFLUSHOPT"]</summary><p>
 
-  * [ ] [`_mm_clflushopt`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_clflushopt)
+  * [x] [`_mm_clflushopt`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_clflushopt)
 </p></details>
 
 
--- a/crates/core_arch/src/x86/clflushopt.rs
+++ b/crates/core_arch/src/x86/clflushopt.rs
@@ -4,7 +4,7 @@
 use stdarch_test::assert_instr;
 
 #[allow(improper_ctypes)]
-unsafe extern "C" {
+unsafe extern "unadjusted" {
     #[link_name = "llvm.x86.clflushopt"]
     fn clflushopt(p: *const u8);
 }
@@ -26,6 +26,10 @@ unsafe extern "C" {
 /// Unlike the prefetch intrinsics, `CLFLUSHOPT` is subject to all the
 /// permission checking and faults associated with a byte load, so `p` must
 /// point to a byte that is valid for reads.
+///
+/// [`_mm_clflush`]: crate::arch::x86::_mm_clflush
+/// [`_mm_sfence`]: crate::arch::x86::_mm_sfence
+/// [`_mm_mfence`]: crate::arch::x86::_mm_mfence
 #[inline]
 #[target_feature(enable = "clflushopt")]
 #[cfg_attr(test, assert_instr(clflushopt))]
--- a/crates/core_arch/src/x86/clflushopt.rs
+++ b/crates/core_arch/src/x86/clflushopt.rs
@@ -26,10 +26,6 @@ unsafe extern "C" {
 /// Unlike the prefetch intrinsics, `CLFLUSHOPT` is subject to all the
 /// permission checking and faults associated with a byte load, so `p` must
 /// point to a byte that is valid for reads.
-///
-/// [`_mm_clflush`]: crate::arch::x86_64::_mm_clflush
-/// [`_mm_sfence`]: crate::arch::x86_64::_mm_sfence
-/// [`_mm_mfence`]: crate::arch::x86_64::_mm_mfence
 #[inline]
 #[target_feature(enable = "clflushopt")]
 #[cfg_attr(test, assert_instr(clflushopt))]
--- a/crates/core_arch/src/x86/clflushopt.rs
+++ b/crates/core_arch/src/x86/clflushopt.rs
@@ -6,7 +6,7 @@ use stdarch_test::assert_instr;
 #[allow(improper_ctypes)]
 unsafe extern "C" {
     #[link_name = "llvm.x86.clflushopt"]
-    fn clflushopt(p: *mut i8);
+    fn clflushopt(p: *const u8);
 }
 
 /// Invalidates from every level of the cache hierarchy the cache line that
@@ -21,6 +21,12 @@ unsafe extern "C" {
 ///
 /// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_clflushopt)
 ///
+/// # Safety
+///
+/// Unlike the prefetch intrinsics, `CLFLUSHOPT` is subject to all the
+/// permission checking and faults associated with a byte load, so `p` must
+/// point to a byte that is valid for reads.
+///
 /// [`_mm_clflush`]: crate::arch::x86_64::_mm_clflush
 /// [`_mm_sfence`]: crate::arch::x86_64::_mm_sfence
 /// [`_mm_mfence`]: crate::arch::x86_64::_mm_mfence
@@ -29,7 +35,7 @@ unsafe extern "C" {
 #[cfg_attr(test, assert_instr(clflushopt))]
 #[unstable(feature = "simd_x86_clflushopt", issue = "157096")]
 pub unsafe fn _mm_clflushopt(p: *const u8) {
-    clflushopt(p as *mut i8);
+    clflushopt(p);
 }
 
 #[cfg(test)]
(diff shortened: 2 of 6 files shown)


crates/core_arch/src/x86/clflushopt.rs
```diff
@@ -0,0 +1,45 @@
...
+/// contains `p`.
+///
+/// Unlike [`_mm_clflush`], `CLFLUSHOPT` is only ordered with respect to older
+/// writes to the flushed cache line and with respect to fence/locked
+/// operations; it is *not* serialized against other `CLFLUSHOPT`/`CLFLUSH`
+/// instructions or unrelated stores. This makes flushing a range of lines
+/// substantially faster, but a fence (e.g. [`_mm_sfence`] or [`_mm_mfence`]) is
+/// required afterward to order the flushes against subsequent operations.
+///
+/// [Intel's documentation](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_clflushopt)
```
folkertdev, line 22: Can you add a `# Safety` section here where you document that (unlike other prefetch functions) this function needs a valid pointer? Or at least that is what I think 

> subject to all permission checking and faults associated with a byte load

means? 

cc @Amanieu @sayantn if I'm misinterpreting that.



*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2141/65d15014c5bc1ff9459a5d6bdb2bff17de69bdb7..48f6b2fbd6c6d8396f220a0624100cdaa27e798a)*

folkertdev: For consistency you can add the same section to the `_mm_clflush` doc comment.

crates/core_arch/missing-x86.md
sayantn, line 1: Please do not modify this file manually, instead set `GENERATE_MISSING_X86_MD = true` in crates/stdarch-verify/tests/x86-intel.rs and run `cargo test -p stdarch-verify --test x86-intel`. That will automatically generate this file for you (I know, we need a better way to do this)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2141/65d15014c5bc1ff9459a5d6bdb2bff17de69bdb7..48f6b2fbd6c6d8396f220a0624100cdaa27e798a)*

crates/stdarch-verify/tests/x86-intel.rs
```diff
@@ -668,7 +668,7 @@ fn pointed_type(intrinsic: &Intrinsic) -> Result<Type, String> {
         {
             // AMX, KEYLOCKER and XSAVE intrinsics should take `*u8`
             U8
-        } else if intrinsic.name == "_mm_clflush" {
+        } else if intrinsic.name == "_mm_clflush" || intrinsic.name == "_mm_clflushopt" {
```
sayantn: IMO this can be better written in the part before this, by adding "CLFLUSHOPT" to the list. I added this specifically because it was not possible to distinguish `_mm_clflush` by a target feature.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2141/65d15014c5bc1ff9459a5d6bdb2bff17de69bdb7..48f6b2fbd6c6d8396f220a0624100cdaa27e798a)*

crates/core_arch/src/x86/clflushopt.rs
```diff
@@ -0,0 +1,45 @@
+//! `CLFLUSHOPT` cache-line flush.
+
+#[cfg(test)]
+use stdarch_test::assert_instr;
+
+#[allow(improper_ctypes)]
+unsafe extern "C" {
+    #[link_name = "llvm.x86.clflushopt"]
+    fn clflushopt(p: *mut i8);
```
sayantn: why use `*mut u8` here?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2141/65d15014c5bc1ff9459a5d6bdb2bff17de69bdb7..48f6b2fbd6c6d8396f220a0624100cdaa27e798a)*

jlpetz:
Thanks for the reviews! All four points addressed in 6f9fbf1, and with dc60b5f hopefully it's also passing CI now:

@folkertdev — Added a # Safety section to _mm_clflushopt documenting that, like you suggested. Unlike the prefetch intrinsics, it's subject to the permission checking and faults of a byte load, so p must point to a byte valid for reads. Added the same section to _mm_clflush for consistency.

@sayantn

- Extern signature *mut i8 → *const u8: Good catch. I'd originally mirrored the LLVM intrinsic's C declaration (void @llvm.x86.clflushopt(i8*)) literally, which forced the as *mut i8 cast at the call site. Mirroring the sibling _mm_clflush (fn clflush(p: *const u8)) is clearly better — const since a flush doesn't mutate the bytes, and u8 matches the public signature so the cast is not needed.
- missing-x86.md: Regenerated via GENERATE_MISSING_X86_MD = true + cargo test -p stdarch-verify --test x86-intel instead of hand-editing. Since _mm_clflushopt was the only intrinsic in the CLFLUSHOPT group, the regenerator seems to drop the whole section rather than marking it [x].
- pointed_type: Moved "CLFLUSHOPT" into the cpuid.iter() branch and reverted the _mm_clflush name match to its original form — agreed it's cleaner to distinguish by target feature where one exists, leaving the name hack only for _mm_clflush which lacks one.

crates/core_arch/src/x86/clflushopt.rs
```diff
@@ -0,0 +1,51 @@
...
+///
+/// # Safety
+///
+/// Unlike the prefetch intrinsics, `CLFLUSHOPT` is subject to all the
+/// permission checking and faults associated with a byte load, so `p` must
+/// point to a byte that is valid for reads.
+///
+/// [`_mm_clflush`]: crate::arch::x86_64::_mm_clflush
+/// [`_mm_sfence`]: crate::arch::x86_64::_mm_sfence
+/// [`_mm_mfence`]: crate::arch::x86_64::_mm_mfence
```
sayantn: I don't think you need these, rustdoc should be able to find these functions

sayantn (commented):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2141/65d15014c5bc1ff9459a5d6bdb2bff17de69bdb7..dc60b5f185fee86ef0ae97c222576530345ca9a1)*

crates/core_arch/src/x86/clflushopt.rs
jlpetz: Thanks — you're right that the explicit definitions ideally shouldn't be needed, but removing them broke the docs build (-D rustdoc::broken-intra-doc-links) on the Build Documentation job:

```
  error: unresolved link to `_mm_clflush`
    --> crates/core_arch/src/x86/clflushopt.rs:15:14
     |
  15 | /// Unlike [`_mm_clflush`], ...
     |              ^^^^^^^^^^^ no item named `_mm_clflush` in scope
```

The reason the bare links resolve in avx.rs/avx512f.rs but not here: those modules do use crate::core_arch::{x86::*}, which brings the intrinsics into scope. clflushopt.rs has no use imports at all, so rustdoc can't find _mm_clflush/_mm_sfence/_mm_mfence.

The reference-definitions I'd removed (crate::arch::x86_64::_mm_clflush etc.) did build green on every target including i686. So options to keep your suggestion's intent:

  1. Restore the original reference-definitions (crate::arch::x86_64::...) — builds everywhere.
  2. super:: definitions — these intrinsics are pub use'd at the x86 module level, so super::_mm_clflush should resolve target-agnostically and reads a bit cleaner than the absolute path.
  3. Add use crate::core_arch::x86::*; to match the other modules and keep the links bare — but pulls in a glob this file otherwise doesn't use.

  Any preference?

folkertdev: The `crate::arch` references are fine. rustdoc only looks at what is in scope.

folkertdev:
r? sayantn 

Given that you know more about our infra here.

crates/core_arch/src/x86/clflushopt.rs
```diff
@@ -0,0 +1,47 @@
+//! `CLFLUSHOPT` cache-line flush.
+
+#[cfg(test)]
+use stdarch_test::assert_instr;
+
+#[allow(improper_ctypes)]
+unsafe extern "C" {
```
sayantn:
```suggestion
unsafe extern "unadjusted" {
```


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2141/65d15014c5bc1ff9459a5d6bdb2bff17de69bdb7..64df57190aa3babb78c37ce8ba0430353b7e6784)*

crates/core_arch/src/x86/clflushopt.rs
sayantn: If that's the case then having these references is fine, but please make the references to crate::arch::x86 instead, otherwise a rustdoc lint will trigger in i686

jlpetz: Re-added with crate::arch::x86

crates/core_arch/src/x86/clflushopt.rs
jlpetz: Updated

sayantn (approved):

PR #2142, opened 2026-05-29
Re-enables & Fixes AArch64 SVE ld/st tests for MSVC target

CrooseGit:
The tests were failing as sometimes the FFR was being clobbered
between the `svsetffr` and `svrdffr`.
When `as_ptr` is called on the `LazyLock` data for the first time one
test thread will initialise it, while the others are blocked and wait.
The wait/block is achieved with a syscall which on windows just so
happens to sometimes clobber the FFR, as it is free to do as the FFR is
caller saved.
 
To fix this, we just ensure that the `as_ptr` is not called between the
set and read.


sayantn:
Sorry I have absolutely no knowledge about windows abi etc, I think this would be better reviewed by someone else.

@rustbot reroll

adamgemmell (approved):

PR #2144, opened 2026-05-31
Correct some wrong uses of LLVM intrinsics

sayantn:
There were some small mistakes in the LLVM intrinsics, these were autoupgraded by LLVM so the builds passed. Also replaces all `extern "C"` LLVM intrinsics blocks with `extern "unadjusted"`

## AArch64
cc @adamgemmell

 - wrong type parameter in `neon.vluti4q.lane`
 - missing type parameter in `prefetch`
 - typo in `sve.{and,bic,eor,orr,orn}.z`
 - extra `.x` in `sve.fcmla.lane` and `sve.frecpx`
 - trailing whitespace in `sve.expa.x`
 - one missing type parameter in `sve.fmmla`
 - splitting `svcvt` in two parts, because same size operands need a different intrinsic

## x86

 - missing type parameter in `prefetch`

## NVPTX
I don't know who to cc here, the file was last modified by @gnzlbg but he doesn't seem active here anymore

 - deprecated intrinsic `llvm.nvvm.barrier0` used

## WASM32
cc @alexcrichton

 - missing type parameter in `llvm.wasm.memory.{grow,size}`

r? @folkertdev

### DIFF
--- a/crates/core_arch/src/mips/msa.rs
+++ b/crates/core_arch/src/mips/msa.rs
@@ -45,7 +45,7 @@ types! {
 }
 
 #[allow(improper_ctypes)]
-unsafe extern "C" {
+unsafe extern "unadjusted" {
     #[link_name = "llvm.mips.add.a.b"]
     fn msa_add_a_b(a: v16i8, b: v16i8) -> v16i8;
     #[link_name = "llvm.mips.add.a.h"]
--- a/crates/core_arch/src/wasm32/memory.rs
+++ b/crates/core_arch/src/wasm32/memory.rs
@@ -2,9 +2,11 @@
 use stdarch_test::assert_instr;
 
 unsafe extern "unadjusted" {
-    #[link_name = "llvm.wasm.memory.grow"]
+    #[cfg_attr(target_pointer_width = "32", link_name = "llvm.wasm.memory.grow.i32")]
+    #[cfg_attr(target_pointer_width = "64", link_name = "llvm.wasm.memory.grow.i64")]
     fn llvm_memory_grow(mem: u32, pages: usize) -> usize;
-    #[link_name = "llvm.wasm.memory.size"]
+    #[cfg_attr(target_pointer_width = "32", link_name = "llvm.wasm.memory.size.i32")]
+    #[cfg_attr(target_pointer_width = "64", link_name = "llvm.wasm.memory.size.i64")]
     fn llvm_memory_size(mem: u32) -> usize;
 }
 
--- a/crates/core_arch/src/aarch64/neon/generated.rs
+++ b/crates/core_arch/src/aarch64/neon/generated.rs
@@ -11903,7 +11903,7 @@ pub unsafe fn vluti4q_lane_s8<const LANE: i32>(a: int8x16_t, b: uint8x8_t) -> in
     unsafe extern "unadjusted" {
         #[cfg_attr(
             any(target_arch = "aarch64", target_arch = "arm64ec"),
-            link_name = "llvm.aarch64.neon.vluti4q.lane.v8i8"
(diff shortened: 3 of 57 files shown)


folkertdev:
Out of curiosity, how did you find these?

crates/core_arch/src/mips/msa.rs
```diff
@@ -45,7 +45,7 @@ types! {
 }
 
 #[allow(improper_ctypes)]
-unsafe extern "C" {
+unsafe extern "unadjusted" {
```
folkertdev, line 48: @bjorn3 I vaguely remember you wanted to get rid of `unadjusted`, do you have issues with this change?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2144/6277e35542667f251dcbdaadcd95e0a2fcc74e79..7a2df97705163fcfcdd8853fbba64a80c461d54f)*

sayantn:
> Out of curiosity, how did you find these?

A modified version of rust-lang/rust#157145

crates/core_arch/src/mips/msa.rs
bjorn3: Now that LLVM intrinsics have an entirely separate code path that ignores the ABI, I'm fine with changing all LLVM intrinsic declarations to `extern "unadjusted"`. I would still like to make `extern "unadjusted"` for non-intrinsics a hard error at some point.

sayantn: > I would still like to make extern "unadjusted" for non-intrinsics a hard error at some point.

It would be absolutely lovely if that and its converse (only allowing unadjusted for intrinsics) happens. I am so tired of seeing extern C anywhere llvm intrinsics are used, where even the abi of vectors is not ideal!

folkertdev (approved):
This looks fine to me, but maybe target maintainers have thoughts?

Also I wonder if this breaks with LLVM 21 at all?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2144/6277e35542667f251dcbdaadcd95e0a2fcc74e79..7a2df97705163fcfcdd8853fbba64a80c461d54f)*

crates/core_arch/src/wasm32/memory.rs
alexcrichton, line 1: Despite the `wasm32` in the path of this filename this file is additionally used on wasm64 targets, so this will technically break those targets because it only works for wasm32. Is there a reason to change these signatures here on wasm?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2144/6277e35542667f251dcbdaadcd95e0a2fcc74e79..7a2df97705163fcfcdd8853fbba64a80c461d54f)*

folkertdev: Good point, we should gate these on the target arch (or pointer width or whatever is convenient).

sayantn: The LLVM intrinsics expect the output type as a type parameter, so I used u32 because I was unsure about usize size. If wasm64 also uses this then we can probably just `cfg_attr` two link_names.

sayantn: I have put the LLVM `link_name`s under `cfg_attr(target_pointer_width`

alexcrichton:
wasm bits lgtm, thanks!

adamgemmell (approved):
Looks good, thanks you!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2144/25138314b4ca69c596ea2d7ac5280911b29f0129..5f21f967726a2744552fa5a33d9b6bbeab4b06d2)*

PR #2145, opened 2026-06-01
Adds `cfg` to ensure little endian for SVE

CrooseGit:
SVE is only compatible with Little Endian architectures.


Amanieu:
r? davidtwco

davidtwco (approved):

PR #2147, opened 2026-06-01
Correct allowed IMM values for `cvtps_ph`

sayantn:
Related: [#t-libs/stdarch > Documentation of _mm256_cvtps_ph seems incorrect](https://rust-lang.zulipchat.com/#narrow/channel/208962-t-libs.2Fstdarch/topic/Documentation.20of.20_mm256_cvtps_ph.20seems.20incorrect/with/598860053)

This corrects all the `cvtps_ph` functions, almost all of them had wrong documentation, and all of them had wrong checks for the rounding imm. Also updates intrinsic-test with the correct values

r? @folkertdev

### DIFF
--- a/crates/core_arch/src/x86/avx512f.rs
+++ b/crates/core_arch/src/x86/avx512f.rs
@@ -15529,17 +15529,17 @@ pub fn _mm512_maskz_cvt_roundps_ph<const ROUNDING: i32>(k: __mmask16, a: __m512)
 
 /// Convert packed single-precision (32-bit) floating-point elements in a to packed half-precision (16-bit) floating-point elements, and store the results in dst using writemask k (elements are copied from src when the corresponding mask bit is not set).\
 /// Rounding is done according to the imm8\[2:0\] parameter, which can be one of:
-/// * [`_MM_FROUND_TO_NEAREST_INT`] | [`_MM_FROUND_NO_EXC`] : round to nearest and suppress exceptions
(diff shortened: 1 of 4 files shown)


sayantn:
There is a ~2m increase in CI time, because we are testing all possible values of `_MM_PERM`, not just till 32

folkertdev:
Hmm, we should start sharding these really...

sayantn:
> Hmm, we should start sharding these really...

Wdym exactly? I am pretty free for the next couple of days

folkertdev:
I mean somehow split the workload so that the total wall time is smaller. 

Maybe just using `cargo nextest` work. CI machines have 4 cores usually, but an individual nextest test runs slower, and we have many small tests so it's not as clear a win as it seems at first, but worth a try perhaps.

Alternatively we split e.g. the `_mm,` `_mm256` and `_mm512` tests into their own CI jobs still using `cargo test`?

sayantn:
The increase is from intrinsic-test, due to SDE being hopelessly slow. We can probably make intrinsic-test emit the tests in two batches, one with SSE, AVX and whatever else qemu can handle, and the other one with AVX512, Then we can probably use qemu for the first case, although I'm not sure how beneficial that would be.

also, interestingly the dev profile tests are slower, with clang being the fastest (~18m) and gcc and icx being similar (~21m). The release profile tests all take ~15m.

PR #2148, opened 2026-06-03
use zero-extension in the loongarch64 crc32 implementation

folkertdev:
Using sign extension caused some issues in miri which actually checks the range of the input.

### DIFF
--- a/crates/core_arch/src/arm_shared/neon/mod.rs
+++ b/crates/core_arch/src/arm_shared/neon/mod.rs
@@ -5326,6 +5326,23 @@ mod tests {
         test_vld3q_lane_f16(f16, 24, 7, float16x8x3_t, vst3q_lane_f16, vld3q_lane_f16);
         test_vld4q_lane_f16(f16, 32, 7, float16x8x4_t, vst4q_lane_f16, vld4q_lane_f16);
     }
+
+    // FIXME: simd_test (which uses is_arm_feature_detected) does not support the v8
+    // feature that we need on arm.
+    #[cfg(any(target_arch = "aarch64", target_arch = "arm64ec"))]
+    #[simd_test(enable = "crc")]
+    fn crc32() {
+        assert_eq!(__crc32b(u32::MAX, u8::MAX), 16777215);
+        assert_eq!(__crc32h(u32::MAX, u16::MAX), 65535);
+        assert_eq!(__crc32w(u32::MAX, u32::MAX), 0);
+
+        assert_eq!(__crc32cb(u32::MAX, u8::MAX), 16777215);
+        assert_eq!(__crc32ch(u32::MAX, u16::MAX), 65535);
+        assert_eq!(__crc32cw(u32::MAX, u32::MAX), 0);
+
+        assert_eq!(__crc32d(u32::MAX, u64::MAX), 3736805603);
+        assert_eq!(__crc32cd(u32::MAX, u64::MAX), 3080238136);
+    }
 }
 
 #[cfg(all(test, target_arch = "arm"))]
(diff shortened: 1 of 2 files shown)


folkertdev:
r? sayantn

crates/core_arch/src/loongarch64/mod.rs
```diff
@@ -64,14 +64,14 @@ unsafe extern "unadjusted" {
 #[inline(always)]
 #[unstable(feature = "stdarch_loongarch", issue = "117427")]
 pub fn crc_w_b_w(a: i8, b: i32) -> i32 {
```
sayantn, line 66: Should the signatures be signed or unsigned? Because typically crypto always uses unsigned

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2148/05ced9d5e41f0a1eb7b3ff2b3de7629a66545366..b383d31e4c23abfbb37d6fb7e3f7d34bc72c992f)*

crates/core_arch/src/arm_shared/neon/mod.rs
```diff
@@ -5326,6 +5326,20 @@ mod tests {
         test_vld3q_lane_f16(f16, 24, 7, float16x8x3_t, vst3q_lane_f16, vld3q_lane_f16);
         test_vld4q_lane_f16(f16, 32, 7, float16x8x4_t, vst4q_lane_f16, vld4q_lane_f16);
     }
+
+    #[cfg_attr(target_arch = "arm", target_feature(enable = "crc,v8"))]
+    #[cfg_attr(not(target_arch = "arm"), target_feature(enable = "crc"))]
```
sayantn: I think you meant to use `simd_test` here instead of `target_feature`

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2148/05ced9d5e41f0a1eb7b3ff2b3de7629a66545366..b383d31e4c23abfbb37d6fb7e3f7d34bc72c992f)*

crates/core_arch/src/loongarch64/mod.rs
folkertdev: Clang uses signed, see also discussion at https://github.com/rust-lang/stdarch/pull/2137#issuecomment-4557687422

folkertdev:
Should be good now, we can't run this on arm at the moment due to the feature detection macro not supporting a feature we need.

sayantn (approved):

heiher:
Would it make sense to handle this in Miri instead? Casting to an unsigned type isn't free, and `crc.w.{b,h}.w` only looks at the relevant low bits, ignoring the upper bits.

```rust
fn crc32_test(a: i8, b:i8) -> i32 {
    let r = crc_w_b_w(a, -1);
    crc_w_b_w(b, r)
}
```
```asm
b4800:   02fffc06    li.d            $a2, -1
b4804:   00241884    crc.w.b.w       $a0, $a0, $a2
b4808:   002410a4    crc.w.b.w       $a0, $a1, $a0
b480c:   4c000020    ret
```

->

```asm
b4800:   0343fc84    andi            $a0, $a0, 0xff
b4804:   02fffc06    li.d            $a2, -1
b4808:   00241884    crc.w.b.w       $a0, $a0, $a2
b480c:   0343fca5    andi            $a1, $a1, 0xff
b4810:   002410a4    crc.w.b.w       $a0, $a1, $a0
b4814:   4c000020    ret
```

heiher:
> Would it make sense to handle this in Miri instead?

https://github.com/rust-lang/miri/pull/5102

folkertdev:
I'd expect LLVM to optimize the mask out when the instruction that uses the result won't look at the upper bits anyway?

heiher:
> I'd expect LLVM to optimize the mask out when the instruction that uses the result won't look at the upper bits anyway?

I've opened a new PR (https://github.com/llvm/llvm-project/pull/203201) to propagate demanded bits for CRC instructions. If there are no objections, I'll remove the unsigned cast before Rust bumps LLVM to a version that includes this PR. Thanks!

PR #2157, opened 2026-06-09
Add stdarch-gen-common: shared check/bless harness for generators

xonx4l:
This PR adds a library crate `stdarch-gen-common` that gives every generator three modes -:

1) Write - Generator writes files straight into the committed location.
2) Check - Generator write into a temporary directory . The crate diffs it against committed and fails on any mismatch without touching the working copy .
3) Bless - Generator writes into a temp directory the crate then mirrors temp over committed. Adding, overwriting, and deleting files as needed.

Mode is picked via the` STDARCH_GEN_MODE` env var. The harness only touches files the generator explicitly lists as its own so hand-written files in the same directory are safe.

This PR also hooks `stdarch-gen-hexagon` to use it as the first example. 

r? @folkertdev

### DIFF
--- a/crates/stdarch-gen-common/src/lib.rs
+++ b/crates/stdarch-gen-common/src/lib.rs
@@ -1,0 +1,352 @@
+//! Shared check/bless harness for stdarch generators.
+
+use std::error::Error as StdError;
+use std::fmt;
+use std::fs;
+use std::io;
+use std::io::Read;
+use std::path::{Path, PathBuf};
+
+/// First-line marker identifying an auto-generated file. Generators emit this
+/// as the first line of every file they produce; the harness uses it to
+/// discover which files in `committed` are owned by the generator.
+pub const GENERATED_MARKER: &str = "// This code is automatically generated. DO NOT MODIFY.";
+
+/// Controls what `run_generator` does with the generator's output.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Mode {
+    /// Verify that the `committed` matches the generator's output for owned files.
+    ///
+    /// Runs the generator into a temp directory, then compares each produced
+    /// file against the committed copy. Returns an error on the first mismatch.
+    Check,
+    /// Update the `committed` to match the generator's output for owned files.
+    ///
+    /// Runs the generator into a temp directory and copies each produced file
+    /// into `committed`. If the generator no longer produces an owned file, the
+    /// committed copy is deleted. Files in `committed` that are not owned
+    /// are left untouched.
+    Bless,
+}
+
+impl Mode {
+    /// Read the mode from the `STDARCH_GEN_MODE` environment variable.
+    ///
+    /// Recognized values:
+    /// - `"check"` → [`Mode::Check`]
+    /// - `"bless"` → [`Mode::Bless`]
+    /// - unset → [`Mode::Bless`]
+    /// - any other value → panic
+    pub fn from_env() -> Self {
+        match std::env::var("STDARCH_GEN_MODE").as_deref() {
+            Ok("check") => Mode::Check,
+            Ok("bless") => Mode::Bless,
+            Ok(other) => panic!("unknown STDARCH_GEN_MODE value: {other:?}"),
+            Err(_) => Mode::Bless,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum Error {
+    Io(io::Error),
+    Mismatch { path: PathBuf, kind: MismatchKind },
+    Generator(Box<dyn StdError + Send + Sync>),
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum MismatchKind {
+    /// Owned file produced by the generator but absent from the `committed`.
+    /// Means the `committed` needs to be regenerated.
+    MissingInCommitted,
+    /// Owned file present in the `committed` but the generator no longer
+    /// produces it. The file must be removed from the `committed` .
+    ExtraInCommitted,
+    /// Owned file exists on both sides but contents differ.
+    ContentsDiffer,
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Error::Io(e) => write!(f, "I/O error: {e}"),
+            Error::Mismatch { path, kind } => match kind {
+                MismatchKind::MissingInCommitted => {
+                    write!(f, "{}: generated but not committed", path.display())
+                }
+                MismatchKind::ExtraInCommitted => {
+                    write!(f, "{}: committed but no longer generated", path.display())
+                }
+                MismatchKind::ContentsDiffer => write!(f, "{}: contents differ", path.display()),
+            },
+            Error::Generator(e) => write!(f, "generator failed: {e}"),
+        }
+    }
+}
+
+impl StdError for Error {
+    fn source(&self) -> Option<&(dyn StdError + 'static)> {
+        match self {
+            Error::Io(e) => Some(e),
+            Error::Generator(e) => Some(&**e),
+            _ => None,
+        }
+    }
+}
+
+impl From<io::Error> for Error {
+    fn from(e: io::Error) -> Self {
+        Error::Io(e)
+    }
+}
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+/// Run a generator under the chosen `mode`, reconciling its output with `committed`.
+///
+/// Arguments:
+/// - `committed` — the directory holding the in-tree (committed) source files.
+///   Files inside `committed` whose first line begins with [`GENERATED_MARKER`]
+///   are treated as owned by the generator. Anything else is treated as
+///   hand-written and is left untouched by `Bless` and ignored by `Check`.
+///   So generated files coexist with hand-written files in the same  directory.
+/// - `mode` — what to do with the generator's output.
+/// - `generate` — closure that writes the generator's output into the
+///   directory it is given. Its error is wrapped in [`Error::Generator`].
+///
+/// Behavior per mode:
+/// - [`Mode::Check`]: runs the generator into a temp dir and compares owned
+///   files against the committed copies. Mismatch returns [`Error::Mismatch`].
+/// - [`Mode::Bless`]: runs the generator into a temp dir and copies owned
+///   files into `committed`, or removes `committed`'s copy if the generator no
+///   longer produces them.
+pub fn run_generator<F, E>(committed: &Path, mode: Mode, generate: F) -> Result<()>
+where
+    F: FnOnce(&Path) -> std::result::Result<(), E>,
+    E: Into<Box<dyn StdError + Send + Sync>>,
+{
+    let scratch = tempfile::tempdir()?;
+    generate(scratch.path()).map_err(|e| Error::Generator(e.into()))?;
+
+    let owned = discover_owned(committed)?;
+    let produced = discover_all(scratch.path())?;
+
+    let mut names: Vec<&String> = owned.iter().chain(produced.iter()).collect();
+    names.sort();
+    names.dedup();
+
+    for name in names {
+        match mode {
+            Mode::Check => compare(scratch.path(), committed, name)?,
+            Mode::Bless => apply_bless(scratch.path(), committed, name)?,
+        }
+    }
+    Ok(())
+}
+
+/// Returns the names of files in `dir` whose first line begins with
+/// [`GENERATED_MARKER`]. Files without the marker are skipped.
+fn discover_owned(dir: &Path) -> Result<Vec<String>> {
+    let mut out = Vec::new();
+    for entry in fs::read_dir(dir)? {
+        let entry = entry?;
+        if !entry.file_type()?.is_file() {
+            continue;
+        }
+        let Ok(mut file) = fs::File::open(entry.path()) else {
+            continue;
+        };
+        let mut buf = [0u8; GENERATED_MARKER.len()];
+        if file.read_exact(&mut buf).is_err() {
+            continue;
+        }
+        if buf == *GENERATED_MARKER.as_bytes()
+            && let Some(name) = entry.file_name().to_str()
+        {
+            out.push(name.to_owned());
+        }
+    }
+    out.sort();
+    Ok(out)
+}
+
+/// Returns every file name in `dir` (scratch dir — all files are generator output).
+fn discover_all(dir: &Path) -> Result<Vec<String>> {
+    let mut out = Vec::new();
+    for entry in fs::read_dir(dir)? {
+        let entry = entry?;
+        if entry.file_type()?.is_file()
+            && let Some(name) = entry.file_name().to_str()
+        {
+            out.push(name.to_string());
+        }
+    }
+    Ok(out)
+}
+
+fn compare(generated_dir: &Path, committed_dir: &Path, filename: &str) -> Result<()> {
+    let rel_path = PathBuf::from(filename);
+    let gen_path = generated_dir.join(&rel_path);
+    let comm_path = committed_dir.join(&rel_path);
+    match (gen_path.exists(), comm_path.exists()) {
+        (true, false) => Err(Error::Mismatch {
+            path: rel_path,
+            kind: MismatchKind::MissingInCommitted,
+        }),
+        (false, true) => Err(Error::Mismatch {
+            path: rel_path,
+            kind: MismatchKind::ExtraInCommitted,
+        }),
+        (false, false) => Ok(()),
+        (true, true) => {
+            if fs::read(&gen_path)? != fs::read(&comm_path)? {
+                Err(Error::Mismatch {
+                    path: rel_path,
+                    kind: MismatchKind::ContentsDiffer,
+                })
+            } else {
+                Ok(())
+            }
+        }
+    }
+}
+
+fn apply_bless(generated_dir: &Path, committed_dir: &Path, filename: &str) -> Result<()> {
+    fs::create_dir_all(committed_dir)?;
+    let rel_path = PathBuf::from(filename);
+    let from = generated_dir.join(&rel_path);
+    let to = committed_dir.join(&rel_path);
+    if from.exists() {
+        if let Some(parent) = to.parent() {
+            fs::create_dir_all(parent)?;
+        }
+        fs::copy(&from, &to)?;
+    } else if to.exists() {
+        fs::remove_file(&to)?;
+    }
+    Ok(())
+}
+
+// Skipped on iOS because these tests fail on the `x86_64-apple-ios-macabi` CI runner
+// with `Os { code: 17, kind: AlreadyExists, message: "File exists" }`.
+#[cfg(all(test, not(target_os = "ios")))]
+mod tests {
+    use super::*;
+
+    fn write_marker(p: &Path, body: &[u8]) {
+        if let Some(d) = p.parent() {
+            fs::create_dir_all(d).unwrap();
+        }
+        let mut bytes = Vec::new();
+        bytes.extend_from_slice(GENERATED_MARKER.as_bytes());
+        bytes.push(b'\n');
+        bytes.extend_from_slice(body);
+        fs::write(p, bytes).unwrap();
+    }
+
+    #[test]
+    fn check_fails_on_byte_diff() {
+        let tmp = tempfile::tempdir().unwrap();
+        let committed = tmp.path().join("c");
+        write_marker(&committed.join("a.txt"), b"hi");
+        let e = run_generator(
+            &committed,
+            Mode::Check,
+            |out| -> std::result::Result<(), io::Error> {
+                write_marker(&out.join("a.txt"), b"HI");
+                Ok(())
+            },
+        )
+        .unwrap_err();
+        assert!(matches!(
+            e,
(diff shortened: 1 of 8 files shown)


Kobzol:
For a bit more context, this is done for the stdarch GSoC project, where we try to integrate stdarch's tests in rust-lang/rust's CI. We want to move the blessing logic out of the CI YAML files and into Rust, to make it easier to reuse, and also share code amongst the various generators.

I'll take a look, but also CC @folkertdev , in case you have in mind who owns this code amongst the stdarch maintainers and who could also chime in.

crates/stdarch-gen-common/src/lib.rs
```diff
@@ -0,0 +1,309 @@
+//! Shared check/bless harness for stdarch generators.
+
+use std::error::Error as StdError;
+use std::fmt;
+use std::fs;
+use std::io;
+use std::path::{Path, PathBuf};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Mode {
```
Kobzol, line 17: How does Write differ from Bless? I'd expect those two modes to be the same. Could you please add some doc comments on the mode to explain what the individual modes mean? Thanks!

crates/stdarch-gen-common/src/lib.rs
```diff
@@ -0,0 +1,309 @@
...
+    Write,
+    Check,
+    Bless,
+}
+
+impl Mode {
+    pub fn from_env() -> Self {
```
Kobzol, line 40: Please add a doc comment here that explains how to use the function, i.e. which env var it expects to read and what are the possible values.

crates/stdarch-gen-common/src/lib.rs
```diff
@@ -0,0 +1,309 @@
...
+    fn from(e: io::Error) -> Self {
+        Error::Io(e)
+    }
+}
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+// owned is the list of relative paths the generator manages. Files in
+// committed not listed here are left untouched by Bless and ignored by Check.
+pub fn run<F, E>(committed: &Path, owned: &[&str], mode: Mode, generate: F) -> Result<()>
```
Kobzol: Could you please add a doc comment that explains in more detail what the function is supposed to be doing? The signature is kinda complex.

I think that we can simplify the function by changing it to only work with a single subdirectory, it doesn't really need to handle slices and multiple paths at once.

crates/stdarch-gen-common/src/lib.rs
```diff
@@ -0,0 +1,309 @@
...
+            Error::Io(e) => write!(f, "I/O error: {e}"),
+            Error::Mismatch { path, kind } => match kind {
+                MismatchKind::MissingInCommitted => {
+                    write!(f, "{}: generated but not committed", path.display())
+                }
+                MismatchKind::ExtraInCommitted => {
+                    write!(f, "{}: committed but no longer generated", path.display())
+                }
+                MismatchKind::ContentsDiffer => write!(f, "{}: contents differ", path.display()),
+                MismatchKind::MissingFromGenerated => write!(
```
Kobzol: When we bless, it can happen that the new blessed output has less files than before, right? 🤔

Kobzol (commented):
Thanks for creating the crate! I left a few comments.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2157/afc40bbc4fde7999324bb58fba692b6a51dc9290..9d9978ab1cd91e4ec018ddc721612bf485dd3c3d)*

crates/stdarch-gen-common/src/lib.rs
xonx4l: Added doc comments .The two modes hit  different code paths in run write skips the owned-list filter entirely.

crates/stdarch-gen-common/src/lib.rs
xonx4l: Added .

crates/stdarch-gen-common/src/lib.rs
xonx4l: Yes there's a case for it already ` MissingFromGenerated`. Added the comment that explains . Thank you!

crates/stdarch-gen-common/src/lib.rs
xonx4l: Added doc comment  explaining the signature .  I kept the slice instead of switching to one folder per generator because generated files are aligned next to hand-written ones in the same folder  . if we owned the whole folder bless would  delete those hand-written files. Thank you!

Kobzol: I don't understand, you can still just take the for loop that is now executed inside `run`, and perform it outside the function, by just passing it a since `owned` string, right?

crates/stdarch-gen-common/src/lib.rs
Kobzol: These comments use a lot of terms and I'm not sure what do they mean 😆

What is target directory? How does "committed tree" differ from "owned list"?

crates/stdarch-gen-common/src/lib.rs
Kobzol: What I meant was that `MissingFromGenerated` shouldn't happen, if something is not present in the generated output, we should delete it from the committed tree.

Kobzol (commented):
I think I'll have to examine how the generator behaves a bit deeper (I'll try to do that on Monday), because I'm still quite surprised from the behavior here 😅 I think that we might wanna do some targeted changes to the generator and the directory layout to make blessing simpler, if the complexity here is caused by the current directory layout.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2157/afc40bbc4fde7999324bb58fba692b6a51dc9290..9aca67e8d3cd933b0e8cded351bf607a6b873551)*

crates/stdarch-gen-common/src/lib.rs
xonx4l: Yes makes sense , the the loop can move out.

crates/stdarch-gen-common/src/lib.rs
xonx4l: Haha I used those in order to differentiate  better and more readable but I think rewording made them confusing . Here target directory  and committed tree mean  the same thing  the folder the generator writes into  to make it less confusing I will simplify it naming to just committed and owned list means the list of file names inside that folder that this generator is responsible for .

crates/stdarch-gen-common/src/lib.rs
xonx4l: Ok will drop.

xonx4l:
> I think I'll have to examine how the generator behaves a bit deeper (I'll try to do that on Monday), because I'm still quite surprised from the behavior here 😅 I think that we might wanna do some targeted changes to the generator and the directory layout to make blessing simpler, if the complexity here is caused by the current directory layout.
> 
> _[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2157/afc40bbc4fde7999324bb58fba692b6a51dc9290..9aca67e8d3cd933b0e8cded351bf607a6b873551)_

Yes targeted changes to the generator will likely make this crate much simpler. Btw we are in a fixing loop :)

crates/stdarch-gen-common/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+    ///
+    /// Recognized values:
+    /// - `"check"` → [`Mode::Check`]
+    /// - `"bless"` → [`Mode::Bless`]
+    /// - anything else, including unset → [`Mode::Write`]
+    pub fn from_env() -> Self {
+        match std::env::var("STDARCH_GEN_MODE").as_deref() {
+            Ok("check") => Mode::Check,
+            Ok("bless") => Mode::Bless,
+            _ => Mode::Write,
```
folkertdev: I'd rather panic on an unknown value?  Unless this is standard practice somehow?

crates/stdarch-gen-common/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+///
+/// Behavior per mode:
+/// - [`Mode::Write`]: invokes `generate(committed)` directly. owned is
+///   not consulted.
+/// - [`Mode::Check`]: runs the generator into a temp dir and compares
+///   `owned` against the committed copy. Mismatch returns [`Error::Mismatch`].
+/// - [`Mode::Bless`]:  runs the generator into a temp dir and copies `owned`
+///   into `committed`, or removes `committed`'s copy if the generator no
+///   longer produces it.
+pub fn run<F, E>(committed: &Path, owned: &str, mode: Mode, generate: F) -> Result<()>
```
folkertdev:
```suggestion
pub fn run_generator<F, E>(committed: &Path, owned: &str, mode: Mode, generate: F) -> Result<()>
```

maybe? `run` itself is quite vague at the use site.

folkertdev (commented):
Overall this seems fine as a design. I'd probably just do the other generators in a separate PR using this as a foundation.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2157/afc40bbc4fde7999324bb58fba692b6a51dc9290..f10694472306c2d2e212c42cc634d5849d6f48ef)*

crates/stdarch-gen-common/src/lib.rs
Kobzol: Good catch, let's panic in case the value is unexpected.

Kobzol:
I looked at the code more and now I understand the issue with blessing; as you said, the `mod.rs` file is manually written, and we sometimes generate more than one file (e.g. `sve/generated.rs` and `sve/ld_st_tests_aarch64.rs`), so we cannot depend on the generated filename to be just `generated.rs`, and we have to deal with multiple files.

I would suggest the following changes and simplifications:

When an output directory is passed to the generator, it will load all files from that directory, and keep a _map_ of files that are automatically generated. These files will be detected by reading their first line and checking if it begins with `// This code is automatically generated. DO NOT MODIFY.`. Note that this line should be stored in a string constant and be automatically added to the start of all generated files; this should be done by `stdarch-gen-common`, but that's for a later PR.

Keep only two modes:
- `Check` will generate files into a temporary directory (or even just into a string, it seems like the existing code works with `Box<dyn Write>`, so it shouldn't be difficult to write to a string instead of having to deal with temporary directories and files) , and compare their contents with the _map_. If there are any extra or missing files in the map, error out. If the contents do not match, error out.
- `Bless` will write the generated output directly into the target directory. Any files from the map that were not (over)written will be deleted.

This assumes that two generators won't ever write into the same directory, but I think that assumption should be fine. In fact I think that it would be easiest to simply move all the generators into a single crate, at which point we could generate everything at once, but let's not complicate it further for now.

Does that make sense to you?

crates/stdarch-gen-common/src/lib.rs
xonx4l: Ok, Understood.

crates/stdarch-gen-common/src/lib.rs
xonx4l: Yeah seems better  will rename to `run_generator` and update the callers. Thanks for the reviews means a lot .

xonx4l:
@Kobzol  Yes , this makes sense to me. 

I will refactor as per the idea you suggested as I think is quite good than per file version( especially the header one is a great way to diff) .  The arm and loongarch generated files already carry the header `// This code is automatically generated. DO NOT MODIFY` . So only for the files generated by  hexagon generators needs the addition of  the header .

And yeah alright two modes look's fine will drop `Write` mode and keep only `Check` and `Bless` as described in the idea.

Thanks for taking the time to dig into this, it's a huge help.

xonx4l:
@Kobzol made all the changes that were suggested and discussed . Thank you!

crates/stdarch-gen-common/src/lib.rs
```diff
@@ -103,43 +105,72 @@ pub type Result<T> = std::result::Result<T, Error>;
...
 {
-    match mode {
-        Mode::Write => {
-            fs::create_dir_all(committed)?;
-            generate(committed).map_err(|e| Error::Generator(e.into()))
+    let scratch = tempfile::tempdir()?;
+    generate(scratch.path()).map_err(|e| Error::Generator(e.into()))?;
+
+    let owned = discover_owned(committed)?;
+    let produced = discover_owned(scratch.path())?;
```
Kobzol: This should just read all files from the scratch directory. It could help find bugs where we actually generate files that do not have the auto-generated marker header.

crates/stdarch-gen-common/src/lib.rs
```diff
@@ -103,43 +105,72 @@ pub type Result<T> = std::result::Result<T, Error>;
...
-            compare(scratch.path(), committed, owned)
+    }
+    Ok(())
+}
+
+/// Returns the names of files in `dir` whose first line begins with
+/// [`GENERATED_MARKER`]. Files without the marker are skipped.
+fn discover_owned(dir: &Path) -> Result<Vec<String>> {
+    let mut out = Vec::new();
+    if !dir.exists() {
```
Kobzol: If the directory does not exist, this should be an error (you don't need to check for it explicitly, it would happen in `fs::read_dir(dir)?` below, so just remove this `if` condition).

crates/stdarch-gen-common/src/lib.rs
```diff
@@ -103,43 +105,72 @@ pub type Result<T> = std::result::Result<T, Error>;
...
+    let owned = discover_owned(committed)?;
+    let produced = discover_owned(scratch.path())?;
+
+    let mut names: Vec<&String> = owned.iter().chain(produced.iter()).collect();
+    names.sort();
+    names.dedup();
+
+    for name in names {
+        match mode {
+            Mode::Check => compare(scratch.path(), committed, name)?,
```
Kobzol, line 140: In the `compare` and `apply_bless` functions, let's rename the arguments to `generated_dir`, `committed_dir`, `filename`.

Kobzol (commented):
Thank you, looks great! Left a few minor comments.

On CI, you should modify the `check-stdarch-gen` job to add the environment variable `STDARCH_GEN_MODE=check`.

The output could be improved a bit, e.g. the "Output: " path could be normalized and the filename of the mismatched file could be made absolute, to make it easier to find where should the file be located:
```
Step 3: Generating v64.rs and v128.rs (mode: Check)...
  Output: /projects/personal/rust/stdarch/crates/stdarch-gen-hexagon/../core_arch/src/hexagon/v64.rs
  Output: /projects/personal/rust/stdarch/crates/stdarch-gen-hexagon/../core_arch/src/hexagon/v128.rs
Error: "v128.rs: contents differ"
```
but those are minor things and I don't think we need to deal with that in this PR.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2157/afc40bbc4fde7999324bb58fba692b6a51dc9290..1fad13781bf8ea4849450d3c1d9aadbfe04793c1)*

crates/stdarch-gen-common/src/lib.rs
xonx4l: Ok, makes sense .

crates/stdarch-gen-common/src/lib.rs
xonx4l: Ok

crates/stdarch-gen-common/src/lib.rs
xonx4l: Ok will rename . Thanks for the reviews!

Kobzol (approved):
Thank you, looks great! I don't have approval rights on this repo, so I'll ask Folkert to take another look.

Edit: nevermind, apparently I do have merge rights :laughing: I'll still wait for Folkert to chime in though.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2157/afc40bbc4fde7999324bb58fba692b6a51dc9290..83725022c4b64647150f249f618c363e4bcf7c76)*

xonx4l:
Damn so fast was just going to ping :) , Yeah we need those  Folkert opinions and feedbacks.

crates/stdarch-gen-common/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+fn discover_owned(dir: &Path) -> Result<Vec<String>> {
+    let mut out = Vec::new();
+    for entry in fs::read_dir(dir)? {
+        let entry = entry?;
+        if !entry.file_type()?.is_file() {
+            continue;
+        }
+        let Ok(contents) = fs::read_to_string(entry.path()) else {
+            continue;
+        };
```
folkertdev: This is perhaps a bit wasteful? you can just read the first `GENERATED_MARKER.len()` bytes and just check those?

crates/stdarch-gen-common/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+    Ok(out)
+}
+
+/// Returns every file name in `dir` (scratch dir — all files are generator output).
+fn discover_all(dir: &Path) -> Result<Vec<String>> {
+    let mut out = Vec::new();
+    for entry in fs::read_dir(dir)? {
+        let entry = entry?;
+        if entry.file_type()?.is_file() {
+            if let Some(name) = entry.file_name().to_str() {
```
folkertdev:
```suggestion
        if entry.file_type()?.is_file() 
            && let Some(name) = entry.file_name().to_str() {
```

you can use `let` in that position these days.

crates/stdarch-gen-common/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+            path: rel_path,
+            kind: MismatchKind::MissingInCommitted,
+        }),
+        (false, true) => Err(Error::Mismatch {
+            path: rel_path,
+            kind: MismatchKind::ExtraInCommitted,
+        }),
+        (false, false) => Ok(()),
+        (true, true) => {
+            if fs::read(&gen_path)? != fs::read(&comm_path)? {
```
folkertdev, line 202: this is probably fine for now, but a more robust and efficient approach is to check the file sizes first, and if they are the same, read the files in chunks and compare those.

crates/stdarch-gen-common/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+            fs::create_dir_all(parent)?;
+        }
+        fs::copy(&from, &to)?;
+    } else if to.exists() {
+        fs::remove_file(&to)?;
+    }
+    Ok(())
+}
+
+#[cfg(all(test, not(target_os = "ios")))]
```
folkertdev, line 232: please add a comment on why ios does not work?

crates/stdarch-gen-hexagon/src/main.rs
```diff
@@ -1691,23 +1693,19 @@ fn main() -> Result<(), String> {
...
-    println!("  Output: {}", v64_path.display());
-
-    // Generate v128.rs (128-byte vector mode)
-    let v128_path = hexagon_dir.join("v128.rs");
-    println!("\nStep 4: Generating v128.rs (128-byte mode)...");
-    generate_module_file(&intrinsics, &v128_path, VectorMode::V128)?;
-    println!("  Output: {}", v128_path.display());
+    let hexagon_dir = crate_dir.join("../core_arch/src/hexagon");
+
+    let mode = Mode::from_env();
```
folkertdev, line 1700: maybe add a comment here on what the mode means, e.g.

```suggestion
    // Either "check" to check the output versus the committed output, or "bless" to update the output.
    let mode = Mode::from_env();
```

folkertdev (commented):
Looks good overall, some final small things.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2157/afc40bbc4fde7999324bb58fba692b6a51dc9290..83725022c4b64647150f249f618c363e4bcf7c76)*

crates/stdarch-gen-common/src/lib.rs
xonx4l: Ok got it.

crates/stdarch-gen-common/src/lib.rs
xonx4l: Okay , makes sense.

crates/stdarch-gen-hexagon/src/main.rs
xonx4l: Ok , got it . Thanks for the reviews.

xonx4l:
@folkertdev made all the requested changes . Thank you!

crates/stdarch-gen-common/src/lib.rs
folkertdev:
```
// Skipped on iOS because these tests fail on the `x86_64-apple-ios-macabi` CI runner
// with `Os { code: 17, kind: AlreadyExists, message: "File exists" }`.
```

So, some sort of bug in `tempdir`? Anyway, probably fine.

folkertdev (approved):
Great, can you squash the commits? 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2157/afc40bbc4fde7999324bb58fba692b6a51dc9290..132e9238cafc74b3194434ad8e25911eb064222a)*

xonx4l:
@folkertdev squashed the commits . Thank you!

PR #2160, opened 2026-06-15
intrinsic-test: sve support 

davidtwco:
This patch contains all the changes necessary for `intrinsic-test` to generate tests for the SVE intrinsics:

- ~~Some of the commits are just general refactors to the tool, motivated by the commits that follow and that I felt were hard to justify on their own~~
  - All of the refactorings prior to this PR have already been landed (rust-lang/stdarch#2121, rust-lang/stdarch#2122, rust-lang/stdarch#2124, rust-lang/stdarch#2126, rust-lang/stdarch#2127)
  - All of the refactorings split out of this PR have landed (rust-lang/stdarch#2169, rust-lang/stdarch#2170, rust-lang/stdarch#2171, rust-lang/stdarch#2172, rust-lang/stdarch#2172, rust-lang/stdarch#2173)
- ~~Some of the commits are small changes I made along the way (e.g. forwarding args from `intrinsic-test.sh` to `cargo test`)~~
  - ~~I could split these out if we wanted~~
  - Merged in rust-lang/stdarch#2164 
- ~~Some of the commits modify the definitions of intrinsics so that they pass tests - this only happened for a handful of intrinsics and the changes required to their definitions were very minor~~
  - ~~I could split these out if we wanted~~
  - Merged in rust-lang/stdarch#2163
- ~~Some of the commits update `arm_intrinsics.json` to add some additional constraints necessary to generate the right tests (we're making sure that these addl. constraints are reflected in the source of truth we use to generate that file internally)~~
  - ~~I could split these out if we wanted~~
  - Merged in rust-lang/stdarch#2162 
- Any remaining commits are changes to make the tool generate the correct tests, primarily:
  - Adding the `arm_sve.h` headers and appropriate target feature flags for C and Rust
  - Using the `svld1` intrinsic when loading an argument of scalable vector type
    - There is a special case for `svbool_t` arguments which do not have an `svld1` intrinsic, we load a `svint8_t` and use `svcmpne_n_s8` against zero to produce a `svbool_t` from it
  - Generating test arrays for intrinsics that take boolean values and `svbool_t` values (though the earlier commits use the same `svptrue_$ty` predicate for these arguments initially)
  - Supporting calls to enum-typed const generic arguments (such as `svpattern`), using simple const functions that map integers from intrinsic-test constraints to the appropriate variant
  - Generating a local containing a predicate that is used by all intrinsics which require a predicate (just `svptrue_$ty()` to enable every lane)
  - Using `svcmpeq_$ty` and `svptest_any` to compare the Rust and C results instead of `assert_eq!`
    - There is a special-case for tuples of vectors, where multiple `svcmpeq_$ty` and `svptest_any` blocks are generated for each vector in the tuple, with `svgetN` calls to extract the appropriate vector from the result tuples
    - There is also a special-case for intrinsics that return `svbool_t`, where there is no equivalent `svcmpeq_$ty` call, so `sveor_b_z` (exclusive OR) and `!svptest_any` is used
    - There is yet another special case for scalable vectors of floats, where the pre-existing `NanEqF*` type cannot be used, so a `(rust == c) || (isnan(rust) && isnan(c))` check is performed instead, via a handful of SVE intrinsics
- I haven't been been able to test this on x86 to check that I haven't broken the test generation, but I don't think I've made any changes to the generated output that weren't in the Arm-specific parts

### DIFF
--- a/crates/intrinsic-test/src/arm/types.rs
+++ b/crates/intrinsic-test/src/arm/types.rs
@@ -72,30 +72,41 @@ impl TypeDefinition for ArmType {
 
     /// Determines the load function for this type.
     fn load_function(&self) -> String {
-        if let IntrinsicType {
-            kind: k,
-            bit_len: Some(bl),
-            vec_len,
-            ..
-        } = **self
-        {
-            let quad = if self.num_lanes() * bl > 64 { "q" } else { "" };
-
-            format!(
-                "vld{len}{quad}_{type}{size}",
-                type = match k {
-                    TypeKind::Int(Sign::Unsigned) => "u",
-                    TypeKind::Int(Sign::Signed) => "s",
-                    TypeKind::Float => "f",
-                    TypeKind::Poly => "p",
-                    x => todo!("get_load_function TypeKind: {x:#?}"),
-                },
-                size = bl,
-                quad = quad,
-                len = vec_len.unwrap_or(1),
-            )
+        if let Some(bl) = self.bit_len {
+            match self.num_lanes() {
+                SimdLen::Scalable => {
+                    format!(
+                        "svld{len}_{type}{bl}",
+                        len = self.num_vectors(),
+                        type = self.rust_intrinsic_name_prefix(),
+                    )
+                }
+                SimdLen::Fixed(num_lanes) => {
+                    format!(
+                        "vld{len}{quad}_{type}{bl}",
+                        quad = if num_lanes * bl > 64 { "q" } else { "" },
+                        len = self.num_vectors(),
+                        type = self.rust_intrinsic_name_prefix(),
+                    )
+                }
+            }
         } else {
             todo!("load_function IntrinsicType: {self:#?}")
+        }
+    }
+}
+
+impl ArmType {
+    /// Returns the Rust prefix for the name of an intrinsic with this type kind (i.e. `s` for
+    /// `i16`, or `u` for `u16`). For type kinds without any bit length at the end (e.g. `bool`),
+    /// returns the whole type name.
+    pub fn rust_intrinsic_name_prefix(&self) -> &str {
+        match self.kind() {
+            TypeKind::Char(Sign::Signed) => "s",
+            TypeKind::Int(Sign::Signed) => "s",
+            TypeKind::Poly => "p",
+            TypeKind::Bool => "s",
+            _ => self.kind.rust_prefix(),
         }
     }
 }
--- a/crates/intrinsic-test/src/arm/types.rs
+++ b/crates/intrinsic-test/src/arm/types.rs
@@ -1,5 +1,9 @@
 use super::intrinsic::ArmType;
-use crate::common::intrinsic_helpers::{IntrinsicType, Sign, SimdLen, TypeDefinition, TypeKind};
+use crate::common::PREDICATE_LOCAL;
+use crate::common::intrinsic_helpers::{
+    IntrinsicType, Sign, SimdLen, TypeDefinition, TypeKind, default_fixed_vector_comparison,
+};
+use itertools::Itertools;
 
 impl TypeDefinition for ArmType {
     /// Gets a string containing the typename for this type in C format.
@@ -92,6 +96,58 @@ impl TypeDefinition for ArmType {
             }
         } else {
             todo!("load_function IntrinsicType: {self:#?}")
+        }
+    }
+
+    fn comparison_function(&self) -> String {
+        match self.num_lanes() {
+            SimdLen::Scalable => {
+                // There isn't a `svcmpeq` for `svbool_t`, so do an XOR instead and test it is
+                // empty..
+                if self.kind() == TypeKind::Bool {
+                    return format!(
+                        r#"
+let eq = sveor_b_z({PREDICATE_LOCAL}, __rust_return_value, __c_return_value);
+assert!(!svptest_any({PREDICATE_LOCAL}, eq), "{{}}", id);
+                    "#,
+                    );
+                }
+
+                // Use `svcmpeq` to compare the return values of Rust and C invocations
+                match self.num_vectors() {
+                    1 => {
+                        format!(
+                            r#"
+let eq = svcmpeq_{ty}{bl}({PREDICATE_LOCAL}, __rust_return_value, __c_return_value);
+assert!(svptest_any(__pred, eq), "{{}}", id);
+                        "#,
+                            ty = self.rust_intrinsic_name_prefix(),
+                            bl = self.inner_size(),
+                        )
+                    }
+                    // For tuples of vectors, do multiple comparisons, each with a `svget` to
+                    // extract the Nth vector.
+                    n @ (2 | 3 | 4) => (0..n)
+                        .format_with("\n", |i, fmt| {
+                            fmt(&format_args!(
+                                r#"
+let eq = svcmpeq_{ty}{bl}(
+    {PREDICATE_LOCAL},
+    svget{n}_{ty}{bl}::<{i}>(__rust_return_value),
+    svget{n}_{ty}{bl}::<{i}>(__c_return_value)
+);
+assert!(svptest_any(__pred, eq), "{{}}-{i_plus_one}/{n}", id);
+                            "#,
+                                ty = self.rust_intrinsic_name_prefix(),
+                                bl = self.inner_size(),
+                                i_plus_one = i + 1, // so that the output is "1/2" and "2/2"
+                            ))
+                        })
+                        .to_string(),
+                    _ => unreachable!(),
+                }
+            }
+            SimdLen::Fixed(num_lanes) => default_fixed_vector_comparison(self, num_lanes),
         }
     }
 }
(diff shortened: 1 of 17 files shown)


crates/intrinsic-test/src/common/intrinsic.rs
```diff
@@ -48,17 +27,11 @@ impl<T: IntrinsicTypeDefinition> Intrinsic<T> {
...
-                .filter_map(|arg| arg.constraint.as_ref()),
-            &mut Vec::new(),
-            &mut f,
-        )
+    pub fn specializations(&self) -> impl Iterator<Item = Vec<i64>> {
+        self.arguments
+            .iter()
+            .filter_map(|arg| arg.constraint.as_ref())
+            .map(|constraint| constraint.into_iter())
+            .multi_cartesian_product()
```
sayantn, line 37: that's pretty neat, didn't know about this function

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2160/6332f19038d35b94f7f091d1ef2a6cce789ceb10..ed132a7308f617515269787c4ef8a6077ae02474)*

crates/intrinsic-test/src/arm/types.rs
```diff
@@ -1,79 +1,200 @@
...
-    fn get_load_function(&self) -> String {
-        if let IntrinsicType {
-            kind: k,
-            bit_len: Some(bl),
-            vec_len,
-            ..
-        } = **self
-        {
-            let quad = if self.num_lanes() * bl > 64 { "q" } else { "" };
+    fn comparison_function(&self) -> String {
```
sayantn, line 112: Is there some way to print the diff for SVE? I would imagine it is difficult due to SVE being non-const sized, but is there some way? That would massively improve debuggability. One approach I can suggest is have some small functions that convert from SVE vectors to `&[T]`, e.g.
```rust
#[target_feature(enable = "sve")]
pub fn svfloat32_to_slice(a: &svfloat32_t) -> &[NanEqF32] {
    unsafe {
        core::slice::from_raw_parts(core::ptr::from_ref(a).cast(), svcntw() as usize)
    }
}
```
this might work, with significantly less complexity

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2160/6332f19038d35b94f7f091d1ef2a6cce789ceb10..ed132a7308f617515269787c4ef8a6077ae02474)*

crates/intrinsic-test/src/common/gen_c.rs
```diff
@@ -14,39 +14,40 @@ use super::intrinsic_helpers::IntrinsicTypeDefinition;
...
-    }
-
-    for intrinsic in intrinsics {
-        intrinsic.iter_specializations(|imm_values| {
-            writeln!(
-                w,
-                "
+{intrinsics}
+"#,
+        intrinsics = intrinsics.iter().format_with("", |intrinsic, fmt| {
```
sayantn, line 33: very minor nit, but you can use something like `intrinsic.specializations().for_each(|imm_values| fmt(...))`

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2160/6332f19038d35b94f7f091d1ef2a6cce789ceb10..ed132a7308f617515269787c4ef8a6077ae02474)*

sayantn:
also, it would be nice if this can be split into a few focused PRs (I can help if you want)

davidtwco:
Submitted rust-lang/rust#158088 to fix the failure in debug mode - should pass CI after that

crates/intrinsic-test/src/arm/types.rs
davidtwco: When I've needed to debug them, I've disabled the invocation of `intrinsic-test` in `intrinsic-test.sh` and added these snippets to the generated tests that I wanted to debug.

For non-bool vectors, just replace the `x0_val` and `x1_val` with the vectors that you want to inspect, e.g. `__rust_return_value` or `__c_return_value`:

```rust
{
    let num_elems = svcnth() as usize;
    let mut x0_buf = Vec::with_capacity(num_elems);
    let mut x1_buf = Vec::with_capacity(num_elems);
    svst1_u16(__pred, x0_buf.as_mut_ptr(), x0_val);
    x0_buf.set_len(num_elems);
    svst1_u16(__pred, x1_buf.as_mut_ptr(), x1_val);
    x1_buf.set_len(num_elems);
    for i in 0..num_elems {
        let x0_val = x0_buf[i];
        dbg!(i, x0_val);
    }
    for i in 0..num_elems {
        let x1_val = x1_buf[i];
        dbg!(i, x1_val);
    }
}
```

Similarly for bool vectors:

```rust
{
    let num_elems = svcntb() as usize;
    let mut _op_val = Vec::with_capacity(num_elems);
    svst1_u8(op_val, _op_val.as_mut_ptr(), svdup_n_u8(1));
    _op_val.set_len(num_elems);
    for i in 0..num_elems {
        let _op_val_el = if _op_val[i] == 1 { true } else { false };
        dbg!(i, _op_val_el);
    }
}
```

I'm not sure if what you've suggested will work, happy to try. I'm more than happy to make some improvements here for debuggability of the tests when they do fail, but could that be left to a follow-up?

davidtwco:
> also, it would be nice if this can be split into a few focused PRs (I can help if you want)

I'll try and split some out tomorrow

davidtwco:
Split out the first batch of changes in #2169

davidtwco:
Split out the second batch of changes in #2170

davidtwco:
Split out the third batch of changes in #2171

davidtwco:
Split out fourth batch of changes in #2172

davidtwco:
Split out final batch of changes in #2173. Shouldn't be much less other than the actual SVE testing implementation here after those are merged, but there will be a few merge conflicts between them as they land.

crates/intrinsic-test/src/common/gen_c.rs
davidtwco: I wouldn't have a `fmt` to call if I used `for_each`

davidtwco:
Rebased after all the prerequisites landed - should pass CI now, works in dev and release locally and the rustc patches are all in nightly - but I'll check back in tomorrow

crates/intrinsic-test/src/arm/types.rs
davidtwco: Ended up doing this in the current patch because it was useful for working out what was going on with CI failures

davidtwco:
Pushed c89cacc to fix the failures with `svtrn{1,2}`, missed these locally as they require `f32mm`/`f64mm` and my local test machine didn't have those, but it's the same fix as rust-lang/stdarch#2163; and also pushed c04381b to print out the contents of scalable vectors when the tests fail.

Only remaining failures are `svcreate` and `svset`. These are curious because I already fixed them locally a while ago with a rustc patch that has long landed. They only fail with the second, third and fourth vectors in the tuple, and when I had investigated this previously, we were generating a `memcpy` that just didn't copy the second, third and fourth vectors. I fixed that in rust-lang/rust#157915 by correcting how rustc understood the size of these types and that fixed these tests locally for me - but they seem to still be failing in a very similar way here, it's very strange.

davidtwco:
> Only remaining failures are `svcreate` and `svset`. These are curious because I already fixed them locally a while ago with a rustc patch that has long landed. They only fail with the second, third and fourth vectors in the tuple, and when I had investigated this previously, we were generating a `memcpy` that just didn't copy the second, third and fourth vectors. I fixed that in [rust-lang/rust#157915](https://github.com/rust-lang/rust/pull/157915) by correcting how rustc understood the size of these types and that fixed these tests locally for me - but they seem to still be failing in a very similar way here, it's very strange.

I think it's just that rust-lang/rust#157915 is wrong - it works by chance for me locally because of the vector length my hardware has (and let me assume that maybe `llvm.memcpy` has some special handling when the type being copied was a scalable vector), but under QEMU w/ `sve512`, it doesn't copy enough.

davidtwco:
This will be fixed once rust-lang/rust#158253 lands and is on nightly

crates/intrinsic-test/src/arm/types.rs
sayantn: Sorry, didn't understand what you meant by `this` and `the current patch`, the GH version shows the `sveor` version. Also, https://godbolt.org/z/a481YdW1P seems to work.

davidtwco: Yeah - by "this", I meant making it easier to debug failures, and "this current patch" being in this PR rather than a follow-up. I did it using the snippets that I shared rather than yours, while that patch does compile, I'm not sure it actually works in practice, see https://godbolt.org/z/ozev8Krqr - once you actually try and use those functions, you get an error from LLVM (that's something we'll need to fix on the rustc side so that we emit an error instead if you do something you shouldn't, but it is still a work-in-progress afterall).

davidtwco:
Finally passing CI!

sayantn (approved):

PR #2161, opened 2026-06-15
PowerPC: Add vec_add/vec_sub/vec_mul intrinsics support for vector_double

lei137:
Add intrinsics support on Linux on Power for:
* vec_add(vector_double, vector_double)
* vec_sub(vector_double, vector_double)
* vec_mul(vector_double, vector_double)

AI Assissted.

### DIFF
--- a/crates/core_arch/src/powerpc/vsx.rs
+++ b/crates/core_arch/src/powerpc/vsx.rs
@@ -10,6 +10,7 @@
 
 use crate::core_arch::powerpc::*;
 use crate::core_arch::simd::*;
+use crate::intrinsics::simd::simd_add;
 
 #[cfg(test)]
 use stdarch_test::assert_instr;
@@ -171,6 +172,27 @@ mod sealed {
     vec_mergeeo! { vector_unsigned_int, mergee, mergeo }
     vec_mergeeo! { vector_bool_int, mergee, mergeo }
     vec_mergeeo! { vector_float, mergee, mergeo }
+
+    #[inline]
+    #[target_feature(enable = "vsx")]
+    #[cfg_attr(test, assert_instr(xvadddp))]
+    pub(crate) unsafe fn vec_add_double_double(
+        a: vector_double,
+        b: vector_double,
+    ) -> vector_double {
+        simd_add(a, b)
+    }
+}
+
+// Implement AltiVec's VectorAdd trait for vector_double to enable vec_add support
+#[unstable(feature = "stdarch_powerpc", issue = "111145")]
+impl crate::core_arch::powerpc::altivec::sealed::VectorAdd<vector_double> for vector_double {
+    type Result = vector_double;
+    #[inline]
+    #[target_feature(enable = "vsx")]
+    unsafe fn vec_add(self, other: vector_double) -> Self::Result {
+        sealed::vec_add_double_double(self, other)
+    }
 }
 
 /// Vector permute.
@@ -255,4 +277,15 @@ mod tests {
     test_vec_xxpermdi! {test_vec_xxpermdi_i64x2, i64x2, vector_signed_long, [0], [-1], [2], [-3]}
     test_vec_xxpermdi! {test_vec_xxpermdi_m64x2, m64x2, vector_bool_long, [false], [true], [false], [true]}
     test_vec_xxpermdi! {test_vec_xxpermdi_f64x2, f64x2, vector_double, [0.0], [1.0], [2.0], [3.0]}
+
+    #[simd_test(enable = "vsx")]
+    fn test_vec_add_f64x2_f64x2() {
+        let a = vector_double::from(f64x2::from_array([1.0, 2.0]));
+        let b = vector_double::from(f64x2::from_array([3.0, 4.0]));
+        let expected = vector_double::from(f64x2::from_array([4.0, 6.0]));
+
+        unsafe {
+            assert_eq!(f64x2::from(vec_add(a, b)), f64x2::from(expected));
+        }
+    }
 }
--- a/crates/core_arch/src/powerpc/vsx.rs
+++ b/crates/core_arch/src/powerpc/vsx.rs
@@ -10,7 +10,7 @@
 
 use crate::core_arch::powerpc::*;
 use crate::core_arch::simd::*;
-use crate::intrinsics::simd::simd_add;
+use crate::intrinsics::simd::{simd_add, simd_sub};
 
 #[cfg(test)]
 use stdarch_test::assert_instr;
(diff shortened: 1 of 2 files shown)


lei137:
This is my first rust patch and is the first of a series of vsx intrinsics support patches.
Feedback on how the stdarch community like these type of implementation patches to be done is greatly appreciated.

crates/core_arch/src/powerpc/vsx.rs
```diff
@@ -171,6 +172,27 @@ mod sealed {
...
+// Implement AltiVec's VectorAdd trait for vector_double to enable vec_add support
+#[unstable(feature = "stdarch_powerpc", issue = "111145")]
+impl crate::core_arch::powerpc::altivec::sealed::VectorAdd<vector_double> for vector_double {
+    type Result = vector_double;
+    #[inline]
+    #[target_feature(enable = "vsx")]
+    unsafe fn vec_add(self, other: vector_double) -> Self::Result {
+        sealed::vec_add_double_double(self, other)
+    }
 }
```
folkertdev: So is the argument here that if you got your hands on a `vector_double` then `vsx` must be enabled? 

Because normally adding stricter target features on a trait method is error-prone: there is now an additional safety requirement on that trait method, which is that if that particular instance is called, the target feature must be enabled.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2161/1b87998d1a3985812a1cd4dc7748bf71520ef2b7..e67122b3e7e261aec84e56981e2afc2d2dabf767)*

crates/core_arch/src/powerpc/vsx.rs
```diff
@@ -171,6 +172,27 @@ mod sealed {
...
+    }
+}
+
+// Implement AltiVec's VectorAdd trait for vector_double to enable vec_add support
+#[unstable(feature = "stdarch_powerpc", issue = "111145")]
+impl crate::core_arch::powerpc::altivec::sealed::VectorAdd<vector_double> for vector_double {
+    type Result = vector_double;
+    #[inline]
+    #[target_feature(enable = "vsx")]
+    unsafe fn vec_add(self, other: vector_double) -> Self::Result {
```
folkertdev: is this the only function you'd want to do this for? I'd guess not?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2161/1b87998d1a3985812a1cd4dc7748bf71520ef2b7..e67122b3e7e261aec84e56981e2afc2d2dabf767)*

crates/core_arch/src/powerpc/vsx.rs
lei137: Yes that was the idea.  On PPC, vec_add(vector_double, vector_double) should result in code gen `xvadddp`, which is only available if vsx feature is enabled.

crates/core_arch/src/powerpc/vsx.rs
lei137: Sorry, I do not understand what you mean.  Can you please elaborate?

folkertdev: well what about `vec_sub` and other `vec_*` methods? Should they not also follow this pattern, or is `vec_add` special somehow?

crates/core_arch/src/powerpc/vsx.rs
folkertdev: Hmm yeah the `s390x` implementation has the same sort of problem with `float` in their case. Many instructions for `f32` were only added in a later target feature. Actually this is fine for `vec_add` because the implementation is just `simd_add`, which will generate `xvadddp` when the right target features are available, or some fallback thing (using scalars, probably) when that target feature is not available.

crates/core_arch/src/powerpc/vsx.rs
lei137: Ah yes!  I do have a few more other ones to add.  I tried to limit this PR to just `vec_add` as it is my first patch so I was just trying to get a feel on how these should be done.  

I have a list of intrinsics that I need to add and the plan was to follow up with sets of 3/4 intrinsics based on type/complexity afterwards.  Is there a preferred PR size/breakdown?

lei137: updated to add support also for vec_sub and vec_mul.

crates/core_arch/src/powerpc/vsx.rs
folkertdev: We discussed this more at [#t-libs/stdarch > target featurs on trait impls](https://rust-lang.zulipchat.com/#narrow/channel/208962-t-libs.2Fstdarch/topic/target.20featurs.20on.20trait.20impls/with/604617807), and the conclusion is that this approach is unsound. 

Which is to say, you cannot (soundly) require more target features on a specific impl than in the trait definition. 

---

Perhaps we should back up a bit here: why are you looking to add this part of the API now?

lei137: Thank-you for looking into this!

What is a better approach?  I thought this was the way to do it but happy to go a different direction.  I basically just want to add vector_double support for various intrinsics that currently are supported for other types.  The restriction is that the ppc instruction they are bound to will be supported in different CPUs.  

Eg. altivec features are pre pwr7, vsx support started in  pwr7,  however alot of our vsx instructions are only available starting in pwr8 so will require `target_feature = "power8-vector"` etc...

folkertdev: Unfortunately, in current rust there is no way to work around this problem. 

We have some (rather speculative) ideas, e.g. struct target features and effectful target feature, to be able to do this in the future.

Given that such a solution is probably years out, I wonder why you are looking to add this part of the API now?

folkertdev: One thing we could do is add the instances, but without the `vsx` target feature. Then, if the feature is not available, LLVM will generate a fallback (likely using scalar operations). But when `vsx` is enabled at compile time, you get the expected codegen.

This only works for functions with a portable fallback (using `core::intrinsics::simd`).

If that path is acceptable, I'd prefer the instances in the altivec file (so we don't need to open up the sealed module), and clearly documented as "this is really a vsx implementation but adding the target feature would be unsound".

lei137: Are you suggesting to just use `enable = "altivec"` for these instances?  I noticed that currently in altivec.rs there is code that use that even though the instruction being generated is vsx and not altivec.

eg.  xxsldwi is a vsx instruciton.
```
    macro_rules! impl_vec_sld {
        ($($ty:ident),+) => { $(
            #[unstable(feature = "stdarch_powerpc", issue = "111145")]
            impl VectorSld for $ty {
                #[inline]
                #[target_feature(enable = "altivec")]
                unsafe fn vec_sld<const UIMM4: i32>(self, b: Self) -> Self {
                    transmute(vsldoi::<UIMM4>(transmute(self), transmute(b)))
                }
                #[inline]
                #[target_feature(enable = "altivec")]
                unsafe fn vec_sldw<const UIMM2: i32>(self, b: Self) -> Self {
                    transmute(xxsldwi::<UIMM2>(transmute(self), transmute(b)))
                }
           }
        ) };
    }
```

As you mentioned this would work for these functions, but falls apart if I want to specify a specific intrinsic that doesn't easily fall under a generic moniker like vec_add.  I have a few draft PRs that will need to do that as well.  What do you suggest for those cases?

PR #2163, opened 2026-06-16
core_arch: redefine `svrev`, `svzip` and `svuzp`

davidtwco:
Split out from rust-lang/stdarch#2160.

Clang uses the `llvm.aarch64.sve.rev.bN` intrinsic for `svrev` with `b16`, `b32` and `b64`. Likewise with `llvm.aarch64.sve.zip.bN` for `svzip` and `llvm.aarch64.sve.uzp.bN` for `svuzp`. This required small generator changes so it knew a bool-to-bool conversion was a no-op and a new blanket identity impl of `SveInto` so the calls generated compile. These intrinsics were just returning the wrong value prior to this change, but obviously this isn't tested until rust-lang/stdarch#2160 lands.

r? @adamgemmell

### DIFF
--- a/crates/core_arch/src/aarch64/sve/generated.rs
+++ b/crates/core_arch/src/aarch64/sve/generated.rs
@@ -35226,19 +35226,6 @@ pub fn svreinterpret_u64_u64(op: svuint64_t) -> svuint64_t {
     unsafe { crate::intrinsics::transmute_unchecked(op) }
 }
 #[doc = "Reverse all elements"]
-#[doc = "[Arm's documentation](https://developer.arm.com/architectures/instruction-sets/intrinsics/svrev_b8)"]
-#[inline]
-#[target_feature(enable = "sve")]
-#[unstable(feature = "stdarch_aarch64_sve", issue = "145052")]
-#[cfg_attr(test, assert_instr(rev))]
-pub fn svrev_b8(op: svbool_t) -> svbool_t {
-    unsafe extern "unadjusted" {
-        #[cfg_attr(target_arch = "aarch64", link_name = "llvm.vector.reverse.nxv16i1")]
-        fn _svrev_b8(op: svbool_t) -> svbool_t;
-    }
-    unsafe { _svrev_b8(op) }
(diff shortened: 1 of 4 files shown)


crates/core_arch/src/aarch64/sve/generated.rs
```diff
@@ -35226,30 +35226,17 @@ pub fn svreinterpret_u64_u64(op: svuint64_t) -> svuint64_t {
...
 #[inline]
 #[target_feature(enable = "sve")]
 #[unstable(feature = "stdarch_aarch64_sve", issue = "145052")]
 #[cfg_attr(test, assert_instr(rev))]
 pub fn svrev_b16(op: svbool_t) -> svbool_t {
     unsafe extern "unadjusted" {
-        #[cfg_attr(target_arch = "aarch64", link_name = "llvm.vector.reverse.nxv8i1")]
-        fn _svrev_b16(op: svbool8_t) -> svbool8_t;
+        #[cfg_attr(target_arch = "aarch64", link_name = "llvm.aarch64.sve.rev.b16")]
+        fn _svrev_b16(op: svbool_t) -> svbool_t;
```
adamgemmell, line 35237: Why the change away from lane-count-specific svbool types? I think LLVM still uses them

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2163/1b87998d1a3985812a1cd4dc7748bf71520ef2b7..a8f7bfc4607ae62174b8f5e3cf462b5a9c1f69ce)*

davidtwco: I was getting a signature mismatch between the LLVM intrinsic and our signature here without this change.

adamgemmell: Fair enough, this https://godbolt.org/z/8v1W7TnYq is little odd considering all the other intrinsics seem to use different lane counts for their predicates

adamgemmell (approved):

PR #2178, opened 2026-07-01
Port stdarch-gen-loongarch to stdarch-gen-common harness

xonx4l:
This PR ports `stdarch-gen-loongarch` to `stdarch-gen-common` harness. 

The change in `main` invokes `run_generator` once per committed dir when called with no args or `lsx`/`lasx`.

The CI switches from `-- <spec>` to `-- lsx` / `-- lasx` with `STDARCH_GEN_MODE=check`.

Swapped the first two comment lines of `lsx/tests.rs` and `lasx/tests.rs` so the auto-generated marker moves to line 2 . So `discover_owned` only checks the first line and  makes the harness skip `tests.rs` which is not produced by `stdarch-gen-loongarch` .

r? @folkertdev

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -328,13 +328,17 @@ jobs:
       run: |
         cargo run --bin=stdarch-gen-arm --release -- crates/stdarch-gen-arm/spec
         git diff --exit-code
-    - name: Check lsx.spec
+    - name: Check loongarch lsx
+      env:
+        STDARCH_GEN_MODE: check
       run: |
-        cargo run --bin=stdarch-gen-loongarch --release -- crates/stdarch-gen-loongarch/lsx.spec
+        cargo run -p stdarch-gen-loongarch --release -- lsx
         git diff --exit-code
-    - name: Check lasx.spec
+    - name: Check loongarch lasx
+      env:
+        STDARCH_GEN_MODE: check
       run: |
-        cargo run --bin=stdarch-gen-loongarch --release -- crates/stdarch-gen-loongarch/lasx.spec
+        cargo run -p stdarch-gen-loongarch --release -- lasx
         git diff --exit-code
     - name: Check hexagon
       env:
--- a/crates/core_arch/src/loongarch64/lsx/tests.rs
+++ b/crates/core_arch/src/loongarch64/lsx/tests.rs
@@ -1,5 +1,5 @@
+// Auto-generated tests. DO NOT MODIFY.
 // See crates/stdarch-gen-loongarch/README.md
-// This code is automatically generated. DO NOT MODIFY.
 
 use crate::{
     core_arch::{loongarch64::*, simd::*},
--- a/crates/core_arch/src/loongarch64/lsx/tests.rs
+++ b/crates/core_arch/src/loongarch64/lsx/tests.rs
@@ -1,5 +1,5 @@
-// This code is automatically generated. DO NOT MODIFY.
 // See crates/stdarch-gen-loongarch/README.md
+// This code is automatically generated. DO NOT MODIFY.
 
 use crate::{
     core_arch::{loongarch64::*, simd::*},
--- a/crates/stdarch-gen-loongarch/src/main.rs
+++ b/crates/stdarch-gen-loongarch/src/main.rs
@@ -1604,10 +1604,11 @@ static void {current_name}(void)
 /// no args or a bare ext name.
 pub fn main() -> Result<(), String> {
     let args: Vec<String> = env::args().collect();
-    let harness_exts: Option<&[&str]> = match args.len() {
-        1 => Some(&["lsx", "lasx"]),
-        2 if args[1] == "lsx" => Some(&["lsx"]),
-        2 if args[1] == "lasx" => Some(&["lasx"]),
+    let arg_strs: Vec<&str> = args.iter().map(String::as_str).collect();
+    let harness_exts: Option<&[&str]> = match arg_strs.as_slice() {
+        [_] => Some(&["lsx", "lasx"]),
+        [_, "lsx"] => Some(&["lsx"]),
+        [_, "lasx"] => Some(&["lasx"]),
         _ => None,
     };
     if let Some(exts) = harness_exts {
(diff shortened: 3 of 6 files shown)


.github/workflows/main.yml
```diff
@@ -328,13 +328,17 @@ jobs:
       run: |
         cargo run --bin=stdarch-gen-arm --release -- crates/stdarch-gen-arm/spec
         git diff --exit-code
-    - name: Check lsx.spec
+    - name: Check loongarch lsx
+      env:
+        STDARCH_GEN_MODE: check
       run: |
-        cargo run --bin=stdarch-gen-loongarch --release -- crates/stdarch-gen-loongarch/lsx.spec
+        cargo run -p stdarch-gen-loongarch --release -- lsx
```
Kobzol, line 335: Ultimately, this will still be a bit annoying to replicate in bootstrap, as we have to ensure that we invoke both the right crates, and with the right arguments.

That is not for this PR, but I'd think about merging everything into a single binary that will generate everything, I think that would be easier to work with.

crates/core_arch/src/loongarch64/lsx/tests.rs
```diff
@@ -1,5 +1,5 @@
-// This code is automatically generated. DO NOT MODIFY.
 // See crates/stdarch-gen-loongarch/README.md
+// This code is automatically generated. DO NOT MODIFY.
```
Kobzol: Since the `tests.rs` files are not generated using the `stdarch-gen-common` machinery, let's use a different header for them, to avoid this line moving hack.

By the way, if you regenerate the Loongarch `tests.rs` files now, are they the same as they committed versions?

Kobzol (commented):
Looks reasonable, thanks! I suppose that we could just generate even the spec and tests using the new `stdarch-gen-common` machinery? Or do they differ somehow from `gen_bind`?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2178/80ade55ace1388cbb634f06174f6c0813747250b..04459874d7d56507710121e4e0ec150e59cd17b0)*

crates/core_arch/src/loongarch64/lsx/tests.rs
xonx4l: Added a different header . As for regenerating the tests.rs i haven't actually run it. Thank you!

.github/workflows/main.yml
xonx4l: Agreed.

folkertdev:
cc @heiher

heiher:
Thanks! LoongArch SIMD intrinsics are being gradually migrated to `intrinsics::simd` to make them easier for Miri to execute. These portable impls are handwritten rather than generated, so the LoongArch SIMD code generator is expected to be phased out.

folkertdev:
> so the LoongArch SIMD code generator is expected to be phased out.

I don't think you can eliminate it entirely. Anyhow this is just a heads-up that we're refactoring it a bit to make it easier to run in `rust-lang/rust` CI.

crates/stdarch-gen-loongarch/src/main.rs
```diff
@@ -1604,28 +1600,52 @@ static void {current_name}(void)
...
     let args: Vec<String> = env::args().collect();
-    let in_file = args.get(1).cloned().expect("Input file missing!");
-    let in_file_path = PathBuf::from(&in_file);
-    let in_file_name = in_file_path
+    let harness_exts: Option<&[&str]> = match args.len() {
+        1 => Some(&["lsx", "lasx"]),
+        2 if args[1] == "lsx" => Some(&["lsx"]),
+        2 if args[1] == "lasx" => Some(&["lasx"]),
+        _ => None,
+    };
```
folkertdev: nitpicky but I think this should work
```suggestion
    let harness_exts: Option<&[&str]> = match &args {
        [_] => Some(&["lsx", "lasx"]),
        [_, "lsx"] => Some(&["lsx"]),
        [_, "lasx"] => Some(&["lasx"]),
        _ => None,
    };
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2178/80ade55ace1388cbb634f06174f6c0813747250b..a375893d7974a2f808a280a3afebab5e79734f24)*

crates/stdarch-gen-loongarch/src/main.rs
```diff
@@ -1604,28 +1600,52 @@ static void {current_name}(void)
...
-
     if in_file_name.ends_with(".h") {
-        gen_spec(in_file, ext_name)
-    } else if args.get(2).is_some() {
-        gen_test(in_file, ext_name)
-    } else {
-        gen_bind(in_file, ext_name)
+        return gen_spec(in_file, ext_name).map_err(|e| e.to_string());
+    }
+    if args.get(2).is_some() {
```
folkertdev:
```suggestion
    if let [_, _lsx_or_lasx, _some_name_for_what_this_is] = &args {
```

I kind of lost what is in `args[2]` and what it's supposed to indicate.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2178/80ade55ace1388cbb634f06174f6c0813747250b..a375893d7974a2f808a280a3afebab5e79734f24)*

heiher:
> > so the LoongArch SIMD code generator is expected to be phased out.
> 
> I don't think you can eliminate it entirely. Anyhow this is just a heads-up that we're refactoring it a bit to make it easier to run in `rust-lang/rust` CI.

LoongArch currently generates tests using random input values, with the expected results produced by running them under QEMU. We don't currently fix the random seed, so regenerating the tests produces a different version each time.

If maintaining the LoongArch generator is making this work more difficult, another option would be to simply check in the currently generated code and keep them as-is. In any case, most of them are expected to be rewritten eventually.

crates/stdarch-gen-loongarch/src/main.rs
xonx4l: `args[2]` is the literal string "test"  it flips this into `gen_test` mode.

folkertdev:
```suggestion
    if let [_, _lsx_or_lasx, "test"] = &args {
```

is much clearer then.

xonx4l: YES .

folkertdev:
Closing and opening again to restart CI

xonx4l:
WOW! it worked!!!

PR #2180, opened 2026-07-02
loongarch: Use `intrinsics::simd` for vbs{l,r}l

heiher:


folkertdev:
Kind of verbose but not worth it to be clever I guess.

PR #2186, opened 2026-07-08
Use mirrored llvm artifacts

jieyouxu:
## Summary

Use llvm artifacts mirrored under `ci-mirrors` instead of going to kernel.org (which presumably is under stress following the outage from a few days ago).

- Needs https://github.com/rust-lang/ci-mirrors/pull/49 (already merged).

Reported in [#t-infra > kernel.org is borked (stdarch variant)](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/kernel.2Eorg.20is.20borked.20.28stdarch.20variant.29/with/608988032)


folkertdev:
Awesome, thanks!

PR #2188, opened 2026-07-13
Simplify scalable vector testing

sayantn:
For SVE types, e.g `svint8_t`, we can convert `&svint8_t` safely to `&[i8]` with length `svcntb()`.
For `svbool_t`, it is equivalent to `<vscale x 16 x i1>`, so the layout is the same as an `i8` array of size `2*vscale`, so we can use `svcntd()` to convert `&svbool_t` to `&[i8]`. I used a small `repr(transparent)` struct with a custom `Debug` implementation to make the debug output nicer

This approach seems to work nicely, at least for now (it was indeed failing a few days ago, maybe some LLVM update fixed it recently)

r? @davidtwco

### DIFF
--- a/crates/intrinsic-test/src/arm/mod.rs
+++ b/crates/intrinsic-test/src/arm/mod.rs
@@ -233,72 +233,91 @@ const fn svprfop_from_i32(value: i32) -> svprfop {
     }
 }
 
-macro_rules! debug_print_integral {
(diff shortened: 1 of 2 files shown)


davidtwco (approved):
If the tests are passing and this seems to work then happy to do this instead

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2188/77c0c7be37e9957842251a17d0a4d359948d68c6..f44dd5b6940d53cd3646e9bd9e24895d5f92959e)*

PR #2189, opened 2026-07-13
Re-enable ICX run

sayantn:
It was failing because Intel updated the version to 2026.1


folkertdev:
Not a blocker but should we request that this is mirrored also to prevent breakage in the future?

sayantn:
That might be difficult, this is an APT package. We can also probably try to not include the version number to prevent future breakages.

PR #2191, opened 2026-07-15
for sse4a functions compare only the lower quadword

### LINKED ISSUE #2190: Invalid sse4a tests
Quoting from the AMD64 manual for EXTRQ:

> Extracts specified bits from the lower 64 bits of the first operand (the destination XMM register). The
extracted bits are saved in the least-significant bit positions of the lower quadword of the destination;
the remaining bits in the lower quadword of the destination register are cleared to 0. The upper
quadword of the destination register is undefined.

However, the test here seems to assume that the upper bits are zero:
https://github.com/rust-lang/stdarch/blob/cb3aaba5a5491073d793688af9e2e8877e9da21f/crates/core_arch/src/x86/sse4a.rs#L154

The comparison can evaluate to true or false as undefined bits are compared.

Same applies for the extracti, insert and inserti tests.

These start failing on i686 with LLVM 23.

folkertdev:
fixes https://github.com/rust-lang/stdarch/issues/2190

the upper quadword is undefined, see the issue.

r? sayantn
cc @nikic

### DIFF
--- a/crates/core_arch/src/x86/sse4a.rs
+++ b/crates/core_arch/src/x86/sse4a.rs
@@ -28,6 +28,11 @@ unsafe extern "unadjusted" {
 ///
 /// If `length == 0 && index > 0` or `length + index > 64` the result is
 /// undefined.
+///
+/// The extracted bits are saved in the least-significant bit positions of the lower
+/// quadword of the destination; the remaining bits in the lower quadword of the
+/// destination register are cleared to 0. The upper quadword of the destination
+/// register is undefined.
 #[inline]
 #[target_feature(enable = "sse4a")]
 #[cfg_attr(test, assert_instr(extrq))]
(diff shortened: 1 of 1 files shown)


sayantn:
Sorry for the late review

crates/core_arch/src/x86/sse4a.rs
```diff
@@ -150,6 +162,11 @@ mod tests {
     use crate::core_arch::x86::*;
     use stdarch_test::simd_test;
 
+    // Normally this requires SSE2, but for tests it does not matter whether we use the instruction.
+    fn _mm_cvtsi128_si64(a: __m128i) -> i64 {
```
sayantn, line 166: Not to hold the queue up, but why did you not use the sse2 function? sse4a does imply sse2

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2191/594ef8b6bab670e28ae425b7d071cc1fdf8d8020..bd6e145a9546dcc60117a2beacd2276401644527)*

folkertdev: Well we already import `use crate::core_arch::x86::*;` here but it couldn't find that function. Maybe some `cfg` is incorrect somewhere? Anyhow this worked.

PR #2197, opened 2026-07-26
Add missing avx512vl intrinsics for u32->f32 conversions

### LINKED ISSUE #158196: Missing AVX-512 intrinsics from the avx512vl extension
The standard library is missing some AVX-512 intrinsics added in the avx512vl extension, introduced in Skylake-X. Not only are they not stabilized, the intrinsics are missing from stdarch entirely.

The VL extension for AVX-512 added the AVX-512 instruction variants for register sizes smaller than AVX-512.

Clang contains the `_mm256_cvtepu32_ps` intrinsic, see https://clang.llvm.org/doxygen/avx512vlintrin_8h_source.html which is defined as 

```c
static __inline__ __m256 __DEFAULT_FN_ATTRS256_CONSTEXPR
_mm256_cvtepu32_ps(__m256i __A) {
  return (__m256)__builtin_convertvector((__v8su)__A, __v8sf);
}
```

But stdarch only contains `_mm512_cvtepu32_ps` from the avx512f set but **not** `_mm256_cvtepu32_ps` from the avx512vl set!

The intrinsic lowers directly into a [single hardware instruction](https://www.felixcloutier.com/x86/vcvtudq2ps) so its absence is odd.

`_mm256_cvtepu32_ps` is also mysteriously absent from the [Intel Intrinsics Guide](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_cvtepu32_ps) as of version 3.6.9, along with `_mm128_cvtepu32_ps`, which makes the whole situation even weirder. But they are present in the [Intel Software Development Manual](https://cdrdv2-public.intel.com/835761/326018-sdm-vol-2c.pdf).

This matters because the instructions behind these intrinsics are very desirable for performance. Implementing these operations with AVX2 primitives takes [a lot of instructions](https://github.com/linebender/fearless_simd/blob/9c5343f6783d0b51eac94bed7bb063ac9818a4eb/fearless_simd/src/generated/avx2.rs#L7435-L7444). And there doesn't seem to be a way to reliably reach these avx512vl instructions from Rust.

While I've confirmed that `_mm256_cvtepu32_ps` and `_mm128_cvtepu32_ps` are missing, there are likely more of them that I simply haven't stumbled upon yet.

cc @folkertdev

Shnatsel:
Fixes https://github.com/rust-lang/rust/issues/158196

Adds 6 intrinsics that were apparently forgotten by Intel and not included into the Intrinsics Guide web viewer, despite being present in the Intel Software Development Manual and implemented in clang. They all lower into a single hardware instruction and there is no other way to access that instruction; the AVX2 emulation is very costly compared to these native conversions.

The implementation mirrors both the existing 512-bit-wide form and the clang implementation for these 128-bit and 256-bit forms.

This is my first contribution so I have no idea what I'm doing. I don't know what the stabilization process for these would be and if these are supposed to be instantly stable or not, so I just made up a new unstable feature name.

### DIFF
--- a/crates/core_arch/src/x86/avx512f.rs
+++ b/crates/core_arch/src/x86/avx512f.rs
@@ -13097,7 +13097,10 @@ pub const fn _mm512_maskz_cvtepu32_ps(k: __mmask16, a: __m512i) -> __m512 {
 /// Documentation on them can be found in the [Intel Software Development Manual](https://www.intel.com/content/www/us/en/developer/articles/technical/intel-sdm.html).
 #[inline]
 #[target_feature(enable = "avx512f,avx512vl")]
-#[unstable(feature = "stdarch_x86_avx512_vl_u32_to_f32_conversions", issue = "161585")]
+#[unstable(
+    feature = "stdarch_x86_avx512_vl_u32_to_f32_conversions",
+    issue = "161585"
+)]
 #[cfg_attr(test, assert_instr(vcvtudq2ps))]
 #[rustc_const_unstable(feature = "stdarch_const_x86", issue = "149298")]
 pub const fn _mm256_cvtepu32_ps(a: __m256i) -> __m256 {
@@ -13113,7 +13116,10 @@ pub const fn _mm256_cvtepu32_ps(a: __m256i) -> __m256 {
 /// Documentation on them can be found in the [Intel Software Development Manual](https://www.intel.com/content/www/us/en/developer/articles/technical/intel-sdm.html).
 #[inline]
 #[target_feature(enable = "avx512f,avx512vl")]
-#[unstable(feature = "stdarch_x86_avx512_vl_u32_to_f32_conversions", issue = "161585")]
+#[unstable(
+    feature = "stdarch_x86_avx512_vl_u32_to_f32_conversions",
+    issue = "161585"
+)]
 #[cfg_attr(test, assert_instr(vcvtudq2ps))]
 #[rustc_const_unstable(feature = "stdarch_const_x86", issue = "149298")]
 pub const fn _mm256_mask_cvtepu32_ps(src: __m256, k: __mmask8, a: __m256i) -> __m256 {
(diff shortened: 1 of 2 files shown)


folkertdev:
Can you rebase this? Our CI was broken when you opened  the PR.

crates/core_arch/src/x86/avx512f.rs
```diff
@@ -13091,6 +13091,104 @@ pub const fn _mm512_maskz_cvtepu32_ps(k: __mmask16, a: __m512i) -> __m512 {
     }
 }
 
+/// Convert packed unsigned 32-bit integers in a to packed single-precision (32-bit) floating-point elements, and store the results in dst.
+///
+/// Due to an omission by Intel, these intrinsics are not documented in the Intrinsics Guide.
+/// Documentation on them can be found in the [Intel Software Development Manual](https://www.intel.com/content/www/us/en/developer/articles/technical/intel-sdm.html).
+#[inline]
+#[target_feature(enable = "avx512f,avx512vl")]
+#[unstable(feature = "stdarch_x86_avx512_vl_updates", issue = "158196")]
```
folkertdev: Can you make a proper tracking issue for these? Something similar to https://github.com/rust-lang/rust/issues/153514.

Picking a more tightly scoped name like `stdarch_x86_avx512_vl_f32_to_u32_conversions` might be helpful when stabilizing.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2197/50afa9f3acca321f55f176107102693e6e67f4fa..849ba38d0fd84b851369ac092ee3a83b1081ff36)*

Shnatsel:
Rebased without conflicts.

Automatic verification fails because [Intel Intrinsics Guide](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_cvtepu32_ps) does not contain these intrinsics as of version 3.6.9. They are present in the [Intel Software Development Manual](https://cdrdv2-public.intel.com/835761/326018-sdm-vol-2c.pdf) and other sources, see https://github.com/rust-lang/rust/issues/158196 for details.

folkertdev:
I think you can add them to `crates/intrinsic-test/missing_x86_clang.txt` etc. for them to be skipped in that check.

Shnatsel:
Thank you for the guidance!

I've added Intrinsics Guide XML check exception, opened the proper tracking issue and changed the stability annotations to point to it. CI is green. This should be ready to go now.

Shnatsel:
Thank you! How do I mark the tracking issue as implemented? I don't think I can edit the labels on https://github.com/rust-lang/rust/issues/161585

folkertdev:
I've fixed it, but you can with `@rustbot label -label-name` I believe.

PR #2201, opened 2026-08-05
fix CI after LLVM bump

folkertdev:


folkertdev:
Hitting https://github.com/rust-lang/rust/issues/160580

sayantn:
can we just remove the `#[derive(Debug)]`?

PR #2202, opened 2026-08-05
Generate lockfile in intrinsic-test

### LINKED ISSUE #2193: Ship lockfile with intrinsic-test
intrinsic-test produces a new workspace, but doesn't ship any lockfile for it, making it impossible to test with `--locked`. It should be possible to dynamically generate a lockfile based on fixed versions of cc and dependencies and add the entries for the packages of each test chunk without touching crates.io at all and thus ensuring reproducibility.

cc https://github.com/rust-lang/rustc_codegen_cranelift/pull/1667/files#r3512505593

bjorn3:
This allows building the resulting workspace with --locked.

Fixes https://github.com/rust-lang/stdarch/issues/2193


crates/intrinsic-test/src/common/gen_rust.rs
```diff
@@ -66,17 +69,57 @@ pub fn write_bin_cargo_toml(
 members = [{members}]
 "#,
         members = (0..module_count).format_with(",", |i, fmt| fmt(&format_args!("\"mod_{i}\"")))
+    )?;
+    write!(
+        lockfile,
+        r#"# This file is automatically @generated by Cargo.
```
folkertdev, line 75: should we really hardcode this (as a string)?

folkertdev (commented):
Needs a rebase

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2202/50afa9f3acca321f55f176107102693e6e67f4fa..257af4414f3af63c684e0f464a8495d27f3481b1)*

crates/intrinsic-test/src/common/gen_rust.rs
bjorn3: This is the exact text that Cargo would produce. I could move it to a separate text file, but it isn't going to be automatically updateable either way. Without the Cargo.toml files generated by intrinsic-test, `cargo generate-lockfile` would just error. So I think updating it would be generate a workspace using intrinsic-test, run `cargo update` in this workspace and then copy the start of Cargo.lock here.

bjorn3:
How did this cause a miscompilation in CI???

```
---- core_arch::x86::avx512fp16::tests::test_mm256_add_ph stdout ----

thread 'core_arch::x86::avx512fp16::tests::test_mm256_add_ph' (5956) panicked at crates\core_arch\src\x86\avx512fp16.rs:17729:9:
assertion `left == right` failed
  left: Simd(0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40)
 right: Simd(0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40, 0x4c40)
```

bjorn3:
Rebased and CI passes just fine now.

PR #2204, opened 2026-08-06
sve: re-enable `sveorv` testing after LLVM upgrade

davidtwco:
This was disabled in rust-lang/stdarch#2160 due to an LLVM bug (llvm/llvm-project#203921) that was fixed in LLVM 23, which rustc now uses on nightly (rust-lang/rust#158734).


davidtwco:
CI failure seems unrelated

Amanieu:
LGTM once CI is fixed.

Issue #2208, opened 2026-08-17
Bad codegen for _mm_packus_epi16


hsivonen:
Consider https://rust.godbolt.org/z/dY8schnG4

Input:

```rust
#![feature(portable_simd)]
use std::simd::ToBytes;
use core::simd::u8x16;
use core::simd::u16x8;
use core::arch::x86_64::_mm_packus_epi16;

#[inline(always)]
pub fn simd_pack_sse2(a: u16x8, b: u16x8) -> u8x16 {
    unsafe {
        _mm_packus_epi16(a.into(), b.into()).into()
    }
}

#[inline(always)]
pub fn simd_pack_portable(a: u16x8, b: u16x8) -> u8x16 {
    let first: u8x16 = a.to_le_bytes();
    let second: u8x16 = b.to_le_bytes();
    let (ret, _) = first.deinterleave(second);
    ret
}

#[inline(always)]
fn split_u16_stride(stride: &[u16; 16]) -> (&[u16; 8], &[u16; 8]) {
    let (chunks, _) = stride.as_chunks::<8>();
    (&chunks[0], &chunks[1])
}

#[unsafe(no_mangle)]
pub fn pack_portable(src: &[u16; 16], dst: &mut [u8; 16]) {
    let (a, b) = split_u16_stride(src);
    let packed = simd_pack_portable((*a).into(), (*b).into());
    *dst = packed.to_array();
}

#[unsafe(no_mangle)]
pub fn pack_sse2(src: &[u16; 16], dst: &mut [u8; 16]) {
    let (a, b) = split_u16_stride(src);
    let packed = simd_pack_sse2((*a).into(), (*b).into());
    *dst = packed.to_array();
}
```

Output:

```asm
.LCPI0_0:
        .short  255
        .short  255
        .short  255
        .short  255
        .short  255
        .short  255
        .short  255
        .short  255
pack_portable:
        movdqu  xmm0, xmmword ptr [rdi]
        movdqu  xmm1, xmmword ptr [rdi + 16]
        movdqa  xmm2, xmmword ptr [rip + .LCPI0_0]
        pand    xmm1, xmm2
        pand    xmm0, xmm2
        packuswb        xmm0, xmm1
        movdqu  xmmword ptr [rsi], xmm0
        ret

.LCPI1_0:
        .short  255
        .short  255
        .short  255
        .short  255
        .short  255
        .short  255
        .short  255
        .short  255
pack_sse2:
[...]

PR #2211, opened 2026-08-19
Revert "x86: use `simd::intrinsics` for saturating packs"

folkertdev:
Reverts rust-lang/stdarch#2033


tgross35:
Any context here?

folkertdev:
Oh, yes https://github.com/rust-lang/stdarch/issues/2208. The pattern doesn't actually optimize well, and we don't have a good way in general to make it work using portable intrinsics right now.

PR #2214, opened 2026-08-20
neon: drop the align requirement on vld1/vst1/vld1 lane loads and stores, dup loads

### LINKED ISSUE #2198: ld1 inserts an alignment requirement
Current implementations of `ld1 { vN.d }[N]`  insert an alignment requirement in debug builds, since a pointer dereference assumes the element is aligned to its type's size, whereas to the best of my knowledge, ld1 doesn't have any particular alignment requirements.

```rust
pub unsafe fn vld1q_lane_u32<const LANE: i32>(ptr: *const u32, src: uint32x4_t) -> uint32x4_t {
    static_assert_uimm_bits!(LANE, 2);
    simd_insert!(src, LANE as u32, *ptr)
}

pub unsafe fn vld1q_lane_u64<const LANE: i32>(ptr: *const u64, src: uint64x2_t) -> uint64x2_t {
    static_assert_uimm_bits!(LANE, 1);
    simd_insert!(src, LANE as u32, *ptr)
}
```

valentynkit:
vld1/vst1 lane functions and vld1 dup functions were de-referencing pointer directly which requires it to be aligned to the element type. (Too strict alignment requirements)
The instructions don't requires alignment, so calling these with unaligned pointer caused UB.

Lane loads and stores, and dup loads were updated to use `read_unaligned()` and
`write_unaligned()` instead. 

Use crate::ptr::read_unaligned(ptr) instead of ptr.read_unaligned() because the same convention is used in neighboring code. Let me know if the other approach seems cleaner and should be used instead.

Also on armv7 the 64-bit loads now compile to `vld1.8` instead of  `vldr` (`vldr` requires an aligned address), so their arm `assert_instr` values updated to match. Without this the tests would fail after getting rid of alignment requirements in the function, and the f32 lane and dup loads LLVM emits `ldr` + `vmov`/`vdup.32` on arm instead of a NEON load form, so those four asserts now expect `ldr`. Is this acceptable, or I am missing something here, it feels pretty unreliable to change assertions, but not sure if there are a better way? 

Align requirement documented in:
[ARMv7-A and ARMv7-R Manual](https://support.arm.com/documentation/ddi0406/latest/) ch A3.2.1

This covers a bit more than the reported issue required to fix it, but there look to be more places with the same category of bug, because implementation requiring stricter alignment that instructions do.
I could take a look at those as follow-ups, if it will be valuable.

Fixes rust-lang/stdarch#2198


veluca93:
Note that the functions as written now are unsound, and as such this PR fixes a soundness bug (or multiple)

PR #1900, opened 2025-08-11
`Intrinsic-test`: Supporting `LoongArch` behavioural testing

madhav-madhusoodanan:

PR #2221, opened 2026-09-09
Stabilize LoongArch64 SIMD Intrinsics

heiher:
Tracking issue: https://github.com/rust-lang/rust/issues/162508

LoongArch64 SIMD intrinsics are already used by rustc itself and by a number of downstream projects. Experience from these real-world uses indicates that the API is ready for stabilization. Stabilizing the intrinsics would remove the nightly requirement for applications that currently depend on them.

## API surface

- [core::arch::loongarch64::lsx_*](https://doc.rust-lang.org/nightly/core/arch/loongarch64/index.html?search=loongarch64%3A%3Alsx_)
- [core::arch::loongarch64::lasx_*](https://doc.rust-lang.org/nightly/core/arch/loongarch64/index.html?search=loongarch64%3A%3Alasx_)

Note for subtree update

Remove this line: https://github.com/rust-lang/rust/blob/1edd55dcfcd573872c727fa3e086369a71661ee0/compiler/rustc_span/src/lib.rs#L20


crates/core_arch/src/loongarch64/lasx/generated.rs
```diff
@@ -909,36 +909,36 @@ unsafe extern "llvm-intrinsic" {
 
 #[inline]
 #[target_feature(enable = "lasx")]
-#[unstable(feature = "stdarch_loongarch", issue = "117427")]
+#[stable(feature = "stdarch_loongarch_simd", since = "CURRENT_RUSTC_VERSION")]
 pub fn lasx_xvsrar_b(a: m256i, b: m256i) -> m256i {
     unsafe { transmute(__lasx_xvsrar_b(transmute(a), transmute(b))) }
 }
```
tgross35, line 915: What's the reason for changing the feature name?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2221/44882e056bffd4584fc4d5bca662adcdd749edec..5f1c38dcbe8c60ff2e31c00386d65e56f270dcc3)*

tgross35:
Note that this stabilizes more cases of `rustc_legacy_const_generics`, which was discussed at https://github.com/rust-lang/rust/issues/149654. Cc @rust-lang/lang

Nominating since that's probably worth clarifying, and also this is a lot of surface area with potential sharp edges.

@rustbot label +I-libs-nominated

crates/core_arch/src/loongarch64/lasx/generated.rs
heiher: > What's the reason for changing the feature name?

There are also some unstable APIs gated by `stdarch_loongarch`. Having both stable and unstable APIs gated by the same feature name causes the following error:

```
error[E0711]: feature `stdarch_loongarch` is declared stable, but was previously declared unstable
```

Also, putting the SIMD intrinsics under `stdarch_loongarch_simd` makes the categorization more explicit, similar to `stdarch_loongarch_crc`.

heiher:
> Note that this stabilizes more cases of `rustc_legacy_const_generics`, which was discussed at [rust-lang/rust#149654](https://github.com/rust-lang/rust/issues/149654).

Thanks for pointing this out. As far as I know, no projects currently depend on the LoongArch SIMD API signatures using the `rustc_legacy_const_generics` form, so there are no legacy compatibility concerns here. If adding new instances of rustc_legacy_const_generics is not desirable, we can simply remove it.

heiher:
`rustc_legacy_const_generics` has been removed from LoongArch64 SIMD in [v2](https://github.com/rust-lang/stdarch/commit/bc04e1499a533a253bc8b1b86a7e27fd4715cc7a).

clarfonthey:
@rfcbot merge libs

Amanieu:
It would be nice to have this fuzz tested against the C intrinsics using intrinsic-test before stabilizing.

@rfcbot concern fuzz testing

adamgemmell:
There was a previous attempt at doing so at https://github.com/rust-lang/stdarch/pull/1900, though it never fully passed CI and will need some rebasing.

heiher:
> There was a previous attempt at doing so at #1900, though it never fully passed CI and will need some rebasing.

Thanks! There's a new impl in #2226 with all related CI checks passing.

traviscross:
With these intrinsics, Miri reports UB in safe code:

```rust
#![feature(stdarch_loongarch)]
use std::arch::is_loongarch_feature_detected;
use std::arch::loongarch64::*;

#[target_feature(enable = "lsx")]
fn probe() -> i32 {
  let ones = lsx_vrepli_b::<1>();
  let zero = lsx_vrepli_b::<0>();
  let r = lsx_vdiv_b(ones, zero);
  lsx_vpickve2gr_b::<0>(r)
}

fn main() {
  if is_loongarch_feature_detected!("lsx") {
    // SAFETY: We just checked: the `lsx` feature is present.
    let lane0 = unsafe { probe() };
    println!("lane 0 = {lane0}");
  }
}
```

[Godbolt link](https://rust.godbolt.org/#z:OYLghAFBqRAWIDGB7AJgUwKKoJYBdkAnAGhxAgCsQBGYgGwEMA7UQgVwGc9iPk3DE6EAHIApACYAzGDCiArACEAZugZ5%2B6CF1QMBcAPp1kyFrsRwAlPIAiogAwBBTugDU2kCDMIQODoeOmevoqahr6GHjoiJGoopIK9k4cru6eeh5GJsBeAGwALB4AVHEJjokSkvIKeLrA6HjBquqEmuhMDABGdK5x1i4S4nQcAB4DVnK2jkpMLgAOhMgdmhYuALRxmC44kuL9AOylDi4u3XguJugc/ZJ9Q8P6AG4ts3Q4%2Bh0ecQDC1BsQVvFEsdTi4AF7oBbXW4jR7PV7vT6SL52P4Aw7A%2BouQhQk4wh64B7vCAXDjEMEQ5BooG4%2B4PWY4RAAawe6HEwEICJA3xRkkwEEIVjKe0mDkS0xcAFsGDgmP99uitkotn5MoFzI1Qi1wvUojEZBABncxvLqccAPRmlwAZQcADFMAAVACaIBcAHVXBROGdzFFGehUK68HAejk7Eawy4Qs1XL45i1kkw8AA6U0nTGMJjoOw4thMDgMFTy%2BOLZb7WyAxzHY7zGV4OhMfWG5iuHO9eWZ7OiYVjErU7sigeJYQWYiMkB5cT0ERyEfEaDgKAIFAS%2BndEhkCArtcQkBMHDAOD1gCexCUODokUIHHIH2EdmExA6Mt0x5E1Efz/ahGPAHkOhQurvo%2BK4Sm0eC/kwdBvvej4YB0bDAF8DB0EMwHEBgUosN0IgPhhOAtNEOAsjesHEOgwxRGwkToXW6B0Lhj6vB0hCvl8aBCGReCEDgErAaOVC0JmrDevQOAdDekCjsgsx4DgJikasqw1IQdRnKq2R6Pkqx5oyTDIAA7kwqyvEwbDDKswBmWsqyIKxkTKcesyuF%2BaxfOcskmegLJ0L04iPsgLKEDxGCSVA7RgSAAb4PJTDEA8KFsJx4h2OIeQjqOvD8IINCjgxwizsQfGTqO8AgAQyB0KQ5C1Ipym1BmASaeY2m6fpRkmTK5mWWZj61QACi2dBWnAhlMIxxDbheEIAHIMJFwjsFwiAuPuh4no%2B0UECFNDEIJRUmAwKBWnguDjcIH48KdMoiI%2BFV0AAkoGEoETgeUzo%2BfFyAAHBYo7jj907CJIGVA/5f3xRCHCxROQA%3D%3D)

From https://github.com/rust-lang/stdarch/pull/2078, see:

- https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/stdarch/pull/2078#discussion_r3121430126
- https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/stdarch/pull/2078#discussion_r3121435644
- https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/stdarch/pull/2078#discussion_r3121452899

cc @sayantn

heiher:
> With these intrinsics, Miri reports UB in safe code:

Good catch!

I double-checked this with the LoongArch ISA developer. It differs from the previous comment in PR [#2078](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/stdarch/pull/2078#discussion_r3122447444). The ISA defines `vdiv`/`vmod` to return 0 if the divisor is 0. (This behavior will be documented in the SIMD reference)

So it seems we can't use portable `simd_div`/`simd_rem` here. I'll revert this part. Any other better approach? Thanks!

Amanieu:
There are 2 options, depending on what the intrinsic behavior should be:
1. Make it unsafe and declare that it is UB if any vector lane of the divisor is 0.
2. Keep it safe and use LLVM target-specific intrinsics or inline asm to implement it.

heiher:
> There are 2 options, depending on what the intrinsic behavior should be:
> 
> 1. Make it unsafe and declare that it is UB if any vector lane of the divisor is 0.
> 2. Keep it safe and use LLVM target-specific intrinsics or inline asm to implement it.

Thanks! A new PR has been created based on option 2: https://github.com/rust-lang/stdarch/pull/2229

PR #2224, opened 2026-09-13
only run `f16` tests when `target_has_reliable_f16`

folkertdev:
In response to [#t-libs/stdarch > CI is broken](https://rust-lang.zulipchat.com/#narrow/channel/208962-t-libs.2Fstdarch/topic/CI.20is.20broken/with/615044157).

In https://triage.rust-lang.org/gha-logs/rust-lang/stdarch/103773369987?pr=2223 we hit a spurious issue, where the `f16` tests sometimes fail on windows. The failures look like an ABI issue, and we know that the `f16` ABI is broken with older windows GNU toolchains. So, guard the `f16` tests with `target_has_reliable_f16` to hopefully prevent this. 

I'm not sure how to prove that this works, but this sounds plausible to me and really more correct: we actually should only run the `f16` tests when `target_has_reliable_f16`.


adamgemmell (approved):
Sounds good, I came to the same conclusion separately. There's other f16 intrinsics on that target but I've only ever seen the avx512 ones fail, so we can expand this in the future as needed.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2224/43304e7a8c5f8ebfc0f0b12dbf42f24e0da2f147..29a8fe8afbb40e8681dc07799206b99a2b18654a)*

PR #2225, opened 2026-09-14
Refactor generation to use a unified formatting approach

Kobzol:
This PR ensures that all generated files are always formatted, with a unified version of rustfmt, and with the same rustfmt config.

This should unblock https://github.com/rust-lang/rust/pull/156674. I also started using Clap in the generator crates, because the argument parsing there was kinda wild, and I wanted to enable passing `--rustfmt-path`, so that we can avoid messing with `PATH` in bootstrap.


Kobzol:
r? folkertdev

Kobzol:
This test failure looks unrelated to this PR, I think.

folkertdev:
A rebase should fix it, a fix for that issue was just merged.

PR #2226, opened 2026-09-15
intrinsic-test: Support testing LoongArch64 SIMD intrinsics

heiher:


crates/intrinsic-test/src/common/mod.rs
```diff
@@ -49,6 +49,8 @@ pub trait SupportedArchitecture: Sized {
     const C_PRELUDE: &str;
     const RUST_PRELUDE: &str;
 
+    const C_NAME_PREFIX: &str;
```
folkertdev, line 54: Can you add a comment on what this is for?

folkertdev (commented):
Otherwise this looks good to me.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/stdarch/2226/899814bf407111631eafd0491cb91149b3941d35..e16405892f929c8a43f35b422f8aa230d466ef11)*

crates/intrinsic-test/src/common/mod.rs
heiher: Done.

PR #2227, opened 2026-09-15
core_arch: Remove redundant LoongArch64 SIMD tests

heiher:
Remove redundant LoongArch64 SIMD intrinsic tests from core_arch, as they are covered by intrinsic-test.


heiher:
Depends on #2226

folkertdev:
For x86_64 we've mostly kept these tests, they are faster to run and deterministic. We can drop tests if really all they check is that the intrinsic is hooked up I guess.

But also the intrinsic tests do some filtering (e.g. skip anything with an argument that is a pointer), so they don't test everything.

heiher:
Thanks! Let's keep these tests.

